PR #1, opened 2020-05-19
Use 2018 Rust and json test example

franziskuskiefer:


src/utils.rs
```diff
@@ -34,6 +35,28 @@ pub fn random_u32() -> u32 {
     OsRng.next_u32()
 }
 
+pub(crate) fn u32_to_bytes(d: &[u32]) -> Vec<u8> {
```
raphaelrobert, line 38: I think we can reuse `codec.rs` for this:

https://github.com/raphaelrobert/maelstrom/blob/master/src/codec.rs#L119

PR #6, opened 2020-06-22
use "secret" in key schedule

franziskuskiefer:
This changed in https://github.com/cfrg/draft-irtf-cfrg-hpke/commit/82422d9fbf5d18bb566104e7d0961b124e5a6429


raphaelrobert:
I guess the test vectors must have changed as well because of that. At least they now fail with that change.

raphaelrobert:
Outdated

PR #10, opened 2020-08-12
Cleanup 1

franziskuskiefer:
Some first cleanup. Mostly moving code around and deleting some unused code.


src/framing.rs
```diff
@@ -862,8 +862,13 @@ fn codec() {
 fn padding() {
     use crate::ciphersuite::*;
     use crate::utils::*;
+    use rand::RngCore;
 
-    let ciphersuite = Ciphersuite::new(Name::MLS10_128_DHKEMX25519_CHACHA20POLY1305_SHA256_Ed25519);
+    fn random_usize() -> usize {
```
raphaelrobert, line 867: We should have that function in `utils.rs`

franziskuskiefer: Nope, it's only used in a test. So if you want it outside it must go into a test_util.

franziskuskiefer: All the testing needs quite some clean-up. But I figured this PR is already big enough.

src/framing.rs
```diff
@@ -245,7 +245,7 @@ impl MLSCiphertext {
                     &epoch_secrets.handshake_secret,
                     "hs key",
                     &mls_plaintext.sender.encode_detached().unwrap(),
-                    CHACHAKEYBYTES,
+                    CHACHA_KEY_BYTES,
```
raphaelrobert, line 248: Since I see this: could you add a TODO for agility here?

src/utils.rs
```diff
@@ -18,39 +18,26 @@ use evercrypt::prelude::*;
...
-pub fn bytes_to_hex(bytes: &[u8]) -> String {
+pub(crate) fn bytes_to_hex(bytes: &[u8]) -> String {
     let mut hex = String::new();
     for b in bytes {
         hex += &format!("{:02X}", *b);
     }
     hex
 }
-
-pub fn hex_to_bytes(hex: &str) -> Vec<u8> {
```
raphaelrobert, line 49: We might still need this for further testing

franziskuskiefer: Then it must go to testing utils as well 🙂

PR #14, opened 2020-08-15
API refactoring

raphaelrobert:


src/client.rs
```diff
@@ -0,0 +1,55 @@
+// maelstrom
```
franziskuskiefer, line 1: So can we delete this? 🙂

src/framing.rs
```diff
@@ -246,9 +225,8 @@ impl MLSCiphertext {
             + 2
             + TAG_BYTES
             + 4;
-        let mut padding_length = (config.padding_block_size as usize)
-            - (padding_offset % (config.padding_block_size as usize));
-        if (config.padding_block_size as usize) == padding_length {
+        let mut padding_length = PADDING_SIZE - (padding_offset % PADDING_SIZE);
+        if PADDING_SIZE == padding_length {
```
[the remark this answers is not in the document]
franziskuskiefer: yeah... this is `let padding_length = (PADDING_SIZE - (padding_offset % PADDING_SIZE)) % PADDING_SIZE;` I guess.
I just complain about this because the `mut` here shouldn't be necessary.

src/group/errors.rs
```diff
@@ -0,0 +1,44 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+
+pub enum WelcomeError {
+    CiphersuiteMismatch,
```
franziskuskiefer, line 18: For the spec these errors should get integer values.

src/group/mls_group/apply_commit.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    // Verify KeyPackage extensions
+    if let Some(path) = commit.path {
+        if !is_own_commit {
+            let parent_hash = provisional_tree.compute_parent_hash(NodeIndex::from(sender));
+            if let Some(received_parent_hash) = path
+                .leaf_key_package
+                .get_extension(ExtensionType::ParentHash)
+            {
+                if let ExtensionPayload::ParentHash(parent_hash_inner) = received_parent_hash {
```
franziskuskiefer, line 163: That's a lot of `if`s 😆

Cargo.toml
```diff
@@ -23,3 +23,6 @@ criterion = "^0.2"
 [[bench]]
 name = "benchmark"
 harness = false
+
+[package.metadata.clippy]
+type-complexity-threshold = 1_000_000
```
franziskuskiefer, line 28: eol

src/group/errors.rs
```diff
@@ -0,0 +1,43 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+
+pub enum WelcomeError {
```
franziskuskiefer, line 17: So this should get into the spec as well as enums.
I have to look at these a little closer but have to make sure that the level of detail is not an issue. At first sight it looks like a lot of different error code.

src/group/mls_group/api.rs
```diff
@@ -0,0 +1,101 @@
...
+        force_self_update: bool,
+    ) -> CreateCommitResult;
+
+    /// Apply a `Commit` message
+    fn apply_commit(
+        &mut self,
+        mls_plaintext: MLSPlaintext,
+        proposals: Vec<(Sender, Proposal)>,
+        own_key_packages: Vec<(HPKEPrivateKey, KeyPackage)>,
+    ) -> Result<(), ApplyCommitError>;
```
franziskuskiefer, line 75: For us (not for the spec API) we could return `self` here in case of success to allow chaining.

franziskuskiefer (commented):
Shall we merge this? 🙂

PR #15, opened 2020-09-08
Cleanup 4

franziskuskiefer:


franziskuskiefer:
There's a followup starting to separate tree logic from MLS logic and tests here https://github.com/franziskuskiefer/maelstrom/pull/1

src/framing.rs
```diff
@@ -718,6 +785,7 @@ impl Codec for MLSCiphertextContent {
         Ok(())
     }
 
+    // FIXME: delete
```
raphaelrobert, line 788: Can this already be deleted?

src/group/mls_group/api.rs
```diff
@@ -17,20 +17,22 @@
...
+use crate::messages::{proposals::*, *};
+use crate::tree::{index::LeafIndex, node::*};
+
+pub trait Api: Sized {
     /// Create a new group.
     fn new(
         group_id: &[u8],
         ciphersuite: Ciphersuite,
-        key_package_bundle: (HPKEPrivateKey, KeyPackage),
+        key_package_bundle: KeyPackageBundle,
```
raphaelrobert, line 28: So far I avoided `KeyPackageBundle` in the API, because it introduces a new type that doesn't exist in the spec. But I agree that it looks better than a tuple.

src/group/mls_group/mod.rs
```diff
@@ -21,44 +21,39 @@ mod new_from_welcome;
...
+use std::cell::{Ref, RefCell};
+
 pub struct MlsGroup {
     ciphersuite: Ciphersuite,
     group_context: GroupContext,
     generation: u32,
     epoch_secrets: EpochSecrets,
-    astree: ASTree,
-    tree: RatchetTree,
+    astree: RefCell<ASTree>,
```
raphaelrobert, line 43: TODO: Have more than one ASTree and keep only the last n trees.

src/group/mls_group/mod.rs
```diff
@@ -200,23 +194,33 @@ impl Api for MlsGroup {
...
+        let generation = astree.get_generation(mls_plaintext.sender.sender);
+        let application_secrets = astree
+            .get_secret(&self.ciphersuite, mls_plaintext.sender.sender, generation)
+            .unwrap();
+        MLSCiphertext::new_from_plaintext(&mls_plaintext, &self, generation, &application_secrets)
     }
+
     fn decrypt(&mut self, mls_ciphertext: MLSCiphertext) -> MLSPlaintext {
-        let context = &self.get_context();
+        let tree = self.tree.borrow();
```
raphaelrobert, line 206: We might use the roster() function elsewhere going forward, any reason it should be inline here?

src/tree/sender_ratchet.rs
```diff
@@ -0,0 +1,125 @@
+use crate::ciphersuite::*;
+use crate::codec::*;
+use crate::tree::{astree::*, index::LeafIndex};
+
+const OUT_OF_ORDER_TOLERANCE: u32 = 5;
```
raphaelrobert, line 5: FYI: This is the number of ASTrees a group should keep.

tests/test_group.rs
```diff
@@ -1,10 +1,69 @@
...
+        &ciphersuite,
+        &alice_identity.get_signature_key_pair().get_private_key(), // TODO: bad API, we shouldn't have to get the private key out here (this function shouldn't exist!)
+        Credential::Basic(alice_credential),
+        None,
+    );
+    let alice_key_package = alice_key_package_bundle.get_key_package();
+
+    let charlie_key_package_bundle = KeyPackageBundle::new(
+        &ciphersuite,
+        &alice_identity.get_signature_key_pair().get_private_key(), // TODO: bad API, we shouldn't have to get the private key out here (this function shouldn't exist!)
```
raphaelrobert, line 40: Agreed about the bad API, but we'll have to touch `Identity` anyway going forward.

src/group/mls_group/api.rs
franziskuskiefer: Yeah I was wondering about that, what's the best way is to handle this.
For this generic API having two separate arguments for key and key package might be the best.

src/group/mls_group/mod.rs
franziskuskiefer: I filed https://github.com/raphaelrobert/maelstrom/issues/16

src/group/mls_group/mod.rs
franziskuskiefer: This is the only place where the `roster` function was used. Also, it's not usable as is (`self` is borrowed immutable and it's returning borrowed values).
I think I'd keep it inlined here for this iteration and see how we can pull it out once we need it elsewhere.

tests/test_group.rs
franziskuskiefer: I filed https://github.com/raphaelrobert/maelstrom/issues/17

src/group/mls_group/mod.rs
raphaelrobert: Fair enough

PR #22, opened 2020-09-17
Extensible credentials

kkohbrok:
Introduce a (proprietary) type of credentials that has an additional field for extensions. Additionally, I introduce the DeviceCapabilities extension, which I adapted from ones present in the MI implementation.


src/creds.rs
```diff
@@ -185,6 +196,53 @@ impl Codec for BasicCredential {
...
+
+impl Codec for ExtensibleCredential {
+    fn encode(&self, buffer: &mut Vec<u8>) -> Result<(), CodecError> {
+        encode_vec(VecSize::VecU16, buffer, &self.identity)?;
+        self.ciphersuite.encode(buffer)?;
+        self.public_key.encode(buffer)?;
+        encode_vec(VecSize::VecU32, buffer, &self.extensions)?;
+        Ok(())
+    }
+    fn decode(cursor: &mut Cursor) -> Result<Self, CodecError> {
```
franziskuskiefer, line 232: Please don't implement `decode` if it's not absolutely necessary.

src/creds.rs
```diff
@@ -185,6 +196,53 @@ impl Codec for BasicCredential {
     // }
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub struct ExtensibleCredential {
```
franziskuskiefer, line 200: What is this struct doing? (comment)

src/extensions.rs
```diff
@@ -71,6 +71,7 @@ pub enum ExtensionType {
     KeyID = 3,
     ParentHash = 4,
     RatchetTree = 5,
+    DeviceCapabilities = 6,
```
franziskuskiefer, line 74: So this is a custom extension. They must go into `0xff00-0xffff`.

src/creds.rs
franziskuskiefer, line 200: Can you put this new credential into its own module (file)? Then you can add doc comments there that describe how it works. Since it's a custom extension this needs a little more description.

src/extensions.rs
```diff
@@ -270,6 +271,78 @@ impl RatchetTreeExtension {
     }
 }
 
+#[repr(u32)]
+#[cfg_attr(rustfmt, rustfmt_skip)]
```
franziskuskiefer, line 275: Hm, I don't really like `rustfmt_skip`. I understand that you like the formatting here. But couldn't we write this in a way that's coherent with formatting?

src/extensions.rs
```diff
@@ -270,6 +271,78 @@ impl RatchetTreeExtension {
...
+    AddCap          = 0b0001u32,
+    RemoveCap       = 0b0010u32,
+    NonRemovableCap = 0b0100u32,
+    SelfUpdateCap   = 0b1000u32,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct DeviceCapabilities(pub u32);
+
+impl DeviceCapabilities {
```
franziskuskiefer, line 286: Can you put the new extension in it's own module?
It'll also need documentation.

src/extensions.rs
```diff
@@ -270,6 +271,78 @@ impl RatchetTreeExtension {
...
+#[cfg_attr(rustfmt, rustfmt_skip)]
+pub enum DeviceCapabilityType {
+    AddCap          = 0b0001u32,
+    RemoveCap       = 0b0010u32,
+    NonRemovableCap = 0b0100u32,
+    SelfUpdateCap   = 0b1000u32,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct DeviceCapabilities(pub u32);
```
franziskuskiefer, line 284: I wonder why we need `DeviceCapabilities` as a struct. Isn't the `enum` enough?

src/extensions.rs
```diff
@@ -270,6 +271,78 @@ impl RatchetTreeExtension {
...
+
+impl DeviceCapabilities {
+    pub fn is_permanent(&self) -> bool {
```
franziskuskiefer, line 287: All these functions should just call a single function that is `match` on `self.0`.

franziskuskiefer (changes requested):
A good first version but a couple things that need to get addressed.
Also we need tests for all of the new code.

src/creds.rs
kkohbrok: Fair enough. The reason being that we need encode for signatures, but decode only if we expect to receive it from the network?

franziskuskiefer: I want to get rid of almost all decodes. Right, it's only to build the object from bytes. And this will only work as long as it's a simple object that doesn't have any references to other objects.

kkohbrok: I see. I think a lot of the time, the objects are 1-to-1 implementations of the structs we have in the spec, so it should be possible for most of them. But of course that doesn't always make sense.

src/extensions.rs
kkohbrok: This is all existing code from the original MI implementation. I think the idea was that the the `u32` is used to encode some combination of `DevicecapabilityTypes`. So if we do it like that, I think we need both the `DeviceCapabilities` struct _and_ the enum. Is that what you meant? I'm not sure if this is the best way to implement this kind of pattern.

kkohbrok: I guess `DeviceCapabilities` should probably use a Set of `DeviceCapabilityType`.

kkohbrok: The disadvantage being that (if I understand correctly), enums can't represent HashSet instances :-(

raphaelrobert: Going forward we should also see to what extent we want to have custom extensions in maelstrom. For testing purposes there is no problem, but ideally maelstrom should be as generic as possible.

kkohbrok: Maybe it's better if I leave all this stuff on my own fork/branch for now. I can still do the PR at a later date.

franziskuskiefer: > Going forward we should also see to what extent we want to have custom extensions in maelstrom. For testing purposes there is no problem, but ideally maelstrom should be as generic as possible.

We should add non-standard things behind feature flags then it's ok to have them in.

src/extensions.rs
kkohbrok: I don't think that's going to be much nicer. The `match` statement would have to return a set of `DeviceTypes` and `DeviceCapabilityTypes`. Then the individual function would have to check if the specific `DeviceType` or `DeviceCapabilityType` is among those returned. Is that what you meant?

raphaelrobert:
I will close this for now due to inactivity. We can always open it again later.

PR #23, opened 2020-09-21
Remove Sender from Group API

### LINKED ISSUE #20: create_commit
I don't understand how to use `create_commit`. There are two different arguments for `KeyPackageBundle` that are unclear how to use and there's no way to create a `Sender` to pass in `proposals`.

https://github.com/raphaelrobert/maelstrom/blob/843120c6821d4e01304ba92127749889f7817ce3/src/group/mls_group/api.rs#L61-L69

franziskuskiefer:
`Sender` should not be used in the group API. This PR removes it from all public group APIs.

Make sure to review this with `Hide whitespace changes` enabled.

Fixes #20

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -17,7 +17,6 @@
 use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
 use std::collections::HashMap;
 use std::convert::*;
-use std::io::Write;
 
 #[derive(Debug)]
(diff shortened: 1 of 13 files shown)


franziskuskiefer:
@raphaelrobert can you add @kkohbrok to the repo?

raphaelrobert (approved):

PR #39, opened 2020-09-29
fix type confusion after hpke update

franziskuskiefer:
HPKE has these types now as well and we have to avoid confusing the types.
They should go away here, but we can only do that after the other open PRs are merged because it requires quite a number of changes.

### DIFF
--- a/src/ciphersuite/mod.rs
+++ b/src/ciphersuite/mod.rs
@@ -21,8 +21,13 @@
 
 use evercrypt::prelude::*;
 use hpke::{
-    aead::Mode as HpkeAeadMode, kdf::Mode as HpkeKdfMode, kem::Mode as KemMode, Hpke, Mode,
+    aead::Mode as HpkeAeadMode, kdf::Mode as HpkeKdfMode, kem::Mode as KemMode,
+    HPKEKeyPair as RealHPKEKeyPair, HPKEPrivateKey as RealHPKEPrivateKey,
+    HPKEPublicKey as RealHPKEPublicKey, Hpke, Mode,
 };
+
+// TODO: re-export for other parts of the library when we can use it
+// pub(crate) use hpke::{HPKEKeyPair, HPKEPrivateKey, HPKEPublicKey};
 
 mod ciphersuites;
 mod codec;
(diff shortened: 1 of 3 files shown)


src/ciphersuite/mod.rs
```diff
@@ -258,7 +267,9 @@ impl Ciphersuite {
     ) -> HpkeCiphertext {
         // TODO: put hpke in the ciphersuite.
         let hpke = Hpke::new(Mode::Base, self.hpke_kem, self.hpke_kdf, self.hpke_aead);
-        let (kem_output, ciphertext) = hpke.seal(&pk_r.value, info, aad, ptxt, None, None, None);
+        let (kem_output, ciphertext) = hpke
+            .seal(&pk_r.into(), info, aad, ptxt, None, None, None)
```
kkohbrok, line 271: Do I understand correctly, that the `pk_r` (or `sk_r` further below) is cloned here? I think you mentioned that we wanted to avoid cloning key material, so I assume that it's promptly consumed by `seal`?

src/ciphersuite/mod.rs
```diff
@@ -51,6 +56,8 @@ pub enum HKDFError {
     InvalidLength,
 }
 
+// TODO: remove these and use the proper types from HPKE.
+
```
kkohbrok, line 60: I was wondering why we weren't doing that. The functions below seem to wrap around the everycrypt HPKE implementation without really doing anything.

kkohbrok (commented):
I assume this is a bit of stop-gap fix? I assume it's too big of a change to remove the wrapper altogether as indicated in the TODO?

src/ciphersuite/mod.rs
franziskuskiefer: We will do that as soon as the other PRs are merged. See https://github.com/raphaelrobert/maelstrom/pull/39#issue-494802567
This needs another round of clean up but there's too much in flight and the change is too intrusive.

src/ciphersuite/mod.rs
franziskuskiefer: Yes, this clones right now. But same argument as below 🙂

franziskuskiefer: Yes, this clones. But same comment as below 🙂 So this should go away soon.

kkohbrok (approved):

franziskuskiefer:
> I assume this is a bit of stop-gap fix? I assume it's too big of a change to remove the wrapper altogether as indicated in the TODO?

Right, we need this, otherwise this will stop building 🙂 
But the `ciphersuite` should get smaller and only be a thin convenience wrapper in the end (if needed at all).

Issue #28, opened 2020-09-24
[Meta] Separate ratchet tree


franziskuskiefer:
Meta ticket to track separation of the ratchet tree.

Depends on 
- [x] #26  
- [x] #25 
- [x] #24

Issue #27, opened 2020-09-24
[Meta] Separate tree logic


franziskuskiefer:
Meta ticket to track separation of the tree logic.

PR #66, opened 2020-10-07
tree reorgv2

### LINKED ISSUE #24: PathKeypairs should not hold key pairs but again only private keys
https://github.com/raphaelrobert/maelstrom/blob/b2b72848c9a656f8ab5bf907249a1833414e5908/src/tree/path_key_pairs.rs#L7-L7

### LINKED ISSUE #25: Replace `Vec` with a HashMap in PathKeypairs
https://github.com/raphaelrobert/maelstrom/blob/b2b72848c9a656f8ab5bf907249a1833414e5908/src/tree/path_key_pairs.rs#L5-L5

### LINKED ISSUE #26: OwnLeaf functions are unrelated to struct
https://github.com/raphaelrobert/maelstrom/blob/b2b72848c9a656f8ab5bf907249a1833414e5908/src/tree/own_leaf.rs#L5-L5

franziskuskiefer:
rebase of #29 

In order to allow for a clean separation of the tree logic, ratchet tree, key schedule, and the group that build on all of this the tree needs to be re-organised.

This is the first of a series of PRs to achieve this.

cc #27 #28
fixes #25 
fixes #24 
fixes #26


franziskuskiefer:

This is the last PR this can be ignored. But because of the re-base we can't really do anything else here ☹️

src/group/errors.rs
```diff
@@ -44,3 +46,14 @@ pub enum ApplyCommitError {
 pub enum CreateCommitError {
     CannotRemoveSelf = 300,
 }
+
```
raphaelrobert, line 49: Not a blocker, but we might want to log exact errors in the initial debugging phase before aggregating errors.

One way would be to have something like

```
WelcomeError::InvalidRatchetTree(TreeError)
```

src/tree/mod.rs
```diff
@@ -289,105 +246,162 @@ impl RatchetTree {
...
 
         // Decrypt the ciphertext of that node
-        let hpke_ciphertext = &direct_path.nodes[common_ancestor_sender_dirpath_index]
-            .encrypted_path_secret[position_in_resolution];
+        let common_ancestor_node = match update_path.nodes.get(common_ancestor_sender_dirpath_index)
+        {
+            Some(node) => node,
+            None => return Err(TreeError::InvalidArguments),
+        };
+        assert_eq!(
```
raphaelrobert, line 296: Shouldn't this be a `debug_assert()`? Otherwise the following code wouldn't get executed.

src/tree/mod.rs
```diff
@@ -289,105 +246,162 @@ impl RatchetTree {
...
+            .private_tree
+            .generate_path_keypairs(&self.ciphersuite, &common_path)?;
+
+        // Check that the public keys are consistent with the update path.
+        for (i, key) in new_path_public_keys
+            .iter()
+            .enumerate()
+            .take(common_path.len())
+        {
+            assert_eq!(&update_path.nodes[sender_path_offset + i].public_key, key);
```
raphaelrobert, line 352: Same here regarding the `assert()`

src/tree/mod.rs
```diff
@@ -474,34 +498,84 @@ impl RatchetTree {
...
+        public_keys: &[HPKEPublicKey],
+        path: &[NodeIndex],
+    ) -> Result<(), TreeError> {
+        assert_eq!(public_keys.len(), path.len());
+        if public_keys.len() != path.len() {
+            return Err(TreeError::InvalidArguments);
+        }
         for i in 0..path.len() {
-            let node = ParentNode::new(keypairs[i].get_public_key().clone(), &[], &[]);
+            // TODO: drop clone
```
raphaelrobert, line 564: Can we drop the clone by using an iterator?

src/tree/mod.rs
```diff
@@ -474,34 +498,84 @@ impl RatchetTree {
...
+            let node = ParentNode::new(public_keys[i].clone(), &[], &[]);
             self.nodes[path[i].as_usize()].node = Some(node);
         }
+        Ok(())
     }
 
+    // pub fn merge_keypairs(&mut self, keypairs: &[HPKEKeyPair], path: &[NodeIndex]) {
```
raphaelrobert, line 571: Remove?

src/tree/private_tree.rs
```diff
@@ -0,0 +1,209 @@
...
+//!
+
+use super::{index::NodeIndex, path_keys::PathKeys, TreeError};
+use crate::ciphersuite::{Ciphersuite, HPKEKeyPair, HPKEPrivateKey, HPKEPublicKey};
+use crate::codec::{Codec, CodecError};
+use crate::messages::CommitSecret;
+use crate::schedule::hkdf_expand_label;
+
+#[derive(Debug)]
+pub(crate) struct OwnLeaf {
```
raphaelrobert, line 12: Since you already introduce the term private tree, shouldn't this be renamed to `PrivateTree`?

src/utils.rs
```diff
@@ -28,12 +29,14 @@ pub(crate) fn random_u32() -> u32 {
...
 }
 
+#[cfg(test)]
+pub(crate) fn random_u8() -> u8 {
+    get_random_vec(1)[0]
+}
+
+#[inline]
 pub(crate) fn zero(length: usize) -> Vec<u8> {
-    let mut result: Vec<u8> = vec![];
```
raphaelrobert, line 32: 🤦‍♀️

raphaelrobert (changes requested):

src/tree/mod.rs
franziskuskiefer: It should!

raphaelrobert: There a few more places where that's the case, should I flag them?

src/tree/mod.rs
franziskuskiefer: This requires a couple more changes. Because we only have a reference here and the caller needs the keys again.
Ideally the caller should take them out of the tree after this action and this should consume. But we should do that in a follow up.

src/tree/private_tree.rs
franziskuskiefer: yeah... I meant to do that ...

src/tree/mod.rs
franziskuskiefer: I replace all in here.

src/group/errors.rs
franziskuskiefer: Yeah, this is a little basic and we should make error handling a little nicer.
I filed #83 to follow up on this.

src/group/mls_group/new_from_welcome.rs
```diff
@@ -89,29 +85,33 @@ impl MlsGroup {
...
+            let common_ancestor_index = treemath::common_ancestor_index(
                 tree.get_own_node_index(),
                 NodeIndex::from(group_info.signer_index),
             );
-            let common_path = treemath::dirpath_root(common_ancestor, tree.leaf_count());
-            let (path_secrets, _commit_secret) = PathKeypairs::continue_path_secrets(
+            let common_path = treemath::direct_path_root(common_ancestor_index, tree.leaf_count());
+
+            // Update the private tree.
+            let own_leaf = tree.get_private_tree_mut();
```
raphaelrobert, line 103: Rename own leaf here as well

src/tree/mod.rs
```diff
@@ -289,105 +246,162 @@ impl RatchetTree {
...
+            Node::new_leaf(Some(update_path.leaf_key_package.clone()));
         self.compute_parent_hash(NodeIndex::from(sender));
-        commit_secret
+
+        // TODO: Do we really want to return the commit secret here?
+        Ok(self.private_tree.get_commit_secret().clone())
     }
+
+    /// Update the private tree with the new `KeyPackageBundle`.
     pub(crate) fn replace_own_leaf(
```
raphaelrobert, line 370: Rename own leaf here as well

src/tree/mod.rs
```diff
@@ -289,105 +246,162 @@ impl RatchetTree {
...
+        let _path_option = self.replace_private_tree(
+            &key_package_bundle,
+            group_context,
+            false, /* without update path */
+        )?;
+        Ok(self.private_tree.get_commit_secret())
     }
+
+    /// Update the private tree.
     pub(crate) fn refresh_own_leaf(
```
raphaelrobert, line 384: Rename own leaf here as well

raphaelrobert (approved):

Issue #85, opened 2020-10-09
Add global config


franziskuskiefer:
There's a [stub](https://github.com/raphaelrobert/maelstrom/blob/master/src/config.rs) for a configuration of MLS already. This should get extended.

- [x] make the config configurable
- [x] add ciphersuite to config
- [x] make sure the config is used everywhere

PR #86, opened 2020-10-09
Extensions

### LINKED ISSUE #32: [Meta] Implement extensions
Meta issue to implement extensions properly.

- [x] #59
- [x] #99 
- [x] #100
- [x] #101

### LINKED ISSUE #59: Make extensions spec compliant (#379)
Extensions aren't fully implemented yet.

One issue is: `8` instead of `12`, see PR.

https://github.com/mlswg/mls-protocol/pull/379

franziskuskiefer:
This PR implements and cleans up the extensions and makes them compliant with the current state of the spec.

Fixes #32 
Fixes #59

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,0 +1,55 @@
+//! This config contains all structs, enums and functions to configure MLS.
+//!
+
+use crate::codec::{Codec, CodecError};
+use crate::errors::ConfigError;
+
+/// # Protocol Version
+///
+/// 7. Key Packages
+///
+/// ```text
+/// enum {
+///     reserved(0),
+///     mls10(1),
+///     (255)
+/// } ProtocolVersion;
+/// ```
+///
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[repr(u8)]
+pub enum ProtocolVersion {
+    Reserved = 0,
+    Mls10 = 1,
+}
+
+/// There's only one version right now, which is the default.
+impl Default for ProtocolVersion {
+    fn default() -> Self {
+        ProtocolVersion::Mls10
+    }
+}
+
+impl Codec for ProtocolVersion {
+    fn encode(&self, buffer: &mut Vec<u8>) -> Result<(), CodecError> {
+        (*self as u8).encode(buffer)?;
+        Ok(())
+    }
+}
+
+impl ProtocolVersion {
+    /// Convert an integer to the corresponding protocol version.
+    ///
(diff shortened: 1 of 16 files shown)


src/extensions/key_package_id_extension.rs
```diff
@@ -0,0 +1,39 @@
...
+
+use super::{Extension, ExtensionType};
+use crate::codec::{decode_vec, encode_vec, Cursor, VecSize};
+
+#[derive(PartialEq, Clone, Debug)]
+pub struct KeyIDExtension {
+    key_id: Vec<u8>,
+}
+
+impl KeyIDExtension {
```
kkohbrok, line 24: I would also expect a `new_from_keypackage` here. TODO?

src/extensions/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+    //     let extension_type = ExtensionType::decode(cursor)?;
+    //     let extension_data = decode_vec(VecSize::VecU16, cursor)?;
+    //     Ok(Extension {
+    //         extension_type,
+    //         extension_data,
+    //     })
+    // }
+}
+
+#[test]
```
kkohbrok, line 124: Should tests always have a description of what they're testing? This one seems simple enough, though.

kkohbrok (approved):

src/extensions/key_package_id_extension.rs
franziskuskiefer: I don't think that's needed. But all this is part of the second PR (or probably the third, given how much is missing here ...)

src/extensions/mod.rs
franziskuskiefer: They'll be removed here in one of the follow up PRs.

src/config.rs
```diff
@@ -0,0 +1,55 @@
...
+            1 => Ok(ProtocolVersion::Mls10),
+            _ => Err(ConfigError::UnsupportedMlsVersion),
+        }
+    }
+
+    /// Returns a list of all supported protocol versions.
+    pub fn supported() -> Vec<Self> {
+        vec![Self::Mls10]
+    }
+}
```
raphaelrobert, line 55: Not a blocker, but having an `is_supported()` function might be neat.

raphaelrobert (approved):

src/config.rs
franziskuskiefer: yeah, that's a separate issue and should be tackled as part of #85

Issue #72, opened 2020-10-07
Use the same tree for Handshake and Application secrets (#401)


kkohbrok:
https://github.com/mlswg/mls-protocol/pull/401

Issue #57, opened 2020-10-06
Use the KDF provided by HPKE (#364)


kkohbrok:
https://github.com/mlswg/mls-protocol/pull/364

franziskuskiefer:
Might be partially covered in #66

raphaelrobert:
This has been addressed by the ciphersuite and secret refactoring.

Issue #46, opened 2020-10-06
Rely on HPKE where possible (not individual primitives) (#338)


kkohbrok:
This might be already the case. If so, feel free to close.

https://github.com/mlswg/mls-protocol/pull/338

franziskuskiefer:
Code review

raphaelrobert:
Rename `ciphersuite.hash_length()`

raphaelrobert:
Closing this, because the hash length we use is the digest size of the ciphersuite in use.

PR #89, opened 2020-10-11
Use the same tree for Handshake and Application secrets

raphaelrobert:


src/tree/secret_tree.rs
```diff
@@ -182,38 +213,63 @@ impl ASTree {
...
+        };
         self.nodes[index_in_tree.as_usize()] = None;
-        self.sender_ratchets[index.as_usize()] = Some(sender_ratchet);
-        application_secret
+        self.handshake_sender_ratchets[index.as_usize()] = Some(handshake_sender_ratchet);
+        self.application_sender_ratchets[index.as_usize()] = Some(application_sender_ratchet);
+        ratchet_secrets
     }
 
     fn hash_down(&mut self, ciphersuite: &Ciphersuite, index_in_tree: NodeIndex) {
```
kkohbrok, line 248: More like `derive_down`?

src/tree/secret_tree.rs
```diff
@@ -182,38 +213,63 @@ impl ASTree {
...
         let left_index = left(index_in_tree);
         let right_index = right(index_in_tree, self.size);
-        let left_secret = derive_app_secret(
+        let left_secret = derive_tree_secret(
             &ciphersuite,
             &node_secret,
             "tree",
             left_index.as_u32(),
             0,
             hash_len,
```
kkohbrok, line 259: This should be the `KDF.Nh`, but that's a bug in the spec as well. (Same a few lines below.)

src/tree/sender_ratchet.rs
```diff
@@ -79,16 +79,16 @@ impl SenderRatchet {
...
             }
             let secret = self.past_secrets.last().unwrap();
-            let application_secrets = self.derive_key_nonce(&secret, generation, ciphersuite);
+            let ratchet_secrets = self.derive_key_nonce(&secret, generation, ciphersuite);
             self.generation = generation;
-            Ok(application_secrets)
+            Ok(ratchet_secrets)
         }
     }
     fn ratchet_secret(&self, secret: &[u8], ciphersuite: &Ciphersuite) -> Vec<u8> {
```
kkohbrok, line 87: If the result is a `secret`, it should probably not be `Vec<u8>`, but some struct encapsulating the scret, right? Not sure where this is mean to be used, as it isn't `pub` and it I can't see it being used anywhere here.

kkohbrok (commented):
Left a few comments. Also: Do we explicitly delete keys anywhere? I feel like we should and I couldn't find anything here.

src/tree/secret_tree.rs
raphaelrobert: Will be fixed with #57 and #46.

src/tree/sender_ratchet.rs
raphaelrobert: Will be addressed in #94

src/framing/mod.rs
```diff
@@ -344,8 +295,16 @@ impl MLSCiphertext {
             )
             .unwrap();
         let sender_data = MLSSenderData::from_bytes(&sender_data_bytes).unwrap();
-        let application_secrets = astree
-            .get_secret(ciphersuite, sender_data.sender, sender_data.generation)
+        // TODO Handle error
```
franziskuskiefer, line 298: TODO

src/framing/mod.rs
```diff
@@ -344,8 +295,16 @@ impl MLSCiphertext {
...
+        let secret_type = SecretType::try_from(&self.content_type).unwrap();
+        // TODO Handle error
```
franziskuskiefer, line 300: TODO

src/tree/test_secret_tree.rs
```diff
@@ -0,0 +1,93 @@
+#[test]
+fn test_boundaries() {
```
franziskuskiefer, line 2: These tests need descriptions what they're supposed to test.

franziskuskiefer (changes requested):
I don't know what this PR is doing (no description) but a little drive by review

kkohbrok:
Would it make sense to also add some documentation stating what it does on a high level and how it relates to the standard? I think that would make it easier to grasp the code structure and find bugs, as there is a verbatim description of what is supposed to happen (on a high level) in the comment, which can then be compared to what is happening (on a low level) in the code.

raphaelrobert:
> I don't know what this PR is doing (no description) but a little drive by review

It addresses #72, which in turn is a spec change in the MLS protocol.

raphaelrobert:
> Open issues
> 
>     * There's no description in this PR what it does

We should align how much we need here. We have a flurry of issues that address changes in the MLS protocol spec and they are tagged as such. The changes in the MLS protocol spec are in turn compartmentalized in PRs, making them easy to digest. We can repeat some of the information in the PR itself, if you think that helps.

>     * No description in #72 either

See above.

>     * No link to #72

GH does that automatically when the PR's name says "Closes #XX".

>     * #72 is unassigned

Fixed.

>     * This PR is too big

The limit was reached during review unfortunately. I'll keep a better eye on the size going forward.

franziskuskiefer (approved):
ok let's get this in and start doing this properly after this one 🙂

PR #90, opened 2020-10-11
Fix compiler warnings

raphaelrobert:


src/framing/mod.rs
```diff
@@ -63,7 +63,7 @@ impl MLSPlaintext {
         mls_plaintext
     }
     // XXX: Only used in tests right now.
-    pub(crate) fn from_bytes(bytes: &[u8]) -> Result<Self, CodecError> {
```
franziskuskiefer, line 66: If it's really used in tests, add a `#[cfg(tests)]` to make the warning go away.

tests/test_group.rs
```diff
@@ -151,27 +139,18 @@ fn basic_group_setup() {
...
+    let group_alice_1234 = MlsGroup::new(&group_id, ciphersuite, alice_key_package_bundle);
 
     // Alice adds Bob
     let bob_add_proposal = group_alice_1234.create_add_proposal(
         group_aad,
         &alice_identity.get_signature_key_pair().get_private_key(),
         bob_key_package.clone(),
     );
-    let commit = match group_alice_1234.create_commit(
+    let _commit = match group_alice_1234.create_commit(
```
franziskuskiefer, line 153: Yeah this test needs to get finished to get rid of the compiler warnings. But this way we get less yellow output until we're there 😄

franziskuskiefer (approved):
lgtm with the one issue addressed.

PR #97, opened 2020-10-12
Extensions 2/?

### LINKED ISSUE #32: [Meta] Implement extensions

### LINKED ISSUE #59: Make extensions spec compliant (#379)

franziskuskiefer:
This is the second part of the extension rewrite and is based on #86.
There will be another one ... This one makes extensions usable and starts tests. But extensions aren't used yet. So that has to happen first in another PR.

There are multiple ways to implement extensions. This PR takes the trait approach. Generics would be another good option.

Open To Dos
- [ ] Tests

Fixes_not_yet #32
Fixes_not_yet #59

### DIFF
--- a/src/extensions/life_time_extension.rs
+++ b/src/extensions/life_time_extension.rs
@@ -66,7 +66,7 @@ impl Extension for LifetimeExtension {
     }
 
     /// Build a new LifetimeExtension from a byte slice.
-    fn new_from_bytes(bytes: &[u8]) -> Result<Box<dyn Extension>, ConfigError>
+    fn new_from_bytes(bytes: &[u8]) -> Result<Self, ConfigError>
     where
         Self: Sized,
     {
@@ -80,7 +80,7 @@ impl Extension for LifetimeExtension {
         if out.is_expired() {
             return Err(ConfigError::ExpiredLifetimeExtension);
         }
-        Ok(Box::new(out))
+        Ok(out)
     }
 
     fn to_extension_struct(&self) -> ExtensionStruct {
--- a/src/extensions/life_time_extension.rs
+++ b/src/extensions/life_time_extension.rs
@@ -19,8 +19,9 @@
 //! uint64 not_after;
 //! ```
 //!
-use super::{Extension, ExtensionType};
+use super::{Extension, ExtensionStruct, ExtensionType};
 use crate::codec::{Codec, Cursor};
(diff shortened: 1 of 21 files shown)


src/extensions/life_time_extension.rs
```diff
@@ -49,30 +50,48 @@ impl LifetimeExtension {
...
         }
     }
-    pub fn new_from_bytes(bytes: &[u8]) -> Self {
+
+    fn is_expired(&self) -> bool {
+        let now = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
+            .as_secs();
+        self.not_before < now && self.not_after > now
```
raphaelrobert, line 59: I think this actually `is_valid()`...

franziskuskiefer: Possible... I'll double-check the logic here. This will also get a test

src/tree/mod.rs
```diff
@@ -487,7 +485,7 @@ impl RatchetTree {
         for (path_secret, copath_node) in path_secrets.iter().zip(copath.iter()) {
             let node_ciphertexts: Vec<HpkeCiphertext> = self
                 .resolve(*copath_node)
-                .par_iter()
+                .iter()
```
raphaelrobert, line 488: Why is the parallelization removed here?

src/tree/mod.rs
```diff
@@ -676,7 +674,7 @@ impl RatchetTree {
         let added_members = if !proposal_id_list.adds.is_empty() {
             let add_proposals: Vec<AddProposal> = proposal_id_list
                 .adds
-                .par_iter()
+                .iter()
```
raphaelrobert, line 677: Parallelization also removed here.

src/tree/mod.rs
franziskuskiefer: Because extensions are a trait now that doesn't implement [Sync](https://doc.rust-lang.org/std/marker/trait.Sync.html) yet. We can add that later. But I don't think this is necessary right now. I filed #108

raphaelrobert (approved):

Issue #83, opened 2020-10-08
Error handling


franziskuskiefer:
Improve error handling.

- [ ] More detailed lifetime extension validation results
- [ ] More detailed capabilities extension validation
- [ ] Make framing `verify` functions return errors

PR #102, opened 2020-10-13
fix tree math

kkohbrok:


src/tree/mod.rs
```diff
@@ -197,8 +198,14 @@ impl RatchetTree {
             return unmerged_leaves;
         }
 
-        let mut left = self.resolve(treemath::left(index));
-        let right = self.resolve(treemath::right(index, size));
+        let mut left = self.resolve(treemath::left(index).expect(
+            "resolve:
+            TreeMath error when computing left child.",
```
raphaelrobert, line 203: I'm not sure there should be line breaks in strings

kkohbrok: Strange. `rustfmt` didn't complain. Maybe it was respecting the newlines in my strings? Anyway. Removed the explicit line breaks and let `rustfmt` do its job.

raphaelrobert (approved):

src/group/mls_group/create_commit.rs
```diff
@@ -168,7 +168,8 @@ impl MlsGroup {
                     let dirpath = treemath::direct_path_root(
                         provisional_tree.get_own_node_index(),
                         provisional_tree.leaf_count(),
-                    );
+                    )
+                    .expect("create_commit_internal: TreeMath error when computing direct path.");
```
franziskuskiefer, line 172: You added `expect` everywhere here. This implies that these errors must not happen because they result in panics. I'm not sure that this is really what we want. AFAIK treemath might return an error when there are inconsistencies in the tree. Couldn't this be recoverable or a test on an incoming tree?

src/tree/test_treemath.rs
```diff
@@ -19,29 +24,33 @@ fn verify_binary_test_vector_treemath() {
     let parent: Vec<u32> = decode_vec(VecSize::VecU32, cursor).unwrap();
     let sibling: Vec<u32> = decode_vec(VecSize::VecU32, cursor).unwrap();
 
+    fn test_result(index: usize, input: &u32, result: Result<NodeIndex, TreeMathError>) {
```
franziskuskiefer, line 27: `u32` doesn't need to be borrowed here.

src/tree/test_treemath.rs
```diff
@@ -19,29 +24,33 @@ fn verify_binary_test_vector_treemath() {
...
+            assert_eq!(NodeIndex::from(*input), result.unwrap());
+        } else {
+            assert!(result.is_err());
+        }
+    }
+
     for (i, r) in root.iter().enumerate() {
         assert_eq!(NodeIndex::from(*r), treemath::root(LeafIndex::from(i + 1)));
     }
     for (i, l) in left.iter().enumerate() {
```
franziskuskiefer, line 39: Make this `for (i, &l)` to get rid of the need to borrow or deref.

src/tree/test_treemath.rs
```diff
@@ -19,29 +24,33 @@ fn verify_binary_test_vector_treemath() {
...
-        assert_eq!(NodeIndex::from(*l), treemath::left(NodeIndex::from(i)));
+        let result = treemath::left(NodeIndex::from(i));
```
franziskuskiefer, line 40: Can you add a comment to each of these test cases to make clear what they are supposed to test?

src/tree/treemath.rs
```diff
@@ -17,6 +17,12 @@
 use crate::tree::index::*;
 use std::cmp::Ordering;
 
+#[derive(Debug)]
+pub enum TreeMathError {
```
franziskuskiefer, line 21: This shouldn't be public.

src/tree/treemath.rs
```diff
@@ -79,80 +89,104 @@ pub(crate) fn parent_step(x: usize) -> usize {
...
-        d.push(p);
-        p = parent(p, size);
+    if index == r {
+        return Ok(vec![]);
     }
-    d
+
+    let mut d = vec![];
+    let mut x = index;
+    x = parent(x, size)?;
```
franziskuskiefer, line 124: Why not `let mut x = parent(index, size)?`?

franziskuskiefer: Also, the `mut` doesn't appear to be necessary here.

src/tree/treemath.rs
```diff
@@ -79,80 +89,104 @@ pub(crate) fn parent_step(x: usize) -> usize {
...
+    while x != r {
+        d.push(x);
+        x = parent(x, size)?;
+    }
+    return Ok(d);
```
franziskuskiefer, line 129:
```suggestion
    Ok(d)
```

src/tree/treemath.rs
```diff
@@ -79,80 +89,104 @@ pub(crate) fn parent_step(x: usize) -> usize {
...
+
+    let mut x = index;
+    let mut d = vec![index];
+    while x != r {
+        x = parent(x, size)?;
+        d.push(x);
     }
-    d.push(r);
-    d
+    return Ok(d);
```
franziskuskiefer, line 149:
```suggestion
    Ok(d)
```

src/tree/treemath.rs
```diff
@@ -79,80 +89,104 @@ pub(crate) fn parent_step(x: usize) -> usize {
...
     let r = root(size);
     if index == r {
-        return vec![p];
+        return Ok(vec![]);
     }
-    while p != r {
-        d.push(p);
-        p = parent(p, size);
+
+    let mut x = index;
```
franziskuskiefer, line 143: `x` again isn't muted I think. (same in the other functions)

src/tree/treemath.rs
```diff
@@ -79,80 +89,104 @@ pub(crate) fn parent_step(x: usize) -> usize {
...
+
+    let mut d = vec![];
+    let mut x = index;
+    while x != r {
+        x = parent(x, size)?;
+        d.push(x);
     }
-    d.push(r);
-    d
+    return Ok(d);
```
franziskuskiefer, line 169:
```suggestion
    Ok(d)
```

franziskuskiefer (changes requested):
Thanks looks pretty good already. A couple comments inline.

src/group/mls_group/create_commit.rs
kkohbrok: I thought so, too, initially. But Raphael pointed out that TreeMath really only operates on the array-representation of the tree, so if we try to, for example, get the child of a leaf, it must be an implementation error, because that will be the case for the representation of any tree, regardless if the content of the nodes make sense.

src/tree/treemath.rs
kkohbrok: The first one works, but `cargo build` complains if `x` isn't `mut`. It gets assigned a different value in line 127.

src/tree/treemath.rs
kkohbrok: It gets assigned a new value in line 145. Rust complains if I don't assign it as `mut`.

src/tree/treemath.rs
```diff
@@ -41,7 +47,11 @@ pub(crate) fn level(index: NodeIndex) -> usize {
 }
 
 pub(crate) fn node_width(n: usize) -> usize {
-    2 * (n - 1) + 1
+    if n == 0 {
+        return 0;
```
franziskuskiefer, line 51: What the bot says ...

src/tree/treemath.rs
franziskuskiefer: Ugh, of course. Sorry 😬

src/group/mls_group/create_commit.rs
franziskuskiefer: Hm, not so sure if this is the best way. But we'll have to tackle error handling again anyway (#83) let's revisit then.

franziskuskiefer (approved):
lgtm

PR #105, opened 2020-10-13
Change API to enforce always MLSCiphertext

raphaelrobert:
This PR changes the API so that it is not possible to create an application message and extract it as an MLSPlaintext.


kkohbrok (approved):

franziskuskiefer (approved):
lgtm. But some tests would've been good.

raphaelrobert:
> lgtm. But some tests would've been good.

I couldn't think of anything. Happy to write some if you have an idea.
I will merge for now but can open an issue if we have an idea of what to test.

Issue #106, opened 2020-10-13
Support handshake messages from externals and new members


raphaelrobert:
Right now only the sender type 'member' is supported, but we need to allow external proposals and external commits as well.

PR #107, opened 2020-10-13
Make group context optional for MLSPlaintextTBS

raphaelrobert:
This PR makes the group context field optional for `MLSPlaintextTBS`. 

Note that currently only messages from internal members are supported. A corresponding issue has been filed in #106.


kkohbrok (dismissed):

src/framing/mod.rs
```diff
@@ -66,7 +66,8 @@ impl MLSPlaintext {
             content,
             signature: Signature::new_empty(),
         };
-        mls_plaintext.sign(ciphersuite, signature_key, context);
+        let serialized_context = context.encode_detached().unwrap();
```
franziskuskiefer, line 69: Shouldn't the error get propagated here instead of panicking? The `context` is input and might not be well-formed.

franziskuskiefer (changes requested):
Generally looks good but I'm missing tests.

src/framing/mod.rs
raphaelrobert: The codec code throws an error when slices are too long to be encoded with a certain length header (e.g. a slice of 257 elements encoded with a `u8` header).
If we ant to make sure the group context doesn't have this problem, we should check it when constructing it, not when serializing it.
Sanitizing input is something we should improve anyway, so that we don't have to worry about it internally. I filed #115 for that.

franziskuskiefer (changes requested):
Generally looks good but I'm missing tests.

src/framing/test_framing.rs
```diff
@@ -0,0 +1,89 @@
...
+    let enc = orig.encode_detached().unwrap();
+    let copy = MLSPlaintext::from_bytes(&enc).unwrap();
+    assert_eq!(orig, copy);
+}
+
+#[test]
+fn context_presence() {
+    use crate::ciphersuite::*;
+
+    let ciphersuite =
```
franziskuskiefer, line 58: Not for this PR but we should write helpers for this. We need this first couple lines in most tests.

src/framing/test_framing.rs
```diff
@@ -0,0 +1,89 @@
...
+        signature: Signature::new_empty(),
+    };
+    let context = GroupContext {
+        group_id: GroupId::random(),
+        epoch: GroupEpoch(1u64),
+        tree_hash: vec![],
+        confirmed_transcript_hash: vec![],
+    };
+    let serialized_context = context.encode_detached().unwrap();
+    let signature_input = MLSPlaintextTBS::new_from(&orig, Some(serialized_context.clone()));
```
franziskuskiefer, line 82: This only exercises the `Some` case. We should also have a test for the `None` case.

franziskuskiefer (changes requested):
One more 🙂

src/framing/test_framing.rs
raphaelrobert: Good point, I filed #121

franziskuskiefer (approved):
thanks

Issue #71, opened 2020-10-07
Use full-size path secrets (#398)


kkohbrok:
https://github.com/mlswg/mls-protocol/pull/398

raphaelrobert:
After reviewing the private tree and key packages these changes seem to have been implemented.

PR #111, opened 2020-10-14
Use HKDF digest length instead of hash length.

### LINKED ISSUE #46: Rely on HPKE where possible (not individual primitives) (#338)

kkohbrok:
Fixes #46 .

### DIFF
--- a/src/ciphersuite/mod.rs
+++ b/src/ciphersuite/mod.rs
@@ -200,6 +200,11 @@ impl Ciphersuite {
         get_digest_size(self.hash)
     }
 
+    /// Get the output length of the kdf.
+    pub(crate) fn hkdf_length(&self) -> usize {
+        get_tag_size(self.hmac)
+    }
+
     /// HKDF extract.
     pub(crate) fn hkdf_extract(&self, salt: &[u8], ikm: &[u8]) -> Vec<u8> {
         hkdf_extract(self.hmac, salt, ikm)
--- a/src/ciphersuite/mod.rs
+++ b/src/ciphersuite/mod.rs
@@ -195,11 +195,6 @@ impl Ciphersuite {
         hash(self.hash, payload)
     }
 
-    /// Get the length of the used hash algorithm.
-    pub(crate) fn hash_length(&self) -> usize {
-        get_digest_size(self.hash)
-    }
-
     /// Get the output length of the kdf.
     pub(crate) fn hkdf_length(&self) -> usize {
         get_tag_size(self.hmac)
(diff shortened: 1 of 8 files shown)


src/ciphersuite/mod.rs
```diff
@@ -195,9 +195,9 @@ impl Ciphersuite {
         hash(self.hash, payload)
     }
 
-    /// Get the length of the used hash algorithm.
-    pub(crate) fn hash_length(&self) -> usize {
-        get_digest_size(self.hash)
+    /// Get the output length of the kdf.
+    pub(crate) fn hkdf_length(&self) -> usize {
+        get_tag_size(self.hmac)
```
raphaelrobert, line 200: The MLS spec mentions both `KDF.Nsk` and `KDF.Nh`. Are the two the same?

kkohbrok: Where does the MLS spec mention `KDF.Nsk`? I couldn't find it. There is an `Nsk` in the HPKE RFC. Is that what you meant? There is also `AEAD.Nk` and `AEAD.Nn`.

raphaelrobert: It is mentioned [here](https://github.com/mlswg/mls-protocol/pull/338/files), but it seems that it got reverted in a later PR. I think it's in #71.

There also seems to be some overlap with #57.

It looks to me like it would be good to combine #46, #57 and #71.

kkohbrok: We could do that, but I don't think it hurts to just merge this PR and then tackle the other issues based on the changes here.

raphaelrobert: It's just that this PR is addressing the spec changes that introduce `KDF.Nsk`, but that's not what's implemented.

kkohbrok: I see. How should we proceed then? Should I use this PR to also tackle the other issues? Alternatively, I create a meta issue that gathers the changes needed for #46, #57 and #71. Then this PR solves one item on that meta-issue todo list.

raphaelrobert: How you want to do it is up to you. Personally I'd extend this PR to cover the two other issues, unless it gets too big.

kkohbrok: It's up to me, but I'm also trying to learn what "good practice" would look like or what good options are to organize work in this style of workflow. But ok, I'll continue work on this PR then and will split it once it gets too big.

raphaelrobert:
@kkohbrok #114 is closed now. Are you still working on this? If not, I suggest we close it for now (you can always re-open it later when you work on it).

kkohbrok:
Closing for now. I'll create a new PR when #176 is merged.

Issue #114, opened 2020-10-14
[Meta] Ciphersuite update


franziskuskiefer:
The ciphersuite needs an overhaul

- [ ] add to config
- [x] rely more on hpke
- [ ] remove unnecessary wrapper

kkohbrok:
Would it be useful to do this in conjunction with #94 ?

PR #118, opened 2020-10-14
Don't allow cloning private hpke keys

franziskuskiefer:
This is the first step towards getting rid of HPKE key definitions in the ciphersuite and #114 .

This removes the ability to clone HPKE private keys, which requires some changes to how state is moved around.

I didn't add any tests here but the paths are tests with the existing group tests and additional tests should be written for the ciphersuite and relevant paths later.

### DIFF
--- a/src/group/mls_group/apply_commit.rs
+++ b/src/group/mls_group/apply_commit.rs
@@ -30,17 +30,11 @@ impl MlsGroup {
         own_key_packages: Vec<KeyPackageBundle>,
     ) -> Result<(), ApplyCommitError> {
         let ciphersuite = self.get_ciphersuite();
+        let mut pending_kpbs = own_key_packages;
 
         // Verify epoch
         if mls_plaintext.epoch != self.group_context.epoch {
             return Err(ApplyCommitError::EpochMismatch);
-        }
-
-        // Create KeyPackageBundles
-        let mut pending_kpbs = vec![];
-        for kpb in own_key_packages {
-            let (pk, kp) = (kpb.private_key, kpb.key_package);
-            pending_kpbs.push(KeyPackageBundle::from_values(kp, pk));
         }
 
         // Extract Commit from MLSPlaintext
@@ -88,11 +82,14 @@ impl MlsGroup {
             }
             if is_own_commit {
                 // Find the right KeyPackageBundle among the pending bundles
-                let own_kpb = pending_kpbs
+                let own_kpb_index = match pending_kpbs
                     .iter()
-                    .find(|&kpb| kpb.get_key_package() == kp)
-                    .unwrap()
-                    .clone();
+                    .position(|kpb| kpb.get_key_package() == kp)
+                {
+                    Some(i) => i,
+                    None => return Err(ApplyCommitError::MissingOwnKeyPackage),
+                };
+                let own_kpb = pending_kpbs.remove(own_kpb_index);
                 provisional_tree
                     .replace_private_tree(own_kpb, &self.group_context.serialize())
                     .unwrap()
--- a/src/group/mls_group/apply_commit.rs
+++ b/src/group/mls_group/apply_commit.rs
@@ -60,7 +60,7 @@ impl MlsGroup {
         // Create provisional tree and apply proposals
         let mut provisional_tree = self.tree.borrow_mut();
         let (membership_changes, _invited_members, group_removed) =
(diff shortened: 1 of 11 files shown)


src/group/mls_group/apply_commit.rs
```diff
@@ -90,11 +84,14 @@ impl MlsGroup {
...
                     .iter()
-                    .find(|&kpb| kpb.get_key_package() == kp)
-                    .unwrap()
-                    .clone();
+                    .position(|kpb| kpb.get_key_package() == kp)
+                {
+                    Some(i) => i,
+                    None => return Err(ApplyCommitError::MissingOwnKeyPackage),
+                };
+                let own_kpb = pending_kpbs.remove(own_kpb_index);
```
raphaelrobert, line 94: I'm not sure we need to remove keys from the pending key packages, because we should drop all of them anyway on epoch changes. The rule is that proposals are not carried over between epochs and therefore I don't see a reason to keep the private keys around.
It has nothing to do with your changes directly, but I thinks this is the right PR to make that change, can you add that at the end of the function?

src/tree/mod.rs
```diff
@@ -649,12 +647,16 @@ impl RatchetTree {
...
+                let own_kpb_index = match pending_kpbs
                     .iter()
-                    .find(|&kpb| kpb.get_key_package() == &update_proposal.key_package)
-                    .unwrap();
+                    .position(|kpb| kpb.get_key_package() == &update_proposal.key_package)
+                {
+                    Some(i) => i,
+                    None => panic!("Handle this error case"),
+                };
+                let own_kpb = pending_kpbs.remove(own_kpb_index);
```
raphaelrobert, line 657: Here it's also not necessary to remove the keys. The parameter `pending_kpbs` is only non-empty when this function is called from `apply_commit.rs`.

src/tree/path_keys.rs
```diff
@@ -27,17 +27,27 @@ impl PathKeys {
             return Err(TreeError::InvalidArguments);
         }
         let mut private_keys = private_keys;
+        println!("private_keys: {:?}", private_keys);
```
raphaelrobert, line 30: Some debug output?

src/tree/path_keys.rs
```diff
@@ -27,17 +27,27 @@ impl PathKeys {
...
+        println!("path: {:?}", path);
 
         for (i, private_key) in private_keys.drain(..).enumerate() {
             let index = path[i];
             if self.keys.insert(index, private_key).is_some() {
+                assert!(false);
```
raphaelrobert, line 36: Also debug code?

src/tree/private_tree.rs
```diff
@@ -52,16 +52,35 @@ impl PrivateTree {
...
+        &mut self,
+        ciphersuite: &Ciphersuite,
+        hpke_private_key: Option<HPKEPrivateKey>,
+        path: &[NodeIndex],
+    ) -> Result<Vec<HPKEPublicKey>, TreeError> {
+        // Set new private key if present.
+        match hpke_private_key {
+            Some(k) => self.hpke_private_key = k,
+            None => (),
+        }
```
raphaelrobert, line 70: Adress the clippy warning

raphaelrobert (changes requested):

src/group/mls_group/apply_commit.rs
franziskuskiefer: We have to remove the pending key here because it is consumed. We didn't do this before because we were able to clone them. Now we have to cleanly handle keys without cloning.
This ensure that we can't re-use them. At the same time does this function consume the vector of key packages such that they can't be used later and there's no need to explicitly remove any.

src/tree/mod.rs
franziskuskiefer: It's again necessary because we move it.

src/tree/path_keys.rs
franziskuskiefer: ugh, yeah. I'll remove this.

src/group/mls_group/apply_commit.rs
raphaelrobert: Agreed, this makes sense. We should still drop all of them at the end of the epoch. It could be that not all keys were consumed, because not all updates are covered by the Commit, those keys should be dropped as well.

franziskuskiefer: I filed #119 to follow up on this.

raphaelrobert (approved):

Issue #101, opened 2020-10-13
test extensions


franziskuskiefer:

franziskuskiefer:
Nothing specific to do here at this point.

PR #123, opened 2020-10-15
key package id extensions

franziskuskiefer:
This implements support for the key package id extension and tests for it.
This works towards #32, #101 .

### DIFF
--- a/src/creds.rs
+++ b/src/creds.rs
@@ -140,13 +140,16 @@ impl Codec for Credential {
         }
         Ok(())
     }
-    // fn decode(cursor: &mut Cursor) -> Result<Self, CodecError> {
-    //     let credential_type = CredentialType::from(u8::decode(cursor)?);
-    //     match credential_type {
-    //         CredentialType::Basic => Ok(Credential::Basic(BasicCredential::decode(cursor)?)),
-    //         _ => Err(CodecError::DecodingError),
-    //     }
-    // }
+    fn decode(cursor: &mut Cursor) -> Result<Self, CodecError> {
+        let credential_type = match CredentialType::try_from(u16::decode(cursor)?) {
+            Ok(c) => c,
+            Err(_) => return Err(CodecError::DecodingError),
+        };
+        match credential_type {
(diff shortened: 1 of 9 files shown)


src/creds.rs
```diff
@@ -173,16 +173,16 @@ impl Codec for BasicCredential {
...
+    fn decode(cursor: &mut Cursor) -> Result<Self, CodecError> {
+        let identity = decode_vec(VecSize::VecU16, cursor)?;
+        let ciphersuite = Ciphersuite::decode(cursor)?;
+        let public_key = SignaturePublicKey::decode(cursor)?;
+        Ok(BasicCredential {
+            identity,
+            ciphersuite,
+            public_key,
+        })
+    }
```
raphaelrobert, line 185: Would be nice to have encoding/decoding tests for both `Credential` and `BasicCredential`

raphaelrobert (changes requested):

src/creds.rs
franziskuskiefer: I agree, I filed #130 to write tests for encoding and decoding in general. This specific case might also get covered before when we look at credentials.

src/key_packages/mod.rs
```diff
@@ -123,6 +137,15 @@ impl KeyPackage {
         None
     }
 
+    /// Get the ID of this key package as byte slice.
+    /// Returns an error if no Key ID extension is present.
+    pub fn get_id<'a>(&'a self) -> Result<&'a [u8], KeyPackageError> {
```
raphaelrobert, line 142: Resolve clippy warning

src/extensions/key_package_id_extension.rs
```diff
@@ -29,6 +29,11 @@ impl KeyIDExtension {
             key_id: id.to_vec(),
         }
     }
+
+    /// Get the value of the key id as byte slice.
+    pub(crate) fn as_slice<'a>(&'a self) -> &'a [u8] {
```
raphaelrobert, line 34: Resolve clippy warning

raphaelrobert (changes requested):

raphaelrobert (approved):

PR #125, opened 2020-10-15
Update BasicCredential

raphaelrobert:
This PR addresses some of the changes of https://github.com/mlswg/mls-protocol/pull/350/files

Since the most important change is reverted in https://github.com/mlswg/mls-protocol/pull/402/files, it is not implemented in this PR.

### DIFF
--- a/src/creds.rs
+++ b/src/creds.rs
@@ -17,6 +17,8 @@
 use crate::ciphersuite::*;
 use crate::codec::*;
 
+use std::convert::TryFrom;
+
 #[derive(Clone)]
 pub struct Identity {
     pub id: Vec<u8>,
@@ -78,7 +80,7 @@ impl Codec for Identity {
 }
 
 #[derive(Copy, Clone)]
-#[repr(u8)]
+#[repr(u16)]
 pub enum CredentialType {
     Reserved = 0,
     Basic = 1,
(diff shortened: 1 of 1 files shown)


src/creds.rs
```diff
@@ -80,8 +80,9 @@ impl Codec for Identity {
 #[derive(Copy, Clone)]
 #[repr(u8)]
 pub enum CredentialType {
-    Basic = 0,
-    X509 = 1,
+    Reserved = 0,
+    Basic = 1,
+    X509 = 2,
     Default = 255,
 }
```
kkohbrok, line 87: I think the `from` implementation a few lines below should be changed accordingly.

kkohbrok (changes requested):
Looks good with one change.

src/creds.rs
```diff
@@ -80,8 +80,9 @@ impl Codec for Identity {
 #[derive(Copy, Clone)]
 #[repr(u8)]
 pub enum CredentialType {
-    Basic = 0,
-    X509 = 1,
+    Reserved = 0,
+    Basic = 1,
+    X509 = 2,
     Default = 255,
```
franziskuskiefer, line 86: Drive-by: Why does this default to `255` and why is a "default" needed?

kkohbrok (approved):

PR #128, opened 2020-10-16
Global config

franziskuskiefer:
This PR adds a global configuration to maelstrom (#85).
Pointing the `OPENMLS_CONFIG` environment variable to a JSON serialized config file (see `tests/config.json` for an example) sets the config used when running the library.

There are currently no tests for this, this needs a little additional infrastructure and isn't super high priority. The default case is tested.


kkohbrok (approved):
Looks good, except for the clippy complaint.

kkohbrok (approved):

PR #132, opened 2020-10-19
welcome messages & ciphersuite cleanup

franziskuskiefer:
There are two parts to this PR

Welcome messages cleanup and tests. This adds tests (`messages/test_welcome`) for welcome messages and makes the struct fields private.

Instead of passing around the entire ciphersuite (this should probably get a better name) we should usually use the ciphersuite name. This will make changing the ciphersuite easier.

### DIFF
--- a/src/messages/test_welcome.rs
+++ b/src/messages/test_welcome.rs
@@ -1,0 +1,122 @@
+use super::{EncryptedGroupSecrets, GroupInfo, Welcome};
+use crate::{
+    ciphersuite::{AeadKey, AeadNonce, Ciphersuite, HPKEKeyPair, Signature},
+    codec::*,
+    config::Config,
+    group::{GroupEpoch, GroupId},
+    tree::index::LeafIndex,
+    utils::*,
+};
+
+macro_rules! test_welcome_msg {
+    ($name:ident, $suite:expr, $version:expr) => {
+        #[test]
+        fn $name() {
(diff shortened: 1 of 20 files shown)


src/messages/test_welcome.rs
```diff
@@ -0,0 +1,122 @@
...
+            let welcome_nonce =
+                AeadNonce::from_slice(&randombytes(ciphersuite.aead_nonce_length()));
+
+            // Generate receiver key pair.
+            let receiver_key_pair = HPKEKeyPair::derive(&[1, 2, 3, 4], &ciphersuite);
+            let hpke_info = b"group info welcome test info";
+            let hpke_aad = b"group info welcome test aad";
+            let hpke_input = b"these should be the group secrets";
+            let secrets = vec![EncryptedGroupSecrets {
+                key_package_hash: vec![0, 0, 0, 0],
```
kkohbrok, line 41: TODO: Provide proper key package hash?

kkohbrok (approved):
Lgtm, although I think there is still a clippy warning.

src/messages/test_welcome.rs
franziskuskiefer: For a welcome unit test we don't need a fully correct input. This should be part of a test for key packages and when we start testing messages on the public API.

kkohbrok (approved):

PR #138, opened 2020-10-21
Remove interim_transcript_hash from GroupInfo

raphaelrobert:
This PR addresses a spec change and removes the `interim_transcript_hash` from `GroupInfo`.

### DIFF
--- a/src/framing/mod.rs
+++ b/src/framing/mod.rs
@@ -800,7 +800,6 @@ impl Codec for MLSPlaintextCommitContent {
 
 pub struct MLSPlaintextCommitAuthData {
     pub confirmation_tag: Vec<u8>,
-    pub signature: Vec<u8>,
 }
 
 impl MLSPlaintextCommitAuthData {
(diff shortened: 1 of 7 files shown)


src/framing/mod.rs
```diff
@@ -800,7 +800,6 @@ impl Codec for MLSPlaintextCommitContent {
 
 pub struct MLSPlaintextCommitAuthData {
     pub confirmation_tag: Vec<u8>,
-    pub signature: Vec<u8>,
```
kkohbrok, line 803: As of yet, the signature is still part of the auth data. See https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#name-group-state.

raphaelrobert: I think you should refresh the page...

kkohbrok: Argh. Now I'm wondering how long I've been staring at an outdated version.

In any case, good to merge, then.

kkohbrok (approved):

src/framing/mod.rs
kkohbrok: Part of my confusion, I think, was because that particular change is not part of the PR pointed to in #428 (the mls-protocol issue).

raphaelrobert: Yes, this is the very latest spec change. I implemented it to validate the spec change early on, obviously there are other pending changes we need to address.

PR #152, opened 2020-10-22
Change Api function names for compliance with the Naming Guidelines

### LINKED ISSUE #122: Unify API function names _ref _ref_mut
Public API function names should be a little clearer and unified. 
* `get_key_package -> KeyPackage`
* `get_key_package_ref -> &KeyPackage`
* `get_key_package_ref_mut -> &mut KeyPackage`

kkohbrok:
If #122 is meant only for the `Api` trait, then it's a no-op. In this PR, I have changed some of the `MlsGroup` implementation function names, as `MlsGroup` also implements the `Api`.

Fixes #122 (I think).

Naming Guidlines: https://rust-lang.github.io/api-guidelines/naming.html

### DIFF
--- a/src/framing/mod.rs
+++ b/src/framing/mod.rs
@@ -189,9 +189,9 @@ impl MLSCiphertext {
     ) -> MLSCiphertext {
         const PADDING_SIZE: usize = 10;
 
-        let ciphersuite = mls_group.get_ciphersuite();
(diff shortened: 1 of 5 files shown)


franziskuskiefer (approved):
Thanks! Looks already way nicer.

Issue #154, opened 2020-10-23
Implement fuzzing tests for the API


kkohbrok:

franziskuskiefer:
The framework and CI is there. Now we only need more targets and make sure they don't fail because of panics.

PR #155, opened 2020-10-23
init fuzzing infrastructure

franziskuskiefer:
This PR adds the basic infrastructure for fuzzing in #154.
I'm skipping the GH action for spot fuzzing for now, the setup is a little weird. I'll add that later.

The key package fuzzer will quickly fail. This will need more work as sketched in the changes to ciphersuite already.

### DIFF
--- a/src/ciphersuite/ciphersuites.rs
+++ b/src/ciphersuite/ciphersuites.rs
@@ -14,7 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see http://www.gnu.org/licenses/.
 
-use crate::ciphersuite::*;
+use crate::{ciphersuite::*, errors::Error};
+pub(crate) use std::convert::TryFrom;
 
 impl From<&CiphersuiteName> for u16 {
     fn from(s: &CiphersuiteName) -> u16 {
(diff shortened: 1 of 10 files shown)


src/ciphersuite/ciphersuites.rs
```diff
@@ -14,24 +14,27 @@
...
-            0x0005 => CiphersuiteName::MLS10_256_DHKEMP521_AES256GCM_SHA512_P521,
-            0x0006 => CiphersuiteName::MLS10_256_DHKEMX448_CHACHA20POLY1305_SHA512_Ed448,
-            _ => panic!("Not implemented."),
+            0x0001 => Ok(CiphersuiteName::MLS10_128_DHKEMX25519_AES128GCM_SHA256_Ed25519),
+            0x0002 => Ok(CiphersuiteName::MLS10_128_DHKEMP256_AES128GCM_SHA256_P256),
+            0x0003 => Ok(CiphersuiteName::MLS10_128_DHKEMX25519_CHACHA20POLY1305_SHA256_Ed25519),
+            0x0004 => Ok(CiphersuiteName::MLS10_256_DHKEMX448_AES256GCM_SHA512_Ed448),
+            0x0005 => Ok(CiphersuiteName::MLS10_256_DHKEMP521_AES256GCM_SHA512_P521),
+            0x0006 => Ok(CiphersuiteName::MLS10_256_DHKEMX448_CHACHA20POLY1305_SHA512_Ed448),
+            _ => Err(Error::DecodingError),
```
kkohbrok, line 37: Do we explicitly want errors when fuzzing as opposed to panics? Or What is the motivation for this change?

kkohbrok (approved):

src/ciphersuite/ciphersuites.rs
franziskuskiefer: We generally don't want to panic. Fuzzing is running into all the error cases. Panicking will be an error.

Issue #134, opened 2020-10-19
implement X509 capabilities for credentials


kkohbrok:

franziskuskiefer:
Also see https://github.com/raphaelrobert/maelstrom/pull/156#discussion_r511791681

PR #156, opened 2020-10-23
Remove Identity in favor of CredentialBundle

kkohbrok:
This PR cleans up the API a little by getting rid of Identity and introducing `CredentialBundle`, which, analogous to `KeyPackageBundle`, contains a `Credential` and the corresponding `PrivateSignatureKey`.

### DIFF
--- a/benches/benchmark.rs
+++ b/benches/benchmark.rs
@@ -26,22 +26,17 @@ use maelstrom::key_packages::*;
 
 fn criterion_kp_bundle(c: &mut Criterion) {
     c.bench_function("KeyPackage create bundle", |b| {
-        let ciphersuite = CiphersuiteName::MLS10_128_DHKEMX25519_AES128GCM_SHA256_Ed25519;
-        let signature_keypair = Ciphersuite::new(ciphersuite).new_signature_keypair();
+        let ciphersuite_name = CiphersuiteName::MLS10_128_DHKEMX25519_AES128GCM_SHA256_Ed25519;
+
         b.iter_with_setup(
             || {
-                let identity = Identity::new(ciphersuite, vec![1, 2, 3]);
-                Credential::from(MLSCredentialType::Basic(BasicCredential::from(&identity)))
+                let id = vec![1, 2, 3];
+                CredentialBundle::new(id.clone(), CredentialType::Basic, ciphersuite_name).unwrap()
             },
-            |credential| {
-                KeyPackageBundle::new(
-                    ciphersuite,
-                    signature_keypair.get_private_key(),
-                    credential,
-                    Vec::new(),
-                );
+            |credential_bundle: CredentialBundle| {
+                KeyPackageBundle::new(ciphersuite_name, &credential_bundle, Vec::new());
             },
-        )
+        );
     });
 }
 
--- a/benches/benchmark.rs
+++ b/benches/benchmark.rs
@@ -30,8 +30,8 @@ fn criterion_kp_bundle(c: &mut Criterion) {
 
         b.iter_with_setup(
             || {
-                let id = vec![1, 2, 3];
-                CredentialBundle::new(id.clone(), CredentialType::Basic, ciphersuite_name).unwrap()
+                CredentialBundle::new(vec![1, 2, 3], CredentialType::Basic, ciphersuite_name)
+                    .unwrap()
             },
             |credential_bundle: CredentialBundle| {
                 KeyPackageBundle::new(ciphersuite_name, &credential_bundle, Vec::new());
--- a/src/ciphersuite/mod.rs
+++ b/src/ciphersuite/mod.rs
@@ -376,7 +376,7 @@ impl SignatureKeypair {
     }
 
     /// Get the private and public key objects
-    pub fn as_tuple(self) -> (SignaturePrivateKey, SignaturePublicKey) {
+    pub fn into_tuple(self) -> (SignaturePrivateKey, SignaturePublicKey) {
         (self.private_key, self.public_key)
     }
 }
--- a/src/ciphersuite/mod.rs
+++ b/src/ciphersuite/mod.rs
@@ -374,6 +374,11 @@ impl SignatureKeypair {
     pub fn get_public_key(&self) -> &SignaturePublicKey {
         &self.public_key
     }
+
+    /// Get the private and public key objects
+    pub fn as_tuple(self) -> (SignaturePrivateKey, SignaturePublicKey) {
+        (self.private_key, self.public_key)
+    }
 }
 
 #[test]
--- a/src/ciphersuite/signable.rs
+++ b/src/ciphersuite/signable.rs
@@ -9,16 +9,12 @@ pub trait Signable: Sized {
     /// Sign the payload with the given `id`.
     ///
     /// Returns a `Signature`.
-    fn sign(
-        &mut self,
-        ciphersuite: &Ciphersuite,
-        signature_key: &SignaturePrivateKey,
-    ) -> Signature {
+    fn sign(&self, credential_bundle: &CredentialBundle) -> Signature {
         let payload = self.unsigned_payload().unwrap();
-        ciphersuite.sign(signature_key, &payload).unwrap()
+        credential_bundle.sign(&payload).unwrap()
     }
 
-    /// Verifies the payload against the given `id` and `signature`.
+    /// Verifies the payload against the given `credential` and `signature`.
     ///
     /// Returns a `true` if the signature is valid and `false` otherwise.
     fn verify(&self, credential: &Credential, signature: &Signature) -> bool {
--- a/src/ciphersuite/signable.rs
+++ b/src/ciphersuite/signable.rs
@@ -21,8 +21,8 @@ pub trait Signable: Sized {
     /// Verifies the payload against the given `id` and `signature`.
     ///
     /// Returns a `true` if the signature is valid and `false` otherwise.
-    fn verify(&self, id: &Identity, signature: &Signature) -> bool {
+    fn verify(&self, credential: &Credential, signature: &Signature) -> bool {
         let payload = self.unsigned_payload().unwrap();
-        id.verify(&payload, signature)
+        credential.verify(&payload, signature)
     }
 }
--- a/src/creds.rs
+++ b/src/creds.rs
@@ -14,6 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see http://www.gnu.org/licenses/.
 
+use evercrypt::prelude::SignatureError;
+
 use crate::ciphersuite::*;
 use crate::codec::*;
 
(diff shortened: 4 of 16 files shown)


benches/benchmark.rs
```diff
@@ -26,22 +26,17 @@ use maelstrom::key_packages::*;
...
-        let ciphersuite = CiphersuiteName::MLS10_128_DHKEMX25519_AES128GCM_SHA256_Ed25519;
-        let signature_keypair = Ciphersuite::new(ciphersuite).new_signature_keypair();
+        let ciphersuite_name = CiphersuiteName::MLS10_128_DHKEMX25519_AES128GCM_SHA256_Ed25519;
+
         b.iter_with_setup(
             || {
-                let identity = Identity::new(ciphersuite, vec![1, 2, 3]);
-                Credential::from(MLSCredentialType::Basic(BasicCredential::from(&identity)))
+                let id = vec![1, 2, 3];
+                CredentialBundle::new(id.clone(), CredentialType::Basic, ciphersuite_name).unwrap()
```
franziskuskiefer, line 34: Why do you need to clone `id` here?

src/ciphersuite/signable.rs
```diff
@@ -21,8 +21,8 @@ pub trait Signable: Sized {
     /// Verifies the payload against the given `id` and `signature`.
```
franziskuskiefer, line 21: The comment needs updating.

src/creds.rs
```diff
@@ -19,65 +19,9 @@ use crate::codec::*;
 
 use std::convert::TryFrom;
 
-#[derive(Clone)]
```
franziskuskiefer, line 22: 👍 for deleting code.

src/creds.rs
```diff
@@ -259,3 +193,48 @@ fn test_protocol_version() {
     assert_eq!(mls10_e[0], 1);
     assert_eq!(default_e[0], 1);
 }
+
+/// This struct contains a credential and the corresponding private key.
+pub struct CredentialBundle {
+    credential: Credential,
+    signature_private_key: SignaturePrivateKey,
```
franziskuskiefer, line 200: Maybe `signature_key`? Is a little shorter.

src/creds.rs
```diff
@@ -259,3 +193,48 @@ fn test_protocol_version() {
...
+}
+
+impl CredentialBundle {
+    /// Create a new `CredentialBundle` of the given credential type for the
+    /// given identity and ciphersuite. The corresponding `SignatureKeyPair` is
+    /// freshly generated.
```
franziskuskiefer, line 206: Generating the signature key pair only makes sense for basic credentials I think. X509 will come with it's own private key. How is this handled?

src/creds.rs
```diff
@@ -259,3 +193,48 @@ fn test_protocol_version() {
...
+    pub fn new(
+        identity: Vec<u8>,
+        credential_type: CredentialType,
+        ciphersuite_name: CiphersuiteName,
+    ) -> Result<Self, CredentialError> {
+        let keypair = Ciphersuite::new(ciphersuite_name).new_signature_keypair();
+        let ciphersuite = Ciphersuite::new(ciphersuite_name);
```
franziskuskiefer, line 213: First create `ciphersuite`, then you can use it to generate the key pair.

src/creds.rs
```diff
@@ -259,3 +193,48 @@ fn test_protocol_version() {
...
+        let mls_credential = match credential_type {
+            CredentialType::Basic => Ok(BasicCredential {
```
franziskuskiefer, line 215: `Ok` not needed.

src/creds.rs
```diff
@@ -259,3 +193,48 @@ fn test_protocol_version() {
...
+                identity,
+                ciphersuite,
+                public_key: keypair.get_public_key().clone(),
+            }),
+            _ => Err(CredentialError::UnsupportedCredentialType),
```
franziskuskiefer, line 220: Just return the error here.

src/creds.rs
```diff
@@ -259,3 +193,48 @@ fn test_protocol_version() {
...
+        }?;
+        let credential = Credential {
+            credential_type,
+            credential: MLSCredentialType::Basic(mls_credential),
+        };
+        Ok(CredentialBundle {
+            credential,
+            signature_private_key: keypair.get_private_key().clone(),
```
franziskuskiefer, line 228: I was hoping we could get rid of cloning the private key. Why do we still need to do that?

src/creds.rs
```diff
@@ -259,3 +193,48 @@ fn test_protocol_version() {
...
+        })
+    }
+
+    /// Get a reference to the signature private key.
+    pub fn signature_private_key(&self) -> &SignaturePrivateKey {
```
franziskuskiefer, line 233: Shouldn't this function go away and the `CredentialBundle` be passed around instead? I really don't like having to pull out a private key to do something.

src/key_packages/mod.rs
```diff
@@ -72,11 +71,11 @@ impl KeyPackage {
             protocol_version: ProtocolVersion::default(),
             cipher_suite: ciphersuite_name,
             hpke_init_key,
-            credential,
+            credential: credential_bundle.credential().clone(),
```
franziskuskiefer, line 74: Let's not pass in a reference. Then we don't have to clone. The caller is then responsible to clone if they need the credential again.

src/key_packages/mod.rs
```diff
@@ -72,11 +71,11 @@ impl KeyPackage {
...
             extensions,
             signature: Signature::new_empty(),
         };
-        key_package.sign(&ciphersuite, signature_key);
+        key_package.sign(&ciphersuite, credential_bundle.signature_private_key());
```
franziskuskiefer, line 78: So we pass in a `CredentialBundle` but then only consume the credential? This doesn't look like what we should be doing.

franziskuskiefer (changes requested):
Thanks for starting this. I like the general approach but I think there needs to be some work around ownership of the credential and the signature key. It's not entirely clear who's owning the bundle or the two components. In particular, the need to clone and getting out the private key should go away.

src/creds.rs
kkohbrok: TODO: Create additional constructor that takes X509 reference.

kkohbrok: I could create a stub for the private counterpart of `Certificate` (which only exists, because it also exists in the spec, but I think we can leave this for issue #134.

src/creds.rs
kkohbrok: For the record: We discussed this in a call and decided to leave the name as-is to avoid confusion with the public part of the signature key. For example in `BasicCredential`, there is a field `signature_key`, which denotes the public part of the signature key.

src/creds.rs
kkohbrok: I don't think this works, as it's a `let ... = match { ... }` construction and I can't just return. As far as I understand, this also means that I do need the `Ok(...)` noted in the comment above.

franziskuskiefer: You can return from in here.

src/key_packages/mod.rs
kkohbrok: This is tricky, because the `credential_bundle` should not be part of the `key_package`, but the `credential` should be. In `key_packages` owned by other parties, we need to be able to update these `key_package`s, as they might change over time and also might not be the same in every group for that party. In `key_package`s belonging to ourselves, I'm not sure if we want to attach the `credential_bundle`, because we might have one for multiple groups, but then again, we might not.

I don't know what's the best way to do ownership here, but I think we should clear this up in a separate PR.

src/ciphersuite/mod.rs
```diff
@@ -374,6 +374,11 @@ impl SignatureKeypair {
     pub fn get_public_key(&self) -> &SignaturePublicKey {
         &self.public_key
     }
+
+    /// Get the private and public key objects
+    pub fn as_tuple(self) -> (SignaturePrivateKey, SignaturePublicKey) {
```
franziskuskiefer, line 379: If we want to stick to Rust naming guidelines this should be `into_`.

src/creds.rs
```diff
@@ -209,32 +219,34 @@ impl CredentialBundle {
...
-    pub fn signature_private_key(&self) -> &SignaturePrivateKey {
-        &self.signature_private_key
-    }
-
     pub fn credential(&self) -> &Credential {
         &self.credential
     }
+
+    /// Sign a `msg` using the private key of the credential bundle.
+    pub(crate) fn sign(&self, msg: &[u8]) -> Result<Signature, SignatureError> {
```
franziskuskiefer, line 247: Hm, not sure if propagating the `SignatureError` up here is the right thing to do. It has to be handled somewhere. But I guess we can leave this to #83 .

src/key_packages/mod.rs
```diff
@@ -195,10 +194,10 @@ impl KeyPackage {
         &self.extensions
     }
 
-    /// Sign the key package.
-    pub(crate) fn sign(&mut self, ciphersuite: &Ciphersuite, signature_key: &SignaturePrivateKey) {
+    /// Populate the `signature` field using the `credential_bundle`.
+    pub(crate) fn sign_self(&mut self, credential_bundle: &CredentialBundle) {
```
franziskuskiefer, line 198: Why `sign_self`? The `&mut self` should give that away 🙂

src/creds.rs
```diff
@@ -157,6 +103,14 @@ impl Credential {
             MLSCredentialType::X509(_) => panic!("X509 certificates are not yet implemented."),
         }
     }
+
+    /// Get the ciphersuite associated with the credential.
+    pub fn ciphersuite(&self) -> &Ciphersuite {
+        match &self.credential {
+            MLSCredentialType::Basic(basic_credential) => &basic_credential.ciphersuite,
```
franziskuskiefer, line 110: yeah... enums...

franziskuskiefer (changes requested):
Ok let's get this landed and follow up on the ownership separately. Only a couple nits left.

src/ciphersuite/mod.rs
kkohbrok: Right, good point!

src/key_packages/mod.rs
kkohbrok: Because if we call it `sign`, then the function is hidden behind the implementation of `signable`, at least that's how I understand it. Is there a better way of doing this?

src/creds.rs
kkohbrok: Right. Credential should be a `trait`, I noticed that while writing this code, but I think we should leave that to a different PR. Should I file an issue?

kkohbrok: I think you already did a lot of the work when I asked for help w.r.t. `Sized` traits. Do you still have that code somewhere?

src/key_packages/mod.rs
kkohbrok: Filed issue #161.

src/creds.rs
franziskuskiefer: Yeah this should be a different PR & issue. I have some code but have to look at it again.

src/key_packages/mod.rs
franziskuskiefer: Ah, this implements signable? Maybe we can call this something like `re_sign` (not resign 😆 ) if it's only used to re-sign. You can still call it even if signable is implemented, but having clashing names is never nice. So pick the name you like best or keep the `sign_self`.

src/creds.rs
kkohbrok: Filed issue #163 and labelled it P3 for now. Feel free to relabel.

src/key_packages/mod.rs
kkohbrok: I think what happens is that the signature field is initialized as empty and then filled with this signature. So it's not (just) for re-signing. Since `Signable` is used nowhere else, I've removed it for now and renamed it to `sign`. I think `Signable` needs to be reworked anyway (see issue #160).

franziskuskiefer (approved):
Thanks. Let's get this landed

Issue #77, opened 2020-10-08
Re-enable group tests


raphaelrobert:
Currently the group tests are disabled and should be re-enabled when the API changes are complete.

franziskuskiefer:
The tests probably have be re-written.

https://github.com/raphaelrobert/maelstrom/blob/06f5130042906b069f9ad5d06e853fc30490d587/tests/test_group.rs#L187

franziskuskiefer:
Looking at the code all tests in `test_group.rs` should currently fail because they use invalid key packages. But they don't.

PR #157, opened 2020-10-26
Re-enable group tests 1/?

raphaelrobert:
This PR re-enables the previously commented out group tests

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -11,7 +11,7 @@ use crate::errors::ConfigError;
 use crate::extensions::ExtensionType;
 
 lazy_static! {
-    static ref CONFIG: Config = {
+    pub static ref CONFIG: Config = {
         if let Ok(path) = env::var("OPENMLS_CONFIG") {
             let file = match File::open(path) {
                 Ok(f) => f,
@@ -50,6 +50,18 @@ pub struct Config {
     protocol_versions: Vec<ProtocolVersion>,
     ciphersuites: Vec<CiphersuiteName>,
     extensions: Vec<ExtensionType>,
+}
+
+impl Config {
+    pub fn get_protocol_version_ref(&self) -> &[ProtocolVersion] {
+        &self.protocol_versions
+    }
+    pub fn get_ciphersuites_ref(&self) -> &[CiphersuiteName] {
+        &self.ciphersuites
+    }
+    pub fn get_extensions_ref(&self) -> &[ExtensionType] {
+        &self.extensions
+    }
 }
 
 /// # Protocol Version
--- a/src/config.rs
+++ b/src/config.rs
@@ -11,7 +11,7 @@ use crate::errors::ConfigError;
 use crate::extensions::ExtensionType;
 
 lazy_static! {
-    pub static ref CONFIG: Config = {
+     static ref CONFIG: Config = {
         if let Ok(path) = env::var("OPENMLS_CONFIG") {
             let file = match File::open(path) {
                 Ok(f) => f,
--- a/src/config.rs
+++ b/src/config.rs
@@ -53,14 +53,19 @@ pub struct Config {
 }
 
 impl Config {
-    pub fn get_protocol_version_ref(&self) -> &[ProtocolVersion] {
-        &self.protocol_versions
+    /// Get a list of the supported extension types.
+    pub fn supported_extensions() -> &'static [ExtensionType] {
+        &CONFIG.extensions
     }
-    pub fn get_ciphersuites_ref(&self) -> &[CiphersuiteName] {
-        &self.ciphersuites
+
+    /// Get a list of the supported cipher suite names.
+    pub fn supported_ciphersuites() -> &'static [CiphersuiteName] {
+        &CONFIG.ciphersuites
     }
-    pub fn get_extensions_ref(&self) -> &[ExtensionType] {
-        &self.extensions
+
+    /// Get a list of the supported protocol versions.
+    pub fn supported_versions() -> &'static [ProtocolVersion] {
+        &CONFIG.protocol_versions
     }
 }
 
@@ -122,22 +127,5 @@ impl CiphersuiteName {
             }
         }
         false
-    }
-}
-
-impl Config {
-    /// Get a list of the supported extension types.
-    pub fn supported_extensions() -> &'static [ExtensionType] {
-        &CONFIG.extensions
-    }
-
-    /// Get a list of the supported cipher suite names.
-    pub fn supported_ciphersuites() -> &'static [CiphersuiteName] {
-        &CONFIG.ciphersuites
-    }
-
-    /// Get a list of the supported protocol versions.
-    pub fn supported_versions() -> &'static [ProtocolVersion] {
-        &CONFIG.protocol_versions
     }
 }
--- a/src/group/mls_group/apply_commit.rs
+++ b/src/group/mls_group/apply_commit.rs
@@ -76,11 +76,8 @@ impl MlsGroup {
         let commit_secret = if let Some(path) = commit.path.clone() {
             // Verify KeyPackage and MLSPlaintext signature
             let kp = &path.leaf_key_package;
-            match kp.verify() {
-                Ok(()) => {}
-                Err(_) => {
-                    return Err(ApplyCommitError::PathKeyPackageVerificationFailure);
-                }
+            if kp.verify().is_err() {
+                return Err(ApplyCommitError::PathKeyPackageVerificationFailure);
             }
             let serialized_context = self.group_context.encode_detached().unwrap();
             if !mls_plaintext.verify(Some(serialized_context.clone()), kp.get_credential()) {
--- a/src/group/mls_group/apply_commit.rs
+++ b/src/group/mls_group/apply_commit.rs
@@ -38,7 +38,7 @@ impl MlsGroup {
         }
 
         // Extract Commit from MLSPlaintext
-        let (commit, confirmation_tag) = match &mls_plaintext.content {
+        let (commit, received_confirmation_tag) = match &mls_plaintext.content {
             MLSPlaintextContentType::Commit((commit, confirmation_tag)) => {
                 (commit, confirmation_tag)
             }
@@ -76,11 +76,14 @@ impl MlsGroup {
         let commit_secret = if let Some(path) = commit.path.clone() {
             // Verify KeyPackage and MLSPlaintext signature
             let kp = &path.leaf_key_package;
-            if !kp.verify() {
-                return Err(ApplyCommitError::PathKeyPackageVerificationFailure);
+            match kp.verify() {
+                Ok(()) => {}
+                Err(_) => {
+                    return Err(ApplyCommitError::PathKeyPackageVerificationFailure);
+                }
             }
             let serialized_context = self.group_context.encode_detached().unwrap();
-            if !mls_plaintext.verify(Some(serialized_context), kp.get_credential()) {
+            if !mls_plaintext.verify(Some(serialized_context.clone()), kp.get_credential()) {
                 return Err(ApplyCommitError::PlaintextSignatureFailure);
             }
             if is_own_commit {
@@ -94,11 +97,11 @@ impl MlsGroup {
                 };
                 let own_kpb = pending_kpbs.remove(own_kpb_index);
                 provisional_tree
-                    .replace_private_tree(own_kpb, &self.group_context.serialize())
+                    .replace_private_tree(own_kpb, &serialized_context)
                     .unwrap()
             } else {
                 provisional_tree
-                    .update_path(sender, &path, &self.group_context.serialize())
+                    .update_path(sender, &path, &serialized_context)
                     .unwrap()
             }
         } else {
@@ -140,12 +143,12 @@ impl MlsGroup {
         );
 
         // Verify confirmation tag
-        if &ConfirmationTag::new(
+        let own_confirmation_tag = ConfirmationTag::new(
             &ciphersuite,
             &provisional_epoch_secrets.confirmation_key,
             &confirmed_transcript_hash,
-        ) != confirmation_tag
-        {
+        );
+        if &own_confirmation_tag != received_confirmation_tag {
             return Err(ApplyCommitError::ConfirmationTagMismatch);
         }
 
--- a/src/group/mls_group/mod.rs
+++ b/src/group/mls_group/mod.rs
@@ -209,12 +209,18 @@ impl Api for MlsGroup {
     fn encrypt(&mut self, mls_plaintext: MLSPlaintext) -> MLSCiphertext {
         let mut secret_tree = self.secret_tree.borrow_mut();
         let secret_type = SecretType::try_from(&mls_plaintext).unwrap();
-        let (generation, ratchet_secrets) = secret_tree.get_secret_for_encryption(
+        let (generation, (ratchet_key, ratchet_nonce)) = secret_tree.get_secret_for_encryption(
             &self.ciphersuite,
             mls_plaintext.sender.sender,
             secret_type,
         );
-        MLSCiphertext::new_from_plaintext(&mls_plaintext, &self, generation, &ratchet_secrets)
+        MLSCiphertext::new_from_plaintext(
+            &mls_plaintext,
+            &self,
+            generation,
+            ratchet_key,
+            ratchet_nonce,
+        )
     }
 
     fn decrypt(&mut self, mls_ciphertext: MLSCiphertext) -> Result<MLSPlaintext, DecryptionError> {
--- a/src/group/mls_group/mod.rs
+++ b/src/group/mls_group/mod.rs
@@ -224,7 +224,8 @@ impl Api for MlsGroup {
         let tree = self.tree.borrow();
         let mut roster = Vec::new();
         for i in 0..tree.leaf_count().as_usize() {
-            let node = &tree.nodes[NodeIndex::from(i).as_usize()];
+            let leaf_index = LeafIndex::from(i);
+            let node = &tree.nodes[NodeIndex::from(leaf_index).as_usize()];
             let credential = if let Some(kp) = &node.key_package {
                 kp.get_credential()
             } else {
--- a/src/group/mls_group/mod.rs
+++ b/src/group/mls_group/mod.rs
@@ -21,6 +21,7 @@ mod new_from_welcome;
 
 use crate::ciphersuite::*;
 use crate::codec::*;
+use crate::creds::CredentialBundle;
 use crate::framing::*;
 use crate::group::*;
 use crate::key_packages::*;
@@ -94,7 +95,7 @@ impl Api for MlsGroup {
     fn create_add_proposal(
         &self,
         aad: &[u8],
-        signature_key: &SignaturePrivateKey,
+        credential_bundle: &CredentialBundle,
         joiner_key_package: KeyPackage,
     ) -> MLSPlaintext {
         let add_proposal = AddProposal {
@@ -103,11 +104,10 @@ impl Api for MlsGroup {
         let proposal = Proposal::Add(add_proposal);
         let content = MLSPlaintextContentType::Proposal(proposal);
         MLSPlaintext::new(
-            &self.ciphersuite,
             self.sender_index(),
             aad,
             content,
-            signature_key,
+            credential_bundle,
             &self.context(),
         )
     }
@@ -119,18 +119,17 @@ impl Api for MlsGroup {
     fn create_update_proposal(
         &self,
         aad: &[u8],
-        signature_key: &SignaturePrivateKey,
+        credential_bundle: &CredentialBundle,
         key_package: KeyPackage,
     ) -> MLSPlaintext {
         let update_proposal = UpdateProposal { key_package };
         let proposal = Proposal::Update(update_proposal);
         let content = MLSPlaintextContentType::Proposal(proposal);
         MLSPlaintext::new(
-            &self.ciphersuite,
             self.sender_index(),
             aad,
             content,
-            signature_key,
+            credential_bundle,
             &self.context(),
         )
     }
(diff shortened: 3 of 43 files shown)


tests/test_group.rs
```diff
@@ -6,6 +6,7 @@ use maelstrom::key_packages::*;
 
 #[test]
 fn create_commit_optional_path() {
+    use maelstrom::extensions::*;
     let ciphersuite_name = CiphersuiteName::MLS10_128_DHKEMX25519_AES128GCM_SHA256_Ed25519;
```
franziskuskiefer, line 10: Tests should cover all supported ciphersuites.

tests/test_group.rs
```diff
@@ -17,97 +18,149 @@ fn create_commit_optional_path() {
...
+    // Alice adds Bob with forced self-update
     let bob_add_proposal = group_alice_1234.create_add_proposal(
         group_aad,
-        &alice_identity.get_signature_key_pair().get_private_key(),
+        alice_signature_key,
         bob_key_package.clone(),
     );
+    let epoch_proposals = vec![bob_add_proposal];
+    let (mls_plaintext_commit, _welcome_bundle_alice_bob_option) = match group_alice_1234
+        .create_commit(group_aad, alice_signature_key, epoch_proposals, true)
```
franziskuskiefer, line 68: This more about the API. But having boolean arguments, especially in public APIs, is not great. It's never clear what the `true` here means. So for this API this should be something like `true /* force self-update */`. But maybe we should consider adding a separate function with this set to `true`, like `create_commit_self_update`.

tests/test_group.rs
```diff
@@ -17,97 +18,149 @@ fn create_commit_optional_path() {
...
-
-    // Only AddProposals
-    let (commit_mls_plaintext, _welcome_option) = match group_alice_1234.create_commit(
-        group_aad,
-        &alice_identity.get_signature_key_pair().get_private_key(),
-        vec![bob_add_proposal.clone()],
-        false,
-    ) {
+    let epoch_proposals = vec![bob_add_proposal];
+    let (mls_plaintext_commit, welcome_bundle_alice_bob_option) = match group_alice_1234
```
franziskuskiefer, line 86: So we add Bob twice to the same group? Shouldn't this fail?

tests/test_group.rs
```diff
@@ -159,109 +212,156 @@ fn basic_group_setup() {
...
         bob_key_package_bundle,
-    );
+    ) {
+        Ok(group) => group,
+        Err(e) => panic!("Error creating group from Welcome: {:?}", e),
+    };
 
-    assert_eq!(group_alice.tree.nodes, group_bob.tree.nodes);
-    group_alice.tree.print(&format!("\n{:?}", ms1));
+    assert_eq!(
```
franziskuskiefer, line 298: This looks like a pretty crucial test. It should probably get an explanation.
Also, the `_print_tree` should only be interesting if this fails. But it won't be reached in the case the assert fails.

tests/test_group.rs
```diff
@@ -159,109 +212,156 @@ fn basic_group_setup() {
     };
 }
 
-/*
 #[test]
 fn group_operations() {
```
franziskuskiefer, line 216: What operations?
This needs a description of what's exactly happening/tested in here.

franziskuskiefer (changes requested):
So this PR should enable the group tests. It's to expect that changes to the code are needed for this. But there's no explanation to which and why. So I only really looked at the test for now.
The description in the PR also doesn't appear to be correct as it only enables a part of the commented out test.

tests/test_group.rs
raphaelrobert: We only create a proposal/commit that we don't apply. This should not fail.

raphaelrobert:
> So this PR should enable the group tests. It's to expect that changes to the code are needed for this. But there's no explanation to which and why. So I only really looked at the test for now.
> The description in the PR also doesn't appear to be correct as it only enables a part of the commented out test.

This PR indeed only covers a part of the tests. Enabling more of them would make the PR grow beyond the limit we imposed ourselves.
This PR fixes all the obvious problems that were introduced in the last weeks and that cause the inner workings to break.
There will be at least one follow-up PR that will close #77.

src/config.rs
```diff
@@ -52,6 +52,18 @@ pub struct Config {
     extensions: Vec<ExtensionType>,
 }
 
+impl Config {
+    pub fn get_protocol_version_ref(&self) -> &[ProtocolVersion] {
```
franziskuskiefer, line 56: Do we need these? The global config can be accessed already with the functions below.
Also, if we want these functions we should put them in the same `impl` block.

tests/test_group.rs
```diff
@@ -17,97 +18,149 @@ fn create_commit_optional_path() {
     let alice_credential = BasicCredential::from(&alice_identity);
     let bob_credential = BasicCredential::from(&bob_identity);
 
+    // Signature keys
+    let alice_signature_key = &alice_identity.get_signature_key_pair().get_private_key();
+    let bob_signature_key = &bob_identity.get_signature_key_pair().get_private_key();
+
+    // Mandatory extensions
+    let capabilities_extension = Box::new(CapabilitiesExtension::default());
```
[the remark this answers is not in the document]
franziskuskiefer: 🤔

tests/test_group.rs
franziskuskiefer: blocked by #168

tests/test_group.rs
```diff
@@ -17,97 +18,149 @@ fn create_commit_optional_path() {
...
+    let lifetime_extension = Box::new(LifetimeExtension::new(60));
```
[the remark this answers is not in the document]
franziskuskiefer: 🤔

tests/test_group.rs
franziskuskiefer: Then there are comments missing explaining what this does (as mentioned in another comment already).

tests/test_group.rs
franziskuskiefer: 🤔

franziskuskiefer (changes requested):
Maybe I'm missing something but I don't find a description of the changes.

tests/test_group.rs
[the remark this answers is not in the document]
raphaelrobert: This is something for #164 then, we don't want to do that here.

tests/test_group.rs
raphaelrobert: That was added, but I think you are looking at the wrong file

franziskuskiefer (changes requested):
I still can't find a description of the changes.

raphaelrobert:
> I still can't find a description of the changes.

The last commit only addresses your comments, as the name suggests. There are no other changes.

franziskuskiefer:
> > I still can't find a description of the changes.
> 
> The last commit only addresses your comments, as the name suggests. There are no other changes.

There is no description of any changes in this PR. That's what I'm waiting for before reviewing. The description is "This PR re-enables the previously commented out group tests". But there are a ton of code changes in this PR. They all need a high level description here.

raphaelrobert:
> > > I still can't find a description of the changes.
> > 
> > 
> > The last commit only addresses your comments, as the name suggests. There are no other changes.
> 
> There is no description of any changes in this PR. That's what I'm waiting for before reviewing. The description is "This PR re-enables the previously commented out group tests". But there are a ton of code changes in this PR. They all need a high level description here.

The context of your question wasn't clear.
There is a high-level description which you didn't quote, but here is a more detailed one:
 - Re-enable part of the group tests to test basic operations like Add & Update
 - Validate public keys from the UpdatePath
 - Change return type of KeyPackage verification from boolean to more specific errors
 - Fix direct path calculation in treemath
 - Fix missing RatchetTree clone in create_commit
 - Refactor part of PrivateTree
 - Fix various smaller issues that made the tests break

src/tree/private_tree.rs
```diff
@@ -134,13 +134,13 @@ impl PrivateTree {
     ) {
         let hash_len = ciphersuite.hash_length();
         let start_secret = match start_secret {
-            Some(secret) => hkdf_expand_label(ciphersuite, secret, "path", &[], hash_len),
+            Some(secret) => secret.to_vec(),
```
franziskuskiefer, line 137: This appears to be wrong. According to spec we have `path_secret[0] = DeriveSecret(leaf_secret, "path")`. So unless you changed the way this function is called the `start_secret` shouldn't be used directly.

src/tree/treemath.rs
```diff
@@ -136,7 +136,7 @@ pub(crate) fn dirpath_long(
 ) -> Result<Vec<NodeIndex>, TreeMathError> {
     let r = root(size);
     if index == r {
-        return Ok(vec![]);
```
franziskuskiefer, line 139: Please add a test for this.

src/tree/treemath.rs
```diff
@@ -156,7 +156,7 @@ pub(crate) fn direct_path_root(
...
```
franziskuskiefer, line 159: And this one as well.

tests/test_group.rs
franziskuskiefer: #168 got merged so we should add tests for all ciphersuites.

tests/test_group.rs
```diff
@@ -15,94 +16,146 @@ fn create_commit_optional_path() {
...
+            true, /* force self-update */
+        ) {
+        Ok(c) => c,
+        Err(e) => panic!("Error creating commit: {:?}", e),
+    };
+    let commit = match &mls_plaintext_commit.content {
+        MLSPlaintextContentType::Commit((commit, _)) => commit,
+        _ => panic!(),
+    };
+    assert!(commit.path.is_some());
```
franziskuskiefer, line 71: Can we test more here than just the fact that there is a path?

tests/test_group.rs
```diff
@@ -15,94 +16,146 @@ fn create_commit_optional_path() {
...
 
-    // Alice updates
-    let alice_update_proposal = group_alice_1234.create_update_proposal(
+    // Alice adds Bob without forced self-update
+    // Since there are on Add Proposals, this not generate a path field on the Commit
```
franziskuskiefer, line 74: "... only Add Proposals this does not generate ..."

tests/test_group.rs
```diff
@@ -15,94 +16,146 @@ fn create_commit_optional_path() {
...
-    // Only AddProposals with forced self update
-    let (commit_mls_plaintext, _welcome_option) = match group_alice_1234.create_commit(
-        group_aad,
-        &alice_credential_bundle,
-        vec![bob_add_proposal],
-        true,
-    ) {
-        Ok(c) => c,
-        Err(e) => panic!("Error creating commit: {:?}", e),
+    // Alice applies Commit
```
franziskuskiefer, line 97: Which one? Looks like the second without forced self-update.

tests/test_group.rs
```diff
@@ -15,94 +16,146 @@ fn create_commit_optional_path() {
...
+        Err(e) => panic!("Error applying commit: {:?}", e),
     };
+    let ratchet_tree = group_alice_1234.tree().get_public_key_tree();
 
-    let (commit, _confirmation_tag) = match commit_mls_plaintext.content {
-        MLSPlaintextContentType::Commit((commit, confirmation_tag)) => (commit, confirmation_tag),
-        _ => panic!(),
+    // Bob creates group from Welcome
+    let mut group_bob_1234 = match MlsGroup::new_from_welcome(
+        welcome_bundle_alice_bob_option.unwrap(),
```
franziskuskiefer, line 106: Shouldn't this `unwrap` be a test case earlier?

tests/test_group.rs
```diff
@@ -15,94 +16,146 @@ fn create_commit_optional_path() {
...
+    let proposals = vec![alice_update_proposal];
+
+    // Only UpdateProposal
     let (commit_mls_plaintext, _welcome_option) = match group_alice_1234.create_commit(
         group_aad,
         &alice_credential_bundle,
-        vec![alice_update_proposal],
-        true,
+        proposals.clone(),
+        false,
```
franziskuskiefer, line 132: `/* force self update */`

tests/test_group.rs
```diff
@@ -15,94 +16,146 @@ fn create_commit_optional_path() {
...
+            commit_mls_plaintext.clone(),
+            proposals.clone(),
+            vec![alice_update_key_package_bundle],
+        ) {
+            Ok(()) => {}
+            Err(e) => panic!("Error applying commit: {:?}", e),
+        }
+    */
+
+    match group_bob_1234.apply_commit(commit_mls_plaintext, proposals, vec![]) {
```
franziskuskiefer, line 155: Maybe use `.expect()`? There are a lot more like this in this file.

tests/test_group.rs
```diff
@@ -150,109 +203,162 @@ fn basic_group_setup() {
     };
 }
 
-/*
 #[test]
+/// This test simulates various group operations like Add, Update, Remove in a small group
```
franziskuskiefer, line 207: To make this more digestible having a high level description of what's supposed to happen would be nice. Something like
* Alice creates group
* Alice adds Bob
* Bob removes Alice 
...

tests/test_group.rs
```diff
@@ -150,109 +203,162 @@ fn basic_group_setup() {
...
+            &alice_credential_bundle,
+            bob_key_package.clone(),
+        );
+        let epoch_proposals = vec![bob_add_proposal];
+        let (mls_plaintext_commit, welcome_bundle_alice_bob_option) = match group_alice_1234
+            .create_commit(
+                group_aad,
+                &alice_credential_bundle,
+                epoch_proposals.clone(),
+                false, // TODO force update seems broken
```
franziskuskiefer, line 263: Shouldn't the `create_commit_optional_path` test fail then?

tests/test_group.rs
```diff
@@ -150,109 +203,162 @@ fn basic_group_setup() {
...
+            ) {
+            Ok(c) => c,
+            Err(e) => panic!("Error creating commit: {:?}", e),
+        };
+        let commit = match &mls_plaintext_commit.content {
+            MLSPlaintextContentType::Commit((commit, _)) => commit,
+            _ => panic!(),
```
franziskuskiefer, line 270: Adding a little more information to panics would help debugging later.

franziskuskiefer (changes requested):
Thanks for the update and additional description. See inline comments

Please also fix the clippy warning.

Other changes in here (that should co into the commit message in the end):
*  new public function `as_application_message` to get a `MLSPlaintext` as application message (byte slice).

src/group/mls_group/apply_commit.rs
```diff
@@ -76,11 +76,14 @@ impl MlsGroup {
...
-                return Err(ApplyCommitError::PathKeyPackageVerificationFailure);
+            match kp.verify() {
+                Ok(()) => {}
+                Err(_) => {
+                    return Err(ApplyCommitError::PathKeyPackageVerificationFailure);
+                }
             }
             let serialized_context = self.group_context.encode_detached().unwrap();
-            if !mls_plaintext.verify(Some(serialized_context), kp.get_credential()) {
+            if !mls_plaintext.verify(Some(serialized_context.clone()), kp.get_credential()) {
```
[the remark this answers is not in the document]
raphaelrobert: https://github.com/raphaelrobert/maelstrom/issues/171

src/group/mls_group/mod.rs
```diff
@@ -221,7 +221,8 @@ impl Api for MlsGroup {
         let tree = self.tree.borrow();
         let mut roster = Vec::new();
         for i in 0..tree.leaf_count().as_usize() {
-            let node = &tree.nodes[i];
+            let leaf_index = LeafIndex::from(i);
```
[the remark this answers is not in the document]
raphaelrobert: The problem was that it was off by a factor of 2. The new indexing wasn't available when I wrote this, I'm adapting it now.

tests/test_group.rs
raphaelrobert: When you use rust-analyzer, you get that for free, that's why I don't make these kind of annotations.

tests/test_group.rs
raphaelrobert: Good point. I'll add this at the end in the last PR, because the order is still very much in flux. These tests need a better system anyway, to be able to cover more cases. For now the idea is to just re-enable the tests that were already written.

tests/test_group.rs
raphaelrobert: That comment was outdated

tests/test_group.rs
franziskuskiefer: Yeah, That's what @kkohbrok is currently working on (no PR yet).

src/key_packages/mod.rs
```diff
@@ -82,7 +86,8 @@ impl KeyPackage {
     /// * verify that the signature on this key package is valid
     /// * verify that all mandatory extensions are present
     /// * make sure that the lifetime is valid
-    pub(crate) fn verify(&self) -> bool {
+    /// Returns `Ok(())` if all checks succeed and `KeyPackageError` otherwise
+    pub fn verify(&self) -> Result<(), KeyPackageError> {
```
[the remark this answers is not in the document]
raphaelrobert: Agreed. The problem right now is that the public error code all have integer values which prevents the nesting. We should probably revert on that decision for now and do this only for FFI bindings.

src/tree/mod.rs
```diff
@@ -49,7 +49,7 @@ mod test_treemath;
 #[cfg(test)]
 mod test_util;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
raphaelrobert: Yes, we have to make a complete copy of the tree, including the of the secrets. There might be a way around it, but it would require a crazy setup of references, and at this point I want to keep it simple.

src/tree/mod.rs
```diff
@@ -338,11 +329,9 @@ impl RatchetTree {
         let common_path = treemath::dirpath_long(common_ancestor_index, self.leaf_count())
             .expect("update_path: Error when computing direct path.");
 
-        debug_assert!(sender_direct_path.len() > common_path.len());
-        if sender_direct_path.len() <= common_path.len() {
+        if sender_direct_path.len() < common_path.len() {
```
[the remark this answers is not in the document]
raphaelrobert: No. the `debug_assert` didn't seem to make much sense, not sure how it ended up there.

src/tree/private_tree.rs
raphaelrobert: You are right and that's described here: https://github.com/raphaelrobert/maelstrom/issues/165
It's already fixed in the follow-up PR.

src/tree/treemath.rs
raphaelrobert: This can only be tested with interop test vectors, recalculating the same thing is a moot point. We should do that during the IETF hackathon.

tests/test_group.rs
raphaelrobert: The follow-up PR does test more. We need the API change for that.

tests/test_group.rs
raphaelrobert: Do you mean we should test if the welcome message exists? I can add that test case.

raphaelrobert:
Thanks for all the feedback @franziskuskiefer!

I addressed it in the last two commits, or through comments.

Unfortunately P256 is still broken so it didn't make it in here. I need to investigate what it is this time.

tests/test_group.rs
```diff
@@ -272,11 +272,12 @@ fn group_operations() {
             _ => panic!("Wrong content type"),
         };
         assert!(commit.path.is_none());
+        // Check that the fucntion returned a Welcome message
```
franziskuskiefer, line 275: "function"

tests/test_group.rs
franziskuskiefer: 🤔

tests/test_group.rs
franziskuskiefer: 🤔

tests/test_group.rs
franziskuskiefer: Yeah, we shouldn't just panic here when it's missing but check for this earlier.

src/tree/treemath.rs
franziskuskiefer: A simple test with predefined tree is certainly possible. That's all that's needed for this for now.
Of course we need more and better tests. But this code change should have a test. Same below.

src/tree/mod.rs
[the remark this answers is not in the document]
franziskuskiefer: Any change like this will cost us a couple weeks to fix down the road. If you think that's worth it, you can keep it as is.

src/tree/mod.rs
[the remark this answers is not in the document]
franziskuskiefer: Then we need a test 🙂

franziskuskiefer (changes requested):
Almost there.

There are always a lot of unresolved and unanswered comments. This costs a lot of time to go through every time and increases the number of roundtrips we need to get this reviewed.

tests/test_group.rs
raphaelrobert: The check is now in L276

src/tree/treemath.rs
raphaelrobert: There's a test now in `treemath_test.rs`

src/tree/mod.rs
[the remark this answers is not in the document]
raphaelrobert: I thought about this some more and concluded that this cannot occur with malformed payload. It must be a library error on the treemath level and I changed the check accordingly.

src/tree/mod.rs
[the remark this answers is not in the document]
raphaelrobert: We can discuss this offline, but I don't see a big issue with copying the tree right now. This is also what the spec is suggesting.

franziskuskiefer (approved):
Ok lets get this in and address everything that's still open later.
Please make sure to squash before merging and make the commit message cover all changes in here.

PR #158, opened 2020-10-26
Some leaf index refactoring

kkohbrok:
This PR aims to reduce some of the conversions from `LeafIndex` to `NodeIndex` by implementing `Index` and `MutIndex`, so we can index `Vec<Node>` directly with `LeafIndex` without an explicit detour via `NodeIndex`

### DIFF
--- a/src/tree/index.rs
+++ b/src/tree/index.rs
@@ -80,6 +80,10 @@ impl From<NodeIndex> for LeafIndex {
 impl Index<LeafIndex> for Vec<Node> {
     type Output = Node;
 
+    /// This converts a `LeafIndex`, which points to a particular leaf in the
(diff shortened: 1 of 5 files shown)


src/tree/index.rs
```diff
@@ -72,6 +77,20 @@ impl From<NodeIndex> for LeafIndex {
     }
 }
 
+impl Index<LeafIndex> for Vec<Node> {
+    type Output = Node;
+
+    fn index(&self, leaf_index: LeafIndex) -> &Self::Output {
```
franziskuskiefer, line 83: Add a comment what a leaf index is and why this happens here (and below).

franziskuskiefer (approved):
Nice! Good first step in making this less ugly.

franziskuskiefer (approved):

PR #170, opened 2020-10-27
Change apply_commit to take a reference instead of the Vector of KeyPackageBundles object

kkohbrok:
As the name suggests, this PR changes the function signature of `apply_commit` to accept a mutable reference to the client's `KeyPackageBundle`s instead of taking complete ownership. I think this is a good idea in general, but specifically, it makes testing a bit more convenient, as I don't have to clone that many things.


franziskuskiefer (commented):
So far the API makes sure that all key packages are consumed by the function and the caller is responsible for any clones. But changing this to mutable references means that there will be some key packages left in the vector at the end of the function, which have to be checked explicitly by the caller. So I'm not sure if we really want this.

kkohbrok:
I think the function shouldn't consume all of the key packages bundles of the caller in the first place. As far as I'm aware, the function only picks out the `KeyPackageBundle`s that are used in case someone else commits an update to the party's own leaf. I thus think, we can just pass in all the Party's `KeyPackageBundle`s and the function will consume those that it needs, ignoring the rest.

@raphaelrobert can you confirm this?

franziskuskiefer:
I think both are valid approaches that differ in the responsibility of the caller. But just changing the API here will break the usage because the function will consume some of the key packages in the vector. So if we want to use a reference here I think we should use an immutable one and clone out the values we need. Or better, have a mutable reference and mark used key packages but don't consume then. Or safer, have an immutable argument and return a list identifying the used key packages to the caller.

kkohbrok:
I'd like to understand how it breaks the usage. If we just pass in a mutable reference and the function consumes the KeyPackageBundles it needs (which we won't need anywhere else afterwards), then that's fine, isn't it? Of course we should be sure that we really don't need them elsewhere, but so far, KeyPackageBundles that become the new leaf of that group are also consumed in other functions.

franziskuskiefer:
That's just how it works right now. And when a function is consuming a part of a vector it must be very explicit about which ones and we have to be certain that we don't need them again outside.

kkohbrok:
But you would prefer the approach where we clone everything and and let the function take ownership just to be safe? I'm ok with that, too, but I thought we were trying to get away from cloning private key material (and cloning in general).

franziskuskiefer:
There will always be the need for some clones. But we shouldn't minimize them, especially for private keys. So let's say we do a mutable reference here. Can you define the contract for the function and add it to the PR, i.e. define which ones are consumed, and write something about why the consumed keys aren't needed outside anymore?

raphaelrobert:
As discussed offline, we want:

 - to only pass the KeyPackageBundles that are related to the epoch, not the whole KeyPackage store
 - pass them by reference, so they can be re-used in case applying the Commit fails
 - have better documentation that explains this, including the fact they should be deleted if applying the Commit succeeds

kkohbrok:
In any case, `KeyPackageBundle` should be clonable, which it currently isn't, because `HPKEPrivateKey` does not implement `Clone`. @franziskuskiefer can you change this? If you send me the link to the HPKE repo, I can also make a PR, but with such a small change, it might be easiest if you make the change.

franziskuskiefer:
I'd prefer not to implement `Clone` for the `HPKEPrivateKey`. We have `hazmat` enabled on the crate so you can use `as_slice` to clone out the value. https://github.com/franziskuskiefer/hpke-rs/blob/0.0.2/src/lib.rs#L545

src/group/mls_group/api.rs
```diff
@@ -72,7 +72,7 @@ pub trait Api: Sized {
         &mut self,
         mls_plaintext: MLSPlaintext,
         proposals: Vec<MLSPlaintext>,
-        own_key_packages: Vec<KeyPackageBundle>,
+        own_key_packages: &Vec<KeyPackageBundle>,
```
franziskuskiefer, line 75: What clippy says. But we might actually need the vector. Please check. Same below.

src/group/mls_group/apply_commit.rs
```diff
@@ -85,14 +84,14 @@ impl MlsGroup {
             }
             if is_own_commit {
                 // Find the right KeyPackageBundle among the pending bundles
-                let own_kpb_index = match pending_kpbs
+                let own_kpb: KeyPackageBundle = match own_key_packages
                     .iter()
-                    .position(|kpb| kpb.get_key_package() == kp)
+                    .find(|kpb| kpb.get_key_package() == kp)
                 {
                     Some(i) => i,
```
franziskuskiefer, line 91: `i` is probably not the best variable here anymore.

src/group/mls_group/apply_commit.rs
```diff
@@ -85,14 +84,14 @@ impl MlsGroup {
             }
             if is_own_commit {
                 // Find the right KeyPackageBundle among the pending bundles
```
franziskuskiefer, line 86: Can you update the comment making clear that we clone out the one we want?

src/key_packages/mod.rs
```diff
@@ -237,6 +237,15 @@ pub struct KeyPackageBundle {
     pub(crate) private_key: HPKEPrivateKey,
 }
 
+impl Clone for KeyPackageBundle {
+    fn clone(&self) -> Self {
+        KeyPackageBundle {
+            key_package: self.key_package.clone(),
+            private_key: HPKEPrivateKey::new(self.private_key.as_slice().to_vec()),
```
franziskuskiefer, line 244: Can you add a comment here that this requires the hpke hazmat feature to make clear that this might be undesirable?

src/tree/mod.rs
```diff
@@ -656,16 +656,15 @@ impl RatchetTree {
             self.nodes[sender_index.as_usize()] = leaf_node;
             // Check if it is a self-update
             if sender_index == self.get_own_node_index() {
```
franziskuskiefer, line 658: This is not the sender index anymore.

franziskuskiefer (changes requested):
Thanks. Can you update the PR description and title to what's actually happening here now?
I have a couple nits.

src/tree/mod.rs
kkohbrok: What do you mean? What else would it be? I'm not changing that variable, I think.

src/group/mls_group/api.rs
kkohbrok: I changed it and everything compiles/tests/checks just fine. Anything else I need to look out for?

src/tree/mod.rs
franziskuskiefer: Well.... I don't know 😆

tests/test_group.rs
```diff
@@ -96,7 +96,7 @@ fn create_commit_optional_path() {
     assert!(commit.path.is_none());
 
     // Alice applies the Commit without the forced self-update
-    match group_alice_1234.apply_commit(mls_plaintext_commit, epoch_proposals, vec![]) {
+    match group_alice_1234.apply_commit(mls_plaintext_commit, epoch_proposals, &vec![]) {
```
franziskuskiefer, line 99: These can all be `&[]`.

franziskuskiefer (dismissed):
Ok lets do this. I'm still not super happy with the cloning here but I don't see another way right now.

src/group/mls_group/api.rs
```diff
@@ -72,7 +72,7 @@ pub trait Api: Sized {
         &mut self,
         mls_plaintext: MLSPlaintext,
         proposals: Vec<MLSPlaintext>,
-        own_key_packages: Vec<KeyPackageBundle>,
+        own_key_packages: &[KeyPackageBundle],
```
raphaelrobert, line 75: This still passes the KeyPackageBundles by value. I think what you intended was

`&[&KeyPackageBundle]`

src/key_packages/mod.rs
```diff
@@ -237,6 +237,18 @@ pub struct KeyPackageBundle {
...
+        KeyPackageBundle {
+            key_package: self.key_package.clone(),
+            // This requires the hpke hazmat feature, which is not something
+            // that should ordinarily be used, as we generally try to avoid
+            // cloning private key material.
+            private_key: HPKEPrivateKey::new(self.private_key.as_slice().to_vec()),
+        }
+    }
+}
+
```
raphaelrobert, line 251: Do we need that? With the `PrivateTree` refactoring KeyPackageBundles can be passed as a reference.

raphaelrobert (changes requested):

src/group/mls_group/api.rs
kkohbrok: Isn't a slice `&[KeyPackageBundle]` already a reference and analogous to `&Vec<KeyPackageBundle>? My intention was for the ownership of the `KeyPackageBundle`s not to change when the function is called. I thought using slices here would do the trick?

src/key_packages/mod.rs
kkohbrok: I'll look into it.

src/group/mls_group/api.rs
franziskuskiefer: https://doc.rust-lang.org/book/ch04-03-slices.html

kkohbrok: @franziskuskiefer so slices are inherently references and we're fine here? I re-read the section on slices and that's how I understand it.

src/key_packages/mod.rs
kkohbrok: The clone is indeed not needed anymore :slightly_smiling_face:

src/group/mls_group/api.rs
franziskuskiefer: Yeah, unless @raphaelrobert sees a reason to have references of references here.

raphaelrobert: You're right, apologies.

src/group/mls_group/apply_commit.rs
```diff
@@ -40,7 +39,7 @@ impl MlsGroup {
         // Create provisional tree and apply proposals
         let mut provisional_tree = self.tree.borrow_mut();
         let (path_required_by_commit, group_removed, _invited_members) = match provisional_tree
-            .apply_proposals(&commit.proposals, proposal_queue, &mut pending_kpbs)
+            .apply_proposals(&commit.proposals, proposal_queue, &own_key_packages)
```
franziskuskiefer, line 42: No need for `&`.

franziskuskiefer (dismissed):
lgtm with one nit

franziskuskiefer (approved):

raphaelrobert (approved):

Issue #159, opened 2020-10-26
Change Credential to include SignatureScheme instead of Ciphersuite (#402)


kkohbrok:
It should only contain the SignatureScheme according to spec. We then need to check when creating KeyPackages that the ciphersuite in the KeyPackage and the one in the Credential actually matches. Depends on #156.

raphaelrobert:
Relates to https://github.com/mlswg/mls-protocol/pull/402/

PR #173, opened 2020-10-28
ciphersuites cleanup: Signature keys

franziskuskiefer:
Move sign and verify into the signature keys.

Note that this requires cloning the ciphersuite into the keys for now. This should be changed later by using a global ciphersuite from the config or borrowing the existing cihpersuite.


kkohbrok (approved):
Nice! Ideally, we would have a `SignatureScheme` in the structs instead the `Ciphersuite` but we can do that in a separate PR. We could probably do it as part of #159.

PR #176, opened 2020-10-28
Introduce a Secrets type and some Ciphersuite refactoring

kkohbrok:
The purpose of this PR is to restrict the handling of "raw" secrets to the `ciphersuite` module. For the moment, this effort is focused specifically on secrets that are inputs or outputs of `KDF.Extrac` or `KDF.Expand`, as well AEAD keys.

Specifically, this PR introduces a `Secrets` type and refactors various functions in the `ciphersuites` module to use it. Additionally, I move the aead functions `aead_seal` and `aead_open` into the `AeadKey` struct.

An alternative design could be making `Secret` a trait, which would enable us to have instances for more specific secrets such as `CommitSecret` (which I've removed for now).

Let me know what you think.


src/group/mls_group/apply_commit.rs
```diff
@@ -127,7 +127,7 @@ impl MlsGroup {
         let own_confirmation_tag = ConfirmationTag::new(
             &ciphersuite,
             &provisional_epoch_secrets.confirmation_key,
-            &confirmed_transcript_hash,
+            &Secret::from(confirmed_transcript_hash),
```
raphaelrobert, line 130: The transcript hashes are not secrets.

src/messages/mod.rs
```diff
@@ -39,10 +39,14 @@ pub struct ConfirmationTag(pub Vec<u8>);
 impl ConfirmationTag {
     pub fn new(
         ciphersuite: &Ciphersuite,
-        confirmation_key: &[u8],
-        confirmed_transcript_hash: &[u8],
+        confirmation_key: &Secret,
+        confirmed_transcript_hash: &Secret,
```
raphaelrobert, line 43: Same as above, this is not a secret.

raphaelrobert (changes requested):
I agree that going forward a Trait would be better to distinguish different secrets, but this is a good starting point!

src/ciphersuite/mod.rs
```diff
@@ -121,7 +150,7 @@ pub struct Ciphersuite {
     hpke_kem: HpkeKemMode,
     hpke_kdf: HpkeKdfMode,
     hpke_aead: HpkeAeadMode,
-    aead: AeadMode,
+    aead_mode: AeadMode,
```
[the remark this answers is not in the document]
kkohbrok: I think we should have it for the other modes as well, but for now I've only refactored the AEAD part of `ciphersuite.rs`.

src/ciphersuite/mod.rs
```diff
@@ -293,34 +305,100 @@ impl Ciphersuite {
...
 }
 
 impl AeadNonce {
     /// Build a new nonce for an AEAD from `bytes`.
+    #[cfg(test)]
     pub(crate) fn from_slice(bytes: &[u8]) -> Self {
         let mut nonce = [0u8; NONCE_BYTES];
         nonce.clone_from_slice(bytes);
         AeadNonce { value: nonce }
     }
```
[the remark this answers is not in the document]
kkohbrok: It was used in some tests, but as it was always used to create nonces from random, I refactored it to a constructor from random.

src/group/mls_group/apply_commit.rs
kkohbrok: I agree. However, I have changed the `extract` and `expand` functions to expect `Secret`s as inputs for both `salt` and `ikm`, so I need to make it a secret here for the moment. Note that this particular instance will go away once the confirmation tag is a MAC.

kkohbrok: I have changed it now such that the `ConfirmationTag` constructor takes `&[u8]` as input and the transformation to `Secret` is handled internally. It's indeed a bit nicer now. Was that what you meant?

src/messages/mod.rs
kkohbrok: See my reply above.

src/ciphersuite/mod.rs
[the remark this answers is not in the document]
raphaelrobert: In that case all of this should be done in one go. It's not good practice to do partial renaming without a good reason. Since it's not really related to this PR, I would recommend doing a dedicated PR.

raphaelrobert (changes requested):

src/ciphersuite/mod.rs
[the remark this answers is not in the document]
kkohbrok: Done!

raphaelrobert (dismissed):

src/ciphersuite/mod.rs
```diff
@@ -296,31 +308,96 @@ impl Ciphersuite {
...
+    }
+
+    /// AEAD decrypt `ciphertext` with `key`, `aad`, and `nonce`.
+    pub(crate) fn aead_open(
+        &self,
+        ciphertext: &[u8],
+        aad: &[u8],
+        nonce: &AeadNonce,
+    ) -> Result<Vec<u8>, AEADError> {
+        // TODO: don't hard-code tag bytes
```
franziskuskiefer, line 370: Any TODO should have an issue. (Not sure if we'll actually do this before there is any PQ scheme.)

src/ciphersuite/mod.rs
```diff
@@ -296,31 +308,96 @@ impl Ciphersuite {
...
     }
 
+    #[cfg(test)]
     /// Get a slice to the key value.
     pub(crate) fn as_slice(&self) -> &[u8] {
         &self.value
     }
+
+    /// Encrypt a payload under the AeadKey given a nonce.
+    pub fn aead_seal(
```
franziskuskiefer, line 342: Does this need to be public?

src/ciphersuite/mod.rs
```diff
@@ -296,31 +308,96 @@ impl Ciphersuite {
             .unwrap()
     }
 
-    /// Generate a new HPKE key pair and return it.
-    pub(crate) fn derive_hpke_keypair(&self, ikm: &[u8]) -> HPKEKeyPair {
-        self.hpke.derive_key_pair(ikm)
+    /// Derive a new HPKE keypair from a given Secret.
+    pub(crate) fn derive_hpke_keypair(&self, ikm: &Secret) -> HPKEKeyPair {
+        self.hpke.derive_key_pair(ikm.value.as_slice())
```
franziskuskiefer, line 313: `&`?

franziskuskiefer (changes requested):
Generally looks, some comments inline. Not sure yet if they were addressed already. The review got a little stale.

src/ciphersuite/codec.rs
```diff
@@ -84,3 +84,15 @@ impl Codec for ReuseGuard {
         Ok(ReuseGuard { value: guard })
     }
 }
+
+impl Codec for Secret {
```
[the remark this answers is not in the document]
kkohbrok: I think we _do_ send secrets over the wire, for example when sending a path secrets to the co-path the course of an update.

src/ciphersuite/mod.rs
```diff
@@ -66,8 +66,37 @@ pub enum AEADError {
     WrongKeyLength,
 }
 
+/// A struct to contain secrets. This is to provide better visibility into where
+/// and how secrets are used and to avoid passing secrets in their raw
+/// representation.
+#[derive(Clone, PartialEq, Eq, Default, Debug)]
+pub struct Secret {
+    value: Vec<u8>,
```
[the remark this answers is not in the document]
kkohbrok: Good point. I thought it was needed, because it was derived for `EpochSecrets`, but it doesn't seem to be needed by anything, so I removed it.

src/ciphersuite/mod.rs
```diff
@@ -66,8 +66,37 @@ pub enum AEADError {
...
+}
+
+impl Secret {
+    /// Create an empty secret.
+    pub(crate) fn new_empty_secret() -> Self {
```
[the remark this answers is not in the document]
kkohbrok: I'm not sure. I prefer to make it explicit that it's generating an empty secret. Someone might think that `default` yields a randomly generated one. I don't have too strong of an opinion here, so if you feel strongly about it, I can change it.

src/tree/mod.rs
```diff
@@ -499,7 +502,7 @@ impl RatchetTree {
                 .map(|&x| {
                     let pk = self.nodes[x.as_usize()].get_public_hpke_key().unwrap();
                     self.ciphersuite
-                        .hpke_seal(&pk, group_context, &[], &path_secret)
+                        .hpke_seal_secret(&pk, group_context, &[], &path_secret)
```
franziskuskiefer, line 505: Why is this a different function to `hpke_seal`?

franziskuskiefer (changes requested):
Ok I think we're almost there. A couple nits and thoughts.

src/ciphersuite/codec.rs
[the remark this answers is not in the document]
kkohbrok: That's probably a good idea. Quite a few Secrets already have their own wrapper struct such as `PathSecret`, but then we'd need to create a few more wrapper structs, this is partially because all Secrets in `EpochSecrets` need to be encoded (although if can get rid of `Codec` for `MlsGroup`, that would disappear).

Also, I think those Secret wrappers would have to live in `ciphersuites/mod.rs` so they can access Secret.value directly.

src/ciphersuite/mod.rs
```diff
@@ -296,31 +308,96 @@ impl Ciphersuite {
...
+    pub(crate) fn from_secret(secret: Secret, aead_mode: AeadMode) -> AeadKey {
+        AeadKey {
+            aead_mode,
+            value: secret.value,
+        }
+    }
+
+    #[cfg(test)]
+    /// Generate a random AEAD Key
+    pub fn new_from_random(aead_mode: AeadMode) -> Self {
```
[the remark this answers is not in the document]
kkohbrok: Good question. We have that pattern in a few other places, but I agree that it's awkward and I've changed it.

I think the thought was that other `from_...` functions always consumed a value and this one doesn't really. Except maybe the randomness.

src/tree/mod.rs
kkohbrok: It's so we can seal Secrets directly without having to transform a secret into a slice first. But I agree that it's no necessary if we can't get rid of `encode` as well. The problem is that we're using `hpke_seal` not just on `Secret`s, but also, for example, on `GroupSecrets`, which contains `Secret`s as well. So one way to get rid of `encode` would be to have dedicated `hpke_seal` functions.

franziskuskiefer (approved):
Ok let's get this landed.

PR #177, opened 2020-10-28
Re-enable group tests 2/?

### LINKED ISSUE #165: PathSecrets are not spec compliant
The HPKE private key is not derived from the node secret. This will probably require a change in KeyPackageBundle and the API.

### LINKED ISSUE #166: Create Commit does not return new KeyPackage
During the creation of a Commit message, a new KeyPackage can be created , if a path field is required for that Commit. This KeyPackage, as well as the secret values should be returned to the caller.

### LINKED ISSUE #181: Check if we ever need to clone private keys
Right now HPKE private keys are cloned in a few places. It would be neat if can avoid that completely.

raphaelrobert:
This is a follow-up PR to #157.

It addresses the following:

**PrivateTree refactoring**

 - split secret generating functions
 - compute commit_secret automatically after secret path generation
 - remove leaf_secret field
 - combine generation of path secrets and keypairs
 - adapt tests

**Fixes**

 - Fixes #165 by no longer derinving path secrets from the HPKE private key
 - Fixes #166 by changing the group API
 - Fixes #181 by no longer cloning the `RatchetTree` and only cloning the `leaf_secret` for `KeyPackageBundle`

**Tests**

 - Add a test for ProposalQueue filtering


src/tree/mod.rs
```diff
@@ -407,53 +397,54 @@ impl RatchetTree {
...
             .unwrap();
 
-        // Compute the parent hash extension and update the KeyPackage
+        // Compute the parent hash extension and update the KeyPackage and sign it
         let parent_hash = self.compute_parent_hash(own_index);
         let key_package = self.get_own_key_package_ref_mut();
         key_package.update_parent_hash(&parent_hash);
+        // Sign the KeyPackage
         key_package.sign(credential_bundle);
+        // STore it in the UpdatePath
```
franziskuskiefer, line 427: "Store"

franziskuskiefer (changes requested):
So my main questions here I think is who's owning the hpke private keys. There should be only one. That's not the case here as far as I can see.

src/tree/private_tree.rs
```diff
@@ -126,17 +150,13 @@ impl PrivateTree {
     /// ```
     ///
     /// Note that this overrides the `path_secrets`.
-    pub(crate) fn generate_path_secrets(
+    pub(crate) fn continue_path_secrets(
```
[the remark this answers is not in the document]
raphaelrobert: The two functions are different semantically, despite some code overlap. Combining them broke the logic because some parts went missing in that process. It's much more readable and easier to audit that way.

src/group/mls_group/create_commit.rs
```diff
@@ -46,7 +48,7 @@ impl MlsGroup {
 
         let sender_index = self.sender_index();
         // Make a copy of the current tree to apply proposals safely
-        let mut provisional_tree = self.tree().clone();
+        let mut provisional_tree = RatchetTree::new_from_public_tree_values(self.tree().borrow());
```
franziskuskiefer, line 51: This is equivalent to `&self.tree()`. Then you also don't need to `use std::borrow::Borrow;`.

src/tree/mod.rs
```diff
@@ -88,6 +81,16 @@ impl RatchetTree {
         }
     }
 
+    /// Create a new `RatchetTree` by cloning the public tree nodes from another tree
+    /// and an empty `PrivateTree`
+    pub(crate) fn new_from_public_tree_values(ratchet_tree: &RatchetTree) -> Self {
```
franziskuskiefer, line 86: Maybe `from_tree` or something? We pass in the entire tree, not only public values.
But this should really be implemented in the `Clone` trait of `RatchetTree` I guess, as that's what this is really.

src/messages/test_proposals.rs
```diff
@@ -0,0 +1,132 @@
...
+    use crate::extensions::*;
+    use crate::framing::*;
+    use crate::group::*;
+    use crate::key_packages::*;
+    use crate::messages::proposals::*;
+    use crate::tree::index::*;
+
+    for ciphersuite_name in Config::supported_ciphersuites() {
+        let ciphersuite = Ciphersuite::new(*ciphersuite_name);
+        //let ciphersuite = Ciphersuite::new(ciphersuite_name);
```
franziskuskiefer, line 18: 🤔

src/messages/test_proposals.rs
```diff
@@ -0,0 +1,132 @@
+/// This test makes sure ProposalQueue works as intented. This functionality is used in `create_commit` to filter the epoch proposals.
+/// Expected result:
+/// `get_filtered_proposals` returns only proposals of a certain type
+#[test]
+fn proposal_queue_functions() {
+    use crate::ciphersuite::*;
```
franziskuskiefer, line 6: The imports should probably go into the file itself in case there are more functions.

src/tree/private_tree.rs
```diff
@@ -36,69 +41,61 @@ impl Clone for PrivateTree {
...
             path_keys: PathKeys::default(),
             commit_secret: self.commit_secret.clone(),
+            leaf_secret: self.leaf_secret.clone(),
             path_secrets: self.path_secrets.clone(),
         }
     }
 }
 
 impl PrivateTree {
     /// Create a minimal `PrivateTree` setting only the private key.
```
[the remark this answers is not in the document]
franziskuskiefer: Well, it's the most significant change in this PR. So it must be clarified what exactly is happening and why it's needed.

src/tree/private_tree.rs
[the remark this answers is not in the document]
franziskuskiefer: Ok, if you want to separate functions then factor out the common code and let both functions call this function to avoid code duplication.

franziskuskiefer (changes requested):
Thanks, getting there.

But the most significant questions about who's owning private keys wasn't answered. Having a follow up is fine but since it's the most significant change in this PR I need to understand what's happening.

src/tree/mod.rs
raphaelrobert: I renamed it. We discussed that ratchet trees should not be clonable and that's why I implemented it like this. Of course I can also implement `Clone` again, but it might be misleading, because you might assume that all values from the tree are then cloned.

src/tree/private_tree.rs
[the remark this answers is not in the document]
raphaelrobert: Done

src/tree/private_tree.rs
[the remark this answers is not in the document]
raphaelrobert: I extended the comments

src/group/mls_group/mod.rs
```diff
@@ -227,7 +227,8 @@ impl Api for MlsGroup {
         let tree = self.tree.borrow();
         let mut roster = Vec::new();
         for i in 0..tree.leaf_count().as_usize() {
-            let node = &tree.nodes[LeafIndex::from(i)];
+            let leaf_index = LeafIndex::from(i);
+            let node = &tree.nodes[NodeIndex::from(leaf_index).as_usize()];
```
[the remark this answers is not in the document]
franziskuskiefer: I don't think so, LeafIndex *must* bei used to index the nodes. The index implementation ensures that the correct node is take. Having the two different types only make sense if they are used to actually make the access safe, not for converting all the time. So again, if this isn't working, there's a bug that needs fixing. Either the index implementation is broken or this shouldn't be a LeafIndex.

src/key_packages/mod.rs
```diff
@@ -278,23 +301,41 @@ impl KeyPackageBundle {
...
+        leaf_secret_option: Option<Vec<u8>>,
+    ) -> Self {
         Self {
             key_package,
             private_key,
+            leaf_secret_option,
         }
     }
 
+    /// Replace the init key in the current KeyPackage with a random one
```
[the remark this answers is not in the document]
franziskuskiefer: 🤔

src/key_packages/mod.rs
```diff
@@ -262,44 +284,76 @@ impl KeyPackageBundle {
...
         &mut self.key_package
     }
 
     /// Get a reference to the `HPKEPrivateKey`.
-    pub fn get_private_key(&self) -> &HPKEPrivateKey {
+    pub(crate) fn get_private_key_ref(&self) -> &HPKEPrivateKey {
         &self.private_key
     }
+
+    /// Get a reference to the `leaf_secret_option`.
```
franziskuskiefer, line 347: Not an `Option`.

src/key_packages/mod.rs
```diff
@@ -262,44 +284,76 @@ impl KeyPackageBundle {
...
+    pub(crate) fn get_leaf_secret(&self) -> &[u8] {
```
franziskuskiefer, line 348: `leaf_secret`

src/group/mls_group/mod.rs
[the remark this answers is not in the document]
raphaelrobert: I reverted the change and I think this works for now. I also filed #185 because I think more work is needed here.

src/tree/index.rs
```diff
@@ -85,13 +85,13 @@ impl Index<LeafIndex> for Vec<Node> {
     /// same leaf, but in the array representing the tree as opposed to the one
     /// only containing the leaves.
     fn index(&self, leaf_index: LeafIndex) -> &Self::Output {
-        &self[leaf_index.as_usize() * 2]
+        &self[NodeIndex::from(leaf_index).as_usize()]
```
franziskuskiefer, line 88: Ha so this was broken. `as_usize` does the `*2` already. The result doesn't look nice though. Let's clean that up in a follow up.

franziskuskiefer (changes requested):
Ok, almost there. Only a couple nits.

raphaelrobert:
@franziskuskiefer I think everything is addressed, can you approve?

franziskuskiefer (approved):
Thanks, let's get this in.

PR #180, opened 2020-10-30
Extend HPKE test

raphaelrobert:
This PR extends the HPKE test to also cover `hpke_open()`.

### DIFF
--- a/src/ciphersuite/test_ciphersuite.rs
+++ b/src/ciphersuite/test_ciphersuite.rs
@@ -3,16 +3,17 @@
 use super::*;
 use crate::config::Config;
 
-// Spot test to make sure hpke seal works.
+// Spot test to make sure hpke seal/open work.
 #[test]
-fn test_hpke_seal() {
+fn test_hpke_seal_open() {
     // Test through ciphersuites.
     for &suite in Config::supported_ciphersuites() {
         println!("Test {:?}", suite);
         let ciphersuite = Ciphersuite::new(suite);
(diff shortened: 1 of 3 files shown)


src/ciphersuite/test_ciphersuite.rs
```diff
@@ -3,16 +3,17 @@
...
+fn test_hpke_seal_open() {
     // Test through ciphersuites.
     for &suite in Config::supported_ciphersuites() {
         println!("Test {:?}", suite);
         let ciphersuite = Ciphersuite::new(suite);
         println!("Ciphersuite {:?}", ciphersuite);
         let kp = ciphersuite.new_hpke_keypair();
-        ciphersuite.hpke_seal(kp.public_key(), &[], &[], &[1, 2, 3]);
+        let ciphertext = ciphersuite.hpke_seal(kp.public_key(), &[], &[], &[1, 2, 3]);
+        ciphersuite.hpke_open(&ciphertext, kp.private_key(), &[], &[]);
```
franziskuskiefer, line 16: Can you make sure that the output here is equivalent to the input we put into seal?

franziskuskiefer (changes requested):
Thanks, but let's actually test that the result is correct at least.

src/ciphersuite/test_ciphersuite.rs
raphaelrobert: Fixed

franziskuskiefer (approved):

franziskuskiefer (approved):

Issue #133, opened 2020-10-19
Validate proposals


raphaelrobert:
- Syntactic checks 
- Semantic checks
- Policy checks

franziskuskiefer:
From #293 

> I took quite some time to investigate and refactor this. There is now a function that verifies both the signature and the membership tag, it is however just used in tests.
> 
> Adding the signature and the membership tag and verifying them is complicated and needs to happen in different places and separately.
> 
> External plaintexts:
> 
> They have no membership tag, just a mandatory signature
> Group plaintexts:
> 
> They have both a signature and a membership tag
> For Commits: The membership tag covers the confirmation tag, which covers the signature. This means the signature and the membership tag cannot be created at the same time, it has to happen sequentially.
> Group ciphertexts:
> 
> They have no membership tag
> Verification:
> 
> The verification also happens in different places. MlsGroup only verifies the signature, because it cannot determine whether a plaintext came from a ciphertext or was sent as a plaintext, therefore it doesn't know whether a membership tag is required.
> I used this opportunity to do some refactoring and fix some bugs as well, but the gist of it is that integrity checks of MLSPlaintext remain complicated.

franziskuskiefer:
https://github.com/openmls/openmls/blob/0a39799516d9c024e803a8088613a58fcdcef03c/src/group/managed_group/mod.rs#L421

> This entire function needs to be refactored... maybe file a follow up.

_Originally posted by @franziskuskiefer in https://github.com/openmls/openmls/pull/397#r620044683_

kkohbrok:
> Can you file a follow up then for a negative test case?

_Originally posted by @franziskuskiefer in https://github.com/openmls/openmls/pull/397#discussion_r625073480_

When (re-)implementing the validation logic, we should make sure to include negative test cases to catch invalid messages.

franziskuskiefer:
With #705 fixed ValSem004 might have changed semantics, depending on how the sender and member are defined.

kkohbrok:
I just closed #984 since there is not much to do from a functional standpoint. I'm just posting this here so that we don't lose sight of the fact that the application needs to validate identity changes. (Although that should probably pop up again when we collect all the MUSTs from the spec.)

PR #182, opened 2020-11-01
Re-enable group tests 3/3

### LINKED ISSUE #77: Re-enable group tests

raphaelrobert:
This is a follow-up PR to #177.

This PR re-enables all remaining group operation tests that were previously commented out and closes #77.

While much more testing will be needed to cover more scenarios, this is not in the scope of this PR. The only goal is to have a baseline for testing.

### DIFF
--- a/tests/test_group.rs
+++ b/tests/test_group.rs
@@ -109,7 +109,7 @@ fn create_commit_optional_path() {
         let ratchet_tree = group_alice.tree().public_key_tree_copy();
 
         // Bob creates group from Welcome
-        let _group_bob = match MlsGroup::new_from_welcome(
+        let group_bob = match MlsGroup::new_from_welcome(
             welcome_bundle_alice_bob_option.unwrap(),
             Some(ratchet_tree),
             bob_key_package_bundle,
@@ -120,7 +120,7 @@ fn create_commit_optional_path() {
 
         assert_eq!(
             group_alice.tree().public_key_tree(),
-            group_alice.tree().public_key_tree()
+            group_bob.tree().public_key_tree()
         );
 
         // Alice updates
@@ -165,7 +165,7 @@ fn basic_group_setup() {
     for ciphersuite in Config::supported_ciphersuites() {
         let group_aad = b"Alice's test group";
 
-        // Define identities
+        // Define credential bundles
         let alice_credential_bundle =
             CredentialBundle::new("Alice".into(), CredentialType::Basic, ciphersuite.name())
                 .unwrap();
@@ -173,20 +173,14 @@ fn basic_group_setup() {
             CredentialBundle::new("Bob".into(), CredentialType::Basic, ciphersuite.name()).unwrap();
 
         // Generate KeyPackages
-        let bob_key_package_bundle = KeyPackageBundle::new(
-            &[ciphersuite.name()],
-            &bob_credential_bundle, // TODO: bad API, we shouldn't have to get the private key out here (this function shouldn't exist!)
-            Vec::new(),
-        )
-        .unwrap();
+        let bob_key_package_bundle =
+            KeyPackageBundle::new(&[ciphersuite.name()], &bob_credential_bundle, Vec::new())
+                .unwrap();
         let bob_key_package = bob_key_package_bundle.get_key_package();
 
-        let alice_key_package_bundle = KeyPackageBundle::new(
-            &[ciphersuite.name()],
-            &alice_credential_bundle, // TODO: bad API, we shouldn't have to get the private key out here (this function shouldn't exist!)
-            Vec::new(),
-        )
-        .unwrap();
+        let alice_key_package_bundle =
+            KeyPackageBundle::new(&[ciphersuite.name()], &alice_credential_bundle, Vec::new())
+                .unwrap();
 
         // Alice creates a group
         let group_id = [1, 2, 3, 4];
@@ -232,7 +226,7 @@ fn group_operations() {
     for ciphersuite in Config::supported_ciphersuites() {
         let group_aad = b"Alice's test group";
 
-        // Define identities
+        // Define credential bundles
         let alice_credential_bundle =
             CredentialBundle::new("Alice".into(), CredentialType::Basic, ciphersuite.name())
                 .unwrap();
--- a/tests/test_group.rs
+++ b/tests/test_group.rs
@@ -2,212 +2,219 @@ use openmls::prelude::*;
 
 #[test]
 fn create_commit_optional_path() {
-    let ciphersuite_name = CiphersuiteName::MLS10_128_DHKEMX25519_AES128GCM_SHA256_Ed25519;
-    let group_aad = b"Alice's test group";
+    for ciphersuite in Config::supported_ciphersuites() {
+        let group_aad = b"Alice's test group";
 
-    // Define identities
-    let alice_credential_bundle =
-        CredentialBundle::new("Alice".into(), CredentialType::Basic, ciphersuite_name).unwrap();
-    let bob_credential_bundle =
-        CredentialBundle::new("Bob".into(), CredentialType::Basic, ciphersuite_name).unwrap();
+        // Define identities
+        let alice_credential_bundle =
+            CredentialBundle::new("Alice".into(), CredentialType::Basic, ciphersuite.name())
+                .unwrap();
+        let bob_credential_bundle =
+            CredentialBundle::new("Bob".into(), CredentialType::Basic, ciphersuite.name()).unwrap();
 
-    // Mandatory extensions, will be fixed in #164
-    let lifetime_extension = Box::new(LifetimeExtension::new(60));
-    let mandatory_extensions: Vec<Box<dyn Extension>> = vec![lifetime_extension];
+        // Mandatory extensions, will be fixed in #164
+        let lifetime_extension = Box::new(LifetimeExtension::new(60));
+        let mandatory_extensions: Vec<Box<dyn Extension>> = vec![lifetime_extension];
 
-    // Generate KeyPackages
-    let alice_key_package_bundle = KeyPackageBundle::new(
-        &[ciphersuite_name],
-        &alice_credential_bundle,
-        mandatory_extensions.clone(),
-    )
-    .unwrap();
+        // Generate KeyPackages
+        let alice_key_package_bundle = KeyPackageBundle::new(
+            &[ciphersuite.name()],
+            &alice_credential_bundle,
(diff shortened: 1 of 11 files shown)


tests/test_group.rs
```diff
@@ -239,36 +241,37 @@ fn group_operations() {
...
             bob_key_package.clone(),
         );
         let epoch_proposals = vec![bob_add_proposal];
-        let (mls_plaintext_commit, welcome_bundle_alice_bob_option, kpb_option) =
-            match group_alice_1234.create_commit(
+        let (mls_plaintext_commit, welcome_bundle_alice_bob_option, kpb_option) = match group_alice
+            .create_commit(
                 group_aad,
                 &alice_credential_bundle,
                 epoch_proposals.clone(),
```
franziskuskiefer, line 269: So this didn't change but I'm not quite sure why we have to clone the proposals here (not sure if I asked this before...). If the proposals don't change and we need them again outside, shouldn't they be passed in as reference?

tests/test_group.rs
```diff
@@ -206,8 +201,15 @@ fn basic_group_setup() {
...
+///  - Bob updates and commits
+///  - Alice updates and commits
+///  - Bob updates and Alice commits
+///  - Bob adds Charlie
+///  - Charlie sends a message to the group
+///  - Charlie updates and commits
+///  - Charlie removes Bob
 fn group_operations() {
     let supported_ciphersuites = vec![
         CiphersuiteName::MLS10_128_DHKEMX25519_AES128GCM_SHA256_Ed25519,
```
franziskuskiefer, line 215: Please add the P256 ciphersuite. You can also use `Config::supported_ciphersuites` now.

tests/test_group.rs
```diff
@@ -312,106 +312,344 @@ fn group_operations() {
...
-    // Bob updates and Alice commits
-    let update_proposal_bob = group_bob.create_update_proposal(None);
-    group_alice_1234.process_proposal(update_proposal_bob);
-    let (commit5, ms5, _) = group_alice_1234.create_commit(None);
+        let update_proposal_bob = group_bob.create_update_proposal(
+            &[],
+            &bob_credential_bundle,
+            bob_update_key_package_bundle.get_key_package().clone(),
+        );
+        let (mls_plaintext_commit, _, kpb_option) = match group_bob.create_commit(
```
franziskuskiefer, line 328: Should we check what's in the `Welcome` as well even if we don't use it? (Same below.)

tests/test_group.rs
```diff
@@ -312,106 +312,344 @@ fn group_operations() {
...
+        group_alice
+            .apply_commit(
+                mls_plaintext_commit.clone(),
+                vec![update_proposal_alice.clone()],
+                vec![kpb_option.unwrap()],
+            )
+            .expect("Error applying commit (Alice)");
+        group_bob
+            .apply_commit(
+                mls_plaintext_commit.clone(),
```
franziskuskiefer, line 396: `clone` shouldn't be necessary here (if at all, see comment above). (Same below.)

franziskuskiefer (changes requested):
Thanks, looking mostly good. A couple comments though.

tests/test_group.rs
```diff
@@ -97,14 +97,14 @@ fn create_commit_optional_path() {
...
+    match group_alice.apply_commit(mls_plaintext_commit, epoch_proposals, vec![]) {
         Ok(_) => {}
         Err(e) => panic!("Error applying commit: {:?}", e),
     };
-    let ratchet_tree = group_alice_1234.tree().public_key_tree_copy();
+    let ratchet_tree = group_alice.tree().public_key_tree_copy();
 
     // Bob creates group from Welcome
-    let _group_bob_1234 = match MlsGroup::new_from_welcome(
+    let _group_bob = match MlsGroup::new_from_welcome(
```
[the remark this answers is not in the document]
raphaelrobert: It would have panicked in case of an error, but the `assert` was indeed wrong.

tests/test_group.rs
raphaelrobert: The reason why the proposals are cloned here is because we simulate different groups. In a real scenario they would be "cloned" by the Delivery Service.

tests/test_group.rs
raphaelrobert: Yes, that check was missing.

tests/test_group.rs
raphaelrobert: Weird that clippy didn't catch that one. In general the cloning is needed because of the group simulation.

tests/test_group.rs
```diff
@@ -2,233 +2,242 @@ use openmls::prelude::*;
...
+        let commit = match &mls_plaintext_commit.content {
+            MLSPlaintextContentType::Commit((commit, _)) => commit,
+            _ => panic!(),
+        };
+        assert!(commit.path.is_some());
+        assert!(commit.path.is_some() && kpb_option.is_some());
+
+        // Alice adds Bob without forced self-update
+        // Since there are only Add Proposals, this does not generate a path field on the Commit
+        // Creating a second proposal to add the same member should not fail, only committing that proposal should fail
```
kkohbrok, line 81: Should we test here, that adding that proposal twice does indeed fail? Or will we do that in the course of #133?

tests/test_group.rs
```diff
@@ -2,233 +2,242 @@ use openmls::prelude::*;
...
-        )
-        .expect("Error applying commit");
+            Ok(group) => group,
+            Err(e) => panic!("Error creating group from Welcome: {:?}", e),
+        };
+
+        assert_eq!(
+            group_alice.tree().public_key_tree(),
+            group_alice.tree().public_key_tree()
+        );
```
kkohbrok, line 124: I'm confused. This seem trivial to be trivially true for me.

It should probably be Bob's tree? Maybe we should have Bob also perform the commit? That should work as well.

tests/test_group.rs
```diff
@@ -2,233 +2,242 @@ use openmls::prelude::*;
...
-        vec![bob_add_proposal],
-        true,
-    ) {
-        Ok(c) => c,
-        Err(e) => panic!("Error creating commit: {:?}", e),
-    };
+    for ciphersuite in Config::supported_ciphersuites() {
+        let group_aad = b"Alice's test group";
+
+        // Define identities
```
kkohbrok, line 168: identities -> CredentialBundles

tests/test_group.rs
```diff
@@ -2,233 +2,242 @@ use openmls::prelude::*;
...
+        let alice_credential_bundle =
+            CredentialBundle::new("Alice".into(), CredentialType::Basic, ciphersuite.name())
+                .unwrap();
+        let bob_credential_bundle =
+            CredentialBundle::new("Bob".into(), CredentialType::Basic, ciphersuite.name()).unwrap();
+
+        // Generate KeyPackages
+        let bob_key_package_bundle = KeyPackageBundle::new(
+            &[ciphersuite.name()],
+            &bob_credential_bundle, // TODO: bad API, we shouldn't have to get the private key out here (this function shouldn't exist!)
```
kkohbrok, line 178: I think that TODO is obsolete.

tests/test_group.rs
```diff
@@ -2,233 +2,242 @@ use openmls::prelude::*;
...
-    let supported_ciphersuites = vec![
-        CiphersuiteName::MLS10_128_DHKEMX25519_AES128GCM_SHA256_Ed25519,
-        CiphersuiteName::MLS10_128_DHKEMX25519_CHACHA20POLY1305_SHA256_Ed25519,
-    ];
-
-    for ciphersuite_name in supported_ciphersuites {
+    for ciphersuite in Config::supported_ciphersuites() {
         let group_aad = b"Alice's test group";
 
         // Define identities
```
kkohbrok, line 235: identities -> CredentialBundles

tests/test_group.rs
```diff
@@ -302,18 +311,15 @@ fn group_operations() {
...
+        // === Alice sends a message to Bob ===
         let message_alice = [1, 2, 3];
-        let mls_ciphertext_alice = group_alice_1234.create_application_message(
-            &[],
-            &message_alice,
-            &alice_credential_bundle,
-        );
+        let mls_ciphertext_alice =
+            group_alice.create_application_message(&[], &message_alice, &alice_credential_bundle);
         let mls_plaintext_bob = match group_bob.decrypt(mls_ciphertext_alice) {
```
kkohbrok, line 323: Maybe test that alice's message is equal to the message decrypted by Bob?

tests/test_group.rs
kkohbrok: Clippy is probably confused that the variable is shadowed later on.

tests/test_group.rs
```diff
@@ -322,106 +328,351 @@ fn group_operations() {
...
+            charlie_key_package_bundle,
+        ) {
+            Ok(group) => group,
+            Err(e) => panic!("Error creating group from Welcome: {:?}", e),
+        };
 
-    group_alice_1234.process_commit(commit8.clone());
-    group_bob.process_commit(commit8.clone());
-    group_charlie.process_commit(commit8);
+        // Make sure that all groups have the same public tree
```
kkohbrok, line 532: Maybe it would be worth writing a `group_agreement` test, that takes a vector of group states and tests that they are equivalent in some ways, e.g. by comparing the public tree, encrypting/decrypting messages, etc?

kkohbrok (dismissed):

tests/test_group.rs
raphaelrobert: As mentioned, we need a lot more testing. I just wanted to get the baseline ready.

tests/test_group.rs
raphaelrobert: Yeah, that's a regression from all the merging...

tests/test_group.rs
raphaelrobert: That's what we are testing in the lines below.

tests/test_group.rs
raphaelrobert: Sure, but any further testing it out of scope for this PR.

tests/test_group.rs
kkohbrok: Ah, sorry, I didn't expand the lines hidden by GH. Right you are!

tests/test_group.rs
kkohbrok: It's not necessarily additional testing, it's just externalizing the "groups-are-equal" assertion into an additional function that does that for us.

kkohbrok (approved):

Issue #79, opened 2020-10-08
Managed groups


raphaelrobert:
- [ ] Define API for managed groups (stubs are there)
- [ ] Implement managed groups

franziskuskiefer:
- [ ] maybe use key package id extension for identifying keys packages?

PR #184, opened 2020-11-02
Introduce prelude

### LINKED ISSUE #84: Public API
Define the public API.

We should also have a prelude for users of the library with the public API.

raphaelrobert:
This PR makes all modules private and introduces a more restricted `prelude` interface.

Fixes #84.


franziskuskiefer (approved):
Nice, thanks! We should probably improve on this later and see if we should for example mention everything explicitly in `prelude.rs` instead of global exports (`*`).

Btw. I think we should also get rid of the `Api` and put it in the doc comment of the `MlsGroup` or `lib` instead.

raphaelrobert:
> Nice, thanks! We should probably improve on this later and see if we should for example mention everything explicitly in `prelude.rs` instead of global exports (`*`).
> 
> Btw. I think we should also get rid of the `Api` and put it in the doc comment of the `MlsGroup` or `lib` instead.

I agree about the `Api`. It doesn't serve a functional purpose anyway.

This might also change when we tackle #79.

PR #188, opened 2020-11-03
gh action for deploying rust docs

franziskuskiefer:
This action builds the rust docs and published them to the gh-pages branch (https://openmls.github.io/openmls).


.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,28 @@
...
+
+jobs:
+  rustdoc:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: hecrj/setup-rust-action@master
+      - uses: actions/checkout@master
+      - name: Build docs
+        run: |
+          cargo doc --all-features --no-deps
```
raphaelrobert, line 16: While openmls is still under heavy development, it could be interesting to use

`--document-private-items`

as well.

franziskuskiefer: Yeah, but I wouldn't want it on the same page. Maybe let's set up two different directories. One with private and with the public API only.

raphaelrobert: Sure, that works too.

franziskuskiefer: ok, not much difference here right now and not worth the additional effort. I added private functions.

raphaelrobert (approved):

PR #195, opened 2020-11-04
Use static ciphersuite references from the config.

franziskuskiefer:
This moves the ciphersuite to config and only pulls it of there, in particular, Ciphersuite::new is forbidden outside of the config.
Instead Config::ciphersuite(name) is used to get a static reference to a ciphersuite object.
* public APIs: use CiphersuiteName
* internally use &'static Ciphersuite for members and args.
* convert CiphersuiteName to Ciphersuite in first public function
* functions using a CiphersuiteName return errors when the ciphersuite is not supported

cc #114 #85


franziskuskiefer:
I'm investigating the macOS issue (I can't reproduce this locally). But this shouldn't influence the PR review (appears to be some weird dependency issue).

kkohbrok (approved):

raphaelrobert (approved):

PR #198, opened 2020-11-05
Add utilities to quickly generate test setups

kkohbrok:
This PR is based on earlier work by @franziskuskiefer and includes the means to create configurations for clients and groups, as well as a function to create a `TestSetup` from these configurations.

As a showcase, I've changed the padding test such that it makes use of the new testing setup.

I've also changed the visibility of some select functions, please let me know if you think there's a better way of handling things.

Not everything is super refined here, so please let me know if there is something I should improve. Keep in mind, though, that this is only meant for tests. I'm sure we'll know more once we've tried using this framework for other tests.


raphaelrobert:
This is still marked as 'draft'. You should mark it as 'ready for review' before requesting a review, otherwise it's unclear if you are still going to make changes to it.

I'm hesitant to make more fields of the group public. If they are needed in integration tests, then the integration tests might be better off as unit tests. Integration tests are there to test the public API primarily.

kkohbrok:
Ah, sorry. I thought GitHub was smart enough to send out the review requests only after it was marked "Ready for Review".

src/group/mls_group/mod.rs
```diff
@@ -288,11 +288,11 @@ impl MlsGroup {
     }
 
     /// Get the ciphersuite implementation used in this group.
-    pub(crate) fn ciphersuite(&self) -> &Ciphersuite {
```
kkohbrok, line 291: I think it's not a bad idea to expose the ciphersuite of a given group to the outside. For example, an application might want to figure out if any of its groups are using deprecated ciphersuites.

src/group/mls_group/mod.rs
```diff
@@ -288,11 +288,11 @@ impl MlsGroup {
...
+    pub fn ciphersuite(&self) -> &Ciphersuite {
         self.ciphersuite
     }
 
-    pub(crate) fn context(&self) -> &GroupContext {
```
kkohbrok, line 295: Getting the context is (as far as I know) the only way to obtain the `GroupId`. I think the outside application should have some way identifier for the group and it might as well be the `GroupId`. This is probably going to be needed in other cases as well, for example if a client wants to let the server know what groups they are a member of.

src/key_packages/mod.rs
```diff
@@ -119,7 +119,7 @@ impl KeyPackage {
     }
 
     /// Compute the hash of the encoding of this key package.
-    pub(crate) fn hash(&self) -> Vec<u8> {
```
kkohbrok, line 122: The client needs to know the hash of a given `KeyPackage`, for example when communicating with the key store (or wherever the `KeyPackages` are uploaded`, so I think it's a good idea to expose it.

src/messages/mod.rs
```diff
@@ -331,7 +331,7 @@ impl Welcome {
     }
 
     /// Get a reference to the ciphersuite in this Welcome message.
-    pub(crate) fn get_secrets_ref(&self) -> &[EncryptedGroupSecrets] {
```
kkohbrok, line 334: The `EncryptedGroupSecrets` contain the hashes of the `KeyPackages`, which we need to determine which `KeyPackageBundle` we need to decrypt and process it. Again, we either need to expose it or expose a function that allows us to handle this case in some other way.

src/group/mls_group/mod.rs
raphaelrobert: Ideally there would be getters for the fields of the group context so that they don't have to be public.

tests/utils/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+                    .borrow_mut()
+                    .remove(kpb_position);
+                // Create the local group state of the new member based on the
+                // Welcome.
+                let new_group =
+                    MlsGroup::new_from_welcome(welcome.clone(), None, key_package_bundle).unwrap();
+                new_group_member
+                    .group_states
+                    .borrow_mut()
+                    .insert(new_group.context().group_id.clone(), new_group);
```
raphaelrobert, line 224: `new_group.get_group_id()` would really be nicer here :)

tests/utils/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+                    .position(|y| y.get_key_package().hash() == member_secret.key_package_hash)
+                    .unwrap();
+                let key_package_bundle = new_group_member
+                    .key_package_bundles
+                    .borrow_mut()
+                    .remove(kpb_position);
+                // Create the local group state of the new member based on the
+                // Welcome.
+                let new_group =
+                    MlsGroup::new_from_welcome(welcome.clone(), None, key_package_bundle).unwrap();
```
raphaelrobert, line 220: New members need a ratchet tree here, since we don't use the Ratchet Tree Extension.

tests/utils/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+                new_group_member
+                    .group_states
+                    .borrow_mut()
+                    .insert(new_group.context().group_id.clone(), new_group);
+            }
+            // Make all members receive and process the commit message.
```
raphaelrobert, line 226: This Commit message was issued before the members joind. Just processing the Welcome message brings them up to speed, there is no need to replay the Commit message.

tests/utils/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+pub(crate) struct TestClient {
+    pub(crate) credential_bundles: HashMap<CiphersuiteName, CredentialBundle>,
+    pub(crate) key_package_bundles: RefCell<Vec<KeyPackageBundle>>,
+    pub(crate) group_states: RefCell<HashMap<GroupId, MlsGroup>>,
+}
+
+/// The state of a test setup, including the state of the clients and the
+/// keystore, which holds the KeyPackages published by the clients.
+pub(crate) struct TestSetup {
+    pub(crate) key_store: RefCell<HashMap<(&'static str, CiphersuiteName), Vec<KeyPackage>>>,
```
raphaelrobert, line 48: This seems unused right now.

raphaelrobert (changes requested):
The `setup` function needs at least one test. I found some place where it would break already now, but there might be more.

tests/utils/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+            key_store.insert((client.name, ciphersuite), Vec::new());
+            // Create a credential_bundle for the given ciphersuite.
+            let credential_bundle = CredentialBundle::new(
+                client.name.as_bytes().to_vec(),
+                CredentialType::Basic,
+                ciphersuite,
+            )
+            .unwrap();
+            // Create a number of key packages.
+            for _ in 1..KEY_PACKAGE_COUNT {
```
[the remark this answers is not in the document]
kkohbrok: Argh, I changed that with one of the last commits. Good point.

tests/utils/mod.rs
kkohbrok: It is right now, but in future tests, the tester will probably want to access the `KeyPackage`s published by the parties. So it should be used in the future.

kkohbrok:
I'm not sure we actually need tests here. Each test that we implement with this testing framework should be a test in and of itself, shouldn't it?

tests/utils/mod.rs
kkohbrok: I thought the process was that new members first get the Welcome message, which would then enable them to process their own Add commit. As depicted [here](https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#section-4).

tests/utils/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+                    .get(group_config.members[client_id].name)
+                    .unwrap()
+                    .borrow_mut();
+                // Figure out which key package bundle we should use. This is
+                // a bit ugly and inefficient.
+                let member_secret = welcome
+                    .get_secrets_ref()
+                    .iter()
+                    .find(|x| {
+                        match new_group_member
```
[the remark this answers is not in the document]
raphaelrobert: Sorry, one more iteration: instead of using `.find(y.get_key_package().hash() == x.key_package_hash).is_some()` this is even more succinct:

`.any(|y| y.get_key_package().hash() == x.key_package_hash)`

tests/utils/mod.rs
raphaelrobert: Then either the spec is wrong or openmls is wrong, let's figure out which on it is.

tests/utils/mod.rs
raphaelrobert: Sure, that's fine. But it triggers a clippy warning so you should at least prefix it with an underscore.

raphaelrobert:
> I'm not sure we actually need tests here. Each test that we implement with this testing framework should be a test in and of itself, shouldn't it?

There is no way to know if the code you wrote behaves correctly if you don't test it. We always need tests for any code. In scenarios where unit tests are difficult to write, it might be ok to go for integration tests only. Here however it's very easy to test this with a unit test:

```
#[test]
fn test_setup() {
    let test_client_config_a = TestClientConfig {
        name: "TestClientConfigA",
        ciphersuites: vec![CiphersuiteName::MLS10_128_DHKEMX25519_AES128GCM_SHA256_Ed25519],
    };
    let test_client_config_b = TestClientConfig {
        name: "TestClientConfigB",
        ciphersuites: vec![CiphersuiteName::MLS10_128_DHKEMX25519_AES128GCM_SHA256_Ed25519],
    };
    let group_config = GroupConfig::default();
    let test_group_config = TestGroupConfig {
        ciphersuite: CiphersuiteName::MLS10_128_DHKEMX25519_AES128GCM_SHA256_Ed25519,
        config: group_config,
        members: vec![test_client_config_a.clone(), test_client_config_b.clone()],
    };
    let test_setup_config = TestSetupConfig {
        clients: vec![test_client_config_a, test_client_config_b],
        groups: vec![test_group_config],
    };
    let _test_setup = setup(test_setup_config);
}
```

The test is not successful, so there is a high likelihood something is wrong in the `setup` function 😄

tests/utils/mod.rs
kkohbrok: The spec is ambiguous. In [Welcoming New Members](https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#section-11.2.1), it says:
> The sender of a Commit message is responsible for sending a Welcome message to any new members added via Add proposals. The Welcome message provides the new members with the current state of the group, after the application of the Commit message. The new members will not be able to decrypt or verify the Commit message, but will have the secrets they need to participate in the epoch initiated by the Commit message.

So you're right and I don't need to send the Commit messages to the new members, but Section 4 of the Spec is wron.g.

kkohbrok:
After some debugging using your `setup_test` function, I think it's not the `setup` function that's at fault here, but the fact that there's no `decode` implemented for `PathSecret`, which we need in `GroupSecrets.decode(...)`.

kkohbrok:
I still think that any test we implement using this test framework is a test of the test framework as well as a test of whatever it is that it is testing. So your test function certainly helped, but we would have noticed the same thing as soon as we started using the `setup` function for other tests, wouldn't we?

raphaelrobert (approved):
> I still think that any test we implement using this test framework is a test of the test framework as well as a test of whatever it is that it is testing. So your test function certainly helped, but we would have noticed the same thing as soon as we started using the setup function for other tests, wouldn't we?

Absolutely. Whoever would have picked this up would have run into the same problem and that's what we have to avoid because you make it *their* responsibility to debug *your* code. You should at least take some reasonable measure to make sure your code works and the bare minimum is a positive test with default values. If you're really eager you would include a negative test, or tests with edge-cases. Another alternative is to file an issue to do the tests later, but for such a small and self-contained piece of code it is preferable to write tests right away. Now these are the best practices, I heard that in IRL shortcuts are sometimes taken.

kkohbrok:
I understand what you're saying and I don't want to dodge the responsibility for my code. However, I think the way to go would have been to pick another test to translate to use the `setup` function, i.e. one that make a bit more thorough use of `setup`. Other than the padding test, that is. That would make that test a test for the `setup` function.

Issue #196, opened 2020-11-04
Community


franziskuskiefer:
The repository needs to be updated to be accessible to the community:
- [x] Update Readme.md
  - [x] Build and test instructions
  - [x] Supported platforms
  - [x] License
- [x] Update Contributing.md
- [ ] ~~Add mailing list and messaging links~~
- [x] Add CODEWONERS and change branch protection accordingly
- [x] Add CODE_OF_CONDUCT.md
- [x] Set up a distribution list
- [ ] ~~Set up a Contributor License Agreement & bot~~

PR #201, opened 2020-11-10
Add metadata to Cargo.toml

raphaelrobert:
This PR adds a few more fields to Cargo.toml.


Cargo.toml
```diff
@@ -3,6 +3,11 @@ name = "openmls"
 version = "0.2.0"
 authors = ["Raphael Robert"]
```
franziskuskiefer, line 4: Authors should change to something like "The openmls developers <some@email.com>"

Cargo.toml
```diff
@@ -3,6 +3,11 @@ name = "openmls"
...
 edition = "2018"
+description = "This is a PoC Rust implementation of the Messaging Layer Security (MLS) protocol based on draft 9+."
```
franziskuskiefer, line 6: I'd not call it PoC but something like WiP or whatever. But it's supposed to be a production library. It's just not there yet.

Cargo.toml
raphaelrobert: Agreed, as I'll change it to that as soon as we have a DL set up.

franziskuskiefer (approved):

Cargo.toml
raphaelrobert: I also added that to #196.

Issue #164, opened 2020-10-26
Auto generate lifetime extension with KeyPackageBundle


raphaelrobert:
According to the spec, both the capabilities extension and the lifetime extension are mandatory. Right now, only the capabilities extension is auto generated.
The default lifetime should be specified in the config of the group.

PR #203, opened 2020-11-10
Add encoding/decoding tests for MLSPlaintexts, MLSCiphertexts and Welcome messages

### LINKED ISSUE #130: Test encoding/decoding
Test encoding and decoding of everything that goes over the wire.

kkohbrok:
This PR fixes #130 and is close to fixing #164.

In the course of fixing some broken code, I added some code checking the `extensions` used to create new `KeyPackageBundle`s, such that a `CapabilitesExtension` and `LifetimeExtension` is now always added. This does not quite solve #164, because I didn't know where to define a default lifetime for the lifetime extension. Let me know what your thoughts are on this, so I can implement `default()` for `LifetimeExtension` and the problem should be fixed.

I also added `Welcome` to the prelude, as it does go over the Wire and external applications might want to decode it.


src/key_packages/mod.rs
```diff
@@ -274,22 +275,69 @@ impl KeyPackageBundle {
     pub fn new_with_keypair(
         ciphersuites: &[CiphersuiteName],
         credential_bundle: &CredentialBundle,
-        extensions: Vec<Box<dyn Extension>>,
+        mut extensions: Vec<Box<dyn Extension>>,
         key_pair: HPKEKeyPair,
         leaf_secret: Vec<u8>,
     ) -> Result<Self, ConfigError> {
         debug_assert!(!ciphersuites.is_empty());
```
franziskuskiefer, line 282: We should return an error `if ciphersuite.is_empty()`.

tests/test_encoding.rs
```diff
@@ -0,0 +1,153 @@
...
+            .unwrap();
+        for _ in 0..100 {
+            // Test encoding/decoding of Application messages.
+            let message = randombytes(random_usize() % 1000);
+            let aad = randombytes(random_usize() % 1000);
+            let encrypted_message =
+                group_state.create_application_message(&aad, &message, &credential_bundle);
+            let encrypted_message_bytes = encrypted_message.encode_detached().unwrap();
+            let encrypted_message_decoded =
+                match MLSCiphertext::decode(&mut Cursor::new(&encrypted_message_bytes)) {
```
franziskuskiefer, line 55: Ugh, I finally have rewrite the `Codec` 😬

tests/test_encoding.rs
```diff
@@ -0,0 +1,153 @@
...
+                    Err(err) => panic!("Error decoding MLSCiphertext: {:?}", err),
+                };
+            assert_eq!(encrypted_message, encrypted_message_decoded);
+        }
+        // Test encoding/decoding of Proposal messages
+
+        // Updates
+
+        // TODO: Leaving out Extensions completely still yields a decoding
+        // error! Should be solved by #164.
```
franziskuskiefer, line 66: Please add a test for this and mark it as `#[should_panic]`.

tests/test_encoding.rs
```diff
@@ -0,0 +1,153 @@
...
+            Ok(a) => a,
+            Err(err) => panic!("Error decoding MPLSPlaintext Update: {:?}", err),
+        };
+
+        assert_eq!(update, update_decoded);
+
+        // Adds
+        let capabilities_extension = Box::new(CapabilitiesExtension::default());
+        let lifetime_extension = Box::new(LifetimeExtension::new(60));
+        let mandatory_extensions: Vec<Box<dyn Extension>> =
```
franziskuskiefer, line 94: You made the same before. Just clone that one.

tests/test_encoding.rs
```diff
@@ -0,0 +1,153 @@
...
+            group_state.create_remove_proposal(&[], credential_bundle, LeafIndex::from(1u32));
+        let remove_encoded = remove.encode_detached().unwrap();
+        let remove_decoded = match MLSPlaintext::decode(&mut Cursor::new(&remove_encoded)) {
+            Ok(a) => a,
+            Err(err) => panic!("Error decoding MPLSPlaintext Remove: {:?}", err),
+        };
+
+        assert_eq!(remove, remove_decoded);
+
+        // Commits
```
franziskuskiefer, line 127: All these different messages should go into their own tests. Move the setup above into a separate function and add individual test functions for the different tests.

franziskuskiefer (changes requested):
Generally looks good. But a couple comments inline.

> This does not quite solve #164, because I didn't know where to define a default lifetime for the lifetime extension.

The `Config` should have a default lifetime for the library. Later we should also add a per-group default. But that requires the group config, which we don't really have yet.

src/framing/mod.rs
```diff
@@ -138,7 +138,7 @@ impl Codec for MLSPlaintext {
     }
 }
 
-#[derive(Clone)]
+#[derive(Debug, PartialEq, Clone)]
```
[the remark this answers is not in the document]
kkohbrok: I currently use it to compare the result of encoding -> decoding with the original object. Is there a better way of doing this?

src/key_packages/mod.rs
```diff
@@ -274,22 +275,69 @@ impl KeyPackageBundle {
...
+        //        return Err(ConfigError::DuplicateExtension);
+        //    }
+        //}
+
+        // Check if the `extensions` already contain the mandatory extensions
+        // and in case one of them is a capabilities extension, add the input
+        // ciphersuites. If not, add the default versions of the mandatory
+        // extensions.
+
+        // First, check if one of the input extensions is a capabilities
```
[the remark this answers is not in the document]
kkohbrok: Is there something you would do differently here if the Vector was sorted?

src/key_packages/mod.rs
```diff
@@ -274,22 +275,69 @@ impl KeyPackageBundle {
...
+            return Err(ConfigError::DuplicateExtension);
+        }
+        //for ext in &extensions {
+        //    if extensions.iter().any(|e|  e.get_type() == ext.get_type()) {
+        //        return Err(ConfigError::DuplicateExtension);
+        //    }
+        //}
+
+        // Check if the `extensions` already contain the mandatory extensions
+        // and in case one of them is a capabilities extension, add the input
```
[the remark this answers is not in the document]
kkohbrok: Would it then not be easier to just remove the `ciphersuites` from the API and just read the ciphersuites off the capabilities extension? And if we have none, we can still take the default one...

src/key_packages/mod.rs
kkohbrok: We should, but we might not need `ciphersuites` in the first place. See my reply above.

src/key_packages/mod.rs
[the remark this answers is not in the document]
franziskuskiefer: Sure that's an option as well as long as the capabilities extension is easy enough to assemble. I can see that ciphersuites is for example one people might want to change often but leave other capabilities as defaults. Maybe the capabilities extension should have a builder pattern for creation or allow modifying parts.

src/key_packages/mod.rs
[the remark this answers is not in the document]
franziskuskiefer: You wouldn't need to search but can just walk through. With the things you want to do above sorting sounds easier to me than searching in the vector all the time.

tests/test_encoding.rs
raphaelrobert: We used to have `decode_detached()` for that purpose

raphaelrobert:
Looking good in general! In theory this should cover all types that are needed for interoperability and therefore you could go ahead and delete all the commented out `decode` functions.

src/key_packages/mod.rs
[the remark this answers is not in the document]
kkohbrok: I'm not sure what you mean by "walk through". Do you mean it improves the efficiency of what we're already doing here? Or do you suggest I do anything different? If that is the case, I'm not sure what I should change.

src/key_packages/mod.rs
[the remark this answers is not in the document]
kkohbrok: That would be handy. You had already suggested that you wanted to rewrite Extensions at some point. Maybe this would be one issue to tackle at that point? For now, I'll just do what you propose initially and fail if the capabilities and the ciphersuites don't match.

src/key_packages/mod.rs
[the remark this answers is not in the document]
franziskuskiefer: It's only about efficiency. This vector will never be super long. So not too important, but I think it will simplify your code.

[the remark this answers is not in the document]
kkohbrok: Sorry, I'm still confused. How will it simplify my code?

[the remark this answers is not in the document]
franziskuskiefer: 😄 well, if your first extension is a `Lifetime` you know that there's no `Capabilities`.

src/key_packages/mod.rs
franziskuskiefer: Cann you add this as long as the `ciphersuite` is still there?

src/key_packages/mod.rs
```diff
@@ -267,29 +267,65 @@ impl KeyPackageBundle {
...
     ) -> Result<Self, ConfigError> {
         debug_assert!(!ciphersuites.is_empty());
-        let capabilities_extension = CapabilitiesExtension::new(None, Some(ciphersuites), None);
-        let mut final_extensions: Vec<Box<dyn Extension>> = vec![Box::new(capabilities_extension)];
 
+        // Remove any duplicate extensions
+        extensions.sort();
+        extensions.dedup();
+
+        // Check if the `extensions` already contain the mandatory extensions
```
franziskuskiefer, line 288: This comment appears to be outdated.

src/key_packages/mod.rs
```diff
@@ -267,29 +267,65 @@ impl KeyPackageBundle {
...
-        extensions: Vec<Box<dyn Extension>>,
+        mut extensions: Vec<Box<dyn Extension>>,
         key_pair: HPKEKeyPair,
         leaf_secret: Secret,
     ) -> Result<Self, ConfigError> {
         debug_assert!(!ciphersuites.is_empty());
-        let capabilities_extension = CapabilitiesExtension::new(None, Some(ciphersuites), None);
-        let mut final_extensions: Vec<Box<dyn Extension>> = vec![Box::new(capabilities_extension)];
 
+        // Remove any duplicate extensions
```
franziskuskiefer, line 284: Hm, so this is an API decision we have to take here. Right now we simply remove duplicates and continue with the lucky extension that we didn't delete. I think this is a little too random for the caller.
I think we should return an error if there is a duplicate in there. So something like 
```
let extensions_len = extensions.len();
extensions.sort();
extensions.dedup();
if extensions.len() != extensions_len {
  return Err(ConfigError::InvalidExtensions);
}
```

src/key_packages/mod.rs
```diff
@@ -267,29 +267,65 @@ impl KeyPackageBundle {
...
+                None,
+            ))),
+        };
+
+        // Check if there is a lifetime extension. If not, add the default one.
+        if !extensions
+            .iter()
+            .any(|e| e.get_type() == ExtensionType::Lifetime)
+        {
+            extensions.push(Box::new(LifetimeExtension::default()));
```
franziskuskiefer, line 324: `Default` is derived here for `LifetimeExtensions`, so it will be `0` and hence an invalid lifetime.

tests/test_encoding.rs
```diff
@@ -0,0 +1,362 @@
...
+            Ok(a) => a,
+            Err(err) => panic!("Error decoding Welcome message: {:?}", err),
+        };
+
+        assert_eq!(welcome, welcome_decoded);
+    }
+}
+
+// TODO: Leaving out Extensions completely still yields a decoding
+// error! Should be solved by #164.
```
franziskuskiefer, line 304: This is probably also because the lifetime extension is invalid (see comment above).

src/key_packages/mod.rs
franziskuskiefer, line 284: In any case, this case needs a test case. Not sure if there's one.

franziskuskiefer (changes requested):
Getting there but I think there's a bug in `new_with_keypair` and a an API issue.

kkohbrok:
Since KeyPackages are now provisioned with the mandatory extensions, I have deleted a number of tests that were testing for that failure.

src/key_packages/test_key_packages.rs
```diff
@@ -35,6 +31,15 @@ fn generate_key_package() {
         .unwrap();
         std::thread::sleep(std::time::Duration::from_secs(1));
         assert!(kpb.get_key_package().verify().is_err());
+
+        // Now with two lifetime extensions, the key package should be valid.
```
franziskuskiefer, line 35: "should be invalid"

src/key_packages/mod.rs
```diff
@@ -251,7 +251,10 @@ impl KeyPackageBundle {
         extensions: Vec<Box<dyn Extension>>,
         leaf_secret: Secret,
     ) -> Result<Self, ConfigError> {
-        debug_assert!(!ciphersuites.is_empty());
+        if ciphersuites.is_empty() {
+            return Err(ConfigError::NoCiphersuitesSupplied);
```
franziskuskiefer, line 255: Let's start logging errors (same below).

franziskuskiefer (changes requested):
Thanks! Two nits (typo and logging) then we're good to go.

franziskuskiefer (dismissed):
Thanks!

franziskuskiefer (approved):

Issue #142, opened 2020-10-22
Logging


raphaelrobert:

franziskuskiefer:
Closing for now. We'll be adding more logging along the way.

PR #209, opened 2020-11-12
Add logging example

franziskuskiefer:
A simple example of how to use logging.
This is only an example of how to do this. Please start adding logging information to new code paths. I'll file a follow up with some more logging for existing code.

The logging level is controlled by the `RUST_LOG` env variable.

cc #142

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -7,6 +7,7 @@ use std::convert::*;
 pub enum CodecError {
     EncodingError,
     DecodingError,
+    Other,
 }
 
 impl From<ConfigError> for CodecError {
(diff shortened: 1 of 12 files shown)


src/codec.rs
```diff
@@ -20,8 +20,7 @@ impl From<Error> for CodecError {
     // TODO: tbd in #83, also this direction shouldn't be necessary.
     fn from(e: Error) -> CodecError {
         match e {
-            Error::DecodingError => CodecError::DecodingError,
-            Error::UnsupportedCiphersuite => CodecError::DecodingError,
+            _ => CodecError::DecodingError,
```
kkohbrok, line 23: Couldn't this also be an `EncodingError`?

kkohbrok (approved):

src/codec.rs
```diff
@@ -20,8 +20,7 @@ impl From<Error> for CodecError {
...
         }
     }
```
raphaelrobert, line 25: clippy warning

raphaelrobert (changes requested):
One clippy warning, the rest looks good!

src/codec.rs
franziskuskiefer: This direction really shouldn't exist.. I guess I'll start on #83 at some point ... this getting annoying.

raphaelrobert (approved):

franziskuskiefer:
I fixed the rebase issue to make the compiler happy. But I also started working on #83 to finally get error handling in better shape. This is getting annoying.

raphaelrobert (approved):

PR #211, opened 2020-11-13
Managed group API

raphaelrobert:
This PR introduces an API for managed groups.
The actual implementation will follow in subsequent PRs, until #79 can be closed 
Details are subject to change during the implementation.
There will also be breaking changes as OpenMLS catches up with spec changes.


src/group/managed_group/mod.rs
```diff
@@ -0,0 +1,276 @@
+use crate::ciphersuite::*;
```
franziskuskiefer, line 1: Can we start to expand these global imports (when it's not too many)?

src/group/managed_group/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+use crate::codec::*;
+use crate::creds::*;
+use crate::errors::ConfigError;
+use crate::framing::{sender::*, *};
+use crate::group::*;
+use crate::key_packages::*;
+use crate::messages::{proposals::*, *};
+use crate::tree::node::*;
+
+/// A `ManagedGroup` represents an `MLSGroup` with an easier, high-level API designed to be used in production. The API
```
franziskuskiefer, line 11: Some of the description should probably go into the module doc in the end.

franziskuskiefer, line 11: We should probably put the other API behind a feature flag later. Maybe mention that here already and make an issue.

src/group/managed_group/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+}
+
+impl ManagedGroup {
+    // === Group creation ===
+
+    /// Creates a new group from scratch with only the creator as a member.
+    pub fn new(
+        credential_bundle: &CredentialBundle,
+        managed_group_config: &ManagedGroupConfig,
+        group_id: GroupId,
```
franziskuskiefer, line 43: Could we choose the `group_id` randomly in here instead?

src/group/managed_group/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+        ciphersuite_name: CiphersuiteName,
```
franziskuskiefer, line 44: Shouldn't the ciphersuite be defined by the `key_package_bundle`?

src/group/managed_group/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+impl From<CodecError> for GroupError {
+    fn from(err: CodecError) -> GroupError {
+        GroupError::Codec(err)
+    }
+}
+
+/// Collection of callback functions that are passed to a `ManagedGroup` as part of the configurations
+/// Callback functions are optional. If no validator function is specified for a certain proposal type, any
+/// semantically valid proposal will be accepted.
+/// Validator fucntions returan a `bool`, epending on whether the proposal is accepted by the application policy.
```
franziskuskiefer, line 244: "epending"

src/group/managed_group/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+        Ok(ManagedGroup {
+            managed_group_config: managed_group_config.clone(),
+            group,
+            proposal_queue: ProposalQueue::new(),
+            own_kpbs: vec![],
+        })
+    }
+
+    /// Creates a new group with the creator and additional members
+    pub fn new_with_members(
```
franziskuskiefer, line 80: We could probably skip this in a first version, or just make it call `new` and `add_members` (so the API should be the same as `new` and `add_members` combined).

src/group/managed_group/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+        unimplemented!()
+    }
+
+    // === Membership management ===
+
+    /// Adds members to the group
+    pub fn add_members(
+        &mut self,
+        credential_bundle: &CredentialBundle,
+        aad: &[u8],
```
franziskuskiefer, line 97: Would it make sense to set `AAD` for the group to re-use it? It's nowhere specified how this is handled. But I'd probably build the `aad` from some group aad and maybe other information such as sender etc. This could make the `aad` here optional and such simpler. (It's already optional as it can be empty, but that's not clear right now.)

src/group/managed_group/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+    ) -> Result<Self, GroupError> {
+        unimplemented!()
+    }
+
+    // === Membership management ===
+
+    /// Adds members to the group
+    pub fn add_members(
+        &mut self,
+        credential_bundle: &CredentialBundle,
```
franziskuskiefer, line 96: Do we need a new `credential_bundle` here? Couldn't we use the one we already have in the group? I'd make updating the credential explicit in a function.

src/group/managed_group/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+        aad: &[u8],
+        key_packages: Vec<KeyPackage>,
```
franziskuskiefer, line 98: We should add some sort of client/member struct that we can use here to consistently identify clients/members. Then we can use the same type for add, remove, etc.

src/group/managed_group/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    /// Processes any incoming message from the DS (MLSPlaintext & MLSCiphertext) and triggers the corresponding callback functions
+    pub fn process_message(&mut self, message: MLSMessage) {
+        unimplemented!()
+    }
+
+    // === Create application messages ===
+
+    /// Create an application message
+    pub fn create_message(
```
franziskuskiefer, line 148: I'd expect this function to return a message.

src/group/managed_group/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    /// Returns a list of proposal
+    pub fn get_pending_proposals(&self) -> Vec<Proposal> {
+        unimplemented!()
+    }
+
+    // === Load & save (WIP) ===
+
+    /// Loads the state from persisted state
+    pub fn load(state: &[u8], managed_group_config: &ManagedGroupConfig) -> ManagedGroup {
```
franziskuskiefer, line 190: The state could get so big that we don't want to have it in a slice/vector. So this should probably either get a `dyn Read` or a path to read from or something similar.

src/group/managed_group/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+        unimplemented!()
+    }
+
+    /// Persists the state
+    pub fn save(&self, state: &mut Vec<u8>) {}
```
franziskuskiefer, line 195: Similar to the `load`, this should probably use some `dyn Write` or use path.

src/group/managed_group/mod.rs
raphaelrobert: Sure, though we should do it after everything is implemented.

src/group/managed_group/mod.rs
raphaelrobert: I was wondering about that. I think we should keep both, because it might well be that you cannot do everything you want with `MlsGroup`, but happy to discuss.

src/group/managed_group/mod.rs
raphaelrobert: Since group IDs have to be globally unique they should not be chosen by clients, otherwise you'd get conflicts.

src/group/managed_group/mod.rs
raphaelrobert: This goes back to the question whether you can clone private keys. If we want to store the credential bundle in the group, we need to clone keys and also persist them redundantly. I'd rather pass it as an argument whenever it is needed.

src/group/managed_group/mod.rs
raphaelrobert: Yes, that's an open question and I had the same idea. Right now you can only specify it for application messages, and you can also just read it when application messages are received. HS messages can also have an AAD, but it's being ignored.
One option would indeed to set it globally and modify it when needed (should it be persisted?). In addition, it also needs to be returned when HS messages are parsed, which is much trickier and doesn't fit into the API right now, unless we make `process_message()` return the AAD.

src/group/managed_group/mod.rs
raphaelrobert: It's tricky, because the amount of information varies:
 - Adds: You need a full KeyPackage
 - Remove: You just need an index
 - Uniquely referencing a member: You need a Credential

We could refactor remove to accept a credential instead of an index, but the problem with Adds remains.

src/group/managed_group/mod.rs
raphaelrobert: Absolutely, this is a total placeholder and should be replaced with whatever we'll use for persistence.

src/group/managed_group/mod.rs
raphaelrobert: Ok, I fixed this now I think.

src/group/managed_group/mod.rs
```diff
@@ -161,6 +150,28 @@ impl ManagedGroup {
         unimplemented!()
     }
 
+    // === Configuration ===
+
+    /// Gets the configuration
+    pub fn get_configuration(&self) -> &ManagedGroupConfig {
```
franziskuskiefer, line 156:
```suggestion
    pub fn configuration(&self) -> &ManagedGroupConfig {
```

src/group/managed_group/mod.rs
```diff
@@ -161,6 +150,28 @@ impl ManagedGroup {
...
+        unimplemented!()
+    }
+
+    /// Sets the configuration
+    pub fn set_configuration(&mut self, managed_group_config: &ManagedGroupConfig) {
+        unimplemented!()
+    }
+
+    /// Gets the AAD used in the framing
+    pub fn get_aad(&self) -> &[u8] {
```
franziskuskiefer, line 166:
```suggestion
    pub fn aad(&self) -> &[u8] {
```

franziskuskiefer (commented):
Ok I think this is a reasonable API to start with. We will probably find a couple things while implementing it that might not work as expected. But let's start and try to implement the functions and see how it goes.

All these clippy warnings make this pretty hard to read 😬

src/group/managed_group/mod.rs
raphaelrobert: I created #213

raphaelrobert:
> Ok I think this is a reasonable API to start with. We will probably find a couple things while implementing it that might not work as expected. But let's start and try to implement the functions and see how it goes.
> 
> All these clippy warnings make this pretty hard to read 😬

Let's make them all disappear in the implementation phase!

src/group/managed_group/mod.rs
raphaelrobert: I created #214 for that.

src/group/managed_group/mod.rs
raphaelrobert: I created https://github.com/openmls/openmls/issues/215 for that.

franziskuskiefer (approved):
Ok, let's get this landed and then implement it.

Please remove the clippy warnings before merging, otherwise we get a lot of noise here (I think it's all just prefixing arguments and removing unused imports).

franziskuskiefer:
You didn't fix the warnings 😞

PR #212, opened 2020-11-16
Fuzzing targets for MLSCiphertext and MLSPlaintext decode

kkohbrok:
This PR introduces new fuzzing targets for the `decode` functions of MLSCiphertext and MLSPlaintext and integrates them into the CI. Thus fixing #154.

It also fixes a few issues uncovered during initial fuzzing, such as missing decode functions, which were previously removed as part of #203. It also fixes a bug, where the `RatchetTreeExtension` is removed from the Welcome message prior to verifying its signature, which lead to the verification failing reliably if a `RatchetTreeExtension` is present.


franziskuskiefer (changes requested):
Thanks! Generally looks good.
But for each of the `decode`s you added back we should add a basic test that exercises it.

kkohbrok:
I found the easiest way to test the decoding is not via an extra test, but to set the `GroupConfig` in the encoding/decoding test for Welcome messages such that a `RatchetTreeExtension` is included. That covers all the new `decode` functions.

tests/test_encoding.rs
```diff
@@ -297,5 +301,15 @@ fn test_welcome_message_encoding() {
...
         assert_eq!(welcome, welcome_decoded);
+
+        let charly = test_clients.get("charly").unwrap().borrow();
+
+        let charly_key_package_bundle =
+            charly.find_key_package_bundle(&charly_key_package).unwrap();
+
+        // This makes Charly decode the internals of the Welcome message, for
+        // example the RatchetTreeExtension.
+        let _charly_group_state =
```
franziskuskiefer, line 312: You should do something with the result to actually have a test.

src/group/mls_group/new_from_welcome.rs
```diff
@@ -59,7 +59,13 @@ impl MlsGroup {
                 return Err(WelcomeError::DuplicateRatchetTreeExtension);
             }
             match extension.to_ratchet_tree_extension_ref() {
-                Ok(e) => Some(e.clone()),
+                Ok(e) => {
```
franziskuskiefer, line 62: I only expected changes in the test. What is this about?
If this is needed, the extension shouldn't be removed in the first place.

franziskuskiefer (changes requested):
Thanks for the tests. But what is this other change?

src/group/mls_group/new_from_welcome.rs
kkohbrok: Ah, good point. I explained it in the commit message, but I should have added it in the PR description as well.

Yeah, I didn't really know how to solve this nicely. The existing code checks for duplicates by removing the existing extension and then checking if there are still `any` left. So I had the option of either implementing a new deduplication strategy or just sticking with what is already there and adding it back afterwards...

franziskuskiefer: Sorry, I guess you should have looked at the commit message a little closer 😬 (I really don't like GH reviews...)
Ok let's keep it like this for now. This looks like something that could use a cleanup in any case later.

franziskuskiefer (approved):
Thanks! lgtm now

src/group/mls_group/new_from_welcome.rs
kkohbrok: No worries. I should have changed the PR description or at least made a comment in the thread.

PR #216, opened 2020-11-16
unified error handling 1/2

franziskuskiefer:
This is the first PR for unifying and cleaning up the error handling.
Each module now has a `errors.rs` that defines errors for this module.
Only errors from this file are allowed to be used in functions of this
module. The `errors.rs` takes care of converting other errors to errors
of the respective module.

Because this is already pretty big I stopped to split this in two.
Not all modules are converted yet.

cc: #83


franziskuskiefer:

I guess I stopped a little too late 😬

src/codec.rs
```diff
@@ -17,24 +17,11 @@ impl From<ConfigError> for CodecError {
...
 impl From<CodecError> for ConfigError {
     // TODO: tbd in #83, also this direction shouldn't be necessary.
     fn from(e: CodecError) -> Self {
         match e {
-            CodecError::DecodingError => ConfigError::DecodingError,
-            CodecError::EncodingError => ConfigError::DecodingError,
-            CodecError::Other => ConfigError::DecodingError,
+            _ => ConfigError::InvalidConfig,
         }
     }
```
raphaelrobert, line 26: Clippy warning

src/extensions/errors.rs
```diff
@@ -0,0 +1,151 @@
...
+
+#[derive(Debug, PartialEq, Clone, Copy)]
+#[repr(u16)]
+pub enum LifetimeExtensionError {
+    /// Invalid lifetime extensions.
+    Invalid = 0,
+
+    /// Lifetime extension is expired.
+    Expired = 1,
+}
```
raphaelrobert, line 54: Do we care about LifetimeExtensions that are not yet valid? (with a future time window)

src/extensions/errors.rs
```diff
@@ -0,0 +1,151 @@
...
+
+#[derive(Debug, PartialEq, Clone, Copy)]
+#[repr(u16)]
+pub enum CapabilitiesExtensionError {
+    /// Invalid capabilities extensions.
+    Invalid = 0,
+
+    /// Capabilities extension is missing a version field.
+    EmptyVersionsField = 1,
```
raphaelrobert, line 63: Could there be unsupported versions?

src/framing/mod.rs
```diff
@@ -101,7 +93,7 @@ impl MLSPlaintext {
     pub fn as_application_message(&self) -> Result<&[u8], MLSPlaintextError> {
```
raphaelrobert, line 93: (not the correct line)

Shouldn't `verify()` return an error as well instead of a `bool`?

src/group/managed_group/mod.rs
```diff
@@ -1,7 +1,6 @@
-use crate::ciphersuite::*;
 use crate::codec::*;
```
raphaelrobert, line 1: Why is this removed?

src/key_packages/errors.rs
```diff
@@ -0,0 +1,92 @@
...
+//! `KeyPackageBundle`s.
+
+use crate::{config::ConfigError, extensions::ExtensionError};
+use std::error::Error;
+
+#[derive(Debug, PartialEq, Clone, Copy)]
+#[repr(u16)]
+pub enum KeyPackageError {
+    /// An unknown extension error occurred.
+    UnknownExtensionError = 0,
```
raphaelrobert, line 13: What is the purpose of this error? It doesn't seem to be used anywhere.

raphaelrobert (changes requested):
Looking good in general! I commented on some specific parts.

src/extensions/errors.rs
franziskuskiefer: We currently don't tell the caller if an extension is expired or has another invalid date. This should get improved in future. But I didn't want to change the logic here in this PR.

src/extensions/errors.rs
franziskuskiefer: Yep, but nothing we check right now. I updated #83 to follow up on these.

src/framing/mod.rs
franziskuskiefer: It probably should, but again a logic change I wouldn't want to do in this PR (updated #83).

src/group/managed_group/mod.rs
franziskuskiefer: Because it's one of the many warnings in here ...

src/key_packages/errors.rs
franziskuskiefer: It is used when the `ExtensionError` is not an `InvalidExtensionType`. It should never happen, but theoretically can because the `match` in the respective from has to be exhaustive.

src/codec.rs
franziskuskiefer: This will be fixed in the next PR when this code is actually doing something useful.

franziskuskiefer:
@raphaelrobert I didn't change anything, only responded to your comments. Let me know if you want me to change any of them anyway.

src/ciphersuite/mod.rs
```diff
@@ -369,17 +349,14 @@ impl AeadKey {
         msg: &[u8],
         aad: &[u8],
         nonce: &AeadNonce,
-    ) -> Result<Vec<u8>, AEADError> {
-        let (ct, tag) = match aead_encrypt(
+    ) -> Result<Vec<u8>, AeadError> {
```
kkohbrok, line 352: Just from a naming standpoint it's not clear why an `AeadError` would not be a `CryptoError`. Maybe convert?

src/extensions/errors.rs
```diff
@@ -0,0 +1,151 @@
...
+            Self::InvalidExtensionType => {
+                "The requested extension type is not supported by OpenMLS."
+            }
+            Self::Capabilities(e) => match e {
+                CapabilitiesExtensionError::Invalid => "Error decoding a capabilities extensions.",
+                CapabilitiesExtensionError::EmptyVersionsField => {
+                    "The versions field in the extension is empty."
+                }
+                CapabilitiesExtensionError::UnsupportedCiphersuite => {
+                    "No supported cihpersuite in the extension."
```
kkohbrok, line 110:
```suggestion
                    "No supported ciphersuite in the extension."
```

src/extensions/errors.rs
```diff
@@ -0,0 +1,151 @@
...
+                }
+            },
+            Self::KeyPackageId(e) => match e {
+                KeyPackageIdError::Invalid => "Error decoding a key package id extensions.",
```
kkohbrok, line 114:
```suggestion
                KeyPackageIdError::Invalid => "Error decoding a key package id extension.",
```

src/extensions/errors.rs
```diff
@@ -0,0 +1,151 @@
...
+            },
+            Self::ParentHash(e) => match e {
+                ParentHashError::Invalid => "Error decoding a parent hash extensions.",
```
kkohbrok, line 117:
```suggestion
                ParentHashError::Invalid => "Error decoding a parent hash extension.",
```

src/extensions/errors.rs
```diff
@@ -0,0 +1,151 @@
...
+            },
+            Self::RatchetTree(e) => match e {
+                RatchetTreeError::Invalid => "Error decoding a ratchet tree extensions.",
```
kkohbrok, line 120:
```suggestion
                RatchetTreeError::Invalid => "Error decoding a ratchet tree extension.",
```

src/extensions/errors.rs
```diff
@@ -0,0 +1,151 @@
...
+            },
+            Self::Lifetime(e) => match e {
+                LifetimeExtensionError::Invalid => "Invalid lifetime extensions.",
```
kkohbrok, line 123:
```suggestion
                LifetimeExtensionError::Invalid => "Invalid lifetime extension.",
```

src/key_packages/errors.rs
```diff
@@ -0,0 +1,92 @@
...
+            Self::InvalidLifetimeExtension => {
+                "The lifetime extension of the key package is not valid."
+            }
+            Self::InvalidSignature => "The key package signature is not valid.",
+            Self::LibraryError => "An unknown OpenMLS library error occurred.",
+            Self::DuplicateExtension => "Duplicate extensions are not allowed.",
+            Self::NoCiphersuitesSupplied => {
+                "Creating a new key package requires at least one ciphersuite."
+            }
+            Self::InvalidCapabilitiesExtension => {
```
kkohbrok, line 87: I know you didn't pick the name here, but now that you've put a message next to it, I notice that the name of the error is a little more general than its description. Maybe we should change the name here to something like: `CiphersuiteMismatch` or similar?

kkohbrok (changes requested):

raphaelrobert:
> @raphaelrobert I didn't change anything, only responded to your comments. Let me know if you want me to change any of them anyway.

Thanks for documenting the missing checks.

One thing you should change for sure: You removed `pub type AppMessageReceived = fn(&ManagedGroup, &[u8], &[u8]);` in the managed API, please add it back.

Otherwise it looks good.

src/ciphersuite/mod.rs
franziskuskiefer: I don't want to wrap any crypto errors in a different crypto error (what would be the point?). `CryptoError` is really only for cases where something in the crypto library goes wrong that shouldn't go wrong. Everything else are "valid" cases that can occur.

src/key_packages/errors.rs
franziskuskiefer: Yeah the name isn't great. I'll change it.

franziskuskiefer:
> One thing you should change for sure: You removed `pub type AppMessageReceived = fn(&ManagedGroup, &[u8], &[u8]);` in the managed API, please add it back.

That's still there 😮 (https://github.com/openmls/openmls/blob/franziskus/error-handling/src/group/managed_group/mod.rs#L286)

kkohbrok (approved):
Thanks! This should make errors quite a bit more relevant in general.

raphaelrobert:
> > One thing you should change for sure: You removed `pub type AppMessageReceived = fn(&ManagedGroup, &[u8], &[u8]);` in the managed API, please add it back.
> 
> That's still there 😮 (https://github.com/openmls/openmls/blob/franziskus/error-handling/src/group/managed_group/mod.rs#L286)

Glitch on GH. It was marked as removed, but I couldn't comment on that specific line. Anyway... looks good then.

raphaelrobert (dismissed):

kkohbrok (approved):

raphaelrobert (approved):

PR #217, opened 2020-11-16
Cleanup

raphaelrobert:
This PR fixes the clippy warnings & introduces `rustfmt.toml`.

Note that `wrap_comments` is currently not stable, so once in a while nightly needs to do a pass on the code base.


rustfmt.toml
```diff
@@ -0,0 +1,4 @@
+comment_width = 80
+wrap_comments = true
+edition = "2018"
+format_code_in_doc_comments = true
```
franziskuskiefer, line 4: newline

rustfmt.toml
```diff
@@ -0,0 +1,4 @@
+comment_width = 80
+wrap_comments = true
```
franziskuskiefer, line 2: Hm so it's a nightly feature... This produces a lot of not so nice warnings.
But I'm actually more concerned about stability. How stable is this feature, i.e. how much churn does it produce?

franziskuskiefer (changes requested):
Generally looks good. But one question for the `wrap_comments`.

While you're at it can you also fix all the warnings in the `managed_group`?

rustfmt.toml
raphaelrobert: It seems to be quite stable, the only problem that was reported was when you use markup tables in comments. I guess for now we have to run it manually once in a while to clean up comments.

raphaelrobert:
> While you're at it can you also fix all the warnings in the `managed_group`?

That's already fixed.

franziskuskiefer (approved):
> > While you're at it can you also fix all the warnings in the `managed_group`?
> 
> That's already fixed.

Ah totally missed that, sorry 🙇 

Ok let's do this.

PR #218, opened 2020-11-18
Integration test for managed group API

raphaelrobert:
This PR adds a first integration test for the managed group API that is modeled after the integration test for `MlsGroup`.

The test is expected to fail because of the missing implementation.

The purpose of this PR is to refine the API in some places before starting to implement it.

### DIFF
--- a/src/framing/sender.rs
+++ b/src/framing/sender.rs
@@ -74,6 +74,9 @@ impl Sender {
     pub(crate) fn as_node_index(self) -> NodeIndex {
         NodeIndex::from(self.sender)
     }
+    pub fn as_leaf_index(self) -> LeafIndex {
+        LeafIndex::from(self.as_node_index())
+    }
 }
 
 impl Codec for Sender {
--- a/src/framing/sender.rs
+++ b/src/framing/sender.rs
@@ -71,11 +71,11 @@ impl Sender {
             sender,
         }
     }
-    pub(crate) fn as_node_index(self) -> NodeIndex {
+    pub(crate) fn to_node_index(self) -> NodeIndex {
         NodeIndex::from(self.sender)
     }
-    pub fn as_leaf_index(self) -> LeafIndex {
-        LeafIndex::from(self.as_node_index())
+    pub fn to_leaf_index(self) -> LeafIndex {
+        LeafIndex::from(self.to_node_index())
     }
 }
 
--- a/src/group/managed_group/mod.rs
+++ b/src/group/managed_group/mod.rs
@@ -1,33 +1,35 @@
 use crate::codec::*;
 use crate::config::ConfigError;
-use crate::creds::*;
+use crate::creds::{Credential, CredentialBundle};
 use crate::framing::{sender::*, *};
 use crate::group::*;
-use crate::key_packages::*;
-use crate::messages::{proposals::*, *};
-use crate::tree::node::*;
+use crate::key_packages::{KeyPackage, KeyPackageBundle};
+use crate::messages::{proposals::*, Welcome};
+use crate::tree::index::LeafIndex;
+use crate::tree::node::Node;
 
+use std::error::Error;
 use std::io::{Read, Write};
 
 /// A `ManagedGroup` represents an `MLSGroup` with an easier, high-level API
 /// designed to be used in production. The API exposes high level functions to
 /// manage a group by adding/removing members, get the current member list, etc.
 ///
-/// The API is modelled such that it can serve as a direct interface to the
-/// Delivery Serive. Functions that modify the public state of the group will
+/// The API is modeled such that it can serve as a direct interface to the
+/// Delivery Service. Functions that modify the public state of the group will
 /// return a `Vec<MLSMessage>` that can be sent to the Delivery Service
 /// directly. Conversely, incoming messages from the Delivery Service can be fed
 /// into `process_nessage()`.
 ///
 /// A `ManagedGroup` has an internal queue of pending proposals that builds up
 /// as new messages are processed. When creating proposals, those messages are
-/// not automatically appended to this queue, instead they have be processed
+/// not automatically appended to this queue, instead they have to be processed
 /// again through `process_message()`. This allows the Delivery Service to
 /// reject them (e.g. if they reference the wrong epoch).
 ///
-/// If incoming messages or applied operations are semantically or syntactily
-/// incorrect, the function will return a corresponding error message and the
-/// state of the group will remain unchanged.
+/// If incoming messages or applied operations are semantically or syntactically
+/// incorrect, a callback function will be called with a corresponding error
+/// message and the state of the group will remain unchanged.
 ///
 /// The application policy for the group can be enforced by implementing the
 /// validator callback functions and selectively allowing/ disallowing each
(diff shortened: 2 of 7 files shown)


src/group/managed_group/mod.rs
```diff
@@ -92,17 +94,15 @@ impl ManagedGroup {
     pub fn add_members(
         &mut self,
         credential_bundle: &CredentialBundle,
-        aad: &[u8],
```
kkohbrok, line 95: Why is aad removed here, as well as in `remove_members`, but not in `propose_add_members`? I think `aad` is something that applications might want to use, so why not expose it?

kkohbrok (approved):
Looks good with one comment.

src/group/managed_group/mod.rs
```diff
@@ -214,6 +213,30 @@ impl ManagedGroup {
 
     /// Persists the state
     pub fn save(&self, writer: Box<dyn Write>) {}
+
+    // === Interface for callbacks ===
+
+    /// Get group ID
+    pub fn group_id(&self) -> GroupId {
```
franziskuskiefer, line 220: Should this return a reference?
```suggestion
    pub fn group_id(&self) -> &GroupId {
```

src/framing/sender.rs
```diff
@@ -74,6 +74,9 @@ impl Sender {
     pub(crate) fn as_node_index(self) -> NodeIndex {
         NodeIndex::from(self.sender)
     }
+    pub fn as_leaf_index(self) -> LeafIndex {
```
franziskuskiefer, line 77: This has been bad in the name above as well, but these functions should be `to_leaf_index` (or `into_` as it's not really expensive). `as_` implies borrows.
But let's not do it in this PR.

src/group/managed_group/mod.rs
raphaelrobert: I forgot to remove it there as well.
AAD is important but will probably not be used often. It can be set with `.set_aad()` instead, so that the API is cleaner.

src/prelude.rs
```diff
@@ -1,9 +1,11 @@
-//! Prelude for openmls.
-//! Include this to get access to all the public functions of openmls.
+//! Prelude for OpenMLS.
```
franziskuskiefer, line 1: We have to clean up the prelude (to only contain the good and necessary stuff) and API (all the public functions). I'll do that after doing the errors.

tests/test_managed_group.rs
```diff
@@ -0,0 +1,341 @@
...
+///  - Bob updates and commits
+///  - Alice updates and commits
+///  - Bob adds Charlie
+///  - Charlie sends a message to the group
+///  - Charlie updates and commits
+///  - Charlie removes Bob
+#[test]
+#[should_panic]
+fn managed_group_operations() {
+    for ciphersuite in Config::supported_ciphersuites() {
```
franziskuskiefer, line 140: Ok for a first version. But we should start making these individual tests, i.e. put the test in a macro and call it for all ciphersuites. Otherwise we'll get test failures that are impossible to debug.

tests/test_managed_group.rs
```diff
@@ -0,0 +1,341 @@
...
+
+        let bob_key_package_bundle =
+            KeyPackageBundle::new(&[ciphersuite.name()], &bob_credential_bundle, vec![]).unwrap();
+        let bob_key_package = bob_key_package_bundle.get_key_package().clone();
+
+        // Define the managed group configuration
+
+        let update_policy = UpdatePolicy::default();
+        let callbacks = ManagedGroupCallbacks::new(
+            Some(validate_add),
```
franziskuskiefer, line 162: We should move all these functions to a util module. We'll be re-using them across tests I guess.

tests/test_managed_group.rs
```diff
@@ -0,0 +1,341 @@
...
+            Some(validate_remove),
+            Some(member_added),
+            Some(member_removed),
+            Some(member_updated),
+            Some(app_message_received),
+            Some(invalid_message_received),
+        );
+        let managed_group_config = ManagedGroupConfig::new(false, update_policy, callbacks);
```
franziskuskiefer, line 170: What's this `false`?

tests/test_managed_group.rs
```diff
@@ -0,0 +1,341 @@
...
+        let (queued_messages, welcome) =
+            match alice_group.add_members(&alice_credential_bundle, &[bob_key_package]) {
+                Ok((qm, welcome)) => (qm, welcome),
+                Err(e) => panic!("Could not add member to group: {:?}", e),
+            };
+
+        alice_group.process_messages(&queued_messages);
+
+        // Check that the group now has two members
+        assert_eq!(alice_group.get_members().len(), 2);
```
franziskuskiefer, line 191: Can you also check that Alice and Bob are the two members?

src/group/managed_group/mod.rs
kkohbrok: Wouldn't it be used to communicate with the DS? I could imagine that deployers will want some way of communicating with the DS and aad could be it. Also, `set_aad()` hints that aad is something that has to be set on a per-group level, rather than on a per-message level, which is a bit misleading.

tests/test_managed_group.rs
```diff
@@ -0,0 +1,341 @@
...
+            alice_group.export_ratchet_tree(),
+            bob_group.export_ratchet_tree(),
+        );
+        assert_eq!(
+            alice_group.export_ratchet_tree(),
+            charlie_group.export_ratchet_tree()
+        );
+
+        // Make sure the group only contains two members
+        assert_eq!(alice_group.get_members().len(), 2);
```
franziskuskiefer, line 339: Here we should also check who's in the group.

tests/test_managed_group.rs
```diff
@@ -0,0 +1,341 @@
...
+        // Make sure that all groups have the same public tree
+        assert_eq!(
+            alice_group.export_ratchet_tree(),
+            bob_group.export_ratchet_tree(),
+        );
+        assert_eq!(
+            alice_group.export_ratchet_tree(),
+            charlie_group.export_ratchet_tree()
+        );
+
```
franziskuskiefer, line 337: What happens if Bob now tries to send something to the group?

franziskuskiefer (changes requested):
Thanks, this API is already way better than the `MlsGroup` 🙂 
A couple issues inline.

Can you start adding comments to all the (public) functions? We'll need that in any case and better doing it now while writing it than having to do it later.

src/group/managed_group/mod.rs
raphaelrobert: That's precisely the design question. Since I'm assuming that AAD will be rarely used, in particular on a per-message basis, I'm stripping it from the functions that will be used often.
If implementation still want to use it, they can change the group's AAD with `.set_aad()`.
If we see that the AAD will be used somewhere we can always revert this decision, but right now we don't have any concrete example of where it would be useful.

tests/test_managed_group.rs
raphaelrobert: Good catch! I forgot to address that. The fix for now is that there is a callback `ErrorOccured` that throws `ManagedGroupError::UseAfterEviction` when this happens. This might not be optimal, the alternative would be to return an error immediately when a group is dead, but it makes the overall API quite ugly.

kkohbrok (approved):

kkohbrok (approved):

kkohbrok (approved):

PR #220, opened 2020-11-19
Add constants in the Config for the Lifetime of KeyPackages

### LINKED ISSUE #164: Auto generate lifetime extension with KeyPackageBundle

kkohbrok:
Fixes #164.

This PR introduces a new `Constants` field in the `Config`, which (for now) only includes some defaults for KeyPackage lifetimes. The PR also includes a small test, as well as some refactoring to make use of the constants.


src/config/mod.rs
```diff
@@ -11,6 +11,15 @@ use crate::extensions::ExtensionType;
 pub mod errors;
 pub(crate) use errors::ConfigError;
 
+/// This value is used as the default lifetime of `KeyPackage`s if no default
+/// lifetime is configured. The value is in seconds and amounts to 3 * 28 Days,
+/// i.e. about 3 months.
+const DEFAULT_KEY_PACKAGE_LIFETIME: u64 = 60 * 60 * 24 * 28 * 3; // in Seconds
+/// This value is sued as the default amount of time (in seconds) the lifetime
```
franziskuskiefer, line 18:
```suggestion
/// This value is used as the default amount of time (in seconds) the lifetime
```

src/config/mod.rs
```diff
@@ -103,6 +130,17 @@ impl Config {
             None => Err(ConfigError::UnsupportedCiphersuite),
         }
     }
+
+    /// Get the default `KeyPackage` lifetime (in seconds).
+    pub fn default_key_package_lifetime() -> &'static u64 {
```
franziskuskiefer, line 135: For primitives we should just return the value.

src/extensions/life_time_extension.rs
```diff
@@ -20,29 +20,29 @@
...
 
 impl LifetimeExtension {
-    /// Create a new lifetime extensions with lifetime `t`.
+    /// Create a new lifetime extensions with lifetime `t` (in seconds).
     /// Note that the lifetime is extended 1h into the past to adapt to skewed
     /// clocks.
     pub fn new(t: u64) -> Self {
-        // TODO: #85 make the margin configurable.
-        const LIFETIME_MARGIN: u64 = 60 * 60;
+        let lifetime_margin: u64 = Config::key_package_lifetime_margin().clone();
```
franziskuskiefer, line 40: This should go away when you change the API and drop the clone. `u64` is copyable.

tests/test_config.rs
```diff
@@ -57,3 +57,12 @@ fn default_ciphersuites() {
         supported_ciphersuites
     );
 }
+
+#[test]
+fn default_constants() {
+    // Make sure the supported ciphersuites are what we expect them to be.
+    let default_key_package_lifetime = Config::default_key_package_lifetime().clone();
```
franziskuskiefer, line 64: No clones needed.

franziskuskiefer (changes requested):
Thanks. lgtm with a couple nits.

src/config/mod.rs
kkohbrok: I was wondering about that. Will do.

raphaelrobert (changes requested):
Looks good except the clippy warning!

raphaelrobert (approved):

franziskuskiefer (approved):

Issue #207, opened 2020-11-11
Codec cleanup


kkohbrok:
Quite a few structs implement `Codec` although they don't have to. Two examples are `MlsGroup` and `EpochSecrets`.

I think (or I hope) that everything that needs codec either because it goes over the wire, it gets encrypted or it gets signed will make tests fail if a relevant `Codec` is missing.

kkohbrok:
I believe this was solved in #203.

kkohbrok:
I was wrong. There are still some unnecessary `encode` functions, for example for `EpochSecrets`.

franziskuskiefer:
@kkohbrok I guess we can close this by now?

kkohbrok:
Yup, this seems to be quite the relic.

PR #222, opened 2020-11-20
Improve FS by dropping the encryption_secret after use

### LINKED ISSUE #91: Enforce FS for secret tree
Make sure `encryption_secret` is deleted as soon as the secret tree is created.

kkohbrok:
Fixes #91 by introducing a dedicated function to retrieve the `encryption_secret` from an `EpochSecrets` object. The function removes the `Secret` such that ownership can be passed to the `SecretTree` constructor, at the end of which the Secret is dropped.

### DIFF
--- a/src/schedule/mod.rs
+++ b/src/schedule/mod.rs
@@ -71,12 +71,12 @@ impl HkdfLabel {
 
 #[derive(Clone, Debug)]
 pub struct EpochSecrets {
-    pub welcome_secret: Secret,
-    pub sender_data_secret: Secret,
+    welcome_secret: Secret,
+    sender_data_secret: Secret,
     encryption_secret: Option<Secret>,
-    pub exporter_secret: Secret,
-    pub confirmation_key: Secret,
-    pub init_secret: Secret,
+    exporter_secret: Secret,
+    confirmation_key: Secret,
+    init_secret: Secret,
 }
 
 impl Default for EpochSecrets {
@@ -158,5 +158,15 @@ impl EpochSecrets {
         // Pass ownership to the `SecretTree` constructor, so that the
         // `encryption_secret` is dropped afterwards.
         Ok(encryption_secret)
+    }
+
+    /// Get the sender_data secret.
+    pub(crate) fn sender_data_secret(&self) -> &Secret {
+        &self.sender_data_secret
+    }
+
+    /// Get the confirmation key.
+    pub(crate) fn confirmation_key(&self) -> &Secret {
+        &self.confirmation_key
     }
 }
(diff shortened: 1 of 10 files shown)


src/schedule/mod.rs
```diff
@@ -76,7 +85,7 @@ impl Default for EpochSecrets {
     fn default() -> Self {
         let welcome_secret = Secret::new_empty_secret();
         let sender_data_secret = Secret::new_empty_secret();
-        let encryption_secret = Secret::new_empty_secret();
+        let encryption_secret = RefCell::new(Some(Secret::new_empty_secret()));
```
raphaelrobert, line 88: Is there a reason you need the `RefCell` here? Why not just an `Option` that you can later consume with`.take()`?

src/schedule/mod.rs
```diff
@@ -139,13 +149,26 @@ impl EpochSecrets {
...
+            Some(es) => es,
+            None => return Err(KeyScheduleError::SecretReuseError),
+        };
+        // Pass ownership to the `SecretTree` constructor, so that the
+        // `encryption_secret` is dropped afterwards.
+        Ok(encryption_secret)
+    }
 }
 
 impl Codec for EpochSecrets {
```
raphaelrobert, line 167: Do we need the encoding at all btw? I think all we need is persistence going forward.

src/schedule/mod.rs
kkohbrok: Good point. That means we need to make `EpochSecrets` `mut`, but that should be just fine.

src/schedule/mod.rs
kkohbrok: Doesn't seem like it. I had already closed #207, but I just reopened it, so the issue is visible. I won't tackle it in this PR, though.

src/schedule/mod.rs
```diff
@@ -152,9 +149,9 @@ impl EpochSecrets {
 
     /// Remove the `encryption_secret` from the `EpochSecrets`, replacing it
     /// with `None` and return it.
-    pub fn remove_encryption_secret(&self) -> Result<Secret, KeyScheduleError> {
+    pub fn remove_encryption_secret(&mut self) -> Result<Secret, KeyScheduleError> {
```
raphaelrobert, line 152: Maybe `consume_encryption_secret` is more appropriate, given that the secret is returned?

kkohbrok:
I moved the actual consumption of the `encryption_secret` into a function of `EpochSecrets` to guarantee that it's actually dropped and to reduce the amount of exposure that secrets get.

raphaelrobert (approved):

PR #226, opened 2020-11-22
Managed group API implementation 1/?

raphaelrobert:
This PR implements the new managed API. All functions needed for the integration test are implemented.

Things that are not covered in this PR:
 - No new tests (integration test was written ahead of time
 - The Update Policy is not evaluated
 - Persistence


src/key_packages/mod.rs
```diff
@@ -237,6 +237,22 @@ impl KeyPackageBundle {
         Self::new_from_leaf_secret(ciphersuites, credential_bundle, extensions, leaf_secret)
     }
 
+    /// Replace the init key in the `KeyPackage` with a random one and return a
+    /// `KeyPackageBundle` with the corresponding secret values
+    pub fn from_rekeyed_key_package(key_package: &KeyPackage) -> Self {
```
kkohbrok, line 242: Why are we not doing this from a `KeyPackageBundle`? It seems like that's what happened before, but now it's done based on a `KeyPackage`. It seems more natural to me to begin with the private part (the bundle) and then generate the public part (the package) based on that.

src/tree/mod.rs
```diff
@@ -213,7 +213,7 @@ impl RatchetTree {
     }
 
     /// Get a reference to the own key package.
-    fn get_own_key_package_ref(&self) -> &KeyPackage {
+    pub fn get_own_key_package_ref(&self) -> &KeyPackage {
```
kkohbrok, line 216: While you're at it, rename to `own_key_package`?

src/utils.rs
```diff
@@ -42,6 +42,48 @@ macro_rules! implement_enum_display {
     };
 }
 
+pub enum ValueOrRef<'a, T: Clone> {
```
kkohbrok, line 45: Can you add a note on what this does and/or what it's for?

tests/test_managed_group.rs
```diff
@@ -144,10 +157,10 @@ fn error_occured(managed_group: &ManagedGroup, error: ManagedGroupError) {
 ///  - Charlie updates and commits
 ///  - Charlie removes Bob
 #[test]
-#[should_panic]
+//#[should_panic]
```
kkohbrok, line 160: Remove?

kkohbrok (commented):
Looks good in general. I think some logging would be good, though. I don't think I saw any in-line.

src/group/managed_group/mod.rs
```diff
@@ -96,7 +98,41 @@ impl ManagedGroup {
         credential_bundle: &CredentialBundle,
         key_packages: &[KeyPackage],
     ) -> Result<(Vec<MLSMessage>, Welcome), ManagedGroupError> {
-        unimplemented!()
+        // Create framed add proposals from key packages
+        let mut plaintext_messages: Vec<MLSPlaintext> = key_packages
+            .iter()
+            .map(|key_package| {
+                self.group
+                    .create_add_proposal(&self.aad, &credential_bundle, key_package.clone())
```
franziskuskiefer, line 106: The way this is currently used I don't see a reason not to consume the key packages in this function. The flow I see here is that the client pulls key packages from the server and then puts them in here to add the corresponding clients. So the question is whether the key packages are need at any point after this.

src/group/managed_group/mod.rs
```diff
@@ -105,16 +141,58 @@ impl ManagedGroup {
         credential_bundle: &CredentialBundle,
         members: &[usize],
```
franziskuskiefer, line 142: I still don't like this API, having to pass in a member index. But let's tackle this after this PR got merged.

src/group/managed_group/mod.rs
```diff
@@ -123,60 +201,221 @@ impl ManagedGroup {
...
+                            plaintext.into_value_or_clone(),
+                            self.pending_proposals.clone(),
+                            &self.own_kpbs,
+                        ) {
+                            Ok(()) => {
+                                // Since the Commit was applied without errors,
+                                // we can can call all corresponding callback
+                                // functions for the whole proposal list
+                                self.send_events();
+                                // We don't meed the pending proposals any longer
```
franziskuskiefer, line 306: "need"

src/group/managed_group/mod.rs
```diff
@@ -105,16 +141,58 @@ impl ManagedGroup {
...
+                self.group.create_remove_proposal(
+                    &self.aad,
+                    &credential_bundle,
+                    LeafIndex::from(*member),
+                )
+            })
+            .collect();
+
+        // Include pending proposals into Commit
+        let mut messages_to_commit = self.pending_proposals.clone();
```
franziskuskiefer, line 156: Why are the remove proposals created here are not put in the pending proposal list? I guess I don't understand what the `pending_proposal` list actually holds...

src/group/managed_group/mod.rs
```diff
@@ -105,16 +141,58 @@ impl ManagedGroup {
...
+
+        // Convert MLSPlaintext messages to MLSMessage and encrypt them if required by
+        // the configuration
+        let mls_messages = self.plaintext_to_mls_messages(plaintext_messages);
+
+        Ok(mls_messages)
     }
 
     /// Creates proposals to add members to the group
     pub fn propose_add_members(
```
franziskuskiefer, line 180: This function doesn't seem to be used and should get a test.

src/group/managed_group/mod.rs
```diff
@@ -105,16 +141,58 @@ impl ManagedGroup {
...
+                    .create_add_proposal(&self.aad, &credential_bundle, key_package.clone())
+            })
+            .collect();
+
+        let mls_messages = self.plaintext_to_mls_messages(plaintext_messages);
+
+        Ok(mls_messages)
     }
 
     /// Creates proposals to remove members from the group
```
franziskuskiefer, line 198: Also not used.

src/group/managed_group/mod.rs
```diff
@@ -123,60 +201,221 @@ impl ManagedGroup {
...
+            })
+            .collect();
+
+        let mls_messages = self.plaintext_to_mls_messages(plaintext_messages);
+
+        Ok(mls_messages)
+    }
+
+    /// Leave the group
+    pub fn leave_group(
```
franziskuskiefer, line 221: Unused

src/group/managed_group/mod.rs
```diff
@@ -123,60 +201,221 @@ impl ManagedGroup {
...
 
     // === Process messages ===
 
     /// Processes any incoming messages from the DS (MLSPlaintext &
     /// MLSCiphertext) and triggers the corresponding callback functions
     pub fn process_messages(&mut self, messages: &[MLSMessage]) {
-        unimplemented!()
+        // Iterate over all incoming messages
+        for message in messages {
+            // Check the type of message we recived
```
franziskuskiefer, line 257: "recived"

src/group/managed_group/mod.rs
```diff
@@ -123,60 +201,221 @@ impl ManagedGroup {
...
+                            (None, Some(aad))
+                        }
+                    }
+                }
+                // If it is a plaintext message we just return the reference
+                MLSMessage::Plaintext(plaintext) => (Some(ValueOrRef::from(plaintext)), None),
+            };
+            // If it was a plaintext message or if the decryption succeeded we continue,
+            // otherwise we move to the next message
+            if let Some(plaintext) = plaintext_option {
```
franziskuskiefer, line 278: If not, we don't do anything? Would be nice to let the caller know that something didn't work out.

src/group/managed_group/mod.rs
```diff
@@ -186,54 +425,223 @@ impl ManagedGroup {
...
+
+        // Create UpdateProposal
+        let mut plaintext_messages = vec![self.group.create_update_proposal(
+            &self.aad,
+            credential_bundle,
+            key_package_bundle.get_key_package().clone(),
+        )];
+
+        // Include pending proposals into Commit
+        let mut messages_to_commit = self.pending_proposals.clone();
```
franziskuskiefer, line 442: Not sure how large the `pending_proposals` will get. I just wonder if we couldn't avoid cloning it every time by only using references (we only pass in references to the group function anyway).

src/group/managed_group/mod.rs
```diff
@@ -186,54 +425,223 @@ impl ManagedGroup {
...
+
+        // Convert MLSPlaintext messages to MLSMessage and encrypt them if required by
+        // the configuration
+        let mls_messages = self.plaintext_to_mls_messages(plaintext_messages);
+
+        Ok(mls_messages)
     }
 
     /// Creates a proposal to update the own leaf node
     pub fn propose_self_update(
```
franziskuskiefer, line 468: Unused

src/group/managed_group/mod.rs
```diff
@@ -186,54 +425,223 @@ impl ManagedGroup {
...
         &mut self,
         credential_bundle: &CredentialBundle,
     ) -> Result<Vec<MLSMessage>, ManagedGroupError> {
-        unimplemented!()
+        let tree = self.group.tree();
+        let existing_key_package = tree.get_own_key_package_ref();
+        let key_package_bundle = KeyPackageBundle::from_rekeyed_key_package(existing_key_package);
```
franziskuskiefer, line 474: I'd expect this to be something like `let new_key_package_bundle = existing_key_package_bundle.rekey()`

src/group/managed_group/mod.rs
```diff
@@ -186,54 +425,223 @@ impl ManagedGroup {
...
+            .credential()
+            .get_identity()
+            .to_vec()
     }
 
     /// Get credential from LeafIndex
-    pub fn member(&self, leaf_index: LeafIndex) -> &Credential {
-        unimplemented!()
+    pub fn member(&self, leaf_index: LeafIndex) -> Option<Credential> {
+        match self.group.tree().nodes[leaf_index]
```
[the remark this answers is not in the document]
franziskuskiefer: This should be a `nodes.get(leaf_index)`.

src/group/managed_group/mod.rs
```diff
@@ -309,9 +717,13 @@ impl From<MLSCiphertext> for MLSMessage {
 
 #[derive(Debug)]
 pub enum ManagedGroupError {
```
franziskuskiefer, line 719: Can we move the errors to a separate file to get the size of this one a little down?
Also, if this is supposed to be a public API integer errors would be nice for FFI.

franziskuskiefer (changes requested):
Generally looks pretty good. I have a couple comments mostly about data flow.

src/group/managed_group/mod.rs
raphaelrobert: Fixed.

src/group/managed_group/mod.rs
raphaelrobert: Fixed.

src/group/managed_group/mod.rs
raphaelrobert: Fixed.

src/group/managed_group/mod.rs
```diff
@@ -123,60 +201,221 @@ impl ManagedGroup {
...
+            &credential_bundle,
+            LeafIndex::from(self.group.tree().get_own_node_index()),
+        );
+
+        let mls_messages = self.plaintext_to_mls_messages(vec![remove_proposal]);
+
+        Ok(mls_messages)
     }
 
     /// Gets the current list of members
```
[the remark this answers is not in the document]
raphaelrobert: No, it does not return the leaf index always. Leaves could be blank, and the blanks are compressed here.
As for the name, the assumption is that a member is fully described by a credential.

src/group/managed_group/mod.rs
```diff
@@ -123,60 +201,221 @@ impl ManagedGroup {
...
-    pub fn get_members(&self) -> &[Credential] {
-        unimplemented!()
+    pub fn get_members(&self) -> Vec<Credential> {
+        let mut members: Vec<Credential> = vec![];
+        let tree = self.group.tree();
+        let leaf_count = self.group.tree().leaf_count();
+        for index in 0..leaf_count.as_usize() {
+            let leaf = &tree.nodes[LeafIndex::from(index)];
+            if let Some(leaf_node) = leaf.get_key_package_ref() {
+                members.push(leaf_node.credential().clone());
```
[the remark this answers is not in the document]
raphaelrobert: Yes, this is a lifetime problem. There might be a way to resolve this, I'm just not sure how at this point.

src/group/managed_group/mod.rs
```diff
@@ -123,60 +201,221 @@ impl ManagedGroup {
...
-        unimplemented!()
+        // Iterate over all incoming messages
+        for message in messages {
+            // Check the type of message we recived
+            let (plaintext_option, aad_option) = match message {
+                // If it is a ciphertext we decrypt it and return the plaintext by value
+                MLSMessage::Ciphertext(ciphertext) => {
+                    let aad = &ciphertext.authenticated_data;
+                    match self.group.decrypt(ciphertext) {
+                        Ok(plaintext) => (Some(ValueOrRef::from(plaintext)), Some(aad)),
```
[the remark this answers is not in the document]
raphaelrobert: I dropped it.

src/group/managed_group/mod.rs
```diff
@@ -123,60 +201,221 @@ impl ManagedGroup {
...
+                            // If there is a callback for that event we should call it
+                            self.invalid_message_event(InvalidMessageError::InvalidCiphertext(
+                                aad.to_vec(),
+                            ));
+                            (None, Some(aad))
+                        }
+                    }
+                }
+                // If it is a plaintext message we just return the reference
+                MLSMessage::Plaintext(plaintext) => (Some(ValueOrRef::from(plaintext)), None),
```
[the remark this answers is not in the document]
raphaelrobert: Yes, I think the `MLSMessage`s can indeed be consumed here. I changed this.

src/group/managed_group/mod.rs
raphaelrobert: We already throw an error earlier if the decryption failed.

src/group/managed_group/mod.rs
```diff
@@ -123,60 +201,221 @@ impl ManagedGroup {
...
+            return Err(ManagedGroupError::PendingProposalsExist);
+        }
+        let ciphertext =
+            self.group
+                .create_application_message(&self.aad, message, credential_bundle);
+        Ok(MLSMessage::Ciphertext(ciphertext))
+    }
+
+    /// Flush pending proposals
+    pub fn flush_proposals(
```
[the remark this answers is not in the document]
raphaelrobert: I changed the name. "Flushing" usually means that something is written to disk or to the network. In this case it is "written" to the DS.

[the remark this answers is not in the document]
raphaelrobert: Because here we just process the pending proposals that were already ratified by the DS. In the other functions you mention we add new proposals that were not yet sent to the DS.

src/group/managed_group/mod.rs
```diff
@@ -186,54 +425,223 @@ impl ManagedGroup {
         &mut self,
         credential_bundle: &CredentialBundle,
     ) -> Result<Vec<MLSMessage>, ManagedGroupError> {
-        unimplemented!()
+        // Create new KeyPackageBundle for own leaf
+        let tree = self.group.tree();
+        let existing_key_package = tree.get_own_key_package_ref();
+        let key_package_bundle = KeyPackageBundle::from_rekeyed_key_package(existing_key_package);
```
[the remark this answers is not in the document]
raphaelrobert: Good point, I added an option for that.

src/group/managed_group/mod.rs
raphaelrobert: Agreed, I creates #234 for that.

src/group/managed_group/mod.rs
raphaelrobert: Fixed.

src/group/managed_group/mod.rs
raphaelrobert: Sure, we could change that. Let's do that in a separate PR though.

src/group/managed_group/mod.rs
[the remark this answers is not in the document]
raphaelrobert: I dropped that function altogether (indexes are no longer needed).

src/group/managed_group/mod.rs
```diff
@@ -186,54 +425,223 @@ impl ManagedGroup {
...
+    }
+}
+
+// Private methods of ManagedGroup
+impl ManagedGroup {
+    /// Converts MLSPlaintext to MLSMessage. Depending on whether handshake
+    /// message should be encrypted, MLSPlaintext messages are encrypted to
+    /// MLSCiphertext first.
+    fn plaintext_to_mls_messages(
+        &mut self,
```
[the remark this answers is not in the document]
raphaelrobert: We could probably do that, but this is not part of the public API, so I'm not sure it's needed.

src/group/managed_group/mod.rs
raphaelrobert: I moved the errors. As for the integers, we should come up with a comprehensive approach for the FFI. Integers don't make sense for Rust.

src/key_packages/mod.rs
raphaelrobert: That's what the function does internally. As @franziskuskiefer mentioned, we could have that function as part of `KeyPackage` instead.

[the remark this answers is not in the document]
raphaelrobert: Fixed.

src/messages/proposals.rs
```diff
@@ -186,7 +186,7 @@ impl ProposalQueue {
         proposals: Vec<MLSPlaintext>,
     ) -> Self {
         let mut proposal_queue = ProposalQueue::new();
-        for mls_plaintext in proposals {
+        for mls_plaintext in proposals.iter() {
```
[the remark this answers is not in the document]
raphaelrobert: Before the `.clone()` was outside, not it is at a later point in time. We cannot consume the proposals for `create_commit`.

src/messages/proposals.rs
```diff
@@ -237,7 +237,7 @@ impl ProposalQueue {
         let mut contains_own_updates = false;
 
         // Parse proposals and build adds and member list
-        for mls_plaintext in proposals {
+        for mls_plaintext in proposals.iter() {
             let queued_proposal = QueuedProposal::new(ciphersuite, mls_plaintext);
```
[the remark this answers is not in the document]
raphaelrobert: It really shouldn't consume proposals, because they are only applied to a provisional tree. I created #234 to improve this.

src/tree/mod.rs
raphaelrobert: Fixed.

src/tree/node.rs
```diff
@@ -116,6 +116,11 @@ impl Node {
         }
     }
 
+    /// Get a reference to the key package in this node.
+    pub(crate) fn get_key_package_ref(&self) -> Option<&KeyPackage> {
```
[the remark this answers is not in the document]
raphaelrobert: Fixed.

src/utils.rs
raphaelrobert: I dropped this.

raphaelrobert (commented):
I addressed all review comments and made some refactoring where needed.

Notes:

 - The concept of whether a group is active is a bit basic, this could get improved down the road
 - The callbacks don't require indexes anymore, everything is based on credentials
 - Removing members is still by index (in order to be compliant with the spec), hopefully this will change
 - The removal callback can now differentiate different variants of a removal
 - Everything is now blockchain-based (just making sure you're still paying attention)
 - All important functions should now be covered by tests

There will be a follow-up PR to this one, to tackle #234 and #79.

raphaelrobert:

Sorry about this, the refactoring increased the PR quite a bit. I don't think it makes sense to split it though, because everything is sort of linked.

raphaelrobert:
> Looks good in general. I think some logging would be good, though. I don't think I saw any in-line.

True, there is no logging. I think we should define the logging levels in general and have a separate PR to add logging everywhere.

src/group/managed_group/mod.rs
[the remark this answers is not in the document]
kkohbrok: Let's leave it like this and see how this shakes out in the spec. I think it should be visible to the outside if a credential is present twice in one group (which is a possibility right now), but hopefully soon that won't be a possibility anymore.

src/group/managed_group/mod.rs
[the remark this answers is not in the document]
kkohbrok: Hm. I remember sein "flush" being used in the context of caches to clear them (without writing anything to disc). But you might be right in most contexts.

kkohbrok (approved):
Thanks! Looks good to me.

src/group/managed_group/mod.rs
```diff
@@ -93,203 +119,641 @@ impl ManagedGroup {
...
+        // Create Commit over all proposals
+        let (commit, welcome_option, kpb_option) = self.group.create_commit(
+            &self.aad,
+            &self.credential_bundle,
+            &messages_to_commit,
+            false,
+        )?;
+
+        // Add the Commit message to the other pending messages
+        plaintext_messages.append(&mut vec![commit]);
```
franziskuskiefer, line 152: `plaintext_messages.push(commit)`? (For all three occurrences.)

src/group/managed_group/mod.rs
franziskuskiefer: 🤔 I don't see this being called anywhere. (That's probably also why codecov is so unhappy with this function.)

src/group/managed_group/mod.rs
franziskuskiefer: #235

src/messages/proposals.rs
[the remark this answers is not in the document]
franziskuskiefer: But the function gets `Vec<MLSPlaintext>`, which consumes such that there's no way to use `proposals` after this for loop.

tests/test_managed_group.rs
```diff
@@ -339,30 +350,143 @@ fn managed_group_operations() {
...
+        // Bob creates a new group
+        let mut bob_group = match ManagedGroup::new_from_welcome(
+            &bob_credential_bundle,
+            &managed_group_config,
+            welcome_option.expect("Welcome was not returned"),
+            Some(alice_group.export_ratchet_tree()),
+            bob_key_package_bundle,
+        ) {
+            Ok(group) => group,
+            Err(e) => panic!("Error creating group from Welcome: {:?}", e),
```
franziskuskiefer, line 430: You could `expect` all of there to avoid codecov complaining.

franziskuskiefer (changes requested):
Almost there.

I'm a little concerned by all the uncovered paths. But let's do a follow-up with more tests and a test framework.

src/group/managed_group/mod.rs
```diff
@@ -37,20 +43,35 @@ use std::io::{Read, Write};
 ///
 /// Changes to the group state are dispatched as events through callback
 /// functions (see ManagedGroupCallbacks).
-pub struct ManagedGroup {
+pub struct ManagedGroup<'a> {
+    // CredentialBundle used to sign messages
+    credential_bundle: &'a CredentialBundle,
+    // The group configuration. See `ManagedGroupCongig` for more information.
```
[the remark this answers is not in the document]
raphaelrobert: Yeah, I wanted to tinker with that anyway. I added some in the actual doc (these are just comments).

src/group/managed_group/mod.rs
raphaelrobert: Why not? The decryption might still fail, in which case we get a `None`.

franziskuskiefer: You said you're throwing an error when the decryption fails. So either we throw an error and `plaintext` is always available here, or `plaintext` is an option and and we have to handle the `None` case.

raphaelrobert: By throwing an error I meant that the corresponding callback function gets called, but it doesn't halt the processing of messages altogether. We just move to the next message in the queue instead.
I guess I can refactor the code to use something like `continue` instead, but I personally don't like these construction because they are hard to understand and review.

franziskuskiefer: If you have trouble with `continue`, you can use labelled continues in Rust to make your life a little easier.
The issue is that there's an `if` now without an `else` branch, which makes me itchy.

src/group/managed_group/callbacks.rs
```diff
@@ -4,11 +4,39 @@ use crate::group::*;
...
 /// function is specified for a certain proposal type, any semantically valid
-/// proposal will be accepted. Validator fucntions returan a `bool`, depending
+/// proposal will be accepted. Validator functions return a `bool`, depending
 /// on whether the proposal is accepted by the application policy.
 ///  - `true` means the proposal should be accepted
 ///  - `false` means the proposal should be rejected
-#[derive(Clone)]
+///
+/// The validator functions are:
+/// ```rust
```
franziskuskiefer, line 13: No need for `rust`, it's the default here. But also doesn't hurt.

src/group/managed_group/callbacks.rs
```diff
@@ -22,33 +50,75 @@ pub struct ManagedGroupCallbacks {
...
+            member_added: None,
+            member_removed: None,
+            member_updated: None,
+            app_message_received: None,
+            invalid_message_received: None,
+            error_occured: None,
         }
     }
+    /// Validator function for AddProposals
+    pub fn with_validate_add(mut self, validate_add: ValidateAdd) -> Self {
```
franziskuskiefer, line 67: This is way better already but pretty inflexible. Not too important for now. But for a builder pattern these functions usually have look like this `fn with_validate_add(&mut self, validate_add: ValidateAdd) -> &mut Self ` and then you have `fn build(&self) -> Self` function to build the object. The `build` isn't always necessary (like in this case I think we could skip if we wanted). But not consuming `self` allows different ways of calling the function.

src/group/managed_group/callbacks.rs
```diff
@@ -22,33 +50,75 @@ pub struct ManagedGroupCallbacks {
...
+    pub fn with_error_occured(mut self, error_occured: ErrorOccured) -> Self {
+        self.error_occured = Some(error_occured);
+        self
+    }
 }
 
 /// This enum lists the 4 different variants of a removal, depending on who the
 /// remover and who the leaver is.
+///
+///  - `WeLeft`: We previously issued a RemoveProposal for ourselves and this
```
franziskuskiefer, line 114: These doc comments should go to the individual elements of the enum to get the documentation displayed correctly.

franziskuskiefer (changes requested):
Only a couple nits left. We don't have to get the builder pattern right for `ManagedGroupCallbacks` now but we should find a solution for `process_message`. What I think would work as well is setting `None` and calling the error callback in the else case (if you don't want to use `continue` or some other control flow mechanisms).

src/group/managed_group/callbacks.rs
raphaelrobert: Yeah, this is just a constructor pattern obviously. Do you have any pointer to documentation on this, so that I can align on that?

raphaelrobert:
> Only a couple nits left. We don't have to get the builder pattern right for `ManagedGroupCallbacks` now but we should find a solution for `process_message`. What I think would work as well is setting `None` and calling the error callback in the else case (if you don't want to use `continue` or some other control flow mechanisms).

I changed it so that it now uses `continue`. I can change the callback constructor in a follow-up PR if you point me to some documentation on what you would like to see exactly.

src/group/managed_group/callbacks.rs
franziskuskiefer: These two are pretty good reads for this https://github.com/rust-unofficial/patterns/blob/master/patterns/builder.md https://www.ameyalokare.com/rust/2017/11/02/rust-builder-pattern.html. But I'd happy with keeping it like this for now and improve later if we see that it's not flexible enough.

franziskuskiefer (approved):

PR #227, opened 2020-11-23
add code coverage

kkohbrok:
config file shamelessly stolen from https://github.com/franziskuskiefer/evercrypt-rust/blob/master/.github/workflows/coverage.yml


franziskuskiefer (approved):
Looks good (well, it's my code ;)). You dropped the artifact. I guess that's fine for now. We can always add it later when we want to analyse the coverage locally.
Let's get this in and see if it works as expected.

Issue #141, opened 2020-10-22
PSK Injection, Group recovery, Re-Init, Sub-group Branching


raphaelrobert:
https://github.com/mlswg/mls-protocol/pull/336

PR #230, opened 2020-11-25
messages refactoring

franziskuskiefer:
As part of #83 (error handling) I did a little cleanup of the messages module.

This PR
* adds PSK dummy structs to the schedule and adds them to the `GroupSecrets` .
* moves codec implementations out of the main messages module file.
* messages only return `CodecErrors`.

### DIFF
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -76,13 +76,13 @@ impl Welcome {
         self.cipher_suite
     }
 
-    /// Get a reference to the ciphersuite in this Welcome message.
-    pub fn get_secrets_ref(&self) -> &[EncryptedGroupSecrets] {
+    /// Get a reference to the encrypted group secrets in this Welcome message.
+    pub fn secrets(&self) -> &[EncryptedGroupSecrets] {
         &self.secrets
     }
 
     /// Get a reference to the encrypted group info.
-    pub(crate) fn get_encrypted_group_info_ref(&self) -> &[u8] {
+    pub(crate) fn encrypted_group_info(&self) -> &[u8] {
         &self.encrypted_group_info
     }
 }
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -4,9 +4,12 @@ use crate::config::Config;
 use crate::config::ProtocolVersion;
 use crate::extensions::*;
 use crate::group::*;
+use crate::schedule::psk::PreSharedKeys;
 use crate::tree::{index::*, *};
 
+mod codec;
 pub(crate) mod proposals;
+pub use codec::*;
 use proposals::*;
 
 #[cfg(test)]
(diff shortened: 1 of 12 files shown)


src/messages/mod.rs
```diff
@@ -15,29 +18,100 @@ mod test_proposals;
...
+            encrypted_group_info,
+        }
+    }
+
+    /// Get a reference to the ciphersuite in this Welcome message.
+    pub(crate) fn ciphersuite(&self) -> &'static Ciphersuite {
+        self.cipher_suite
+    }
+
+    /// Get a reference to the ciphersuite in this Welcome message.
```
kkohbrok, line 79: Copy-paste error? :smile:

src/messages/mod.rs
```diff
@@ -15,29 +18,100 @@ mod test_proposals;
...
+    pub fn get_secrets_ref(&self) -> &[EncryptedGroupSecrets] {
```
kkohbrok, line 80:
```suggestion
    pub fn secrets(&self) -> &[EncryptedGroupSecrets] {
```
or maybe `secrets_ref`, because it returns a slice?

src/messages/mod.rs
```diff
@@ -15,29 +18,100 @@ mod test_proposals;
...
+        &self.secrets
+    }
+
+    /// Get a reference to the encrypted group info.
+    pub(crate) fn get_encrypted_group_info_ref(&self) -> &[u8] {
```
kkohbrok, line 85:
```suggestion
    pub(crate) fn encrypted_group_info(&self) -> &[u8] {
```
Same as above.

src/messages/mod.rs
```diff
@@ -15,29 +18,100 @@ mod test_proposals;
...
-        self.path.encode(buffer)?;
-        Ok(())
-    }
-    fn decode(cursor: &mut Cursor) -> Result<Self, CodecError> {
-        let proposals = decode_vec(VecSize::VecU32, cursor)?;
-        let path = Option::<UpdatePath>::decode(cursor)?;
-        Ok(Commit { proposals, path })
+impl Commit {
+    /// Returns `true` if the commit contains an update path. `false` otherwise.
+    pub fn has_path(&self) -> bool {
```
kkohbrok, line 98: Flag this as `#[cfg(test)]`? It doesn't seem to be used elsewhere. Or do we want this to be public for some other reason?

src/schedule/psk.rs
```diff
@@ -0,0 +1,65 @@
...
+#[derive(Debug, PartialEq, Clone, Copy)]
+#[repr(u8)]
+pub(crate) enum PSKType {
+    Reserved = 0,
+    External = 1,
+    Reinit = 2,
+    Branch = 3,
+}
+
+struct ExternalPsk {
```
kkohbrok, line 20: Maybe add a TODO here, pointing to #141? `ExternalPSK` should probably be public and/or expose a public constructor. Also, some more documentation would be nice for the individual PSK types, but we can also leave that to #141.

kkohbrok (changes requested):
Thanks, this looks good with some nits.

src/messages/mod.rs
franziskuskiefer: Yeah, this is all just c&p I didn't actually touch the code. But I can do these changes (as we want them).

src/messages/mod.rs
franziskuskiefer: Yeah the `Commit` struct itself is a little weird. This function only makes it possible to remove the `pub` on `path`. The entire thing needs to be rewritten.
Unfortunately `#[cfg(test)]` doesn't work because this is the library and that's not built with `-cfg test`.

src/schedule/psk.rs
franziskuskiefer: Yeah I'll add a TODO for the entire file. These are all really just dummy structs to actually be able to use `PreSharedKeys` as a type.

src/messages/mod.rs
franziskuskiefer: #236

kkohbrok (approved):
Looks good to me!

raphaelrobert (commented):
I didn't check the encoding/decoding. If it is just copy & pasted, it should be good to go.

PR #237, opened 2020-12-01
Harmonise some naming

raphaelrobert:
- Get rid of `get_*` and `*_ref`

### DIFF
--- a/src/extensions/mod.rs
+++ b/src/extensions/mod.rs
@@ -104,7 +104,7 @@ impl<'a> ExtensionStruct {
 
     /// Get the type of this extension struct.
     // Not needed now, but will be when we support this extension
-    pub(crate) fn _get_extension_data(&'a self) -> &'a [u8] {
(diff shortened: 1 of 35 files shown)


franziskuskiefer (dismissed):
I scrolled through this and only saw textual replacement. I trust this is good.

src/extensions/mod.rs
```diff
@@ -228,7 +228,7 @@ pub trait Extension: Debug + ExtensionHelper {
     /// Get a reference to the `RatchetTreeExtension`.
     /// Returns an `InvalidExtensionType` error if called on an `Extension`
     /// that's not a `RatchetTreeExtension`.
-    fn to_ratchet_tree_extension_ref(&self) -> Result<&RatchetTreeExtension, ExtensionError> {
```
kkohbrok, line 231: Wouldn't this be `as_ratchet_tree_extension`, as it's just a cast of a reference?

kkohbrok (approved):
Thanks for picking this up! Now everything looks a bit prettier :slightly_smiling_face:

PR #238, opened 2020-12-01
Draft for a new RatchetTree API

kkohbrok:
The API is built around the idea that changes to the `RatchetTreeBundle` can only be made by creating a `ProvisionalTree` first, on which changes are made. The changes detailed in the `ProvisionalTree` can then be verified and merged into the original `RatchetTreeBundle`. 

There are a few other design changes as well: 
* `BinaryTree` is introduced as a (relatively) generic binary tree implementation with `RatchetTree` as a wrapper with MLS-specific functions.
* `CommitSecret` and `PathSecrets` are not part of the private part of the tree anymore. They should be used for derivation or returned immediately for use in other places.

EDIT: Sorry about the mess that are the commit messages.


src/tree/newtree/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// Create a new `RatchetTreeBundle` from a `KeyPackageBundle`. This
+    /// function should only be used when creating a new group.
+    pub(crate) fn new(key_package_bundle: KeyPackageBundle) -> Self {
+        unimplemented!()
+    }
+
+    /// This function should be used when creating a new group from a `Welcome`
+    /// message. It requires a copy of the group's public tree, as well as our
+    /// own starting `KeyPackageBundle`, as well as our position in the tree.
+    pub(crate) fn from_provisional_tree(public_tree: BinaryTree<Node>) -> Self {
```
franziskuskiefer, line 46: Hm... you say this requires a number of things but the function only gets a binary tree.

src/tree/newtree/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    /// The public part of the ratchet tree.
+    public_tree: RatchetTree,
+    /// This is the HPKE private key corresponding to the `HPKEPublicKey` in the
+    /// node with index `node_index`.
+    private_key: HPKEPrivateKey,
+
+    /// A HashMap from NodeIndex to HPKEPrivateKeys in the (potentially partial)
+    /// path from this leaf.
+    path_keys: PathKeys,
```
franziskuskiefer, line 32: Naming of these members isn't great. It's close to what we have right now, which isn't great.
`path_keys`/`PathKeys` is what this is called right now. I don't actually think that this name makes a lot of sense. What path? What type of keys?
`node_index` doesn't make clear that it's about the owner of the tree. Same for `private_key`.

franziskuskiefer (commented):
This is definitely the direction we have to go. I like the separtion of the binary tree and having a clearer definition of the ratchet tree. (It's actually pretty close to what I did in my own implementation.) I'm not entirely sure yet about the provisional tree. But I think it has most things we need. Lets talk about it later.

src/tree/newtree/provisional_tree.rs
```diff
@@ -0,0 +1,162 @@
...
+use super::node::NodeContents;
+use super::tree_secrets::CommitSecret;
+use super::tree_secrets::PathSecret;
+use super::RatchetTreeBundle;
+
+#[allow(dead_code)]
+pub(crate) struct ProvisionalTree<'a> {
+    /// A reference to the RatchetTreeBundle from which the `ProvisionalTree`
+    /// was created.
+    original_tree: Option<&'a RatchetTreeBundle>,
```
raphaelrobert, line 25: Why is it an option? I can not think of a scenario where a provisional tree would exist without an original tree.

src/tree/newtree/provisional_tree.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    /// If the `original_tree` is `Some`, the `node_index` should be equal to
+    /// `original_tree.node_index`. If the `original_tree` is `None`, it has to
+    /// be provided when creating the provisional tree.
+    own_node_index: NodeIndex,
```
raphaelrobert, line 30: This should also come from the original tree, since members cannot move around in the tree.

src/tree/newtree/provisional_tree.rs
```diff
@@ -0,0 +1,162 @@
...
+            own_node_index: original_tree.node_index,
+            provisional_public_tree,
+            provisional_private_key,
+            provisional_path_keys,
+        }
+    }
+
+    /// Create a `ProvisionalTree` from a `KeyPackageBundle`. This function is
+    /// used when creating a new group.
+    pub(crate) fn from_key_package_bundle(key_package_bundle: KeyPackageBundle) -> Self {
```
raphaelrobert, line 70: What is the purpose of having a provisional tree when creating a new group? Is it just because it exposes more functions than a ratchet tree?

src/tree/newtree/provisional_tree.rs
```diff
@@ -0,0 +1,162 @@
...
+            own_node_index: node_index,
+            provisional_public_tree,
+            provisional_private_key,
+            provisional_path_keys,
+        }
+    }
+
+    /// Create a `ProvisionalTree` from a `KeyPackageBundle` and some
+    /// information from a `Welcome` message.
+    pub(crate) fn from_welcome_info(
```
raphaelrobert, line 88: Same question as above. Maybe it would be good to add some explanation what the exact prupose of a provisional tree is. So far we only discussed in the context of `create_commit` and `apply_commit`.

src/tree/newtree/ratchet_tree.rs
```diff
@@ -0,0 +1,35 @@
...
+        unimplemented!()
+    }
+
+    /// Blank the node with index `index`.
+    #[allow(unused_variables)]
+    fn blank_member(&mut self, index: NodeIndex) {
+        unimplemented!()
+    }
+
+    fn free_leaves(&self) -> Vec<NodeIndex> {
```
raphaelrobert, line 25: Maybe add a description: This returns the list of blank leaf nodes in the tree.

src/tree/newtree/ratchet_tree.rs
```diff
@@ -0,0 +1,35 @@
...
+        unimplemented!()
+    }
+
+    /// Get the size of the tree.
+    pub(crate) fn size(&self) -> usize {
```
raphaelrobert, line 30: (total number of nodes)

src/tree/newtree/tree_secrets.rs
```diff
@@ -0,0 +1,95 @@
...
+impl LeafSecret {
+    /// Randomly sample a fresh `LeafSecret`
+    pub(crate) fn random(length: usize) -> Self {
+        let secret = Secret::from(get_random_vec(length));
+        LeafSecret { secret }
+    }
+
+    /// Derive a `PathSecret` and (via a an intermediate secret) an
+    /// `HPKEKeyPair` from a `LeafSecret`, consuming it in the process.
+    pub(crate) fn to_key_pair_and_path_secret(
```
raphaelrobert, line 86: to -> into

raphaelrobert (changes requested):
This looks like a good start!
I commented on some specific functions. In general it would be good to understand a bit better in which scenarios a ratchet tree a provisional tree are used.

kkohbrok:
Thanks for the feedback! As proposed by Franziskus in our standup today, I'm in the process of creating a PR to separate the `BinaryTree` structure from the existing RatchetTree as a first step. In the process, I'll figure out what functions will be needed and what they are going to look like. Let's discuss the binary-tree related concerns when I submit the BinaryTree PR.

kkohbrok:
This is pretty stale at this point. Closing in favour of #323.

Issue #214, opened 2020-11-16
Define a keystore


raphaelrobert:
It would be nice to have a generic keystore in place to handle KeyPackages and CredentialBundles

franziskuskiefer:
@kkohbrok can you update this with what we have and we still need. I think there might also be more issues around this that should get linked and cleaned up.

kkohbrok:
Ah, at some point I created #337, because I was not aware of this issue. Should I use this one, or the other one? #337 would also have to be updated.

franziskuskiefer:
Closing in favour of #337.

PR #239, opened 2020-12-01
Managed Group API 2/2

### LINKED ISSUE #79: Managed groups

### LINKED ISSUE #234: Refactor create_commit
The function `create_commit` should support a list of references instead a list of values for the proposals.

### LINKED ISSUE #235: KeyPackageBundle: add rekey function
I'd expect this to be something like `let new_key_package_bundle = existing_key_package_bundle.rekey()`

_Originally posted by @franziskuskiefer in https://github.com/openmls/openmls/pull/226#r529358962_

raphaelrobert:
This PR fixes open issues; 

Fixes #234. This makes `create_commit` more flexible and efficient.
Fixes #235. No new functionality, just following up on a previously discussed change.

Finally, this PR should also fix #79.

### DIFF
--- a/src/key_packages/mod.rs
+++ b/src/key_packages/mod.rs
@@ -211,23 +211,6 @@ impl KeyPackage {
     pub(crate) fn cipher_suite(&self) -> &Ciphersuite {
         self.cipher_suite
     }
-
-    /// Replace the init key in the `KeyPackage` with a random one and return a
-    /// `KeyPackageBundle` with the corresponding secret values
-    pub fn rekey(&self) -> KeyPackageBundle {
-        let ciphersuite = self.cipher_suite();
-        // Generate a new leaf secret and derive the key pair
-        let leaf_secret = Secret::from(get_random_vec(ciphersuite.hash_length()));
-        let leaf_node_secret = KeyPackageBundle::derive_leaf_node_secret(ciphersuite, &leaf_secret);
-        let (private_key, public_key) = ciphersuite
-            .derive_hpke_keypair(&leaf_node_secret)
-            .into_keys();
-
-        // Repackage everything as a KeyPackageBundle
-        let mut new_key_package = self.clone();
-        new_key_package.set_hpke_init_key(public_key);
-        KeyPackageBundle::new_from_values(new_key_package, private_key, leaf_secret)
-    }
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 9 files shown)


src/key_packages/mod.rs
```diff
@@ -211,6 +211,23 @@ impl KeyPackage {
     pub(crate) fn cipher_suite(&self) -> &Ciphersuite {
         self.cipher_suite
     }
+
+    /// Replace the init key in the `KeyPackage` with a random one and return a
+    /// `KeyPackageBundle` with the corresponding secret values
+    pub fn rekey(&self) -> KeyPackageBundle {
```
kkohbrok, line 217: This is a bit of a weird construction. I think it's ok for now, but with #214, we should really update the `KeyPackageBundle` directly, not the `KeyPackage`.

kkohbrok (approved):

src/key_packages/mod.rs
franziskuskiefer: Yeah I think we should just start tackling #214. This pops up quite a bit lately.

src/key_packages/mod.rs
```diff
@@ -211,6 +211,23 @@ impl KeyPackage {
...
+        let ciphersuite = self.cipher_suite();
+        // Generate a new leaf secret and derive the key pair
+        let leaf_secret = Secret::from(get_random_vec(ciphersuite.hash_length()));
```
franziskuskiefer, line 220: Couldn't you use `Secret::random` here?

src/key_packages/mod.rs
```diff
@@ -211,6 +211,23 @@ impl KeyPackage {
...
+        let leaf_node_secret = KeyPackageBundle::derive_leaf_node_secret(ciphersuite, &leaf_secret);
```
franziskuskiefer, line 221: I'd actually move all this into the `KeyPackageBundle` and have a `random` function there.

franziskuskiefer (changes requested):
Yay, looks much cleaner.
Only one small thing.

src/key_packages/mod.rs
raphaelrobert: 😆 
That's where it was. Anyway, I moved it back now and fixed the `random` part.
I think it's better there anyway.

franziskuskiefer (approved):
Thanks, lgtm

PR #202, opened 2020-11-10
JSON serialization draft

franziskuskiefer:

PR #240, opened 2020-12-01
Persistence for groups

### LINKED ISSUE #33: [Meta] Persistence
We need to be able to persistent MLS groups and key material to disk and load it again.

also see #80

### LINKED ISSUE #80: Define an interface for persistence
Currently the TLS encoding is only used for structs that need interoperability (which is good).

The rest of the state, that needs to be persisted between operations does not have an interface for persistence.

- [ ] Decide on the format. Serde?
- [ ] Decide what should be persisted (probably all of the group state for managed groups? see #79)
- [ ] Implement the interface

raphaelrobert:
This draft is based on #202. It continues the work on persistence/serialization.

For now the format is JSON only (we should add more later).

 - API for persistence for both `MlsGroup` and `ManagedGroup`
 - Tests

Fixes #33.
Fixes #80.


src/extensions/mod.rs
```diff
@@ -284,3 +286,25 @@ impl Ord for dyn Extension {
         self.extension_type().cmp(&other.extension_type())
     }
 }
+// // For persistence we want to implement serialize and deserialize for extensions.
```
kkohbrok, line 289: Obsolete code?

src/group/managed_group/ser.rs
```diff
@@ -0,0 +1,51 @@
...
+    pending_proposals: Vec<MLSPlaintext>,
+    own_kpbs: Vec<KeyPackageBundle>,
+    aad: Vec<u8>,
+    active: bool,
+}
+
+impl<'a> SerializedManagedGroup {
+    pub(crate) fn into_managed_group(
+        self,
+        credential_bundle: &'a CredentialBundle,
```
kkohbrok, line 20: Why is the credential bundle not part of the serialized group? Is this in preparation for #214?

kkohbrok (commented):
`mls_group/ser.rs` appears to be empty. Also, I can't really judge the macro implementation, not having done much with macros myself. Other than that (with the exception of the variable-time secret comparison), it looks good to me.

src/group/managed_group/ser.rs
raphaelrobert: Yes. We don't want to serialize the signing in every group.

src/ciphersuite/mod.rs
```diff
@@ -176,13 +183,15 @@ impl ExporterSecret {
     }
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug)]
```
franziskuskiefer, line 186: Let's not derive `Debug` for keys and secrets but only when people really know what they are doing. I filed #247 for this.

src/ciphersuite/mod.rs
```diff
@@ -7,7 +7,11 @@ use log::error;
 
 use evercrypt::prelude::*;
 use hpke::prelude::*;
-use serde::{Deserialize, Serialize};
+use serde::{
```
franziskuskiefer, line 10: We should `pub(crate) use` these somewhere centrally to not having to include all of this every time.

src/config/mod.rs
```diff
@@ -99,6 +99,11 @@ impl From<PersistentConfig> for Config {
 }
 
 impl Config {
+    /// Get a reference to the global Config object.
+    pub(crate) fn _get() -> &'static Self {
```
franziskuskiefer, line 103: Unused, let's remove it.

src/utils.rs
```diff
@@ -31,6 +32,145 @@ fn _bytes_to_hex(bytes: &[u8]) -> String {
     hex
 }
 
+// Pretty ugly helper to count the number of arguments.
+#[macro_export]
```
franziskuskiefer, line 36: These `macro_export`s here are pretty ugly and not what we should do. (I know it's my code 😬 but this was the quick and dirty way.). We only want to be able to use this macro in other parts of the library not export it to the outside world. There's already a `macro_use` on `utils` now and the order is correct (I did that for another macro). So you should be able to just remove the export and the corresponding `use`s.

src/group/managed_group/mod.rs
```diff
@@ -80,7 +86,7 @@ impl<'a> ManagedGroup<'a> {
     ) -> Result<Self, ManagedGroupError> {
         let group = MlsGroup::new(
             &group_id.as_slice(),
-            key_package_bundle.key_package().cipher_suite().name(),
+            key_package_bundle.key_package().ciphersuite().name(),
```
franziskuskiefer, line 89: `ciphersuite_name()`?

src/group/managed_group/mod.rs
```diff
@@ -547,7 +578,7 @@ impl<'a> ManagedGroup<'a> {
             &self.aad,
             &self.credential_bundle,
             &messages_to_commit,
-            false,
+            true,
```
franziskuskiefer, line 581: comment

src/group/managed_group/mod.rs
```diff
@@ -750,6 +797,13 @@ impl<'a> ManagedGroup<'a> {
         }
     }
 
+    /// Auto-save function
+    fn auto_save(&self) {
+        if let Some(auto_save) = self.managed_group_config.callbacks.auto_save {
+            auto_save(&self);
```
franziskuskiefer, line 803: We should test this.

src/group/managed_group/test_managed_group.rs
```diff
@@ -0,0 +1,50 @@
+use crate::prelude::*;
```
franziskuskiefer, line 1: Why is this a unit test? Shouldn't all of this be usable from the outside?

src/group/managed_group/test_managed_group.rs
```diff
@@ -0,0 +1,50 @@
...
+
+#[test]
+fn test_managed_group_persistence() {
```
franziskuskiefer, line 4: This is good for a first test. But we really need to extend our testing framework and make persistence always a part of it.

src/group/mls_group/new_from_welcome.rs
```diff
@@ -26,7 +26,7 @@ impl MlsGroup {
         } else {
             return Err(WelcomeError::JoinerSecretNotFound);
         };
-        if ciphersuite.name() != key_package_bundle.key_package().cipher_suite().name() {
+        if ciphersuite.name() != key_package_bundle.key_package().ciphersuite().name() {
```
franziskuskiefer, line 29: `ciphersuite_name()`?

Cargo.toml
```diff
@@ -17,6 +17,7 @@ lazy_static = "1.4"
 serde = { version = "^1.0", features = ["derive"] }
 serde_json = "^1.0"
 log = { version = "0.4", features = ["std"] }
+typetag = "0.1"
```
franziskuskiefer, line 20: Shall we put persistence behind a feature flag? It requires quite some dependencies that I can see might be problematic for some users.

franziskuskiefer (changes requested):
Thanks, generally looks good. Only a couple nits. The main point I think is that it looks like we need a couple more tests for this.

There are a couple overlaps with #244. I didn't comment on them. But they should probably not be in here?

I wonder if we should allow an option to encrypt/decrypt the state with a key that's passed in. So no logic there, just encrypting and decrypting. To facilitate securer storage. But we can do this in a follow up if we want. Just came to mind.

Cargo.toml
raphaelrobert: That's true, but I'm struggling to come up with useful examples for a memory-only use of OpenMLS.

src/group/managed_group/mod.rs
raphaelrobert: I added that at the end of the integration tests.

src/group/managed_group/test_managed_group.rs
raphaelrobert: `PartialEq` is only derived for tests, we can't compare secrets from the outside.

src/group/managed_group/test_managed_group.rs
raphaelrobert: Going forward we also should not use actual files for serialization, but rather in-memory buffers.

franziskuskiefer: either that our use temporary files (there's a tempfile crate for something like this)

franziskuskiefer (approved):
Thanks. I think there's a little more cleanup we can do. But we generally have to get includes cleaned up. That's not specific to this PR.

PR #241, opened 2020-12-02
some minor changes to the API that allow using the mls group API

franziskuskiefer:
This commit changes the visibility of some functions and adds
some convenience functions.

* Codec for Vec<T: Codec> using 32-bit lengths
* Codec for String for lengths up to 16-bit
* Add convenience function MLSPlaintext::is_handshake_message to
  check whether a message is a handshake or application message
* Add MLSPlaintext::sender to get the LeafIndex of the sender of
  the message
* Enable ratchet tree extension in groups that were created from
  a welcome message with one. This isn't really necessary but
  makes life easier for now.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -147,6 +147,33 @@ impl Codec for u64 {
     }
 }
 
+impl Codec for String {
+    fn encode(&self, buffer: &mut Vec<u8>) -> Result<(), CodecError> {
+        let string_bytes = self.as_bytes();
+        (string_bytes.len() as u16).encode(buffer)?;
+        buffer.extend_from_slice(&string_bytes);
+        Ok(())
+    }
+
+    fn decode(cursor: &mut Cursor) -> Result<Self, CodecError> {
+        let string_length = u16::decode(cursor)?;
+        match std::str::from_utf8(cursor.consume(string_length.into())?) {
+            Ok(s) => Ok(s.to_string()),
+            Err(_) => Err(CodecError::DecodingError),
+        }
+    }
+}
+
+impl<T> Codec for Vec<T> where T: Codec {
+    fn encode(&self, buffer: &mut Vec<u8>) -> Result<(), CodecError> {
+        encode_vec(VecSize::VecU32, buffer, self)
+    }
+
+    fn decode(cursor: &mut Cursor) -> Result<Self, CodecError> {
+        decode_vec(VecSize::VecU32, cursor)
+    }
+}
+
 impl<T: Codec> Codec for Option<T> {
     fn encode(&self, buffer: &mut Vec<u8>) -> Result<(), CodecError> {
         match self {
--- a/src/framing/mod.rs
+++ b/src/framing/mod.rs
@@ -18,8 +18,8 @@ mod test_framing;
 
 #[derive(Debug, PartialEq, Clone)]
 pub struct MLSPlaintext {
-    pub(crate) group_id: GroupId,
-    pub(crate) epoch: GroupEpoch,
+    pub group_id: GroupId,
+    pub epoch: GroupEpoch,
     pub(crate) sender: Sender,
     pub(crate) authenticated_data: Vec<u8>,
     pub(crate) content_type: ContentType,
@@ -74,10 +74,18 @@ impl MLSPlaintext {
             signature,
         })
     }
+
     /// Returns a reference to the `content` field
     pub fn content(&self) -> &MLSPlaintextContentType {
         &self.content
     }
+
+    /// Get the sender leaf index of this message.
+    pub fn sender(&self) -> LeafIndex {
+        self.sender.to_leaf_index()
+    }
+
+    /// Sign this `MLSPlaintext`.
     pub fn sign(
         &mut self,
         credential_bundle: &CredentialBundle,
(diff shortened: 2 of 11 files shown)


franziskuskiefer:
cargo fmt sucks

src/codec.rs
```diff
@@ -147,6 +147,33 @@ impl Codec for u64 {
     }
 }
 
+impl Codec for String {
+    fn encode(&self, buffer: &mut Vec<u8>) -> Result<(), CodecError> {
+        let string_bytes = self.as_bytes();
+        (string_bytes.len() as u16).encode(buffer)?;
```
kkohbrok, line 153: What happens here if the string is too long? Should we catch that explicitly with an encoding error?

kkohbrok (commented):
Looks good to me with one comment.

Not immediately related to this PR, but just as a general reflection regarding ratchet tree extensions in groups: I'm not even sure that it should be a per-group setting. I think the spec doesn't say that it has to be handled uniformly within a group. It's just that new joiners need one and that it _can_ be added to a `Welcome` message. So within one group, clients can decide to either include it in a `Welcome` message or not. Then it's up to the joiner to get hold of a copy of the group state.

src/codec.rs
franziskuskiefer: yeah we should 👍

franziskuskiefer:
> Looks good to me with one comment.
> 
> Not immediately related to this PR, but just as a general reflection regarding ratchet tree extensions in groups: I'm not even sure that it should be a per-group setting. I think the spec doesn't say that it has to be handled uniformly within a group. It's just that new joiners need one and that it _can_ be added to a `Welcome` message. So within one group, clients can decide to either include it in a `Welcome` message or not. Then it's up to the joiner to get hold of a copy of the group state.

I agree. There needs to be some application layer logic that toggles this later. But in the PoC we don't have a different way of delivering the tree. And this is the easiest way to get it working.

src/codec.rs
```diff
@@ -147,6 +147,43 @@ impl Codec for u64 {
     }
 }
 
+impl Codec for String {
```
raphaelrobert, line 150: No particular objection but I'm curious where this would be needed since we don't use `String`s in any of the MLS structs.

src/codec.rs
```diff
@@ -147,6 +147,43 @@ impl Codec for u64 {
...
+    fn decode(cursor: &mut Cursor) -> Result<Self, CodecError> {
+        let string_length = u16::decode(cursor)?;
+        match std::str::from_utf8(cursor.consume(string_length.into())?) {
+            Ok(s) => Ok(s.to_string()),
+            Err(_) => Err(CodecError::DecodingError),
+        }
+    }
+}
+
+impl<T> Codec for Vec<T>
```
raphaelrobert, line 171: So this is essentially an alias for `encode_vec(VecSize::VecU32, ...)`?

I find it somewhat dangerous because it might just be used in places where the vec size is not 32 bit.

src/framing/mod.rs
```diff
@@ -18,8 +18,8 @@ mod test_framing;
 
 #[derive(Debug, PartialEq, Clone)]
 pub struct MLSPlaintext {
-    pub(crate) group_id: GroupId,
-    pub(crate) epoch: GroupEpoch,
+    pub group_id: GroupId,
```
raphaelrobert, line 21: Maybe `pub fn group_id(&self) -> GroupId` would be nicer?

src/framing/mod.rs
```diff
@@ -18,8 +18,8 @@ mod test_framing;
...
+    pub epoch: GroupEpoch,
```
raphaelrobert, line 22: Same as above.

raphaelrobert (commented):
Looks like minor changes indeed. Just a couple of nits I wanted to clarify.

src/codec.rs
franziskuskiefer: Well as soon as we want to send real messages we have strings. Also names are strings. This doesn't have to be handled here but if we want to treat all encoding and decoding the same way this has to go here. We should rethink encoding later anyway then we should be able to move this somewhere else.

src/codec.rs
franziskuskiefer: Yeah the `encode_vec` functions aren't very ergonomic. When redoing encoding we should add the vecsize to the trait functions. More importantly does this allow to encode vectors of vectors without having to handle the length manually.

src/framing/mod.rs
franziskuskiefer: Yeah this was quick and dirty. I agree that having functions is nicer. I'll change that.

src/codec.rs
raphaelrobert: Sure, we can keep this here for now. In general I'd argue that since the protocol draft only mentions octet strings, anything else should be on the application layer and not in the library.

src/codec.rs
raphaelrobert: I agree on the ergonomy. I think the ideal solution would be to write a `serde` plug-in, or to use something similar where you can annotate fields of structs about how they should be serialised. And then you can just use `derive` for those structs most of the time. But this is for another day.

franziskuskiefer: Right, that's why I put the two implementations in here. Expecting that this will be changed in future anyway.

kkohbrok (approved):

raphaelrobert (approved):

Issue #243, opened 2020-12-06
Improve Forward Secrecy for MLSCiphertext encryption


raphaelrobert:
We need a better way of handling encryption secrets for both application messages and HS messages.

raphaelrobert:
Blocked by #659.

PR #244, opened 2020-12-06
Quick fix for handshake message encryption

raphaelrobert:
This PR introduces a quick fix for handshake message encryption. It also fixes a missing `KeyPackageBundle` signature in Managed Groups.

Note that this fix just enables HS encryption for now, it does not address #243.

### DIFF
--- a/src/framing/mod.rs
+++ b/src/framing/mod.rs
@@ -6,6 +6,7 @@ use crate::messages::{proposals::*, *};
 use crate::schedule::*;
 use crate::tree::{index::*, secret_tree::*};
 
+use std::collections::HashMap;
 use std::convert::TryFrom;
 
 pub mod errors;
@@ -244,7 +245,7 @@ impl MLSCiphertext {
     pub(crate) fn to_plaintext(
         &self,
         ciphersuite: &Ciphersuite,
-        roster: &[&Credential],
+        indexed_members: HashMap<LeafIndex, &Credential>,
         epoch_secrets: &EpochSecrets,
         secret_tree: &mut SecretTree,
         context: &GroupContext,
@@ -317,7 +318,13 @@ impl MLSCiphertext {
             content: mls_ciphertext_content.content,
             signature: mls_ciphertext_content.signature,
         };
-        let credential = roster.get(sender_data.sender.as_usize()).unwrap();
+        let credential = match indexed_members.get(&sender_data.sender) {
+            Some(c) => c,
+            None => {
+                return Err(MLSCiphertextError::UnknownSender);
+            }
+        };
+
         let serialized_context = context.encode_detached().unwrap();
         assert!(mls_plaintext.verify(Some(serialized_context), credential));
         Ok(mls_plaintext)
@@ -520,27 +527,6 @@ impl Codec for MLSPlaintextTBS {
         self.payload.encode(buffer)?;
         Ok(())
     }
-    /*
-    fn decode(cursor: &mut Cursor) -> Result<Self, CodecError> {
-        let context = GroupContext::decode(cursor)?;
-        let group_id = GroupId::decode(cursor)?;
-        let epoch = GroupEpoch::decode(cursor)?;
-        let sender = LeafIndex::from(u32::decode(cursor)?);
-        let authenticated_data = decode_vec(VecSize::VecU32, cursor)?;
-        let content_type = ContentType::decode(cursor)?;
-        let payload = MLSPlaintextContentType::decode(cursor)?;
-
-        Ok(MLSPlaintextTBS {
-            context,
-            group_id,
-            epoch,
-            sender,
-            authenticated_data,
-            content_type,
-            payload,
-        })
-    }
-    */
 }
 
 #[derive(Clone)]
--- a/src/group/managed_group/mod.rs
+++ b/src/group/managed_group/mod.rs
@@ -517,7 +517,7 @@ impl<'a> ManagedGroup<'a> {
         if !self.active {
             return Err(ManagedGroupError::UseAfterEviction);
         }
-        
+
         // If a KeyPackageBundle was provided, create an UpdateProposal
         let mut plaintext_messages = if let Some(key_package_bundle) = key_package_bundle_option {
             let update_proposal = self.group.create_update_proposal(
--- a/src/group/managed_group/mod.rs
+++ b/src/group/managed_group/mod.rs
@@ -517,23 +517,19 @@ impl<'a> ManagedGroup<'a> {
         if !self.active {
             return Err(ManagedGroupError::UseAfterEviction);
         }
-        // Use provided KeyPackageBundle or create a new one
-        let tree = self.group.tree();
-        let key_package_bundle = match key_package_bundle_option {
-            Some(kpb) => kpb,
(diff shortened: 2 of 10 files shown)


src/framing/mod.rs
```diff
@@ -317,7 +318,13 @@ impl MLSCiphertext {
             content: mls_ciphertext_content.content,
             signature: mls_ciphertext_content.signature,
         };
-        let credential = roster.get(sender_data.sender.as_usize()).unwrap();
+        let credential = match indexed_members.get(&sender_data.sender) {
+            Some(c) => c,
+            None => {
+                return Err(MLSCiphertextError::UnknownSender);
```
kkohbrok, line 324: Maybe include a test case that triggers this error, so that codecov is a little happier?

src/group/managed_group/mod.rs
```diff
@@ -517,23 +517,19 @@ impl<'a> ManagedGroup<'a> {
...
-        drop(tree);
 
-        // Create UpdateProposal
-        let mut plaintext_messages = vec![self.group.create_update_proposal(
-            &self.aad,
-            &self.credential_bundle,
-            key_package_bundle.key_package().clone(),
-        )];
+        // If a KeyPackageBundle was provided, create an UpdateProposal
+        let mut plaintext_messages = if let Some(key_package_bundle) = key_package_bundle_option {
```
kkohbrok, line 522: What are you changing here in terms of functionality? Why wouldn't you just create an update proposal with a re-keyed key package?

src/group/managed_group/mod.rs
```diff
@@ -547,7 +543,7 @@ impl<'a> ManagedGroup<'a> {
             &self.aad,
             &self.credential_bundle,
             &messages_to_commit,
-            false,
+            true,
```
kkohbrok, line 546: Can you add a note here on what this boolean does?

src/group/managed_group/mod.rs
```diff
@@ -579,7 +575,12 @@ impl<'a> ManagedGroup<'a> {
         let existing_key_package = tree.own_key_package();
         let key_package_bundle = match key_package_bundle_option {
             Some(kpb) => kpb,
-            None => KeyPackageBundle::from_rekeyed_key_package(existing_key_package),
+            None => {
+                let mut key_package_bundle =
+                    KeyPackageBundle::from_rekeyed_key_package(existing_key_package);
```
kkohbrok, line 580: Shouldn't we handle the signature internally in the `from_rekeyed_key_package` function? It seems like this is something we want to do every time we use that function.

src/framing/mod.rs
```diff
@@ -244,7 +245,7 @@ impl MLSCiphertext {
     pub(crate) fn to_plaintext(
         &self,
         ciphersuite: &Ciphersuite,
-        roster: &[&Credential],
+        indexed_members: HashMap<LeafIndex, &Credential>,
```
kkohbrok, line 248: Is there a particular reason we using a `HashMap` here as opposed to a slice?

kkohbrok (commented):
The actual fix to the handshake secret management looks like an good construction for now and I also understand the fix for the missing signature (good catch!). I'm not sure though, what you are trying to do with the other changes.

src/tree/sender_ratchet.rs
```diff
@@ -63,10 +63,16 @@ impl SenderRatchet {
     }
     /// Gets a secret from the SenderRatchet and ratchets forward
     pub fn secret_for_encryption(&mut self, ciphersuite: &Ciphersuite) -> (u32, RatchetSecrets) {
-        let current_path_secret = self.past_secrets[0].clone();
+        let current_path_secret = self.past_secrets.last().unwrap().clone();
```
franziskuskiefer, line 66: If you're using `last`, you should also handle the error and don't panic here.

tests/test_managed_group.rs
```diff
@@ -175,7 +175,7 @@ fn managed_group_operations() {
             .with_invalid_message_received(invalid_message_received)
             .with_error_occured(error_occured);
         let managed_group_config =
-            ManagedGroupConfig::new(HandshakeMessageFormat::Plaintext, update_policy, callbacks);
+            ManagedGroupConfig::new(HandshakeMessageFormat::Ciphertext, update_policy, callbacks);
```
franziskuskiefer, line 178: Do we still have a test for `Plaintext` after this?

franziskuskiefer (changes requested):
I actually noticed this when doing the client but didn't get around to file anything. Konrad mentioned a couple good things already.

src/framing/mod.rs
raphaelrobert: We could use a slice as well, but then we would have to account for blank leaves and work with options. A `HashMap` seemed the better fit here.

src/framing/mod.rs
raphaelrobert: Done. It's in the framing unit tests.

src/group/managed_group/mod.rs
raphaelrobert: The `UpdateProposal` was unnecessary, because we will overwrite it anyway with a full Commit. This way we save some cost on crypto primitives and enforce a full Commit instead.

src/group/managed_group/mod.rs
raphaelrobert: I was going to go for that, but we also use the function when we do full Commits. There we need to first add the `ParentHashExtension` to the `KeyPackage` and then sign. So we can either split it into two functions or just remember to sign later.
PS: I think it used to be called `..._unsigned()` but I removed it as a response to PR reviews. We could also re-introduce that, but then the name would be awfully long.

tests/test_managed_group.rs
raphaelrobert: Not explicitly, but since all `MLSCiphertext` get converted to `MLSPlaintext` and vice-versa, `MLSPlaintext` are still covered that way.
In fact, this is actually the preferred way, because in this case the `MLSPlaintext`s also get serialized. If we just test with `MLSPlaintext` and ommit the serialization out of convenience (as it was the case), we don't trigger the validity checks. This is how I discovered that the signature was missing.

raphaelrobert (commented):
I added the test, fixed some minor issues I found and also discovered #251.

src/group/managed_group/mod.rs
kkohbrok: I see. Then let's leave it as it is for now.

tests/test_managed_group.rs
franziskuskiefer: Sure, but either the control flow differs depending on this value, then we need a test for this, or it doesn't, then we don't need the value in the first place.

src/framing/mod.rs
kkohbrok: I see. Good point!

kkohbrok (approved):

franziskuskiefer (changes requested):
lgtm but we should have a test MLSPlaintext as content type to improve code coverage. Just parameterize the test on it.

franziskuskiefer (approved):
😄 just made it over the magic number.
Let's get this landed.

PR #248, opened 2020-12-07
Proposals by value

### LINKED ISSUE #146: Allow proposals by value in Commit (#414)
See https://github.com/mlswg/mls-protocol/pull/414.

raphaelrobert:
This PR addresses the MLS spec change described in #146.

 - New structs are introduced
 - The Commit struct is modified
 - `apply_commit` and `create_commit` are changed, the API is adapted to be more similar on both
 - The new proposals by value are used in `add_members` and `remove_members` in the Managed Group
 - The new proposals are tested implicitly as part of the existing integration tests

Fixes #146.


src/messages/proposals.rs
```diff
@@ -117,39 +134,47 @@
...
-        Ok(ProposalID { value })
-    }
-}
-
 /// Alternative representation of a Proposal, where the sender is extracted from
 /// the encapsulating MLSPlaintext and the ProposalID is attached.
 #[derive(Clone)]
-pub struct QueuedProposal {
-    proposal: Proposal,
+pub struct QueuedProposal<'a> {
```
franziskuskiefer, line 140: Does this need to be public?

src/messages/proposals.rs
```diff
@@ -117,39 +134,47 @@
...
-    pub(crate) fn new(ciphersuite: &Ciphersuite, mls_plaintext: &MLSPlaintext) -> Self {
+    pub(crate) fn from_mls_plaintext(
+        ciphersuite: &Ciphersuite,
+        mls_plaintext: &'a MLSPlaintext,
+    ) -> Self {
         debug_assert!(mls_plaintext.content_type == ContentType::Proposal);
         let proposal = match &mls_plaintext.content {
-            MLSPlaintextContentType::Proposal(p) => p.clone(),
+            MLSPlaintextContentType::Proposal(p) => p,
             _ => panic!("API misuse. Only proposals can end up in the proposal queue"),
```
franziskuskiefer, line 156: Let's return an error instead of panicking. Now that the function isn't called `new` anymore it's ok to return a `Result`.

src/messages/proposals.rs
```diff
@@ -169,11 +194,11 @@
 /// Proposal queue that helps filtering and sorting the Proposals from one
 /// epoch.
 #[derive(Default)]
-pub struct ProposalQueue {
-    queued_proposals: HashMap<ProposalID, QueuedProposal>,
+pub struct ProposalQueue<'a> {
```
franziskuskiefer, line 197: `pub(crate)`?

src/messages/proposals.rs
```diff
@@ -293,38 +363,49 @@
...
+        self.queued_proposals
+            .iter()
+            .map(
+                |(proposal_id, queued_proposal)| match queued_proposal.por_type {
+                    ProposalOrRefType::Proposal => {
+                        ProposalOrRef::Proposal(queued_proposal.proposal.clone())
+                    }
+                    ProposalOrRefType::Reference => ProposalOrRef::Reference(proposal_id.clone()),
+                    _ => {
+                        panic!("Library error. Wrong Queued Proposals type.")
```
franziskuskiefer, line 394: I don't see that we check the type when inserting proposal into the queue. So there's a real chance that we run into this panic. Either make sure that we check the type when inserting or return an error here instead.

src/messages/test_proposals.rs
```diff
@@ -120,10 +119,39 @@ fn proposal_queue_functions() {
...
         for filtered_proposal in filtered_proposals {
             assert!(filtered_proposal.proposal().is_type(ProposalType::Add));
         }
     }
 }
+
+/// This test encodes and decodes the `ProposalOrRef` struct and makes sure the
+/// decoded values are the same as the original
+#[test]
+fn proposals_codec() {
```
franziskuskiefer, line 132: We should add a fuzzing target for this as well.

src/config/mod.rs
```diff
@@ -99,6 +99,11 @@
 }
 
 impl Config {
+    /// Get a reference to the global Config object.
```
franziskuskiefer, line 102: Didn't we just delete this?

franziskuskiefer (changes requested):
Mostly looks pretty good already. A couple comments.

src/group/managed_group/mod.rs
```diff
@@ -186,6 +186,9 @@ impl<'a> ManagedGroup<'a> {
         // Since the state of the group was changed, call the auto-save function
         self.auto_save();
 
+        // Since the state of the group was changed, call the auto-save function
```
kkohbrok, line 189: Twice?

src/messages/proposals.rs
kkohbrok: Shouldn't it then be called `try_from`?

src/messages/proposals.rs
```diff
@@ -182,17 +207,42 @@ impl ProposalQueue {
     }
     /// Returns a new `ProposalQueue` from proposals that were committed and
```
kkohbrok, line 208: That comment is not accurate here, is it? Shouldn't these be proposals that are not committed and do need filtering?

src/messages/proposals.rs
```diff
@@ -182,17 +207,42 @@ impl ProposalQueue {
...
     ) -> Self {
         let mut proposal_queue = ProposalQueue::new();
         for mls_plaintext in proposals {
-            let queued_proposal = QueuedProposal::new(ciphersuite, &mls_plaintext);
+            let queued_proposal = QueuedProposal::from_mls_plaintext(ciphersuite, &mls_plaintext);
             proposal_queue.add(queued_proposal);
         }
         proposal_queue
     }
+    /// Returns a new `ProposalQueue` from proposals that were committed and
```
kkohbrok, line 221: Not sure if this is the place, but shouldn't we sanity-check that committed proposals don't need to be filtered? Or is that taken care of by #133?

kkohbrok (changes requested):
Looks good to me already. Just a few nits.

src/messages/proposals.rs
```diff
@@ -29,10 +31,54 @@
...
+///
+/// struct {
+///   ProposalOrRefType type;
+///   select (ProposalOrRef.type) {
+///     case proposal:  Proposal proposal;
+///     case reference: opaque hash<0..255>;
+///   }
+/// } ProposalOrRef;
+
+#[derive(PartialEq, Clone, Copy, Debug)]
```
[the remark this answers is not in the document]
raphaelrobert: There will be more comments when we tackle #141.

src/messages/proposals.rs
```diff
@@ -75,38 +121,9 @@
...
-            ProposalType::Remove => Ok(Proposal::Remove(RemoveProposal::decode(cursor)?)),
-            _ => Err(CodecError::DecodingError),
-        }
-    }
-}
-
 #[derive(Debug, Eq, PartialEq, Hash, Clone, Serialize, Deserialize)]
 pub struct ProposalID {
-    value: Vec<u8>,
+    pub(crate) value: Vec<u8>,
```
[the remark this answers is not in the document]
raphaelrobert: Yeah, that only works for ancestors unfortunately.

src/messages/proposals.rs
raphaelrobert: It's publicly exposed in `apply_commit` unfortunately.

src/messages/proposals.rs
raphaelrobert: This just means they are not filtered now.

src/messages/proposals.rs
raphaelrobert: Yes, there is no validation yet and we should tackle that in #133.

src/messages/test_proposals.rs
raphaelrobert: Added.

franziskuskiefer (changes requested):
Looks good to me. Only a couple nits and the new fuzzing target won't compile.

kkohbrok (approved):

.gitignore
```diff
@@ -1,5 +1,7 @@
 /target
+delivery-service/ds/target
```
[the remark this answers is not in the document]
raphaelrobert: There is no `.gitignore` file in the DS, that's why we need it.

fuzz/fuzz_targets/proposal_decode.rs
```diff
@@ -0,0 +1,9 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+
+use openmls::prelude::*;
+
+fuzz_target!(|data: &[u8]| {
+    let mut cursor = Cursor::new(data);
+    let _ = Proposal::decode(&mut cursor);
```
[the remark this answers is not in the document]
raphaelrobert: Yeah, I since it's broken on macOS I can't really test it. Hopefully CI will go through.

.gitignore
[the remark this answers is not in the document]
franziskuskiefer: If you don't do anything funny, cargo only builds in the workspace `target`. So this directory is never created.

cli/src/main.rs
```diff
@@ -58,8 +58,7 @@ fn main() {
         // Register a client.
         // There's no persistence. So once the client app stops you have to
         // register a new client.
-        if op.starts_with("register ") {
-            let client_name = &op[9..];
+        if let Some(client_name) = op.strip_prefix("register ") {
```
franziskuskiefer, line 61: ugh, unrelated changes... please move them to a separate PR.

franziskuskiefer (commented):
lgtm except for the unrelated changes

cli/src/main.rs
raphaelrobert: I removed clippy warnings that were not caught before. This is just cleanup to make the CI happy since I touched that file.

franziskuskiefer (approved):
ok, issues are only in PoC code. Let's get it in.

PR #249, opened 2020-12-07
OpenMLS PoC Cli

franziskuskiefer:
A proof of concept cli for openmls that uses the PoC DS to deliver messages.

Note that this PR is against the DS PoC.

### DIFF
--- a/cli/src/identity.rs
+++ b/cli/src/identity.rs
@@ -19,7 +19,7 @@ impl Identity {
         }
     }
     pub fn update(&mut self) -> KeyPackageBundle {
-        let ciphersuite = self.kpb.key_package().ciphersuite();
+        let ciphersuite = self.kpb.key_package().ciphersuite_name();
         let key_package_bundle =
             KeyPackageBundle::new(&[ciphersuite], &self.credential, vec![]).unwrap();
 
--- a/cli/src/identity.rs
+++ b/cli/src/identity.rs
@@ -1,0 +1,29 @@
+use std::mem::replace;
+
+use openmls::prelude::*;
+
+pub struct Identity {
+    pub(crate) kpb: KeyPackageBundle,
+    pub(crate) credential: CredentialBundle,
+}
+
+impl Identity {
+    pub(crate) fn new(ciphersuite: CiphersuiteName, id: &[u8]) -> Self {
+        let credential_bundle =
+            CredentialBundle::new(id.to_vec(), CredentialType::Basic, ciphersuite).unwrap();
+        let key_package_bundle =
+            KeyPackageBundle::new(&[ciphersuite], &credential_bundle, vec![]).unwrap();
+        Self {
+            kpb: key_package_bundle,
+            credential: credential_bundle,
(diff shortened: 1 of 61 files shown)


cli/src/identity.rs
```diff
@@ -0,0 +1,29 @@
...
+pub struct Identity {
+    pub(crate) kpb: KeyPackageBundle,
+    pub(crate) credential: CredentialBundle,
+}
+
+impl Identity {
+    pub(crate) fn new(ciphersuite: CiphersuiteName, id: &[u8]) -> Self {
+        let credential_bundle =
+            CredentialBundle::new(id.to_vec(), CredentialType::Basic, ciphersuite).unwrap();
+        let key_package_bundle =
```
kkohbrok, line 14: only one key package bundle?

cli/src/main.rs
```diff
@@ -0,0 +1,234 @@
...
+    // Create one client
+    let mut client_1 = user::User::new("Client1".to_string());
+
+    // Create another client
+    let mut client_2 = user::User::new("Client2".to_string());
+
+    // Create another client
+    let mut client_3 = user::User::new("Client3".to_string());
+
+    // Update the clients to know about the other client.
```
kkohbrok, line 188:
```suggestion
    // Update the clients to know about the other clients.
```

kkohbrok (approved):
Honestly a bit hard to read with all the codecov warnings. I don't have much experience with client development, but it generally looks ok for a PoC.

cli/src/identity.rs
franziskuskiefer: Yeah, since we only take one key package bundle from another client there's no point in creating more 😉 
But this certainly needs more logic to handle key packages properly (like deleting on the server again etc.). Currently essentially there's only a last resort key package right now...

franziskuskiefer (commented):
> Honestly a bit hard to read with all the codecov warnings. I don't have much experience with client development, but it generally looks ok for a PoC.

yeah, I should've disable codecov fo this...

cli/src/main.rs
```diff
@@ -0,0 +1,234 @@
...
+    let mut stdout = stdout.lock();
+    let stdin = stdin();
+    let mut stdin = stdin.lock();
+
+    stdout
+        .write_all(b" >>> Welcome to the MLS CLI :)\nType help to get a list of commands\n\n")
+        .unwrap();
+    let mut client = None;
+
+    loop {
```
raphaelrobert, line 54: For future iterations: It would be nice to have an event loop here that polls new messages regularly.

raphaelrobert (changes requested):

franziskuskiefer (commented):
thanks, a couple valid points. I'd prefer tackling the bigger things in a follow up to also get some basic tests in avoiding totally breaking everything.

kkohbrok (approved):
I'd say we get this in and do the remaining points in follow-ups as suggested.

raphaelrobert (approved):

Issue #213, opened 2020-11-16
Unify how members are addressed in the managed API


raphaelrobert:
Form a previous discussion:

> It's tricky, because the amount of information varies:

    Adds: You need a full KeyPackage
    Remove: You just need an index
    Uniquely referencing a member: You need a Credential

 > We could refactor remove to accept a credential instead of an index, but the problem with Adds remains.

Alternatively we could reference clients by ID (TBD).

kkohbrok:
Since the managed API will need to be initialized (since it needs to maintain state), we could require that it be initialized with a callback function that allows us to retrieve a `KeyPackage` remotely given an ID.

Uniquely referencing a member in a group is indeed a bit tricky. Having to use a Credential is a bit suboptimal, since Credentials are per-group. Ideally, MLS require IDs to be unique per group, i.e. it would be forbidden to add a member if a member with the same ID already exists.

franziskuskiefer:
Blocked on https://mailarchive.ietf.org/arch/msg/mls/t6rJX4MbsNVYvt3qKdNBBw4OzAI/

raphaelrobert:
This has been addressed with introducing the key package hash references recently.

PR #250, opened 2020-12-07
MLS PoC Delivery Service

franziskuskiefer:
A proof of concept delivery service for MLS.

The corresponding CLI PoC is in #249.

### DIFF
--- a/delivery-service/ds-lib/src/lib.rs
+++ b/delivery-service/ds-lib/src/lib.rs
@@ -75,7 +75,7 @@ impl MLSMessage {
     pub fn group_id(&self) -> Vec<u8> {
         match self {
             MLSMessage::MLSCiphertext(m) => m.group_id.as_slice(),
-            MLSMessage::MLSPlaintext(m) => m.group_id.as_slice(),
+            MLSMessage::MLSPlaintext(m) => m.group_id().as_slice(),
         }
     }
 
@@ -83,7 +83,7 @@ impl MLSMessage {
     pub fn epoch(&self) -> u64 {
         match self {
             MLSMessage::MLSCiphertext(m) => m.epoch.0,
-            MLSMessage::MLSPlaintext(m) => m.epoch.0,
+            MLSMessage::MLSPlaintext(m) => m.epoch().0,
         }
     }
 
--- a/delivery-service/ds-lib/src/lib.rs
+++ b/delivery-service/ds-lib/src/lib.rs
@@ -42,7 +42,7 @@ impl ClientInfo {
     /// The identity of a client is defined as the identity of the first key
     /// package right now.
     pub fn id(&self) -> &[u8] {
-        self.key_packages.0[0].1.credential().identity()
+        self.id.as_slice()
     }
 }
 
--- a/delivery-service/ds-lib/src/lib.rs
+++ b/delivery-service/ds-lib/src/lib.rs
@@ -1,0 +1,248 @@
+//! # OpenMLS Delivery Service Library
+//!
+//! This library provides structs and necessary implementations to interact with
+//! the OpenMLS DS.
+//!
+//! Clients are represented by the `ClientInfo` struct.
+
+use openmls::prelude::*;
+
+/// Information about a client.
+/// To register a new client create a new `ClientInfo` and send it to
+/// `/clients/register`.
+#[derive(Debug, Default, Clone, PartialEq)]
+pub struct ClientInfo {
+    pub client_name: String,
+    pub key_packages: ClientKeyPackages,
+    pub id: Vec<u8>,
+    pub msgs: Vec<MLSMessage>,
+    pub welcome_queue: Vec<Welcome>,
+}
+
+/// The DS returns a list of key packages for a client as `ClientKeyPackages`.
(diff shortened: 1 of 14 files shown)


delivery-service/ds-lib/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+#[derive(Debug, Default, Clone, PartialEq)]
+pub struct ClientKeyPackages(pub Vec<(Vec<u8>, KeyPackage)>);
+
+impl ClientInfo {
+    /// Create a new `ClientInfo` struct for a given client name and vector of
+    /// key packages with corresponding hashes.
+    pub fn new(client_name: String, key_packages: Vec<(Vec<u8>, KeyPackage)>) -> Self {
+        Self {
+            client_name,
+            id: key_packages[0].1.credential().identity().to_vec(),
```
kkohbrok, line 35: Wouldn't it be cleaner to have the client provide an identity rather than fishing it out of the first key package?

delivery-service/ds-lib/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+            key_packages: ClientKeyPackages(key_packages),
+            msgs: Vec::new(),
+            welcome_queue: Vec::new(),
+        }
+    }
+
+    /// The identity of a client is defined as the identity of the first key
+    /// package right now.
+    pub fn id(&self) -> &[u8] {
```
kkohbrok, line 44: I'm confused. Why not just return the `id` field here?

delivery-service/ds-lib/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+    /// An OpenMLS `MLSPlaintext`.
+    MLSPlaintext(MLSPlaintext),
+}
+
+impl MLSMessage {
+    /// Get the group ID as plain byte vector.
+    pub fn group_id(&self) -> Vec<u8> {
+        match self {
+            MLSMessage::MLSCiphertext(m) => m.group_id.as_slice(),
+            MLSMessage::MLSPlaintext(m) => m.group_id().as_slice(),
```
kkohbrok, line 78: Is there a particular reason, that we access the ciphertext by field and the plaintext by getter? Same for the `epoch` below.

delivery-service/ds/src/main.rs
```diff
@@ -0,0 +1,305 @@
...
+) -> impl Responder {
+    let data = data.lock().unwrap();
+
+    let id = match base64::decode_config(id, base64::URL_SAFE) {
+        Ok(v) => v,
+        Err(_) => return actix_web::HttpResponse::BadRequest().finish(),
+    };
+    log::debug!("Getting key packages for {:?}", id);
+
+    let client = data.clients.get(&id).unwrap();
```
kkohbrok, line 131: This should probably return an error if there is no match for that id.

delivery-service/ds/src/main.rs
```diff
@@ -0,0 +1,305 @@
...
+/// Send a welcome message to a client.
+/// This takes a serialised `Welcome` message and stores the message for all
+/// clients in the welcome message.
+#[post("/send/welcome")]
+async fn send_welcome(mut body: Payload, data: web::Data<Mutex<DsData>>) -> impl Responder {
+    let mut bytes = web::BytesMut::new();
+    while let Some(item) = body.next().await {
+        bytes.extend_from_slice(&item.unwrap());
+    }
+    let welcome_msg = Welcome::decode(&mut Cursor::new(&bytes)).unwrap();
```
kkohbrok, line 146: Not sure how PoC-ish this should be in the end, but this would also be a good place to not panic :smile:

kkohbrok (commented):
Looks nice!

I didn't mark all the places, but it would be nice for the ds to survive a bad decode or some other input errors and just return an error rather than panicking.

delivery-service/ds-lib/src/lib.rs
franziskuskiefer: Well, this gets us back to the issue that there's no good notion of clients anywhere... I'd keep this and fix it with #213

delivery-service/ds-lib/src/lib.rs
franziskuskiefer: it didn't exist when I wrote this function 😉 will change...

delivery-service/ds-lib/src/lib.rs
franziskuskiefer: Because that's how it currently is in the library... the API needs to be cleaned up.

delivery-service/ds/src/main.rs
franziskuskiefer: yeah, I'll remove some of the unwraps and actually handle them

kkohbrok (approved):
Thanks for adding the error handling!

raphaelrobert (dismissed):
Let's get this landed! Just a few nits I commented on.

kkohbrok (approved):

franziskuskiefer:
MacOS fuzzing is broken (probably because a broken Rust nightly), let's ignore it for now and look into it separately.

PR #253, opened 2020-12-09
ignore PoC code for codecov

franziskuskiefer:


codecov.yml
```diff
@@ -0,0 +1,2 @@
+ignore:
+  - "delivery-service/**/*"  # ignore PoC DS server and lib
```
raphaelrobert, line 2: Let's also add the `cli` right away.

raphaelrobert (changes requested):

kkohbrok (dismissed):

codecov.yml
franziskuskiefer: `cli` doesn't exist yet... but sure I can try to add it and see what happens 😉

kkohbrok (approved):

PR #254, opened 2020-12-09
Errors for OpenMLS

franziskuskiefer:
This PR changes the way errors are defined and handled. I wasn't
really happy with the way this worked before.

There are two different types of errors allowed in OpenMLS now:
* Simple errors
```
  enum MyErrorType {
    ErrorVariant = "description",
  }
```
* Complex errors
```
  enum MyOtherErrorType {
    ErrorVariant(MoreDetaildError) = "description",
  }
```
Every definition of an error must be defined in the `implement_error`
macro, which ensures that conversion and all necessary functions
for the error are defined.

The top level error.rs also provides to basic error types for
complex errors that are not so complex.
`ErrorString(String)` and `ErrorPayload(Vec<u8>)`.

The errors for the group and managed group are changed to work in
the new way. All new errors shoudl be defined in this way and a
follow up PR will change all other existing errors to follow this
pattern.

### DIFF
--- a/src/group/errors.rs
+++ b/src/group/errors.rs
@@ -14,7 +14,7 @@ implement_error! {
         WelcomeError(WelcomeError) =
             "See [`WelcomeError`](`WelcomeError`) for details",
         ApplyCommitError(ApplyCommitError) =
-            "See [`WelcomeError`](`ApplyCommitError`) for details",
+            "See [`ApplyCommitError`](`ApplyCommitError`) for details",
         CreateCommitError(CreateCommitError) =
             "See [`WelcomeError`](`CreateCommitError`) for details",
         ConfigError(ConfigError) =
@@ -31,7 +31,7 @@ implement_error! {
         CiphersuiteMismatch =
             "Ciphersuites in the Welcome message and the corresponding key package bundle don't match.",
         JoinerSecretNotFound =
-            "No joiner secret found the Welcome message.",
+            "No joiner secret found in the Welcome message.",
         MissingRatchetTree =
             "No ratchet tree available to build initial tree after receiving a Welcome message",
         TreeHashMismatch =
@@ -56,7 +56,7 @@ implement_error! {
 implement_error! {
     pub enum ApplyCommitError {
         EpochMismatch =
-            "Couldn't apply commit. The epoch of the group context and MLSPlaintext didn't match.",
+            "Couldn't apply Commit. The epoch of the group context and MLSPlaintext didn't match.",
         WrongPlaintextContentType =
             "apply_commit_internal was called with an MLSPlaintext that is not a Commit.",
         SelfRemoved =
@@ -66,7 +66,7 @@ implement_error! {
         NoParentHashExtension =
             "Parent hash extension is missing.",
         ParentHashMismatch =
-            "Parent has values don't match.",
+            "Parent hash values don't match.",
         PlaintextSignatureFailure =
             "MLSPlaintext signature is invalid.",
         RequiredPathNotFound =
@@ -74,7 +74,7 @@ implement_error! {
         ConfirmationTagMismatch =
             "Confirmation tag is invalid.",
         MissingOwnKeyPackage =
-            "No key package to apply own commit.",
+            "No key package provided to apply own commit.",
         MissingProposal =
             "The proposal queue is missing a proposal for the commit.",
         OwnKeyNotFound =
--- a/src/group/errors.rs
+++ b/src/group/errors.rs
@@ -16,7 +16,7 @@ implement_error! {
         ApplyCommitError(ApplyCommitError) =
             "See [`ApplyCommitError`](`ApplyCommitError`) for details",
         CreateCommitError(CreateCommitError) =
-            "See [`WelcomeError`](`CreateCommitError`) for details",
+            "See [`CreateCommitError`](`CreateCommitError`) for details",
         ConfigError(ConfigError) =
             "See [`ConfigError`](`crate::config::ConfigError`) for details",
         ExporterError(ExporterError) =
--- a/src/group/errors.rs
+++ b/src/group/errors.rs
@@ -4,102 +4,107 @@
 //! `CreateCommitError`.
 
 use crate::config::ConfigError;
-use crate::extensions::ExtensionError;
 use crate::framing::errors::MLSCiphertextError;
-use crate::tree::TreeError;
+use crate::tree::{secret_tree::SecretTypeError, TreeError};
 
-#[derive(PartialEq, Debug)]
-#[repr(u16)]
-pub enum GroupError {
-    DecryptionError = 0,
+implement_error! {
+    pub enum GroupError {
+        MLSCiphertextError(MLSCiphertextError) =
+            "See [`MLSCiphertextError`](`crate::framing::errors::MLSCiphertextError`) for details",
+        WelcomeError(WelcomeError) =
+            "See [`WelcomeError`](`WelcomeError`) for details",
+        ApplyCommitError(ApplyCommitError) =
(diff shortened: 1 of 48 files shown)


src/group/errors.rs
```diff
@@ -4,102 +4,107 @@
...
-pub enum ExporterError {
-    KeyLengthTooLong = 400,
+implement_error! {
+    pub enum GroupError {
+        MLSCiphertextError(MLSCiphertextError) =
+            "See [`MLSCiphertextError`](`crate::framing::errors::MLSCiphertextError`) for details",
+        WelcomeError(WelcomeError) =
+            "See [`WelcomeError`](`WelcomeError`) for details",
+        ApplyCommitError(ApplyCommitError) =
+            "See [`WelcomeError`](`ApplyCommitError`) for details",
```
raphaelrobert, line 17: Is the discrepancy in the link on purpose?

src/group/errors.rs
```diff
@@ -4,102 +4,107 @@
...
-        match e {
-            TreeError::DuplicateIndex => WelcomeError::InvalidRatchetTree,
-            TreeError::InvalidArguments => WelcomeError::InvalidRatchetTree,
-            TreeError::InvalidUpdatePath => WelcomeError::InvalidRatchetTree,
-            TreeError::UnknownError => WelcomeError::UnknownError,
-        }
+implement_error! {
+    pub enum ApplyCommitError {
+        EpochMismatch =
+            "Couldn't apply commit. The epoch of the group context and MLSPlaintext didn't match.",
```
raphaelrobert, line 59: Spelling: If there is an actual `struct` of the same name, it should be capitalized the same way.
commit -> Commit

src/group/managed_group/errors.rs
```diff
@@ -2,74 +2,54 @@
...
-        ManagedGroupError::Config(err)
-    }
-}
-
-impl From<CodecError> for ManagedGroupError {
-    fn from(err: CodecError) -> ManagedGroupError {
-        ManagedGroupError::Codec(err)
+implement_error! {
+    pub enum ManagedGroupError {
+        LibraryError(ErrorString) =
```
raphaelrobert, line 13: We should define some rules on when we want to return a `LibraryError`. We don't have to get rid of all `.unwrap()`s if can clearly document that it is safe, but we should define where the threshold is.

src/lib.rs
```diff
@@ -13,25 +13,31 @@
...
 #[macro_use]
 mod utils;
 
+#[macro_use]
+pub mod error;
+
 mod ciphersuite;
 mod codec;
-mod config;
+pub mod config;
```
raphaelrobert, line 32: Why do we make this public here and not through `prelude.rs`?

raphaelrobert (changes requested):
Thanks! This refactoring was much needed.

src/group/errors.rs
```diff
@@ -4,102 +4,107 @@
...
 
-impl From<MLSCiphertextError> for GroupError {
-    fn from(_e: MLSCiphertextError) -> Self {
-        GroupError::DecryptionError
+implement_error! {
+    pub enum WelcomeError {
+        CiphersuiteMismatch =
+            "Ciphersuites in the Welcome message and the corresponding key package bundle don't match.",
+        JoinerSecretNotFound =
+            "No joiner secret found the Welcome message.",
```
kkohbrok, line 34:
```suggestion
            "No joiner secret found in the Welcome message.",
```

src/group/errors.rs
```diff
@@ -4,102 +4,107 @@
...
+        WrongPlaintextContentType =
+            "apply_commit_internal was called with an MLSPlaintext that is not a Commit.",
+        SelfRemoved =
+            "Tried to apply a commit to a group we are not a part of.",
+        PathKeyPackageVerificationFailure =
+            "Unable to verify the key package signature.",
+        NoParentHashExtension =
+            "Parent hash extension is missing.",
+        ParentHashMismatch =
+            "Parent has values don't match.",
```
kkohbrok, line 69:
```suggestion
            "Parent hash values don't match.",
```

src/group/errors.rs
```diff
@@ -4,102 +4,107 @@
...
+        PlaintextSignatureFailure =
+            "MLSPlaintext signature is invalid.",
+        RequiredPathNotFound =
+            "Unable to determine commit path.",
+        ConfirmationTagMismatch =
+            "Confirmation tag is invalid.",
+        MissingOwnKeyPackage =
+            "No key package to apply own commit.",
```
kkohbrok, line 77:
```suggestion
            "No key package provided to apply own commit.",
```

src/group/managed_group/mod.rs
```diff
@@ -687,20 +709,20 @@ impl<'a> ManagedGroup<'a> {
...
-                |plaintext| match self.configuration().handshake_message_format {
-                    HandshakeMessageFormat::Plaintext => MLSMessage::Plaintext(plaintext),
-                    HandshakeMessageFormat::Ciphertext => {
-                        let ciphertext = self.group.encrypt(plaintext);
-                        MLSMessage::Ciphertext(ciphertext)
-                    }
-                },
-            )
-            .collect()
+        mut plaintext_messages: Vec<MLSPlaintext>,
```
kkohbrok, line 712: Just to confirm: You just re-factored this so that you can throw an error?

kkohbrok (dismissed):
Lgtm with a few typos.

src/group/errors.rs
franziskuskiefer: c&p ftw

src/group/managed_group/errors.rs
franziskuskiefer: Yeah this is a generally contagious issue. I'm generally in favor of either using errors or not using errors. Since we want to use errors I prefer always using them. `unwrap` should only be used when there's obviously no way that the error could happen. But this is rarely the case. In any other case we should be using errors. With proper handling like I started here, this should never be more than a `?`, which is actually more readable in code than `unwrap()`.

src/group/managed_group/mod.rs
franziskuskiefer: yes 🙂 The other option is to collect a `Vec<Result<...>>`, which I'm not a huge fan off.

src/lib.rs
franziskuskiefer: The prelude is only supposed to re-export a subset of the public API that's most commonly used to make usage simpler. Most modules here will become public again to expose their API. I only touched the ones necessary for now (docs won't work otherwise for example as you'd have to point to the prelude, which is not what you want).

src/group/errors.rs
```diff
@@ -14,7 +14,7 @@ implement_error! {
         WelcomeError(WelcomeError) =
             "See [`WelcomeError`](`WelcomeError`) for details",
         ApplyCommitError(ApplyCommitError) =
-            "See [`WelcomeError`](`ApplyCommitError`) for details",
+            "See [`ApplyCommitError`](`ApplyCommitError`) for details",
         CreateCommitError(CreateCommitError) =
             "See [`WelcomeError`](`CreateCommitError`) for details",
```
kkohbrok, line 19: Same here, I think.
```suggestion
            "See [`CreateCommitError`](`CreateCommitError`) for details",
```

kkohbrok (changes requested):

src/group/managed_group/errors.rs
raphaelrobert: I filed #261 for that, I think this is good enough as a starting point.

raphaelrobert (approved):

raphaelrobert (approved):

kkohbrok (approved):

raphaelrobert (approved):

PR #257, opened 2020-12-10
Enforce correct order of proposals in ProposalQueue

### LINKED ISSUE #251: Add proposals don't have a fixed order

kkohbrok:
This PR fixes #251.

 - `ProposalQueue` now has an internal list of `ReferenceProposal` to track the right order of insertion
 - An error in managed API was fixed, where `add_member` and `remove_member` would throw the wrong error when provided with empty input
 - `ProposalID` is now renamed everywhere to `ProposalReference`
 - The is a stub test for group fuzzing

### DIFF
--- a/cli/src/backend.rs
+++ b/cli/src/backend.rs
@@ -24,7 +24,8 @@ impl Backend {
         Ok(String::from_utf8(response).unwrap())
     }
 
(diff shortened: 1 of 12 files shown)


kkohbrok:
I'll use this branch to continue working on the managed API. I'll close this PR now in favor of #264 and will re-open when it's ready.

PR #258, opened 2020-12-10
Introduce issue templates

raphaelrobert:
This PR introduces issue templates. This is working towards #196.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,30 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: "[Bug]"
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Use the following API call/function: ...
+2. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots or debug log**
+If applicable, add screenshots or debug logs to help explain your problem.
+
+**Platform (please complete the following information):**
+ - OS: [e.g. macOS]
+ - Architecture [e.g. x86_64]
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,30 @@
...
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Use the following API call/function: ...
+2. See error
+
+**Expected behavior**
```
kkohbrok, line 18: In addition to "Expected Behaviour", maybe have a field called "Actual Behaviour"?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,33 @@
...
+A clear and concise description of what you expected to happen.
+
+**Actual behavior**
+A brief explanation of what happens instead.
+
+**Screenshots or debug log**
+If applicable, add screenshots or debug logs to help explain your problem.
+
+**Platform (please complete the following information):**
+ - OS: [e.g. macOS]
```
franziskuskiefer, line 28: OS with version would be even better.

franziskuskiefer (commented):
Thanks! A couple thought, no strong opinions.

Maybe give all of them a header with something like 
"Issue tracker is **ONLY** used for reporting bugs, feature requests, and spec changes. For support or discussions please use https://github.com/openmls/openmls/discussions (or some chat service)"
Also, I'm not sure we want feature requests in here. They are usually better handled in discussions before filing an issue.

There are different approaches to how to put the text in. I like when they are commented out (`<!--- -->`) because they can stay in when writing the ticket.

raphaelrobert:
I addressed your remarks.
I left feature requests in, because not everything fits into the bug/spec change categories. We can try this for ourselves and change this later if we see it doesn't work.

franziskuskiefer (approved):

PR #262, opened 2020-12-14
Redundant ciphertexts

### LINKED ISSUE #148: Avoid redundant ciphertexts (#422)
See https://github.com/mlswg/mls-protocol/pull/422.

raphaelrobert:
This PR fixes #148.

The `resolve()` function is modified so that it accepts an exclusion list of `NodeIndex`es to filter out from the resolution.
The exclusion list is created in `apply_proposal()` when new members are added to a provisional tree and passed to the `resolve()` function.
There is a test to verify that the filtering works.

### DIFF
--- a/src/group/mls_group/apply_commit.rs
+++ b/src/group/mls_group/apply_commit.rs
@@ -46,14 +46,14 @@ impl MlsGroup {
 
         // Create provisional tree and apply proposals
         let mut provisional_tree = self.tree.borrow_mut();
-        let (path_required_by_commit, group_removed, _invited_members, new_leaves_indexes) =
+        let apply_proposals_values =
             match provisional_tree.apply_proposals(proposal_queue, own_key_packages) {
                 Ok(res) => res,
                 Err(_) => return Err(ApplyCommitError::OwnKeyNotFound),
             };
 
         // Check if we were removed from the group
-        if group_removed {
+        if apply_proposals_values.self_removed {
             return Err(ApplyCommitError::SelfRemoved);
         }
 
@@ -85,11 +85,16 @@ impl MlsGroup {
             } else {
                 // Collect the new leaves indexes so we can filter them out in the resolution later
                 provisional_tree
-                    .update_path(sender, &path, &serialized_context, new_leaves_indexes)
+                    .update_path(
+                        sender,
+                        &path,
+                        &serialized_context,
+                        apply_proposals_values.exclusion_list(),
+                    )
                     .unwrap()
             }
         } else {
-            if path_required_by_commit {
+            if apply_proposals_values.path_required {
                 return Err(ApplyCommitError::RequiredPathNotFound);
             }
             &zero_commit_secret
--- a/src/group/mls_group/apply_commit.rs
+++ b/src/group/mls_group/apply_commit.rs
@@ -83,7 +83,7 @@ impl MlsGroup {
                 };
                 provisional_tree.replace_private_tree(own_kpb, &serialized_context)
             } else {
-                // Collect the new leaves indexes so we can filter them out in the resolution later
+                // Collect the new leaves' indexes so we can filter them out in the resolution later.
                 provisional_tree
                     .update_path(
                         sender,
--- a/src/group/mls_group/apply_commit.rs
+++ b/src/group/mls_group/apply_commit.rs
@@ -46,7 +46,7 @@ impl MlsGroup {
 
         // Create provisional tree and apply proposals
         let mut provisional_tree = self.tree.borrow_mut();
-        let (path_required_by_commit, group_removed, _invited_members) =
+        let (path_required_by_commit, group_removed, _invited_members, new_leaves_indexes) =
             match provisional_tree.apply_proposals(proposal_queue, own_key_packages) {
                 Ok(res) => res,
                 Err(_) => return Err(ApplyCommitError::OwnKeyNotFound),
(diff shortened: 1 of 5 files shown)


src/tree/mod.rs
```diff
@@ -639,8 +649,8 @@ impl RatchetTree {
         &mut self,
         proposal_queue: ProposalQueue,
         updates_key_package_bundles: &[KeyPackageBundle],
-        // (path_required, self_removed, invitation_list)
-    ) -> Result<(bool, bool, InvitationList), TreeError> {
+        // (path_required, self_removed, invitation_list, new_leaves_indexes)
+    ) -> Result<(bool, bool, InvitationList, HashSet<NodeIndex>), TreeError> {
```
franziskuskiefer, line 653: These return values get pretty wild... at some point we should introduce structs to hold these complex values.

src/tree/mod.rs
```diff
@@ -706,7 +716,20 @@ impl RatchetTree {
         // Determine if Commit needs a path field
         let path_required = has_updates || has_removes || !has_adds;
 
-        Ok((path_required, self_removed, invitation_list))
+        // Collect the new leaves indexes so we can filter them out in the resolution later
+        let new_leaves_indexes: HashSet<NodeIndex> = HashSet::from_iter(
+            invitation_list
```
franziskuskiefer, line 721: The question is though whether we really need to return this here. `new_leaves_indexes` can be computed from the return value `invitation_list`. It has to happen somewhere, it's just that there are already so many return values here.

src/tree/test_resolution.rs
```diff
@@ -0,0 +1,66 @@
+use super::*;
+
+#[test]
+
+fn test_exclusion_list() {
+    for ciphersuite in Config::supported_ciphersuites() {
+        const NODES: usize = 31;
+        const FULL_RESOLUTION: &[usize] =
+            &[0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22, 24, 26, 28, 30];
```
franziskuskiefer, line 9: Can you add some description that tells me what exactly the test is (supposed to be) doing?

src/tree/mod.rs
```diff
@@ -706,7 +716,20 @@ impl RatchetTree {
...
+                .iter()
+                .map(|(index, _)| *index)
+                .collect::<Vec<NodeIndex>>(),
```
franziskuskiefer, line 724: No need to collect.

franziskuskiefer (changes requested):
Thanks. Generally looks good to me. Only a couple nits.

src/tree/mod.rs
```diff
@@ -864,7 +854,24 @@ impl RatchetTree {
...
+}
+
+impl ApplyProposalsValues {
+    pub fn exclusion_list(&self) -> HashSet<&NodeIndex> {
+        // Collect the new leaves indexes so we can filter them out in the resolution later
+        let new_leaves_indexes: HashSet<&NodeIndex> = HashSet::from_iter(
+            self.invitation_list
+                .iter()
+                .map(|(index, _)| index)
+                .collect::<Vec<&NodeIndex>>(),
```
franziskuskiefer, line 870: Still no need to `collect` 🙂 (you call `from_iter`).

src/tree/test_resolution.rs
```diff
@@ -46,21 +55,26 @@ fn test_exclusion_list() {
...
+            .iter()
+            .map(|index| NodeIndex::from(*index))
+            .collect::<Vec<NodeIndex>>();
         let exclusion_list = HashSet::from_iter(
-            EXCLUSION_LIST
+            exclusion_list_node_indexes
                 .iter()
-                .map(|index| NodeIndex::from(*index))
-                .collect::<Vec<NodeIndex>>(),
+                .collect::<Vec<&NodeIndex>>(),
```
franziskuskiefer, line 69: Also no need to `collect` 🙂

franziskuskiefer (changes requested):
Thanks, lgtm but we should skip the unnecessary collects.

src/tree/mod.rs
```diff
@@ -841,7 +854,24 @@ impl RatchetTree {
...
-pub type InvitationList = Vec<(NodeIndex, AddProposal)>;
+pub struct ApplyProposalsValues {
+    pub path_required: bool,
+    pub self_removed: bool,
+    pub invitation_list: Vec<(NodeIndex, AddProposal)>,
+}
+
+impl ApplyProposalsValues {
+    pub fn exclusion_list(&self) -> HashSet<&NodeIndex> {
+        // Collect the new leaves indexes so we can filter them out in the resolution later
```
kkohbrok, line 865:
```suggestion
        // Collect the new leaves' indexes so we can filter them out in the resolution later
```

src/tree/mod.rs
```diff
@@ -841,7 +854,24 @@ impl RatchetTree {
...
 
-pub type InvitationList = Vec<(NodeIndex, AddProposal)>;
+pub struct ApplyProposalsValues {
+    pub path_required: bool,
+    pub self_removed: bool,
+    pub invitation_list: Vec<(NodeIndex, AddProposal)>,
+}
+
+impl ApplyProposalsValues {
+    pub fn exclusion_list(&self) -> HashSet<&NodeIndex> {
```
kkohbrok, line 864: Can you add some documentation on what this is meant to do? The name is not very explanatory.

src/group/mls_group/apply_commit.rs
```diff
@@ -83,12 +83,18 @@ impl MlsGroup {
                 };
                 provisional_tree.replace_private_tree(own_kpb, &serialized_context)
             } else {
+                // Collect the new leaves indexes so we can filter them out in the resolution later
```
kkohbrok, line 86:
```suggestion
                // Collect the new leaves' indexes so we can filter them out in the resolution later.
```

kkohbrok (changes requested):
Nice! Just a few nits, mostly documentation of `exclusion_list`.

Also, I'd much prefer `indices` over `indexes`, but I recognize that both are technically correct.

franziskuskiefer (approved):
lgtm

kkohbrok (approved):
👍🏻

PR #264, opened 2020-12-15
Enforce correct order of proposals in ProposalQueue (take 2)

### LINKED ISSUE #251: Add proposals don't have a fixed order

kkohbrok:
This PR fixes #251.

- ProposalQueue now has an internal list of ReferenceProposal to track the right order of insertion. 
- In the course of testing, an error in managed API was found and fixed ad-hod, where `add_member` and `remove_member` would throw the wrong error when provided with empty input list of members to add/remove.
- ProposalID is now renamed everywhere to ProposalReference
- The test for this fix is simply creating a larger group with more than one Add at a time. The test is also the groundwork for a testing framework, which we can use to subject the managed API to more thorough tests.
- I removed the test file `tests/test_managed_api.rs` from codecov, as I don't see why coverage there should matter. I'll add it back once the testing framework is developed further. The coverage should increase once we use it in more tests.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,39 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: "[Bug]"
+labels: bug
+assignees: ''
+
+---
+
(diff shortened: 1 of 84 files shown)


kkohbrok:
Closing this as we already merged #275.

PR #266, opened 2020-12-15
Return optional Welcome messages

### LINKED ISSUE #263: Potentially return Welcome messages whenever committing
The functions `add_members` and `remove_members`, as well as `self_update` result in the client doing a Commit. In the course of the Commit, an Add Proposal could be processed. However, the functions don't return any `Welcome` messages.

raphaelrobert:
This PR fixes #263.

There was a bug in the managed API where potential add proposals in the pending proposals queue could be applied without the API returning a corresponding `Welcome` message for the new joiners to consume. This has been fixed.

This is mostly covered by existing tests for creating `Commit` messages, but a negative was added to the integration test for the managed API.

### DIFF
--- a/src/group/managed_group/mod.rs
+++ b/src/group/managed_group/mod.rs
@@ -617,7 +617,7 @@ impl<'a> ManagedGroup<'a> {
     /// be provided. If not, a new one will be created on the fly.
     ///
     /// If successful, it returns a `Vec` of
-    /// [`MLSMessage`](crate::prelude::MLSMessage) and a n optional
+    /// [`MLSMessage`](crate::prelude::MLSMessage) and an optional
     /// [`Welcome`](crate::prelude::Welcome) message if there were add proposals
     /// in the queue of pending proposals.
     pub fn self_update(
--- a/src/group/managed_group/mod.rs
+++ b/src/group/managed_group/mod.rs
@@ -204,7 +204,7 @@ impl<'a> ManagedGroup<'a> {
     /// Members are removed by providing the index of their leaf in the tree.
     ///
     /// If successful, it returns a `Vec` of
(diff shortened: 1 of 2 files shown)


src/group/managed_group/mod.rs
```diff
@@ -194,10 +200,17 @@ impl<'a> ManagedGroup<'a> {
     }
 
     /// Removes members from the group
+    ///
+    /// Members are removed by providing the index of their leaf in the tree.
+    ///
+    /// If successful, it returns a `Vec` of
+    /// [`MLSMessage`](crate::prelude::MLSMessage) and a n optional
```
kkohbrok, line 207:
```suggestion
    /// [`MLSMessage`](crate::prelude::MLSMessage) and an optional
```

src/group/managed_group/mod.rs
```diff
@@ -599,10 +612,18 @@ impl<'a> ManagedGroup<'a> {
     }
 
     /// Updates the own leaf node
+    ///
+    /// A [`KeyPackageBundle`](crate::prelude::KeyPackageBundle) can optionally
+    /// be provided. If not, a new one will be created on the fly.
+    ///
+    /// If successful, it returns a `Vec` of
+    /// [`MLSMessage`](crate::prelude::MLSMessage) and a n optional
```
kkohbrok, line 620:
```suggestion
    /// [`MLSMessage`](crate::prelude::MLSMessage) and an optional
```

kkohbrok (changes requested):
Nice one! Thanks for adding the documentation as well! I noticed that it was missing when while building the testing framework, but I forgot to file an issue. Just two typos, then we can land it, I think.

kkohbrok (approved):

franziskuskiefer (approved):
lgtm

PR #268, opened 2020-12-15
Bump HPKE and evercrypt

franziskuskiefer:
There are a couple fixes in HPKE and evercrypt.
This should also build on Windows (it's not under CI yet, so no guarantees).

replacing #260


kkohbrok (approved):

raphaelrobert (commented):
Nice! Can you add the windows support to the README file?

Then we can prepare the v0.3 release for OpenMLS.

franziskuskiefer:
I'd like to have windows on CI before advertising support.

PR #272, opened 2020-12-16
introduce HpkeError::DecryptionError for failed hpke_open calls

### LINKED ISSUE #270: [Bug] hpke_open should throw an error instead of panicking
**Describe the bug**
`hpke_open` panics instead of throwing an error.

**To Reproduce**
Call `hpke_open` with an invalid ciphertext.

**Expected behavior**
`hpke_open` throws an error.

**Actual behavior**
`hpke_open` panics.

kkohbrok:
Fixes #270.

This PR introduces `HpkeError` and more specifically `HpkeError::DecryptionError` for cases where `hpke_open` fails due to an invalid ciphertext. This PR bubbles the error up where necessary and includes a few tests.


src/ciphersuite/errors.rs
```diff
@@ -9,6 +9,11 @@ pub(crate) enum HKDFError {
     InvalidLength,
 }
 
+#[derive(Debug, PartialEq)]
+pub(crate) enum HpkeError {
```
franziskuskiefer, line 13: Please use `implement_error`. With #274 landed the others in here will use it already.

franziskuskiefer: But the question is whether we need a separate error here. I'm not too happy with the errors in here anyway. HPKE and evercrypt have their own errors that we get in here. The question is now how we want to expose them to the rest of the library. We could simply forward them (not great because they don't follow the same error handling logic). So maybe the best thing would be to have something like a `CryptoError` that holds the underlying non-openmls error values or just their error message.

src/group/errors.rs
```diff
@@ -108,3 +113,11 @@ impl From<TreeError> for WelcomeError {
         }
     }
 }
+
+impl From<HpkeError> for WelcomeError {
```
franziskuskiefer, line 117: Shouldn't we have an `HpkeError(HpkeError)` in the enum instead? (Possible with #274 landed)

src/group/mls_group/test_mls_group.rs
```diff
@@ -37,3 +41,84 @@ fn test_mls_group_persistence() {
...
+#[test]
+fn test_failed_groupinfo_decryption() {
+    for version in Config::supported_versions() {
+        for ciphersuite in Config::supported_ciphersuites() {
+            let epoch = GroupEpoch(123);
+            let group_id = GroupId::random();
+            let tree_hash = vec![1, 2, 3, 4, 5, 6, 7, 8, 9];
+            let confirmed_transcript_hash = vec![1, 1, 1];
+            let extensions = Vec::new();
+            let confirmation_tag = vec![6, 6, 6];
```
franziskuskiefer, line 54: 🤘

src/tree/mod.rs
```diff
@@ -931,3 +931,11 @@ impl From<ConfigError> for TreeError {
         }
     }
 }
+
+impl From<HpkeError> for TreeError {
```
franziskuskiefer, line 935: This again shouldn't be necessary if the errors are well defined.

franziskuskiefer (changes requested):
Panicking here is pretty bad 😄 Thanks for fixing it.
A couple comments about the way the errors are defined.

src/ciphersuite/errors.rs
raphaelrobert: Maybe also keep in mind that we want to abstract the whole crypto layer some day, so we shouldn't propagate errors from other libraries.

src/group/errors.rs
raphaelrobert: You mean `WelcomeError(HpkeError)`?

franziskuskiefer: yeah... 🙈

src/group/mls_group/test_mls_group.rs
kkohbrok: I didn't even notice 😄. Credit goes to @raphaelrobert whom (I think) hid that in an existing test that I copy-pasted.

kkohbrok:
Ok, please take a look again. I tried to follow the comments and simply introduced a new `CryptoError` instance that `hpke_open` creates from any underlying crypto failure while decrypting.

franziskuskiefer (approved):
thanks, lgtm

src/ciphersuite/test_ciphersuite.rs
```diff
@@ -14,8 +15,31 @@ fn test_hpke_seal_open() {
...
+            ciphertext: broken_ciphertext,
+        };
+        let decryption_error1 = ciphersuite
+            .hpke_open(&broken_ciphertext1, kp.private_key(), &[], &[])
+            .expect_err("Erroneously correct ciphertext decryption of broken ciphertext.");
+        let decryption_error2 = ciphersuite
+            .hpke_open(&broken_ciphertext2, kp.private_key(), &[], &[])
+            .expect_err("Erroneously correct ciphertext decryption of broken ciphertext.");
+        assert_eq!(decryption_error1, CryptoError::HpkeDecryptionError);
+        assert_eq!(decryption_error2, CryptoError::HpkeDecryptionError);
```
raphaelrobert, line 42: You can make it more compact:

```
assert_eq!(ciphersuite
            .hpke_open(&broken_ciphertext2, kp.private_key(), &[], &[])
            .expect_err(),
    CryptoError::HpkeDecryptionError);
```

raphaelrobert (approved):

kkohbrok:
While hunting for better codecov coverage (my new favourite hobby (not)), I noticed that there was no test for a commit, where one of the ciphertexts is broken, i.e. where `update_path` fails, so I added that test. Also, I got rid of the `unwrap()` of the result of `update_path`, so we can properly propagate the error (which in turn needed a new `ApplyCommitError` variant.

src/ciphersuite/test_ciphersuite.rs
kkohbrok: Done

src/group/mls_group/test_mls_group.rs
```diff
@@ -125,3 +127,183 @@ fn test_failed_groupinfo_decryption() {
         }
     }
 }
+
+#[test]
+/// Test what happens if the KEM ciphertext for the receiver in the UpdatePath is broken.
+fn test_update_path() {
+    for ciphersuite in Config::supported_ciphersuites() {
+        // Basic group setup.
```
franziskuskiefer, line 135: Don't we have a test framework to do all this setup stuff?

src/group/mls_group/test_mls_group.rs
```diff
@@ -125,3 +127,183 @@ fn test_failed_groupinfo_decryption() {
...
+        let path = commit.path.unwrap();
+
+        // For simplicity, let's just break all the ciphertexts.
+        let mut new_nodes = Vec::new();
+        for node in path.nodes {
+            let mut new_eps = Vec::new();
+            for c in node.encrypted_path_secret {
+                let mut c_encoded = c.encode_detached().unwrap();
+                let c_last_bits = c_encoded.pop().unwrap();
+                c_encoded.push(c_last_bits.reverse_bits());
```
franziskuskiefer, line 267: I've seen this code snippet before. It should probably go into a helper function.

franziskuskiefer (changes requested):
thanks generally lgtm but the tests seems overly complex.

src/group/mls_group/test_mls_group.rs
```diff
@@ -37,3 +43,267 @@ fn test_mls_group_persistence() {
...
+                group_aad,
+                &alice_credential_bundle,
+                epoch_proposals,
+                &[],
+                false,
+            )
+            .expect("Error creating commit");
+        let commit = match mls_plaintext_commit.content() {
+            MLSPlaintextContentType::Commit((commit, _)) => commit,
+            _ => panic!("Wrong content type"),
```
raphaelrobert, line 199: Use `expect()` here as well.

src/group/mls_group/test_mls_group.rs
```diff
@@ -37,3 +43,267 @@ fn test_mls_group_persistence() {
...
+            welcome_bundle_alice_bob_option.unwrap(),
+            Some(ratchet_tree),
+            bob_key_package_bundle,
+        )
+        .unwrap();
+
+        // Make sure that both groups have the same public tree
+        if group_alice.tree().public_key_tree() != group_bob.tree().public_key_tree() {
+            _print_tree(&group_alice.tree(), "Alice added Bob");
+            panic!("Different public trees");
```
raphaelrobert, line 220: `expect()`

src/group/mls_group/test_mls_group.rs
```diff
@@ -37,3 +43,267 @@ fn test_mls_group_persistence() {
...
+            .unwrap();
+
+        // Now we break Alice's HPKE ciphertext in Bob's commit by breaking
+        // apart the commit, manipulating the ciphertexts and the piecing it
+        // back together.
+        let (commit, confirmation_tag) = match &mls_plaintext_commit.content {
+            MLSPlaintextContentType::Commit((commit, confirmation_tag)) => {
+                (commit, confirmation_tag)
+            }
+            _ => panic!("Bob created a commit, which does not contain an actual commit."),
```
raphaelrobert, line 253: `expect()`

raphaelrobert (changes requested):
Looking good, just some nits.

kkohbrok:
Good point all around. I was struggling to figure out how to actually test it properly. I'll try to refine the test some more.

src/group/mls_group/test_mls_group.rs
kkohbrok: We do, but it lives in `./tests/utils` and is more suited for integration testing. The test I'm doing is more of a unit test and requires a deep-dive into various data structures that I'm not sure I can easily access from outside.

src/group/mls_group/test_mls_group.rs
kkohbrok: I don't think there is an existing function that either gives me a `Commit(..)` or an error. I'm using `match` here (and I copied this from another test btw.), so I can deconstruct the `MLSPlaintextContentType` and I'm not sure how I would use `expect()` here.

src/group/mls_group/test_mls_group.rs
kkohbrok: Again, this copied from another test and I might as well remove it.

src/group/mls_group/test_mls_group.rs
kkohbrok: Again, I'm deconstructing. I did go through and used `expect()`, where I thought I could. But I'm not sure it applies here.

src/group/mls_group/test_mls_group.rs
kkohbrok: Done

src/ciphersuite/mod.rs
```diff
@@ -65,6 +65,16 @@ pub struct HpkeCiphertext {
     ciphertext: Vec<u8>,
 }
 
+impl HpkeCiphertext {
+    #[cfg(test)]
```
franziskuskiefer, line 69: You can put this a line higher.

src/group/mls_group/test_mls_group.rs
```diff
@@ -194,13 +173,6 @@ fn test_update_path() {
                 false,
             )
             .expect("Error creating commit");
```
franziskuskiefer, line 175: Why did you remove the checks here? Not super important but they are still nice to have.

franziskuskiefer (changes requested):
Thanks, looks good. Only two nits.

src/group/mls_group/test_mls_group.rs
kkohbrok: The exact same procedure was copied from another test, which does those checks. I thought I'd simplify the group setup, but I'm happy to add them back.

franziskuskiefer (approved):
thanks!

src/group/mls_group/test_mls_group.rs
raphaelrobert: True, we would need a function to extract a Commit from MLSPlaintext for that to work.

raphaelrobert (approved):

PR #275, opened 2020-12-16
Enforce proposal ordering in the ProposalQueue

### LINKED ISSUE #251: Add proposals don't have a fixed order

kkohbrok:
Fixes #251.

- Refactor the `ProposalQueue` s.t. the iterator returned by `filtered_by_type` iterates over the Proposals contained in the queue in order of their insertion into the queue.
- Add a test for both constructors that tests this.


src/messages/test_proposals.rs
```diff
@@ -128,6 +130,123 @@ fn proposal_queue_functions() {
...
+            &alice_credential_bundle,
+            &group_context,
+        );
+
+        // This should set the order of the proposals.
+        let proposals = &[
+            &mls_plaintext_add_alice1,
+            &mls_plaintext_add_alice2,
+            &mls_plaintext_add_bob1,
+        ];
```
raphaelrobert, line 210: Can you specify what the expected outcome is, when you add the same proposal twice?

raphaelrobert (changes requested):

src/messages/proposals.rs
```diff
@@ -202,13 +202,16 @@ impl<'a> QueuedProposal<'a> {
 /// epoch.
 #[derive(Default)]
 pub struct ProposalQueue<'a> {
+    // We keep the references in a separate `Vec` to enforce the correct order
+    proposal_references: Vec<ProposalReference>,
     queued_proposals: HashMap<ProposalReference, QueuedProposal<'a>>,
```
franziskuskiefer, line 207: The hash map here also holds `ProposalReference`. So I'm not sure I understand the difference. Maybe the struct just needs better comments. Since this is public each member should get a doc comment anyway.

src/messages/proposals.rs
```diff
@@ -202,13 +202,16 @@ impl<'a> QueuedProposal<'a> {
...
 }
 
 impl<'a> ProposalQueue<'a> {
     // Returns a new empty `ProposalQueue`
     pub(crate) fn new() -> Self {
```
franziskuskiefer, line 212: This should be done by deriving `Default` instead unless I'm missing something.

src/messages/proposals.rs
```diff
@@ -233,17 +236,33 @@ impl<'a> ProposalQueue<'a> {
...
         ciphersuite: &Ciphersuite,
         committed_proposals: &'a [ProposalOrRef],
-        proposals_by_reference: &'a ProposalQueue<'a>,
+        proposals_by_reference: &[&'a MLSPlaintext],
         sender: Sender,
     ) -> Result<Self, ProposalQueueError> {
+        // Feed the `proposals_by_reference` in a `HashMap` so that we can easily
+        // extract then by reference later
+        let mut proposals_by_reference_queue: HashMap<ProposalReference, QueuedProposal> =
+            HashMap::new();
```
franziskuskiefer, line 245: You can put the `<ProposalReference, QueuedProposal>` here to the new and save yourself a `HashMap`.

src/messages/proposals.rs
```diff
@@ -233,17 +236,33 @@ impl<'a> ProposalQueue<'a> {
...
+        for mls_plaintext in proposals_by_reference {
+            let queued_proposal =
+                QueuedProposal::from_mls_plaintext(ciphersuite, mls_plaintext).unwrap();
```
franziskuskiefer, line 248: The error should be propagated.

src/messages/proposals.rs
```diff
@@ -233,17 +236,33 @@ impl<'a> ProposalQueue<'a> {
...
+            proposals_by_reference_queue.insert(
+                queued_proposal.proposal_reference().clone(),
```
franziskuskiefer, line 250: Wouldn't it make more sense to put the reference in here?

src/messages/proposals.rs
```diff
@@ -325,8 +344,9 @@ impl<'a> ProposalQueue<'a> {
         for queued_proposal in queued_proposal_list {
             match queued_proposal.proposal.proposal_type() {
                 ProposalType::Add => {
-                    adds.insert(queued_proposal.proposal_id().clone());
-                    proposal_queue.add(queued_proposal);
+                    let proposal_reference = queued_proposal.proposal_reference().clone();
```
franziskuskiefer, line 347: Looks like we always clone the output of this value... Shouldn't the function do this then?

franziskuskiefer (changes requested):
So I'm not sure I understand this PR. It looks pretty complex for achieving the relatively simple goal of having an ordered proposal queue. Why not just use a vector for the entire queue?

src/messages/test_proposals.rs
kkohbrok: I don't think adding the same proposal twice does anything here test-wise. I've removed it, as that part of the functionality is already covered in the other test.

src/messages/proposals.rs
kkohbrok: Sorry, I'm not sure what you're suggesting here.

franziskuskiefer: the following, equivalent, is a little shorter
`let mut proposals_by_reference_queue = HashMap::<ProposalReference, QueuedProposal>::new();`

src/messages/proposals.rs
kkohbrok: I'm not sure what you mean. We do put the reference in a HashMap there. Where would you suggest we put it instead?

kkohbrok:
As I mentioned OOB, the `ProposalQueue` could do with some refactoring. This PR is solely focused on having the iterator returned by `filtered_by_type` iterate over the `ProposalQueue` in order of the `Proposal`'s insertion.

franziskuskiefer (changes requested):
As already said, this is way to complex for what it is doing we can't land this as is.

src/messages/proposals.rs
```diff
@@ -355,57 +372,67 @@ impl<'a> ProposalQueue<'a> {
...
         }
-        // Only retain valid proposals
-        proposal_queue.retain(|k, _| valid_proposals.get(k).is_some() || adds.get(k).is_some());
-        (proposal_queue, contains_own_updates)
+        // Only retain `adds` and `valid_proposals`
+        let mut proposal_queue = ProposalQueue::default();
+        for proposal_reference in adds.iter().chain(valid_proposals.iter()) {
+            proposal_queue.add(match proposal_pool.get(proposal_reference) {
+                Some(queued_proposal) => queued_proposal.clone(),
+                None => return Err(ProposalQueueError::ProposalNotFound),
```
franziskuskiefer, line 387: Looks like there's no negative test case here. We should add this to the follow up.

franziskuskiefer (approved):
Ok, let's get this in to fix the bug and do a follow up to pull out the data structure we want for the queue to separate MLS logic from the data structure.

raphaelrobert (approved):

Issue #147, opened 2020-10-22
Include the signature in the confirmation tag (#416)


kkohbrok:
See https://github.com/mlswg/mls-protocol/pull/416.

Issue #145, opened 2020-10-22
Add membership tag to MLSPlaintext (#396)


kkohbrok:
This PR also changes the `confirmation_tag` computation from `KDF.Extract` to `HMAC`.

See https://github.com/mlswg/mls-protocol/pull/396.

Issue #55, opened 2020-10-06
Implement changes to sender data encryption (#360)


kkohbrok:
https://github.com/mlswg/mls-protocol/pull/360

PR #276, opened 2020-12-16
Framing refactor 1/X

raphaelrobert:
This is the first of a series of PRs to refactor framing and tackle #145, #147 and potentially #55.

This PR only splits the code of `framing/mod.rs` among multiple files for easier future changes.
There are no changes to functionality, except for minor visibility adjustments.

Since this PR is already so big, all other changes will be conducted in follow-up PRs (including better comments).


franziskuskiefer (approved):
I only skimmed over the changes but it looks like it's really only splitting the code into multiple files. lgtm

PR #281, opened 2020-12-21
Throw an error when trying to add or remove members with an empty input

kkohbrok:
What it says on the title, plus a test.

This was originally committed as part of
280a81eab31608de537683a9aa7370d06945e5a4 by @raphaelrobert.


raphaelrobert (dismissed):

franziskuskiefer (commented):
Generally lgtm. I just wonder why this is a unit test. Since it's a public API shouldn't the test go somewhere in `tests/`?

kkohbrok:
> Generally lgtm. I just wonder why this is a unit test. Since it's a public API shouldn't the test go somewhere in `tests/`?

Fair point. I'll move the test to `tests/`.

franziskuskiefer (approved):

PR #293, opened 2021-01-02
Catching up with draft-11

### LINKED ISSUE #35: Why does verify_integrity require any arguments?
https://github.com/raphaelrobert/maelstrom/blob/df3d83d45961d47a94fc5a21a8978ee8acc2421c/src/group/mls_group/new_from_welcome.rs#L94-L94

cc @raphaelrobert

### LINKED ISSUE #55: Implement changes to sender data encryption (#360)

### LINKED ISSUE #98: Node parent_hash should return byte slice
https://github.com/raphaelrobert/maelstrom/blob/39ba3ac8d027f8641e23b092087bbc054dfa210d/src/tree/node.rs#L84-L84

### LINKED ISSUE #145: Add membership tag to MLSPlaintext (#396)

### LINKED ISSUE #147: Include the signature in the confirmation tag (#416)

### LINKED ISSUE #159: Change Credential to include SignatureScheme instead of Ciphersuite (#402)

### LINKED ISSUE #283: [MLS Spec change] Authentication Secret
Derive an authentication secret from the key schedule and make it available to the application.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/419

### LINKED ISSUE #290: [MLS Spec change] Strong tree signing
Changes how the parent hash is calculated

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/436

### LINKED ISSUE #292: [MLS Spec change] Move joiner_secret to the end of the first KDF cycle in the Key Schedule
Move joiner_secret to the end of the first KDF cycle in the Key Schedule

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/446

raphaelrobert:
This PR addresses most of the spec changes of draft-11 and a few leftovers from draft-10.

The 3 main features that are not yet addressed are: External Commits, PSK-related things and AppAck. 
There also a few issues left that relate to spec compliance.

**MLS spec changes**

 - Fixes #145 (Add membership tag to MLSPlaintext)

 - Fixes #147 (Include the signature in the confirmation tag)

 - Fixes #55 (Implement changes to sender data encryption)

 - Partially addresses #141 (Passive support for PSKs: They can be received but no action is taken.)

 - Fixes #159 (Change Credential to include SignatureScheme instead of Ciphersuite)

 - Fixes #290 (Strong tree signing) (also closes #35 and closes #98)

 - Fixes #283 (Authentication Secret)

 - Fixes #292 (Move joiner_secret to the end of the first KDF cycle in the Key Schedule)

**Misc improvements**

 - Error handling for MLSCiphertext
 - Harmonize codec code
 - Configurable padding block size for managed groups
 - Converting from NodeIndex to LeafIndex now returns an error if the node index is an odd number
 - General code cleanup


src/tree/mod.rs
```diff
@@ -737,121 +752,73 @@ impl RatchetTree {
...
+                if node.node_type.is_leaf() || node.node.is_none() || node.key_package.is_some() {
+                    return false;
+                }
+                let left = treemath::left(index.into()).unwrap();
+                let right = treemath::right(index.into(), self.leaf_count()).unwrap();
+                // Unwrapping here is safe, because we know it is a full parent node
+                let parent_hash_field = &node.parent_hash().unwrap();
+                // Calculate the current
+                let current_hash =
+                    ParentHashInput::new(&self, index.into(), right.into(), parent_hash_field)
```
kkohbrok, line 783: Doesn't the hash we want to check against depend on which child we're looking at? I would expect it to be calculated ad-hoc when comparing with the `parent_hash` fields of the children as outlined [here](https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#section-7.4.2).

kkohbrok: As it currently stands, I think the procedure below does not quite match what's described in the spec.

kkohbrok (changes requested):
Great job on integrating all those changes! This is only a partial review focusing on the spec changes. Everything looks good so far with the exception of the `verify_parent_hashes` algorithm, which doesn't quite match my understanding of what it says in the spec.

src/tree/secret_tree.rs
```diff
@@ -34,7 +34,6 @@ impl TryFrom<&ContentType> for SecretType {
             ContentType::Application => Ok(SecretType::ApplicationSecret),
```
franziskuskiefer, line 34: If this is a complete match now, we could implement `From` instead here.

franziskuskiefer (changes requested):
Ok that's a big one and we'll not be able to do a super thorough review. But that's fine for the sake of getting it in. With the upcoming test framework and interop we should find most logic issues.  And we'll have to refactor lots of this anyway.ConfirmationTag

The code generally looks good. Comments inline.

A pattern I've seen a lot in the code here is that you `match` on results and then return an error. We should be able to avoid this most of the time by proper implementation of `From`/errors.

There are also a bunch of clippy warnings.

raphaelrobert (commented):
Thanks for the reviews! I addressed the comments or left a comment when I didn't.

src/framing/plaintext.rs
```diff
@@ -9,34 +39,92 @@
...
         let mut mls_plaintext = MLSPlaintext {
-            group_id: context.group_id.clone(),
-            epoch: context.epoch,
+            group_id: context.group_id().clone(),
+            epoch: context.epoch(),
             sender,
             authenticated_data: authenticated_data.to_vec(),
             content_type: ContentType::from(&content),
             content,
             signature: Signature::new_empty(),
```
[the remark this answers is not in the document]
franziskuskiefer: Well, so the question is whether we want to make our code ugly because the spec says this. I'd strongly object. If the struct in the spec doesn't make sense for our implementation, we either have to make sure that the encoding/decoding accounts for the differences, or have a spec compliant struct and a "usable" struct. (I'd go for the former approach.)
But since this didn't change here, please file a follow-up issue.

src/framing/plaintext.rs
```diff
@@ -47,23 +135,60 @@
...
         serialized_context_option: Option<Vec<u8>>,
         credential: &Credential,
     ) -> bool {
         let signature_input = MLSPlaintextTBS::new_from(&self, serialized_context_option);
         signature_input.verify(credential, &self.signature)
     }
+
+    /// Verify the membership tag of the `MLSPlaintext`. Returns `true` if the
+    /// verification is successful and `false` otherwise.
+    pub fn verify_membership_tag(
```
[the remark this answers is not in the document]
franziskuskiefer: The membership tag can be optional but that doesn't you need two functions. Just make it optional.

src/framing/plaintext.rs
```diff
@@ -142,32 +275,116 @@
...
+    ) -> Self {
+        Self {
+            tbs_payload,
+            signature: &mls_plaintext.signature,
+            confirmation_tag: &mls_plaintext.confirmation_tag,
+        }
+    }
+    fn into_bytes(self) -> Vec<u8> {
+        let mut buffer = self.tbs_payload.payload;
+        buffer.extend(self.signature.encode_detached().unwrap().iter());
```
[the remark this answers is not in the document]
franziskuskiefer: 🤔

src/messages/codec.rs
```diff
@@ -87,15 +107,20 @@ impl Codec for Welcome {
...
-        Ok(())
-    }
     fn decode(cursor: &mut Cursor) -> Result<Self, CodecError> {
         let joiner_secret = JoinerSecret::decode(cursor)?;
         let path_secret = Option::<PathSecret>::decode(cursor)?;
-        Ok(GroupSecrets::new(joiner_secret, path_secret))
+        let _psks = Option::<PreSharedKeys>::decode(cursor)?;
+        if _psks.is_some() {
+            log::error!(
+                "Error ocurred while decoding GroupSecrets: Found PSKs that are not yet supported"
```
franziskuskiefer, line 116: "ocurred"

src/schedule/mod.rs
```diff
@@ -94,22 +93,27 @@
...
     /// `CommitSecret`. The `InitSecret` is randomly generated. TODO:
     /// For now, this takes a reference to a `CommitSecret` as input. This
     /// should change with #224.
-    fn from_commit_secret(ciphersuite: &Ciphersuite, commit_secret: &CommitSecret) -> Self {
+    pub(crate) fn from_commit_secret(
+        ciphersuite: &Ciphersuite,
+        commit_secret: &CommitSecret,
+    ) -> Self {
         let initial_init_secret = InitSecret::random(ciphersuite.hash_length());
         JoinerSecret {
```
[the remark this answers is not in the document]
franziskuskiefer: 🤔

src/schedule/psk.rs
```diff
@@ -55,13 +96,34 @@
...
-struct PreSharedKeyID {
-    psktype: PSKType,
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+pub struct PreSharedKeyID {
+    pub(crate) psktype: PSKType,
+    pub(crate) psk: Psk,
+    pub(crate) psk_nonce: Vec<u8>,
+}
+
+impl PreSharedKeyID {
```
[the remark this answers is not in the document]
franziskuskiefer: I again don't think that's not necessary a good idea 🙂

src/tree/hash_input.rs
```diff
@@ -10,64 +18,64 @@
...
-            parent_node,
-            left_hash,
-            right_hash,
+            public_key,
+            parent_hash,
+            original_child_resolution,
         }
     }
-    pub fn hash(&self, ciphersuite: &Ciphersuite) -> Vec<u8> {
+    pub(crate) fn hash(&self, ciphersuite: &Ciphersuite) -> Vec<u8> {
```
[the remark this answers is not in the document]
franziskuskiefer: I think we should again implement it sensible, not religiously sticking to spec structs where they make our life hard or the code ugly.

src/tree/hash_input.rs
```diff
@@ -55,44 +109,136 @@
...
+/// The list of HPKEPublicKey values of the nodes in the resolution of `index`
+/// but with the `unmerged_leaves` of the parent node omitted.
+pub(crate) fn original_child_resolution(
+    tree: &RatchetTree,
+    index: NodeIndex,
+) -> Vec<&HPKEPublicKey> {
+    // Build the exclusion list that consists of the unmerged leaves of the parent
+    // node
+    let mut unmerged_leaves = vec![];
+    // If the current index is not the root, we collectthe unmerged leaves of the
```
franziskuskiefer, line 129: "collectthe"

src/tree/mod.rs
```diff
@@ -104,12 +107,16 @@
         fn find_kp_in_tree(
             key_package: &KeyPackage,
             nodes: &[Option<Node>],
-        ) -> Result<NodeIndex, TreeError> {
+        ) -> Result<LeafIndex, TreeError> {
             for (i, node_option) in nodes.iter().enumerate() {
```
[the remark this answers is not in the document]
franziskuskiefer: 🤔

src/messages/test_codec.rs
```diff
@@ -0,0 +1,67 @@
...
+        },
+    };
+    let encoded = orig.encode_detached().unwrap();
+    let decoded = PreSharedKeyProposal::decode_detached(&encoded).unwrap();
+    assert_eq!(decoded, orig);
+
+    // External
+    let orig = PreSharedKeyProposal {
+        psk: PreSharedKeyID {
+            psktype: PSKType::External,
```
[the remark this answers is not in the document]
franziskuskiefer: That's why we have comments 😀

franziskuskiefer (changes requested):
Thanks, a couple unresolved comments and a couple more comments.
Also the tests don't compile.

src/tree/mod.rs
```diff
@@ -737,121 +750,83 @@ impl RatchetTree {
...
+                    } else {
+                        // Iterate over children until we find a node that is neither blank
+                        // nor a leaf node
+                        let mut child = right;
+                        while !self.nodes[child].is_full_parent() {
+                            let right_option = treemath::left(child);
+                            match right_option {
+                                // Keep iterating
+                                Ok(r) => child = r,
+                                // We didn't find any full parent node in this path
```
kkohbrok, line 813: I don't think that's how it works. `right_option` (which is not really an option) doesn't say if it's a blank or not. Just if there is actually an index pointing to the left child of `child`. For example, you'd get an error if `child` was a leaf.

src/tree/mod.rs
```diff
@@ -737,121 +750,83 @@ impl RatchetTree {
...
-                    NodeType::Leaf => {
-                        if let Some(kp) = &node.key_package {
-                            if i % 2 != 0 {
-                                return false;
-                            }
-                            if kp.verify().is_err() {
-                                return false;
-                            }
+                        // Unwrapping here is safe, because we know it is a full parent node
+                        if self.nodes[child].parent_hash().unwrap() == current_hash_left {
```
kkohbrok, line 818: We know it's not a leaf, but it could still be a blank, so we can't `unwrap` here, I think.

src/tree/mod.rs
```diff
@@ -737,121 +750,83 @@ impl RatchetTree {
...
-                            .expect("verify_integrity: Error when computing left child of node.");
-                        let right_index = treemath::right(NodeIndex::from(i), size.into())
-                            .expect("verify_integrity: Error when computing right child of node.");
-                        if right_index >= node_count {
-                            return false;
+
+    /// Verify the parent hashes of the tree nodes. Returns `true` if all parent
+    /// hashes have successfully been verified and `false` otherwise.
+    pub fn verify_parent_hashes(&self) -> bool {
+        for (index, node) in self.nodes.iter().enumerate() {
```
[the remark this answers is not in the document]
kkohbrok: Also, according to the spec, we have to check the parent hash of the committer's leaf with every (full) commit, so we might be able to re-use that function there.

src/tree/hash_input.rs
```diff
@@ -55,44 +109,136 @@ impl<'a> ParentNodeHashInput<'a> {
...
-impl<'a> Codec for ParentNodeHashInput<'a> {
-    fn encode(&self, buffer: &mut Vec<u8>) -> Result<(), CodecError> {
-        self.node_index.encode(buffer)?;
-        self.parent_node.encode(buffer)?;
-        encode_vec(VecSize::VecU8, buffer, &self.left_hash)?;
-        encode_vec(VecSize::VecU8, buffer, &self.right_hash)?;
-        Ok(())
+/// Computes the parent hashes for a leaf node and returns the parent hash for
+/// the parent hash extension
+pub(crate) fn compute_parent_hashes(tree: &mut RatchetTree, index: LeafIndex) -> Vec<u8> {
```
kkohbrok, line 154: Is there a particular reason that this is not part of `impl RatchetTree`? Same goes for `compute_tree_hash`.

kkohbrok (changes requested):
Getting there. Just a few comments and one more bug (I think) in the algorithm.

raphaelrobert (commented):
I hope I addressed all comments this time. This includes a bit more refactoring, as mentioned in one of the comments. It became also more apparent that the functionality from #133 is still missing.

src/framing/plaintext.rs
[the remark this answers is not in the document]
kkohbrok: I have to say, I agree with @franziskuskiefer here. There are quite a few places, where our data structures model the spec 1:1 at the cost of code complexity and I don't think that's a good trade off in most cases. Does the Signature being an `Option` type really make it that much less auditable?

src/framing/plaintext.rs
```diff
@@ -47,23 +128,125 @@ impl MLSPlaintext {
...
         credential: &Credential,
-    ) -> bool {
-        let signature_input = MLSPlaintextTBS::new_from(&self, serialized_context_option);
-        signature_input.verify(credential, &self.signature)
+    ) -> Result<(), VerificationError> {
+        let tbs_payload = MLSPlaintextTBS::new_from(&self, Some(serialized_context));
+        tbs_payload.verify(credential, &self.signature)
+    }
+
+    /// Verify the signature of an `MLSPlaintext` sent from an external party.
```
kkohbrok, line 188: I think this description doesn't quite match the function.

src/framing/plaintext.rs
```diff
@@ -47,23 +128,125 @@ impl MLSPlaintext {
...
+        membership_key: &MembershipKey,
+    ) -> Result<(), VerificationError> {
+        // Verify the signature first
+        let tbs_payload = MLSPlaintextTBS::new_from(&self, Some(serialized_context));
+        tbs_payload.verify(credential, &self.signature)?;
+
+        let tbm_payload =
+            MLSPlaintextTBMPayload::new(&tbs_payload, &self.signature, &self.confirmation_tag)
+                .map_err(VerificationError::CodecError)?;
+        let expected_membership_tag =
```
kkohbrok, line 232: Since we have a `MembershipKey` type, this whole logic should probably be in `membership_key.verify(membership_tag, ...)`.

src/tree/mod.rs
```diff
@@ -737,122 +747,70 @@ impl RatchetTree {
...
-                        &node.node,
-                        &left_hash,
-                        &right_hash,
-                    );
-                    parent_node_hash.hash(ciphersuite)
+            } else {
+                // Iterate over children until we find a node that is neither blank
+                // nor a leaf node
+                let mut child = right;
+                while child.is_parent() && !self.nodes[child].is_blank() {
```
kkohbrok, line 786: Shouldn't the condition here be exactly the inverse? Currently, we're iterating until we find a node that is either not a parent or blank.

src/tree/mod.rs
```diff
@@ -737,122 +747,70 @@ impl RatchetTree {
...
-                        }
-                    }
 
-                    NodeType::Default => {}
-                }
+    /// Verify the parent hashes of the tree nodes. Returns `true` if all parent
+    /// hashes have successfully been verified and `false` otherwise.
+    pub fn verify_parent_hashes(&self) -> bool {
+        self.nodes.iter().enumerate().all(|(index, node)| {
+            if index % 2 == 1 && node.is_full_parent() {
```
kkohbrok, line 808: Just a nit, but I think casting to a `NodeIndex` and using `is_parent` should make it a bit clearer what the condition is.

kkohbrok (changes requested):
I think there's one more bug in the parent hash verification. Is there a test where we tamper with the tree and then make sure that the parent hash fails in that case? The PR has gotten so big I've lost the overview a little bit.

src/framing/plaintext.rs
```diff
@@ -47,23 +128,125 @@
...
-        let signature_input = MLSPlaintextTBS::new_from(&self, serialized_context_option);
-        signature_input.verify(credential, &self.signature)
+    ) -> Result<(), VerificationError> {
+        let tbs_payload = MLSPlaintextTBS::new_from(&self, Some(serialized_context));
+        tbs_payload.verify(credential, &self.signature)
+    }
+
+    /// Verify the signature of an `MLSPlaintext` sent from an external party.
+    /// Returns `Ok(())` if successful or `VerificationError` otherwise.
+    pub fn verify_membership_tag(
```
franziskuskiefer, line 190: Thanks. Can you add this context to #133 for future reference please?

src/framing/plaintext.rs
```diff
@@ -145,47 +142,105 @@ impl MLSPlaintext {
...
+    pub fn verify_from_member(
+        &self,
+        ciphersuite: &Ciphersuite,
+        serialized_context: &[u8],
+        credential: &Credential,
+        membership_key: &MembershipKey,
+    ) -> Result<(), VerificationError> {
+        // Verify the signature first
+        let tbs_payload = MLSPlaintextTBS::new_from(&self, Some(serialized_context));
+        tbs_payload.verify(credential, &self.signature)?;
```
franziskuskiefer, line 227: So this is something that I can see leading to a timing oracle. because we return as soon as a verification fails. (This is another reason why there should be one function.) I'm fine with leaving it like this but again please add the potential issue to #133 so we don't forget about it.

src/framing/plaintext.rs
```diff
@@ -145,47 +142,105 @@ impl MLSPlaintext {
...
+            }
+        } else {
+            Err(VerificationError::MissingMembershipTag)
+        }
+    }
 
+    /// Verify the signature and the membership tag of an `MLSPlaintext` sent
+    /// from a group member. Returns `Ok(())` if successful or
+    /// `VerificationError` otherwise.
+    pub fn verify_from_member(
```
franziskuskiefer, line 218: Looks like there are no negative test cases for this function?

src/tree/index.rs
```diff
@@ -16,6 +16,12 @@ impl NodeIndex {
     pub fn as_usize(self) -> usize {
         self.0 as usize
     }
+    pub fn is_leaf(&self) -> bool {
+        self.0 % 2 == 0
```
franziskuskiefer, line 20: If codecov is right (and I think it is), these two functions need tests.

src/tree/mod.rs
franziskuskiefer: The entire function needs more tests, this should probably be one test case.

franziskuskiefer (changes requested):
This is getting too big to make useful reviews here (well it's been from the beginning but we tried...).
Let's add a couple more tests (see inline comments) and then get it in.

src/tree/mod.rs
raphaelrobert: I took another stab at this and refactored in a way that is as close as possible to the spec.

The test we have in place right now is not good enough to cover all code paths. Before I write (pretty complex) test cases to cover everything I need to check with Joel, if this really needs to be so complex. The iteration down the tree seems redundant to me. In the meantime I created #313.

src/tree/mod.rs
raphaelrobert: Good point, I replaced the modulo arithmetic everywhere.

src/framing/plaintext.rs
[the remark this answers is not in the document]
raphaelrobert: Sure, we can agree to be less strict in the implementation in general. In this particular case I'm not really sure what the best solution is. If we make this an optional field, it will only be set to `None` for a very short time when the struct is created. In all other instances we would have to use `unwrap()`, which is ugly and also doesn't make it much safer. I guess a better way would be to rewrite the creation part of the struct entirely.

franziskuskiefer (commented):
This slew of codecov warnings at the end makes me nervous 😬 But with the test vectors and the test framework I hop that we can get better here in future.
Only commenting here for now as you'll need to rebase anyway

src/schedule/mod.rs
```diff
@@ -238,13 +194,11 @@ pub(crate) struct WelcomeSecret {
...
-        member_secret: &MemberSecret,
-        ciphersuite: &Ciphersuite,
-    ) -> Self {
+    pub(crate) fn new(ciphersuite: &Ciphersuite, intermediate_secret: &IntermediateSecret) -> Self {
+        // Unwrapping here is safe, because we know the key is not empty
         let secret = ciphersuite
             .hkdf_expand(
-                &member_secret.secret,
+                &intermediate_secret.secret,
                 b"mls 1.0 welcome",
```
franziskuskiefer, line 202: This should be "welcome" only.

src/tree/mod.rs
```diff
@@ -737,122 +749,69 @@ impl RatchetTree {
...
-                NodeType::Default => panic!("Default node type not supported in tree hash."),
+
+    /// Verify the parent hash of a tree node. Returns `true` if the parent
+    /// hash has successfully been verified and `false` otherwise.
+    pub fn verify_parent_hash(&self, index: NodeIndex, node: &Node) -> bool {
+        // "Let L and R be the left and right children of P, respectively"
+        let left = treemath::left(index).unwrap();
+        let right = treemath::right(index, self.leaf_count()).unwrap();
+        // Unwrapping here is safe, because we know it is a full parent node
+        let parent_hash_field = &node.parent_hash().unwrap();
```
kkohbrok, line 760: I'm not sure we can say that we know that. The function is public and just takes a `&Node` and an `NodeIndex`. We don't know if it's actually a parent node (or if the unwrapping of the left and right child nodes above will panic for that matter).

src/tree/mod.rs
```diff
@@ -737,122 +749,69 @@ impl RatchetTree {
...
+        // Current hash with right child resolution
+        let current_hash_right = ParentHashInput::new(&self, index, right, parent_hash_field)
+            // It is ok to use `unwrap()` here, since we can be sure the node is not
```
kkohbrok, line 763: Again, I don't think we know that here.

kkohbrok (changes requested):
A few `unwrap` related nits, but the algorithm looks solid now! 👍🏻

src/tree/kat_encryption.rs
```diff
@@ -253,6 +261,7 @@ fn generate_test_vectors() {
         let mut group = group(ciphersuite);
         *group.epoch_secrets_mut().sender_data_secret_mut() =
             SenderDataSecret::from(sender_data_secret_bytes.as_slice());
+        *group.secret_tree_mut() = group_secret_tree;
```
franziskuskiefer, line 264: Ugh, this should probably be a `set_secret_tree` function. But looks like it's only used in tests. That's fine.

src/tree/kat_encryption.rs
```diff
@@ -329,12 +328,10 @@ fn generate_test_vectors() {
     }
 
     for ciphersuite in Config::supported_ciphersuites() {
-        for n_leaves in 1u32..50 {
+        for n_leaves in 1u32..9 {
```
franziskuskiefer, line 331: I'd prefer keeping more in here. But I see that it creates issues when re-generating. I'll make these things configurable.

src/tree/kat_encryption.rs
```diff
@@ -416,19 +411,8 @@ fn run_test_vectors() {
...
-                // The decryption doesn't check the plaintext in test mode!
                 let mls_plaintext_application = mls_ciphertext_application
-                    .to_plaintext(
-                        ciphersuite,
-                        indexed_members,
-                        epoch_secret,
-                        &mut secret_tree,
-                        context,
-                    )
+                    .to_plaintext(ciphersuite, group.epoch_secrets(), &mut secret_tree)
```
franziskuskiefer, line 415: This looks way nicer now 🙂

franziskuskiefer (changes requested):
The rebase now finally makes this impossible to review... But we sort of gave up anyway 🤷🏻 🤣 
Only one nit in the schedule (my fault, I hard coded this in the first place...)

kkohbrok (dismissed):
Looks good from my side now!

franziskuskiefer (approved):
Lets get it landed 🚢

PR #298, opened 2021-01-07
Fix CI: Forcing CI to always checkout the HEAD commit

raphaelrobert:
Fixing this issue: https://github.com/actions/checkout/issues/299

### DIFF
--- a/.github/workflows/benches.yml
+++ b/.github/workflows/benches.yml
@@ -23,5 +23,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
       - name: Benchmarks
         run: cargo bench --verbose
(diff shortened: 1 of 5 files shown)


franziskuskiefer (commented):
Hm, but this only works for PRs, not pushes to main.

raphaelrobert:
> Hm, but this only works for PRs, not pushes to main.

Were do we fix that?

franziskuskiefer:
Well, we could use the appropriate hash each time. But that probably ends up being pretty ugly. Some other comments say v1 might work (not sure if we need anything that's v2 specific)

raphaelrobert:
Why don't we try this? If it doesn't work we can always change later.

franziskuskiefer:
> Why don't we try this? If it doesn't work we can always change later.

do it!

raphaelrobert:
Can you approve? :) It will auto-merge.

franziskuskiefer (approved):

PR #299, opened 2021-01-08
Test vector generation and testing - TreeMath and Encryption

franziskuskiefer:
This implements test vector generation and testing for treemath
and encryption. It follows the basic idea of
https://github.com/mlswg/mls-implementations/blob/87e47925116b95c01db0e7b2a03b71b5a44b8a30/test-vectors.md
but slightly changes the test vector definition and uses JSON
serialised representations.
This version should be used to discuss the exact test vectors we
want for MLS and will probably change later on.

Note that generating and testing for the encryption is currently
pretty cumbersome and ugly due to OpenMLS' code structure. This
should change in future when the code is refactored to respect
logic boundaries between the different parts of MLS.


franziskuskiefer:
The codecov complains are about code that's not being compiled because of the `test-vector` feature. We'll have to improve coverage analysis at some point or wait for the refactoring to make the the feature obsolete.

.github/workflows/tests.yml
```diff
@@ -25,6 +25,8 @@ jobs:
       - uses: actions/checkout@v2
       - name: Tests
         run: cargo test --verbose
+      - name: KAT
```
kkohbrok, line 28: Why "KAT" as opposed to "Test Vectors"? Is there a difference?

src/framing/ciphertext.rs
```diff
@@ -163,6 +163,11 @@ impl MLSCiphertext {
             content: mls_ciphertext_content.content,
             signature: mls_ciphertext_content.signature,
         };
+
+        if cfg!(feature = "test-vectors") {
```
kkohbrok, line 167: If we're changing the code for test vector generation, why not just remove the code below? As far as I can see it's just for an assertion that shouldn't be there anyway.

src/tree/kat_encryption.rs
```diff
@@ -0,0 +1,425 @@
+//! # Known Answer Tests for encrypting to tree nodes
+//!
+//! This currently differs from the test vectors in https://github.com/mlswg/mls-implementations/blob/master/test-vectors.md
```
kkohbrok, line 3: I think it would be helpful here to give a brief statement of how they differ.

src/tree/kat_encryption.rs
```diff
@@ -0,0 +1,425 @@
...
+    cipher_suite: u16,
+    n_leaves: u32,
+    encryption_secret: Vec<u8>,
+    sender_data_secret: Vec<u8>,
+    leaves: Vec<LeafSequence>,
+}
+
+fn group(ciphersuite: &Ciphersuite) -> MlsGroup {
+    let credential_bundle =
+        CredentialBundle::new("Kreator".into(), CredentialType::Basic, ciphersuite.name()).unwrap();
```
kkohbrok, line 114: 🤘🏻

src/tree/kat_encryption.rs
```diff
@@ -0,0 +1,425 @@
...
+                        generation,
+                    )
+                    .expect("Error getting decryption secret");
+                handshake_keys.push(KeyAndNonce {
+                    key: handshake_secret_key.as_slice().to_vec(),
+                    nonce: handshake_secret_nonce.as_slice().to_vec(),
+                });
+
+                // Create sender_data_key/secret
+                let ciphertext = randombytes(77);
```
kkohbrok, line 263: Any particular reason why 77?

src/tree/kat_encryption.rs
```diff
@@ -0,0 +1,425 @@
...
+                );
+                assert_eq!(sender_data_info.secrets.key, sender_data_key.as_slice());
+                assert_eq!(sender_data_info.secrets.nonce, sender_data_nonce.as_slice());
+
+                // Decrypt and check message
+                let indexed_members = HashMap::new();
+                let epoch_secret = group.epoch_secrets();
+                let context = group.context();
+
+                // The decryption doesn't check the plaintext in test mode!
```
kkohbrok, line 406: What do you mean by check? Is this alluding to the early return introduced earlier, that skips the last assert?

src/tree/kat_treemath.rs
```diff
@@ -0,0 +1,130 @@
+//! # Known Answer Tests for treemath
+//!
+//! This test file generates and read test vectors for tree math.
+//! This currently differs from the test vectors in https://github.com/mlswg/mls-implementations/blob/master/test-vectors.md
```
kkohbrok, line 4: Again, a brief explanation as to how might be helpful here.

kkohbrok (commented):
Nice. It would be nice to annotate why you can just unwrap after calls to treemath that now throw errors, but this is going to be refactored with the new binary tree abstraction anyway. LGTM with a few minor suggestions.

src/tree/treemath.rs
```diff
@@ -1,10 +1,19 @@
...
-    RootHasNoParent,
+implement_error! {
+    pub enum TreeMathError {
+        LeafHasNoChildren = "Leaf nodes don't have children.",
+        RootHasNoParent = "Root nodes don't have paretns.",
+        InvalidTreeSize = "Invalid tree size.",
+    }
+}
+
+fn check_tree_size(size: LeafIndex) -> Result<(), TreeMathError> {
```
raphaelrobert, line 12: What does this function do exactly?

src/tree/treemath.rs
```diff
@@ -38,10 +47,11 @@ pub(crate) fn node_width(n: usize) -> usize {
     }
 }
 
-pub(crate) fn root(size: LeafIndex) -> NodeIndex {
+pub(crate) fn root(size: LeafIndex) -> Result<NodeIndex, TreeMathError> {
+    check_tree_size(size)?;
```
raphaelrobert, line 51: I don't understand what this check does. The `root()` function should never fail. It takes the number of leaves as an argument and any positive number of leaves constitutes a valid tree.

raphaelrobert (changes requested):
This is just a partial review of the treemath, I'll do a more thorough review of the vector generation later.

.github/workflows/tests.yml
franziskuskiefer: This is a known answer test, hence the name. The test vector generation should be moved out of this task at some point when everything is more stable.

src/framing/ciphertext.rs
franziskuskiefer: The code here is bad. But there are other parts of the code and tests that rely on this. I don't want to refactor code in this PR but generate test vectors. Changing the logic here requires more code changes.

src/tree/kat_encryption.rs
franziskuskiefer: I'll link to the issue with the discussion

src/tree/kat_encryption.rs
franziskuskiefer: not really...

src/tree/kat_encryption.rs
franziskuskiefer: Yep, i.e. the decryption here is all fine but the sender doesn't exist, hence the decryption should fail in the openmls logic. But we really want to test encryption/decryption here. After refactoring there should be a decryption function on the tree that we can call here.

src/tree/treemath.rs
franziskuskiefer: Well, the function doesn't always produce a valid output. This should never happen but the math here doesn't know this.

raphaelrobert: Can you give an example of invalid output?

If the output is invalid, the math should be fixed.

src/tree/treemath.rs
raphaelrobert: 👀

kkohbrok (approved):
Looks good to me!

src/tree/secret_tree.rs
```diff
@@ -203,6 +209,7 @@ impl SecretTree {
     ) -> Result<RatchetSecrets, SecretTreeError> {
         // Check tree bounds
         if index >= self.size {
+            println!("index: {:?} | size: {:?}", index, self.size);
```
raphaelrobert, line 212: debug output

src/tree/secret_tree.rs
```diff
@@ -7,11 +7,17 @@ use crate::tree::{index::*, sender_ratchet::*, treemath::*};
...
-    IndexOutOfBounds,
+implement_error! {
+    pub enum SecretTreeError {
+        Simple {
+            TooDistantInThePast = "Generation is too old to be processed.",
+            TooDistantInTheFuture = "Generation is too far in the future to be processed.",
+            IndexOutOfBounds = "Index out of bounds",
+        }
+        Complex {
+            TreeMath(TreeMathError) = "An error in the tree arithmetic occurred.",
```
raphaelrobert, line 18: Is this error thrown anywhere? I don't see it in this PR.

src/framing/ciphertext.rs
raphaelrobert: Yeah, this looks like it's fixing the assertion only. The assertion will be removed in #293 anyway.

raphaelrobert (dismissed):
Generally fine, although I didn't review the test vector generation in depth, given that it is an ongoing discussion on the `mlswg/mls-implementations`.

We used to have the treemath test vectors as part of the repository, but I don't think this makes sense for interop. The test vectors should be hosted on `mlswg/mls-implementations` instead and pulled from there for testing.

franziskuskiefer:
> Generally fine, although I didn't review the test vector generation in depth, given that it is an ongoing discussion on the `mlswg/mls-implementations`.
> 
> We used to have the treemath test vectors as part of the repository, but I don't think this makes sense for interop. The test vectors should be hosted on `mlswg/mls-implementations` instead and pulled from there for testing.

Yeah once the test vectors have stabilised we should put them in `mlswg/mls-implementations`. Until then we should keep a copy here in the repo.

kkohbrok (approved):

PR #305, opened 2021-01-13
Compute the correct index to retrieve the decryption key from

kkohbrok:
Previously, we used the wrong index (`common_ancestor_copath_index`) when trying to retrieve non-leaf decryption keys from the `PrivateTree`. This PR corrects the index computation, by checking the resolution and finding an index that is on the direct path of the decryptor.

This PR also adds a new function to the test framework that allows removal of group members by index.


src/tree/mod.rs
```diff
@@ -328,12 +339,13 @@ impl RatchetTree {
         // Get the HPKE private key.
         // It's either the own key or must be in the path of the private tree.
         let private_key = if resolution[position_in_resolution] == own_index {
+            println!("Decrypting using leaf key at index {:?}", own_index);
```
franziskuskiefer, line 342: Remove debug output. If you want logging, add logging.

franziskuskiefer (changes requested):
This needs a test to ensure that the change is correct.

src/group/mls_group/new_from_welcome.rs
```diff
@@ -121,7 +121,7 @@ impl MlsGroup {
                 tree.own_node_index(),
                 NodeIndex::from(group_info.signer_index()),
             );
-            let common_path = treemath::direct_path_root(common_ancestor_index, tree.leaf_count())
+            let common_path = treemath::dirpath_long(common_ancestor_index, tree.leaf_count())
```
raphaelrobert, line 124: The common path now also includes the leaf with that change. That seems wrong conceptually, since the own leaf node cannot be on the same path as any other node.

raphaelrobert (changes requested):
I'm wondering if it makes sense to extract this part of the logic, because it relies so heavily on treemath. If it wasn't for the resolution itself, it could move there altogether.
That would make it much easier to test.

src/group/mls_group/new_from_welcome.rs
kkohbrok: The common path is computed from the `comon_ancestor_index`, not from the leaf. So it's meant to include that node as well, which I think is correct. I'd like to keep this here for now. I'll see if I can extract it into the new binary tree interface.

tests/test_decryption_key_index.rs
```diff
@@ -0,0 +1,60 @@
...
+    let handshake_message_format = HandshakeMessageFormat::Plaintext;
+    let update_policy = UpdatePolicy::default();
+    let callbacks = default_callbacks::default_callbacks();
+    let managed_group_config =
+        ManagedGroupConfig::new(handshake_message_format, update_policy, callbacks);
+    let number_of_clients = 20;
+    let setup = ManagedTestSetup::new(managed_group_config, number_of_clients);
+    setup.create_clients();
+
+    for ciphersuite in Config::supported_ciphersuites() {
```
franziskuskiefer, line 19: Please make one test for each cipher suite.

tests/test_decryption_key_index.rs
```diff
@@ -0,0 +1,60 @@
...
+        // Create a basic group with more than 4 members to create a tree with intermediate nodes.
+        let group_id = setup.create_random_group(10, ciphersuite).unwrap();
+        let mut groups = setup.groups.borrow_mut();
+        let group = groups.get_mut(&group_id).unwrap();
+
+        // Now we have to create a situation, where the resolution is neither
+        // the leaf, nor the common ancestor closest to the root. To do that, we
+        // first have the member at index 0 remove the one at index 2, thus
+        // populating its own parent node.
+        let (_, remover_id) = &group.members[0].clone();
```
franziskuskiefer, line 29: This looks like it might be nice to be able to remove a client with an index instead of the id. If we need something like this more often, we might want to add this to the test framework.

tests/test_decryption_key_index.rs
```diff
@@ -0,0 +1,60 @@
...
+            .remove_clients(
+                ActionType::Commit,
+                group,
+                &remover_id,
+                vec![target_id.clone()],
+            )
+            .unwrap();
+
+        // Then we have the member at index 7 (it's just index 6 in the member
+        // list) remove the one at index 3 (index 2 at the member list). This
```
franziskuskiefer, line 41: Why are these numbers so weird? It wasn't like this above.

franziskuskiefer (changes requested):
Generally lgtm. Only two nits.

tests/test_decryption_key_index.rs
kkohbrok: Is there a particular reason why this would make sense in this test in particular (or in general)? For example, in  `test_group`, we also have a single test function that simply iterates over each ciphersuite in the same way. Doesn't the loop make sense for tests that we don't expect to be particularly ciphersuite-dependent?

tests/test_decryption_key_index.rs
kkohbrok: True. I'll add it.

tests/test_decryption_key_index.rs
kkohbrok: The index in the tree don't change, but the membership list becomes shorter and the position of the members change. I'll change it such that I retrieve the identities before I start removing members.

tests/test_decryption_key_index.rs
franziskuskiefer: Having them in one function makes debugging way harder (you have no idea which ciphersuite breaks). I want to have all tests parameterized on the cipher suite eventually. It just needs some boilerplate macros.

kkohbrok: Macros would make sense. But until then, can't we just stick with loops in places where it's not ciphersuite specific? We have to touch the code anyway when we have the macro.

For example, in follow-up PR #312 , I'm going to add the tests for the interop scenarios. Doing one test per ciphersuite would be quite a bit more code.

tests/test_decryption_key_index.rs
kkohbrok: I've now changed it such that the `remover_id` is explicitly retrieved via its index. It's not a lot prettier, but it's more explicit and less confusing.

tests/test_decryption_key_index.rs
kkohbrok: Done!

tests/test_decryption_key_index.rs
```diff
@@ -26,30 +26,32 @@ fn test_decryption_key_index_computation() {
         // the leaf, nor the common ancestor closest to the root. To do that, we
         // first have the member at index 0 remove the one at index 2, thus
         // populating its own parent node.
-        let (_, remover_id) = &group.members[0].clone();
-        let (_, target_id) = &group.members[2].clone();
+
+        // Find the identity of the member with leaf index 0.
+        let (_, remover_id) = &group
```
franziskuskiefer, line 31: This should probably a function on the test framework (`member_by_index` or something). But let's leave it to a follow up.

franziskuskiefer (dismissed):
One nit, but feel free to leave it to a follow up.

src/group/mls_group/new_from_welcome.rs
raphaelrobert: The part that I don't understand is that you now have a `common_path` that is longer than previously. Why does that not lead to errors?

While you are here, could you also get rid of the `expect()` and propagate the error? Or convert it to an `unwrap()` with a proper explanation.

raphaelrobert (changes requested):
Looks good in general, just one question and a clippy warning.

src/group/mls_group/new_from_welcome.rs
kkohbrok: It's not necessarily longer. If I understand correctly, it's only longer if the common ancestor is not the root. And since the bug essentially occurs when that is the case, I think that it now works correctly and it previously didn't (or can't have).

kkohbrok (commented):
I hope I addressed all the comments. The clippy warning is in a part of `tree/mod.rs` that I didn't touch. I can fix it, if you want, but it's unrelated to fixing this bug.

src/tree/mod.rs
```diff
@@ -293,11 +293,11 @@ impl RatchetTree {
         group_context: &[u8],
         new_leaves_indexes: HashSet<&NodeIndex>,
     ) -> Result<&CommitSecret, TreeError> {
-        let own_index = self.own_node_index();
+        let own_index = NodeIndex::from(self.own_node_index());
```
franziskuskiefer, line 296: `own_node_index` is not `NodeIndex`? 🤔

franziskuskiefer (approved):
lgtm, the comment is something we might want to fix in a follow up though...

src/tree/mod.rs
raphaelrobert: The own index is a leaf index, which is correct.

raphaelrobert (approved):

src/tree/mod.rs
kkohbrok: I think the issue is with the name.

Issue #219, opened 2020-11-19
Create testing framework for managed group API


kkohbrok:
It would be useful to have a testing framework for the group API similar to the existing one for the low-level API. That would make testing of more complex cases easier and it should also be relatively easy to do.

franziskuskiefer:
Collecting a couple follow up items here. Please move to a separate issue when you're done:
* Make cipher suite support more variable
* Allow using `group_config` and `public_tree` from `Welcome` message
* Allow delivering `public_tree` out of band

Can we try to get [these scenarios](https://github.com/mlswg/mls-implementations/blob/master/test-scenarios.md) implemented with the framework (follow-up)?

PR #308, opened 2021-01-15
Introduce a test framework for the managed group API

kkohbrok:
This PR introduces a framework to make testing the managed group API easier. Note, that I re-oranized the test utils, so `mls_utils` is now the module housing the existing test framework for the `mls_group` API. That is also the reason why the PR looks so big.

I added `Node` to the prelude, which I think is useful, because we export the public tree as `Vec<Option<Node>>`.

I've also added the test frameworks to Codecov's `ignore` list. FWIW, I have two follow-up PRs lined up to do some testing with the test framework, which should test the testing framework somewhat.

Sorry about the mess that are the commits. They'll get squashed anyway before merging to `main`.


tests/utils/managed_utils/mod.rs
```diff
@@ -0,0 +1,583 @@
...
+pub struct KeyStore {
+    // Maps a client Id and a ciphersuite to a CredentialBundle.
+    credential_bundles: HashMap<(Vec<u8>, CiphersuiteName), CredentialBundle>,
+}
+
+impl<'ks> KeyStore {
+    pub(crate) fn store_credentials(
+        &mut self,
+        client_id: Vec<u8>,
+        credential_bundles: Vec<(CiphersuiteName, CredentialBundle)>,
```
franziskuskiefer, line 26: This could take a slice, but vec is fine I guess

tests/utils/managed_utils/mod.rs
```diff
@@ -0,0 +1,583 @@
...
+#[derive(Debug)]
+pub struct KeyStore {
+    // Maps a client Id and a ciphersuite to a CredentialBundle.
+    credential_bundles: HashMap<(Vec<u8>, CiphersuiteName), CredentialBundle>,
+}
+
+impl<'ks> KeyStore {
+    pub(crate) fn store_credentials(
+        &mut self,
+        client_id: Vec<u8>,
```
franziskuskiefer, line 25: This shuold be a slice.

tests/utils/managed_utils/mod.rs
```diff
@@ -0,0 +1,583 @@
...
+        credential_bundles: Vec<(CiphersuiteName, CredentialBundle)>,
+    ) {
+        for (cn, cb) in credential_bundles {
+            self.credential_bundles.insert((client_id.clone(), cn), cb);
+        }
+    }
+
+    pub(crate) fn get_credential(
+        &self,
+        client_id: &Vec<u8>,
```
franziskuskiefer, line 35: slice

tests/utils/managed_utils/mod.rs
```diff
@@ -0,0 +1,583 @@
...
+    pub group_config: ManagedGroupConfig,
+    pub public_tree: Vec<Option<Node>>,
+}
+
+impl Group {
+    pub fn random_group_member(&self) -> Vec<u8> {
+        let index = (OsRng.next_u32() as usize) % self.members.len();
+        // We can unwrap here, because the index is scoped with the size of the
+        // HashSet.
+        let (_, identity) = self.members.iter().nth(index).unwrap().clone();
```
franziskuskiefer, line 57: `self.members.get(index)`? Or because we know it must exist `self.members[index]`?

tests/utils/managed_utils/mod.rs
```diff
@@ -0,0 +1,583 @@
...
+        // TODO: Do both things.
+        let action_type = match (OsRng.next_u32() as usize) % 2 {
+            0 => ActionType::Proposal,
+            1 => ActionType::Commit,
+            _ => return Err(SetupError::Unknown),
+        };
+
+        // TODO: Do multiple things.
+        let operation_type = match (OsRng.next_u32() as usize) % 3 {
+            0 => OperationType::Update,
```
franziskuskiefer, line 517: If you don't use the enum anywhere else, this looks a little overcomplicated. Why not just match the random number?

tests/utils/managed_utils/mod.rs
```diff
@@ -0,0 +1,583 @@
...
+            .ok_or(SetupError::UnknownClientId)?
+            .borrow();
+        let mut member_groups = client.groups.borrow_mut();
+        let member_group_state = member_groups
+            .get_mut(&group.group_id)
+            .ok_or(SetupError::ClientNotInGroup)?;
+        let (messages, welcome_option) = match action_type {
+            ActionType::Commit => {
+                // Let the function generate the key package bundle.
+                member_group_state.self_update(None)?
```
franziskuskiefer, line 375: So no way to test with an externally generated key package bundle?

tests/utils/managed_utils/mod.rs
```diff
@@ -0,0 +1,583 @@
...
+
+pub enum OperationType {
+    Update,
+    Add,
+    Remove,
+}
+
+pub struct ManagedTestSetup<'client_lifetime> {
+    pub number_of_clients: usize,
+    // The clients identity is its position in the vector in be_bytes.
```
franziskuskiefer, line 75: 😬 I don't think I like this approach for identities. But it's as good or bad as any other thing we can do here.
We shoudl have a function to get the identity for a client though, i.e. don't duplicate `i.to_be_bytes().to_vec()`.

tests/utils/managed_utils/mod.rs
```diff
@@ -0,0 +1,583 @@
...
+    /// reasons, `create_clients` has to be called in addition with the same
+    /// number of clients.
+    pub fn new(default_mgc: ManagedGroupConfig, number_of_clients: usize) -> Self {
+        let mut key_store = KeyStore {
+            credential_bundles: HashMap::new(),
+        };
+        // Create credentials first to avoid borrowing issues.
+        for i in 0..number_of_clients {
+            let identity = i.to_be_bytes().to_vec();
+            // For now, everyone supports all ciphersuites.
```
franziskuskiefer, line 116: Another thing to extend later.

tests/utils/managed_utils/mod.rs
```diff
@@ -0,0 +1,583 @@
...
+        let key_package = client.get_fresh_key_package(ciphersuite)?;
+        println!("Storing key package with hash: {:?}", key_package.hash());
+        self.waiting_for_welcome
+            .borrow_mut()
+            .insert(key_package.hash(), client.identity.clone());
+        Ok(key_package)
+    }
+
+    /// Deliver a Welcome message to group `group` to the intended recipients.
+    pub fn deliver_welcome(&self, welcome: Welcome, group: &Group) -> Result<(), SetupError> {
```
franziskuskiefer, line 182: You pass in the `group` here and pull information from it. If that's intended in you test setup, it needs description of how it is supposed to be used and how it allows testing things. In the real world the `group_config` must be extracted from the `Welcome` and the `public_tree` might be in there as well, or delivered separately. (All things that are impossible to test with this function.)

tests/utils/managed_utils/mod.rs
```diff
@@ -0,0 +1,583 @@
...
+        group.members = sender
+            .get_members_of_group(&group.group_id)?
+            .iter()
+            .map(|(index, cred)| (index.clone(), cred.identity().clone()))
+            .collect();
+        group.public_tree = sender_group.export_ratchet_tree();
+        drop(sender_group);
+        drop(sender_groups);
+        println!("Group members after distribution:");
+        // Check that the group states of all members match.
```
franziskuskiefer, line 239: Why? I can see that this is a nice thing to check in many cases. But there are at least as many interesting cases where this really shouldn't be the case.

tests/utils/managed_utils/mod.rs
```diff
@@ -0,0 +1,583 @@
...
+        for (index, m_id) in &group.members {
+            println!("Id: {:?}, Index: {:?}", m_id, index);
+            let m = clients.get(m_id).unwrap().borrow();
+            let group_states = m.groups.borrow_mut();
```
franziskuskiefer, line 243: Why do you need this mutably?

tests/utils/managed_utils/mod.rs
```diff
@@ -0,0 +1,583 @@
...
+        group: &Group,
+        number_of_members: usize,
+    ) -> Result<Vec<Vec<u8>>, SetupError> {
+        let clients = self.clients.borrow();
+        if number_of_members + group.members.len() >= clients.len() {
+            return Err(SetupError::NotEnoughClients);
+        }
+        let mut new_member_ids = Vec::new();
+        for _ in 0..number_of_members {
+            let new_member_id = clients
```
franziskuskiefer, line 266: Ugh... So you want to move some of the closures here out to functions with names?

tests/utils/managed_utils/mod.rs
```diff
@@ -0,0 +1,583 @@
...
+                        .is_none())
+                        && (new_member_ids
+                            .iter()
+                            .find(|&member_id| member_id == client_id)
+                            .is_none())
+                })
+                .unwrap();
+            new_member_ids.push(new_member_id.clone());
+        }
+        drop(clients);
```
franziskuskiefer, line 282: Why do you need this? 🤔

tests/utils/managed_utils/mod.rs
```diff
@@ -0,0 +1,583 @@
...
+                        .members
+                        .iter()
+                        .find(|&(_, member_id)| client_id == member_id)
+                        .is_none())
+                        && (new_member_ids
+                            .iter()
+                            .find(|&member_id| member_id == client_id)
+                            .is_none())
+                })
+                .unwrap();
```
franziskuskiefer, line 279: So this panics if there's no client we could add?

franziskuskiefer (changes requested):
I skipped some parts in `managed_utils/mod.rs` because I'm not sure how all this is
supposed to work.

We need at least one very basic example here that uses the test framework to
know how you intend it to be used. (You can add doc tests if you don't want to
add "real" ones.)
A description of how it works and how it enables test coverage would be also nice.

You generate random numbers and then compute them modulo to get something in the
desired range. This will get you skewed results. That's not super important here
but it will certainly lead to some things being tested more than others.

Some things can't be (nicely) tested with this. I made comments where I see issues
(I'm sure there are more). We don't have to fix all of them but file a follow
up issue with everything we know would be good to have in here but is currently
missing.

All the `drop`s make me think that the code is probably not ideal. But we can
refactor this later.

tests/utils/mls_utils/mod.rs
```diff
@@ -0,0 +1,297 @@
...
+
+pub(crate) fn random_usize() -> usize {
+    OsRng.next_u64() as usize
+}
+
+pub(crate) fn randombytes(n: usize) -> Vec<u8> {
+    get_random_vec(n)
+}
+
+// Not currently used.
```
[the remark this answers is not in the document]
kkohbrok: I only moved the `mls_utils`part and did not otherwise touch it. I'd rather not change anything there.

tests/utils/mls_utils/mod.rs
```diff
@@ -0,0 +1,297 @@
...
+        // Create the client.
+        let test_client = TestClient {
+            credential_bundles,
+            key_package_bundles: RefCell::new(key_package_bundles),
+            group_states: RefCell::new(HashMap::new()),
+        };
+        test_clients.insert(client.name, RefCell::new(test_client));
+    }
+    // Initialize all of the groups, each group gets assigned a sequential group
+    // id. TODO: Depending on the use case, it might be hard to figure out which
```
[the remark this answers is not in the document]
kkohbrok: See above. I only moved the `mls_utils` code and would rather not touch it any further in this PR.

tests/utils/mls_utils/mod.rs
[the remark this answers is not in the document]
franziskuskiefer: If you don't change anything in here, please move it to a separate PR.

src/prelude.rs
```diff
@@ -25,6 +25,7 @@ pub use crate::messages::{
     proposals::{AddProposal, RemoveProposal, UpdateProposal},
     Welcome,
 };
+pub use crate::tree::node;
```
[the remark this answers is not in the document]
kkohbrok: I can also put it in `lib.rs`. But I think the DS will want access to this. Not only for indices, but also, for example, to pull credentials from the public tree.

tests/utils/managed_utils/mod.rs
```diff
@@ -0,0 +1,583 @@
+#![allow(dead_code)]
+/// We allow dead code here due to the following issue:
+/// https://github.com/rust-lang/rust/issues/46379, which would otherwise create
```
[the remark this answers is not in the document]
kkohbrok: When I remove this and I run `cargo test`, I get a ton of `function is never used` warnings for functions that are used either in other functions of the framework or in functions of existing tests (that are not (yet) part of this PR). I think that's related to the issue I linked, but may have other causes. Can you do me a favor and test this on your machine? It could also just be a local issue.

tests/utils/managed_utils/mod.rs
kkohbrok: Right. This is a relic of when `self.members` used to be a `HashMap`.

tests/utils/managed_utils/mod.rs
kkohbrok: I used to match the random number, but I thought using the enum would make it a bit more readable. If you prefer matching the number, I'm happy to restore that.

tests/utils/managed_utils/mod.rs
kkohbrok: Good point, we should probably allow that.

tests/utils/managed_utils/mod.rs
kkohbrok: Feel free to suggest another approach. I'm not particularly attached to this one, so I'm happy to replace it with something else. This just seemed convenient to implement.

tests/utils/managed_utils/mod.rs
kkohbrok: Yes, I'll add an issue for that once this PR lands.

tests/utils/managed_utils/mod.rs
kkohbrok: Fair point. I guess I wrote this function mostly with `pub(crate)` use in mind, but it's certainly something we will want to use outside. I should probably switch from passing in `&Group` everywhere to passing in `GroupId`.

tests/utils/managed_utils/mod.rs
kkohbrok: Fair point. Maybe it would be worth passing in a switch that enables/disables this behaviour or moving it to an external function of `Group`.

tests/utils/managed_utils/mod.rs
kkohbrok: Right, this is very ugly, but I didn't really know how to do it better. I think moving out the closures won't do that much. Maybe it would be worth using `HashSet`s here?

tests/utils/managed_utils/mod.rs
kkohbrok: Well, there really should be one. We check in the beginning if enough clients exist.

kkohbrok:
Thanks for the thorough review! You raise some very good points.

> I skipped some parts in `managed_utils/mod.rs` because I'm not sure how all this is
> supposed to work.

I have more documentation on the way. Sorry about that.

> We need at least one very basic example here that uses the test framework to
> know how you intend it to be used. (You can add doc tests if you don't want to
> add "real" ones.)
> A description of how it works and how it enables test coverage would be also nice.

No problem. I have a few lined up already and will add one of them to the PR.

> You generate random numbers and then compute them modulo to get something in the
> desired range. This will get you skewed results. That's not super important here
> but it will certainly lead to some things being tested more than others.

Good point. I agree that this is pretty ad-hoc. Is there a more way standard way of doing this?

> Some things can't be (nicely) tested with this. I made comments where I see issues
> (I'm sure there are more). We don't have to fix all of them but file a follow
> up issue with everything we know would be good to have in here but is currently
> missing.

Another good point. So far, I've mostly geared the framework to enable testing if managed group works in cases it should work. I agree that there should be room for more direct manipulation and testing failure cases as well.

> All the `drop`s make me think that the code is probably not ideal. But we can
> refactor this later.

I'm also very unhappy with all the drops, but from how I understand it, dealing with the lifetimes necessitates the `RefCell`s, which in turn necessitate dropping state every now and again. I'll try to see if I can get rid of some more of them.

tests/utils/mls_utils/mod.rs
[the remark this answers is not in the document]
kkohbrok: Done!

tests/utils/mls_utils/mod.rs
```diff
@@ -0,0 +1,297 @@
+#![allow(dead_code)]
```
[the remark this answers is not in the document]
kkohbrok: The way I understand https://github.com/rust-lang/rust/issues/46379, non-test modules in `tests` get compiled individually with each test. At least on my machine, this means I get warnings, because not every test uses every function of the framework and reports the non-used functions as dead code. Not sure if there is a really good solution to this.

tests/utils/mod.rs
```diff
@@ -1,296 +1,2 @@
...
-        config: group_config,
-        members: vec![test_client_config_a.clone(), test_client_config_b.clone()],
-    };
-    let test_setup_config = TestSetupConfig {
-        clients: vec![test_client_config_a, test_client_config_b],
-        groups: vec![test_group_config],
-    };
-    let _test_setup = setup(test_setup_config);
-}
+pub mod managed_utils;
```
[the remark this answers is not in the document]
kkohbrok: Done!

src/prelude.rs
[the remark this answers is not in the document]
kkohbrok: Done!

tests/utils/managed_utils/client.rs
```diff
@@ -0,0 +1,142 @@
+use std::{cell::RefCell, collections::HashMap};
```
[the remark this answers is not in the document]
kkohbrok: Done!

tests/utils/managed_utils/client.rs
```diff
@@ -0,0 +1,142 @@
...
+use openmls::prelude::{node::Node, *};
+
+use super::{errors::ClientError, KeyStore};
+
+#[derive(Debug)]
+pub struct Client<'managed_group_lifetime> {
+    /// Name of the client.
+    pub(crate) identity: Vec<u8>,
+    /// Ciphersuites supported by the client.
+    pub(crate) _ciphersuites: Vec<CiphersuiteName>,
```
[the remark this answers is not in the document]
kkohbrok: Done!

tests/utils/managed_utils/client.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub(crate) key_store: &'managed_group_lifetime KeyStore,
```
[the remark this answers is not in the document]
kkohbrok: Done!

tests/utils/managed_utils/client.rs
```diff
@@ -0,0 +1,142 @@
...
+    // Map from key package hash to the corresponding bundle.
+    pub(crate) key_package_bundles: RefCell<HashMap<Vec<u8>, KeyPackageBundle>>,
+    pub(crate) groups: RefCell<HashMap<GroupId, ManagedGroup<'managed_group_lifetime>>>,
+}
+
+impl<'managed_group_lifetime> Client<'managed_group_lifetime> {
+    pub fn get_fresh_key_package(
```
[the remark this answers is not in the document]
kkohbrok: Done!

tests/utils/managed_utils/client.rs
```diff
@@ -0,0 +1,142 @@
...
+        &self,
+        ciphersuite: &Ciphersuite,
+    ) -> Result<KeyPackage, ClientError> {
+        let credential_bundle = self
+            .key_store
+            .get_credential(&self.identity, ciphersuite.name())
+            .ok_or(ClientError::CiphersuiteNotSupported)?;
+        let mandatory_extensions = Vec::new();
+        let key_package_bundle: KeyPackageBundle = KeyPackageBundle::new(
+            &[ciphersuite.name()],
```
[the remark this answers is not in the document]
kkohbrok: Done! Now actually uses the first ciphersuite in the list and just announces the rest.

tests/utils/managed_utils/client.rs
```diff
@@ -0,0 +1,142 @@
...
+            .insert(key_package_bundle.key_package().hash(), key_package_bundle);
+        Ok(key_package)
+    }
+
+    pub fn create_group(
+        &self,
+        group_id: GroupId,
+        managed_group_config: ManagedGroupConfig,
+        ciphersuite: &Ciphersuite,
+    ) -> Result<Vec<Option<Node>>, ClientError> {
```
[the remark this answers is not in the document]
kkohbrok: Yes, that didn't really make sense. Fixed!

tests/utils/managed_utils/default_callbacks.rs
```diff
@@ -0,0 +1,57 @@
...
+                "A Commit message with one ore more invalid proposals was received: {:?}",
+                e
+            );
+        }
+        InvalidMessageError::GroupError(e) => {
+            println!("An error in the managed group occurred: {:?}", e);
+        }
+    }
+}
+fn error_occured(managed_group: &ManagedGroup, error: ManagedGroupError) {
```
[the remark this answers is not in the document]
kkohbrok: Done!

tests/utils/managed_utils/mod.rs
```diff
@@ -0,0 +1,583 @@
...
+pub mod default_callbacks;
+pub mod errors;
+
+use self::client::*;
+use self::errors::*;
+
+#[derive(Debug)]
+pub struct KeyStore {
+    // Maps a client Id and a ciphersuite to a CredentialBundle.
+    credential_bundles: HashMap<(Vec<u8>, CiphersuiteName), CredentialBundle>,
```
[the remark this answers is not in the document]
kkohbrok: Done!

tests/utils/managed_utils/mod.rs
kkohbrok: Done!

tests/utils/managed_utils/mod.rs
kkohbrok: Done!

tests/utils/managed_utils/mod.rs
kkohbrok: Done!

tests/utils/managed_utils/mod.rs
kkohbrok: Made the documentation more precise.

tests/utils/managed_utils/mod.rs
kkohbrok: Moved it to an external function that can be called independently.

tests/utils/managed_utils/mod.rs
kkohbrok: Done!

tests/utils/managed_utils/mod.rs
kkohbrok: Done!

tests/utils/managed_utils/mod.rs
kkohbrok: Removed!

kkohbrok:
I made a few passes and (I think I) addressed all reviewer comments. In particular, I added a lot of documentation and changed the structure to get rid of quite a few `drop`s. I also added a brief example that used the framework. Bigger tests using the framework will follow along with a few bugfixes 🙂

tests/test_managed_api.rs
```diff
@@ -0,0 +1,38 @@
...
+        setup
+            .add_clients(ActionType::Commit, group, &adder_id, new_members)
+            .unwrap();
+
+        // Remove a member
+        let (_, remover_id) = group.members[2].clone();
+        let (_, target_id) = group.members[3].clone();
+        setup
+            .remove_clients(ActionType::Commit, group, &remover_id, vec![target_id])
+            .unwrap();
```
franziskuskiefer, line 36: Ok so this is how I use the framework to create a group and add/remove members to/from a group. But how do I test something. I guess we always want to verify the group state after an operation or at the end. Could you do this here so we know how to then write other tests?

tests/utils/managed_utils/client.rs
```diff
@@ -108,18 +123,26 @@ impl<'managed_group_lifetime> Client<'managed_group_lifetime> {
...
+    /// messages.
+    pub fn receive_messages_for_group(&self, messages: &[MLSMessage]) -> Result<(), ClientError> {
         let mut group_states = self.groups.borrow_mut();
-        let group_state = group_states
-            .get_mut(group_id)
-            .ok_or(ClientError::NoMatchingGroup)?;
-        Ok(group_state.process_messages(messages)?)
+        for message in messages {
+            let group_id = GroupId {
+                value: message.group_id(),
```
franziskuskiefer, line 133: This will not be exposed in future (when we finally cleaned up the API). You can use `GroupId::from_slice`. Or you make your `self.groups` use the raw byte vector instead of the `GroupId`.

tests/utils/managed_utils/client.rs
```diff
@@ -139,4 +162,73 @@ impl<'managed_group_lifetime> Client<'managed_group_lifetime> {
...
+    /// Have the client either propose or commit (depending on the
+    /// `action_type`) removing the clients with the given indices from the
+    /// group with the given group id. Returns an error if no group with the
+    /// given group id can be found or if an error occurs while performing the
+    /// remove operation.
+    pub fn remove_members(
+        &self,
+        action_type: ActionType,
+        group_id: &GroupId,
+        target_indices: &[usize],
```
franziskuskiefer, line 222: 🙈

tests/utils/managed_utils/mod.rs
```diff
@@ -232,26 +263,34 @@ impl<'ks> ManagedTestSetup<'ks> {
...
-            drop(group_states);
         }
-        Ok(())
     }
 
     /// Get `number_of_members` new members for the given group `group` for use
-    /// with the `add_members` function.
+    /// with the `add_members` function. If not enough clients are left that are
+    /// not already members of this group, this function will return an error.
+    /// TODO: Make this function ensure that the given members support the
```
franziskuskiefer, line 292: Add issue number or something.

franziskuskiefer (changes requested):
A couple more things.

The easiest thing for generating random numbers in a range is [gen_range](https://docs.rs/rand/0.8.2/rand/trait.Rng.html#method.gen_range).

I added a comment to #219 with some of the things we should add to the framework later.

tests/utils/managed_utils/mod.rs
kkohbrok: Done! #310.

kkohbrok (commented):
I made another pass and (hopefully) addressed all comments.

tests/utils/managed_utils/mod.rs
franziskuskiefer: Yeah, you'd have to consume it.

tests/utils/managed_utils/client.rs
```diff
@@ -0,0 +1,142 @@
...
+            &[ciphersuite.name()],
+            &credential_bundle,
+            mandatory_extensions,
+        )
+        .unwrap();
+        let key_package = key_package_bundle.key_package().clone();
+        self.key_package_bundles
+            .borrow_mut()
+            .insert(key_package_bundle.key_package().hash(), key_package_bundle);
+        Ok(key_package)
```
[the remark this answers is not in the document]
franziskuskiefer: You already have the lifetimes and such. So should work. But lets keep it like this for now.

franziskuskiefer (approved):
Ok let's get it in! 🙂

PR #312, opened 2021-01-21
Introduce tests corresponding to the Interop test scenarios

### LINKED ISSUE #311: [Feature request] Implement interop test scenarios
Follow-up for #308. Documented [here](https://github.com/mlswg/mls-implementations/blob/master/test-scenarios.md).

kkohbrok:
Includes parts for #311.

It does not fix: 
* no X509 credentials are used
* No tests for
  * External join
  * External PSK
  * Resumption
  * Re-Init

The large group lifecycle test currently still panics. That should go away with #305.

This PR also includes some minor fixes for the test setup.


tests/test_interop_scenarios.rs
```diff
@@ -0,0 +1,263 @@
+use openmls::prelude::*;
+use std::convert::TryFrom;
+use test_macros::ctest;
+
+mod utils;
+
+use utils::managed_utils::*;
+
+// The following tests are test corresponding the the interop test scenarios
```
franziskuskiefer, line 9: " tests are test"

tests/test_interop_scenarios.rs
```diff
@@ -0,0 +1,263 @@
...
+// using BasicCredentials. We can change the test setup once #134 is fixed.
+
+fn default_managed_group_config() -> ManagedGroupConfig {
+    let handshake_message_format = HandshakeMessageFormat::Plaintext;
+    let update_policy = UpdatePolicy::default();
+    let callbacks = default_callbacks::default_callbacks();
+    ManagedGroupConfig::new(handshake_message_format, update_policy, 10, callbacks)
+}
+
+/// # 1:1 join
```
franziskuskiefer, line 22: You can't do doc comments here. Just make all of them normal ones.

tests/test_interop_scenarios.rs
```diff
@@ -0,0 +1,263 @@
...
+    let group = groups.get_mut(&group_id).unwrap();
+
+    let (_, alice_id) = group.members.first().unwrap().clone();
+
+    // A vector including Bob's id.
+    let bob_id = setup.random_new_members_for_group(group, 1).unwrap();
+
+    // Create the add commit and deliver the welcome.
+    setup
+        .add_clients(ActionType::Commit, group, &alice_id, bob_id)
```
franziskuskiefer, line 86: So the way I understand the test case there shouldn't be a commit here but only after everyone is added. Not sure what exactly the difference then is to the next test case though. Either the test or the description need to be updated. I guess the Welcome is supposed to also imply commit.

tests/test_interop_scenarios.rs
```diff
@@ -0,0 +1,262 @@
...
+    // "Large" is 20 for now.
+    let number_of_clients = 20;
+    let setup = ManagedTestSetup::new(default_managed_group_config(), number_of_clients);
+    setup.create_clients();
+
+    // Create a group with twenty members. The process includes creating a
+    // one-person group and then adding new members in bunches of up to 5,
+    // each bunch by a random group member.
+    let group_id = setup
+        .create_random_group(20, ciphersuite)
```
franziskuskiefer, line 232: `number_of_clients`

tests/test_interop_scenarios.rs
```diff
@@ -0,0 +1,262 @@
...
+/// * All members update
+/// * While the group size is >1, a randomly-chosen group member removes a
+///   randomly-chosen other group member
+ctest!(large_group_lifecycle {
+    let ciphersuite_name = CiphersuiteName::try_from(_ciphersuite_code).unwrap();
+    println!("Testing ciphersuite {:?}", ciphersuite_name);
+    let ciphersuite = Config::ciphersuite(ciphersuite_name).unwrap();
+
+    // "Large" is 20 for now.
+    let number_of_clients = 20;
```
franziskuskiefer, line 224: If i make this 200 it fails quickly. Something appears to be broken.

franziskuskiefer (changes requested):
Generally looks good. A couple nits.

tests/test_interop_scenarios.rs
kkohbrok: You only get a `Welcome` if you make a commit, so I took `A->B: Welcome` as instruction to actually do an `Add` + `Commit`. But I agree, the instructions could be more precise.

tests/test_interop_scenarios.rs
kkohbrok: What seems to fail is the `from_utf8` call in the call back function announcing who has been added by whom. It seems like it which can't handle the group and client ids. I'll fix that. But other than that, it seems to work fine .

And by fine, I mean that it's been running on my machine for a while now. But at least it hasn't crashed :-)

tests/test_interop_scenarios.rs
kkohbrok: Good point. I've rewritten that paragraph a bit.

tests/test_interop_scenarios.rs
kkohbrok: Done!

tests/test_interop_scenarios.rs
kkohbrok: Done!

franziskuskiefer (approved):
:shipit:

PR #315, opened 2021-01-25
Proc macro to test all ciphersuites

franziskuskiefer:
This PR adds a new crate `test_macros` that exposes a proc-macro
`ctest` that allows defining tests for all supported ciphersuites
as follows

```
ctest!(test_name {
  // test body
})
```

This is used to replace the existing tests in tests/test-key_packages.rs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ members = [
     "delivery-service/ds-lib",
     "delivery-service/ds",
     "cli",
+    "test_macros",
 ]
 default-members = [ "." ]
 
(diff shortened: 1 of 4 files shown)


kkohbrok (dismissed):
This looks pretty cool! I'll try it out in the decryption index test.

raphaelrobert:
I like the idea! I'm just concerned this might throw off `rust-analyzer` and `cargo check/clippy`. I can test this in a bit if it hasn't already be done.

kkohbrok:
I also sometimes get tooling issues when working with macros. Autocomplete won't work properly and clippy gets confused. But for development, we can always copy the code out and convert it to a macro afterwards.

franziskuskiefer:
These tools aren't perfect and can't handle the entirety of Rust unfortunately. But you also shouldn't rely too much on them 😉

kkohbrok (approved):
Looks good! Looking forward to using this for the interop test scenarios!

PR #317, opened 2021-01-25
Move tree hashing code

raphaelrobert:
This is a small follow-up PR to #293.

It moves all functions related parent & tree hashes to `hashes.rs`. I didn't do this earlier, so that it wouldn't impair reviews.

No code is changed in this PR.


franziskuskiefer (approved):
lgtm

Issue #81, opened 2020-10-08
Implement Write and Read for Codec


franziskuskiefer:
The codec should use `std::io::Write` and `std::io::Read`.

franziskuskiefer:
Closing in favour of #318

PR #318, opened 2021-01-26
TLS codec crate & derive

franziskuskiefer:
This PR implements the TLS codec in a separate crate as well as corresponding derive macros for serialization and de-serialization.

This will allow us to remove (almost) all `Codec` related code. I'll do so in follow ups.

### DIFF
--- a/tls-codec/src/hpke.rs
+++ b/tls-codec/src/hpke.rs
@@ -1,0 +1,18 @@
+//! TLS Codec for HPKE primitives
+
+use super::*;
+use hpke_rs::HPKEPublicKey;
+
+impl Serialize for HPKEPublicKey {
+    fn tls_serialize(&self, buffer: &mut Vec<u8>) -> Result<(), Error> {
+        let value = TlsVecU16::from(self.as_slice());
+        value.tls_serialize(buffer)
+    }
+}
+
+impl Deserialize for HPKEPublicKey {
+    fn tls_deserialize(cursor: &Cursor) -> Result<Self, Error> {
+        let value = TlsVecU16::<u8>::tls_deserialize(&cursor)?;
+        Ok(Self::new(value.as_slice().to_vec()))
+    }
+}
--- a/tls-codec/src/lib.rs
+++ b/tls-codec/src/lib.rs
@@ -1,0 +1,80 @@
+use std::cell::Cell;
+
+mod arrays;
+#[cfg(feature = "hpke")]
+mod hpke;
+mod primitives;
+mod tls_vec;
+pub use tls_vec::{TlsVecU16, TlsVecU32, TlsVecU64, TlsVecU8};
+
+#[cfg(feature = "derive")]
+pub use tls_codec_derive::{TlsDeserialize, TlsSerialize};
+
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum Error {
+    EncodingError,
+    InvalidVectorLength,
+    InvalidInput,
+    DecodingError,
+}
+
+pub trait Serialize {
+    fn tls_serialize(&self, buffer: &mut Vec<u8>) -> Result<(), Error>;
+
+    fn serialize_detached(&self) -> Result<Vec<u8>, Error> {
+        let mut buffer = Vec::new();
+        self.tls_serialize(&mut buffer)?;
+        Ok(buffer)
+    }
+}
+
+pub trait Deserialize {
+    fn tls_deserialize(cursor: &Cursor) -> Result<Self, Error>
+    where
+        Self: Sized;
+
+    fn deserialize_detached(bytes: &[u8]) -> Result<Self, Error>
+    where
+        Self: Sized,
+    {
+        Self::tls_deserialize(&mut Cursor::new(bytes))
+    }
+}
+
+#[derive(Debug)]
+pub struct Cursor {
+    bytes: Vec<u8>,
+    position: Cell<usize>,
+}
+
+impl Cursor {
+    pub fn new(bytes: &[u8]) -> Cursor {
+        Cursor {
+            bytes: bytes.to_vec(),
+            position: Cell::new(0),
+        }
+    }
+
+    pub(crate) fn read(&self, length: usize) -> Result<&[u8], Error> {
+        let unread_bytes = self.bytes.len() - self.position.get();
+        if unread_bytes < length {
+            return Err(Error::InvalidInput);
+        }
+
+        let position = self.position.get();
+        self.position.replace(self.position.get() + length);
+        Ok(&self.bytes[position..position + length])
+    }
+
+    pub(crate) fn sub_cursor(&self, length: usize) -> Result<Cursor, Error> {
+        self.read(length).map(|buffer| Cursor::new(buffer))
+    }
+
+    pub(crate) fn is_empty(&self) -> bool {
+        self.position.get() >= self.bytes.len()
+    }
+
+    pub(crate) fn has_more(&self) -> bool {
+        !self.is_empty()
+    }
+}
--- a/tls-codec/src/lib.rs
+++ b/tls-codec/src/lib.rs
@@ -5,11 +5,11 @@
 //!
 //! With the feature `derive` `TlsSerialize` and `TlsDeserialize` can be derived.
 //!
-//! Tis crate provides the following data structures that implement TLS serialization
+//! This crate provides the following data structures that implement TLS serialization
 //! * `u8`, `u16`, `u32`, `u64`
 //! * `TlsVecU8`, `TlsVecU16`, `TlsVecU32`
 //! * `[u8; 2]`, `[u8; 4]`, `[u8; 8]`, `[u8; 16]`, `[u8; 32]`, `[u8; 64]`
-//! * If the `hpke` features is enabled, the TLS codec is implemented for
+//! * If the `hpke` feature is enabled, the TLS codec is implemented for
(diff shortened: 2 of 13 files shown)


tls-codec/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+pub trait Deserialize {
+    fn tls_deserialize(cursor: &Cursor) -> Result<Self, Error>
+    where
+        Self: Sized;
+
+    fn deserialize_detached(bytes: &[u8]) -> Result<Self, Error>
+    where
+        Self: Sized,
+    {
+        Self::tls_deserialize(&mut Cursor::new(bytes))
```
raphaelrobert, line 40: Clippy warning

tls-codec/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub struct Cursor {
+    bytes: Vec<u8>,
+    position: Cell<usize>,
```
raphaelrobert, line 47: I know this was also a problem before, but on 32bit platforms `usize` cannot index 64 bit vectors.

tls-codec/src/primitives.rs
```diff
@@ -0,0 +1,73 @@
...
+impl Serialize for u64 {
+    fn tls_serialize(&self, buffer: &mut Vec<u8>) -> Result<(), Error> {
+        buffer.extend_from_slice(&self.to_be_bytes());
+        Ok(())
+    }
+}
+
+impl Deserialize for u8 {
+    fn tls_deserialize(cursor: &Cursor) -> Result<Self, Error> {
+        match cursor.read(1) {
```
raphaelrobert, line 35: Wouldn't it be more succinct to use `map()` here? (also applies to the following functions)

tls-codec/src/tls_vec.rs
```diff
@@ -0,0 +1,225 @@
...
+                Ok(result)
+            }
+        }
+    };
+}
+
+impl_tls_vec!(u8, TlsVecU8);
+impl_tls_vec!(u16, TlsVecU16);
+impl_tls_vec!(u32, TlsVecU32);
+impl_tls_vec!(u64, TlsVecU64);
```
raphaelrobert, line 213: See the comment about `usize` earlier. The MLS spec doesn't use 64 bit vectors afaik. Maybe dropping this would just resolve the problem.

tls-codec/src/tls_vec.rs
```diff
@@ -0,0 +1,225 @@
...
+            where
+                S: serde::Serializer,
+            {
+                let mut state = serializer.serialize_struct(stringify!($name), 1)?;
+                state.serialize_field("vec", &self.vec)?;
+                state.end()
+            }
+        }
+
+        impl<'de, T> serde::de::Deserialize<'de> for $name<T>
```
raphaelrobert, line 88: serde is really annoying for these simple cases...

tls-codec/tests/decode.rs
```diff
@@ -0,0 +1,31 @@
...
+    let b = [1, 4, 77, 88, 1, 99];
+    let cursor = Cursor::new(&b);
+
+    let a = u8::tls_deserialize(&cursor).expect("Unable to tls_deserialize");
+    assert_eq!(1, a);
+    let v = TlsVecU8::<u8>::tls_deserialize(&cursor).expect("Unable to tls_deserialize");
+    assert_eq!(&[77, 88, 1, 99], v.as_slice());
+
+    // It's empty now.
+    assert!(u8::tls_deserialize(&cursor).is_err())
```
raphaelrobert, line 30: While you are at it you can also test the `has_more()` function

tls-codec/tests/encode.rs
```diff
@@ -0,0 +1,23 @@
+use tls_codec::{Serialize, TlsVecU16};
+
+#[test]
+fn serialize_primitives() {
+    let mut v = Vec::new();
+    77u8.tls_serialize(&mut v).expect("Error encoding u8");
+    88u8.tls_serialize(&mut v).expect("Error encoding u8");
+    355u16.tls_serialize(&mut v).expect("Error encoding u8");
```
raphaelrobert, line 8: "Error encoding u16"

tls-codec/src/primitives.rs
```diff
@@ -0,0 +1,73 @@
...
+            Err(e) => Err(e),
+        }
+    }
+}
+
+impl From<std::array::TryFromSliceError> for Error {
+    fn from(_e: std::array::TryFromSliceError) -> Self {
+        Self::InvalidInput
+    }
+}
```
raphaelrobert, line 73: `Option` is also an official primitive in the MLS spec that we need here

raphaelrobert (changes requested):
Looks good in general, thanks for tackling this overdue refactoring, just a few nits and remarks:

 - There might be more stuff missing for `Vec`, but you can always add it later.
 -  The DS uses tuples TLS serialisation as well, but I'd rather change the DS because we don't need tuples in the MLS spec
 -  Could the two modules not be combined into one?

tls-codec/src/hpke.rs
```diff
@@ -0,0 +1,18 @@
+//! TLS Codec for HPKE primitives
+
+use super::*;
+use hpke_rs::HPKEPublicKey;
+
+impl Serialize for HPKEPublicKey {
```
kkohbrok, line 6: Isn't the purpose of this macro that we can just derive `TlsSerialize` for `HPKEPublicKey`? Or am I missing something?

kkohbrok (changes requested):
I really like the generic construction, so from a functional standpoint it looks good to me. But I think some documentation would be helpful, at least describing the general structure. Also, clippy is unhappy in a couple of places.

raphaelrobert:
@franziskuskiefer can you please also check if you want to tackle #81 with this PR? If not, we can probably close that ticket.

tls-codec/src/hpke.rs
franziskuskiefer: `HPKEPublicKey` is define in the HPKE crate that doesn't know about this (and doesn't really care). So we can't derive but have to implement it manually.

tls-codec/src/lib.rs
franziskuskiefer: Indeed, we should rework `Cursor` anyway later. I filed #319 to keep track of this specific issue.

tls-codec/src/primitives.rs
franziskuskiefer: Here it would be indeed. The others use `try_into`, which doesn't work in the closure.

tls-codec/src/primitives.rs
franziskuskiefer: Indeed. I'm sure there's more I forgot 😬 We can add them along the way when we start using this.

tls-codec/src/tls_vec.rs
franziskuskiefer: Yeah I drop it here with a reference to #319

tls-codec/tests/decode.rs
franziskuskiefer: The function is not public so I can't 😞  But I agree that we should add some unit tests for the `Cursor`. I'm deferring this to the general improvement of `Cursor`.

franziskuskiefer (commented):
> There might be more stuff missing for Vec, but you can always add it later.

yep, that's the plan. I started using it but I'm sure there are things that are missing that we can add along the way.

> The DS uses tuples TLS serialisation as well, but I'd rather change the DS because we don't need tuples in the MLS spec

👍🏻 

> Could the two modules not be combined into one?

Unfortunately we can't. Proc macro crates can't expose anything else.

tls-codec/src/lib.rs
```diff
@@ -1,60 +1,99 @@
...
+//! This crate implements the TLS codec as defined in [RFC 8446](https://tools.ietf.org/html/rfc8446)
+//! as well as some extensions required by MLS.
+//!
+//! With the feature `derive` `TlsSerialize` and `TlsDeserialize` can be derived.
+//!
+//! Tis crate provides the following data structures that implement TLS serialization
+//! * `u8`, `u16`, `u32`, `u64`
+//! * `TlsVecU8`, `TlsVecU16`, `TlsVecU32`
+//! * `[u8; 2]`, `[u8; 4]`, `[u8; 8]`, `[u8; 16]`, `[u8; 32]`, `[u8; 64]`
+//! * If the `hpke` features is enabled, the TLS codec is implemented for
```
kkohbrok, line 12:
```suggestion
//! * If the `hpke` feature is enabled, the TLS codec is implemented for
```

tls-codec/src/lib.rs
```diff
@@ -1,60 +1,99 @@
+//! # TLS Codec
+//!
+//! This crate implements the TLS codec as defined in [RFC 8446](https://tools.ietf.org/html/rfc8446)
+//! as well as some extensions required by MLS.
+//!
+//! With the feature `derive` `TlsSerialize` and `TlsDeserialize` can be derived.
+//!
+//! Tis crate provides the following data structures that implement TLS serialization
```
kkohbrok, line 8:
```suggestion
//! This crate provides the following data structures that implement TLS serialization
```

kkohbrok (dismissed):
Thanks for adding the documentation! Two typos, otherwise it looks good!

kkohbrok (approved):

raphaelrobert (approved):
LGTM, but the CI isn't happy yet...

PR #320, opened 2021-01-27
Make results of sign/verify consistent

### LINKED ISSUE #297: Make results of sign/verify consistent
We are inconsistent in the use of validate/verify functions. Not only the naming but also whether they return an error or a bool.

_Originally posted by @franziskuskiefer in https://github.com/openmls/openmls/pull/293#r551845150_

raphaelrobert:
Fixes #297. Follow-up to #293.

Some of the return values of verification functions were still `bool`. Now the return types are consistently `Result<(), Error>`.

### DIFF
--- a/src/ciphersuite/signable.rs
+++ b/src/ciphersuite/signable.rs
@@ -16,7 +16,7 @@ pub trait Signable: Sized {
 
     /// Verifies the payload against the given `credential` and `signature`.
     ///
-    /// Returns a `Ok(())` if the signature is valid and `CredentialError::InvalidSignature` otherwise.
+    /// Returns `Ok(())` if the signature is valid and `CredentialError::InvalidSignature` otherwise.
     fn verify(
         &self,
         credential: &Credential,
--- a/src/ciphersuite/signable.rs
+++ b/src/ciphersuite/signable.rs
@@ -16,8 +16,12 @@ pub trait Signable: Sized {
 
     /// Verifies the payload against the given `credential` and `signature`.
     ///
-    /// Returns a `true` if the signature is valid and `false` otherwise.
-    fn verify(&self, credential: &Credential, signature: &Signature) -> bool {
+    /// Returns a `Ok(())` if the signature is valid and `CredentialError::InvalidSignature` otherwise.
+    fn verify(
+        &self,
+        credential: &Credential,
+        signature: &Signature,
+    ) -> Result<(), CredentialError> {
         let payload = self.unsigned_payload().unwrap();
         credential.verify(&payload, signature)
     }
--- a/src/key_packages/mod.rs
+++ b/src/key_packages/mod.rs
@@ -92,7 +92,7 @@ impl KeyPackage {
         self.credential
             .verify(&self.unsigned_payload().unwrap(), &self.signature)
             .map_err(|_| {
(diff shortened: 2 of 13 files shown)


kkohbrok (commented):
From a cryptographic standpoint it's a bit strange not to return bool from a `verify` function. To me, an error indicates that something went wrong during the execution, but here, it didn't. The inputs just didn't verify (which is fine from a functional standpoint). I understand how it can be useful for debugging purposes, but can't we just return a bool and then use logging instead to provider more insight into why the verification didn't go through? I mean the calling function doesn't really care why the verification failed, does it?

src/ciphersuite/signable.rs
```diff
@@ -16,8 +16,12 @@ pub trait Signable: Sized {
 
     /// Verifies the payload against the given `credential` and `signature`.
     ///
-    /// Returns a `true` if the signature is valid and `false` otherwise.
-    fn verify(&self, credential: &Credential, signature: &Signature) -> bool {
+    /// Returns a `Ok(())` if the signature is valid and `CredentialError::InvalidSignature` otherwise.
```
franziskuskiefer, line 19:
```suggestion
    /// Returns `Ok(())` if the signature is valid and `CredentialError::InvalidSignature` otherwise.
```

src/key_packages/mod.rs
```diff
@@ -89,15 +89,12 @@ impl KeyPackage {
...
+        self.credential
             .verify(&self.unsigned_payload().unwrap(), &self.signature)
-        {
-            Ok(())
-        } else {
-            log::error!("Key package signature is empty.");
-            Err(KeyPackageError::InvalidSignature)
-        }
+            .map_err(|_| {
+                log::error!("Key package signature is empty.");
```
franziskuskiefer, line 95: Hm, this log message appears to be wrong.

franziskuskiefer, line 95:
```suggestion
                log::error!("Key package signature is invalid.");
```

franziskuskiefer (dismissed):
lgtm with two nits.
We should probably introduce some `Result` types at some point to make these things less verbose.

raphaelrobert:
> From a cryptographic standpoint it's a bit strange not to return bool from a `verify` function. To me, an error indicates that something went wrong during the execution, but here, it didn't. The inputs just didn't verify (which is fine from a functional standpoint). I understand how it can be useful for debugging purposes, but can't we just return a bool and then use logging instead to provider more insight into why the verification didn't go through? I mean the calling function doesn't really care why the verification failed, does it?

In wider sense, something did go wrong: We expected a valid signature/hash but found something else. This also allows to bubble up what went wrong exactly (in the case of parent hash verification) and that might be useful for the consumer of OpenMLS.

kkohbrok:
> In wider sense, something did go wrong: We expected a valid signature/hash but found something else. This also allows to bubble up what went wrong exactly (in the case of parent hash verification) and that might be useful for the consumer of OpenMLS.

Something went wrong from the point of view of the caller, sure, but the verify function itself functioned as expected. I see how it's useful in the context of the implementation, though, so I'm not opposed to landing this. The signature verification functions are essentially boolean anyway.

franziskuskiefer (approved):

PR #321, opened 2021-01-27
Consolidate direct_path functions in TreeMath

### LINKED ISSUE #96: Consolidate direct_path functions in TreeMath
We currently have three variations of the same function, we probably only need one.

raphaelrobert:
Fixes #96.

I cleaned up the direct path mess in `treemath`. Now there are only two functions, depending on whether the direct path starts from a leaf or an intermediary node. Adds some type safety as well, because `LeafIndex` is used more often now.

### DIFF
--- a/src/tree/treemath.rs
+++ b/src/tree/treemath.rs
@@ -5,6 +5,7 @@ implement_error! {
     pub enum TreeMathError {
         LeafHasNoChildren = "Leaf nodes don't have children.",
         RootHasNoParent = "Root nodes don't have parents.",
+        NotAParentNode = "Node index was not a parent node.",
     }
 }
 
@@ -96,56 +97,20 @@ pub(crate) fn sibling(index: NodeIndex, size: LeafIndex) -> Result<NodeIndex, Tr
     }
 }
 
-// Ordered from leaf to root
-// Includes neither leaf nor root
-pub(crate) fn dirpath(index: NodeIndex, size: LeafIndex) -> Result<Vec<NodeIndex>, TreeMathError> {
-    let r = root(size);
-    if index == r {
-        return Ok(vec![]);
-    }
-
-    let mut d = vec![];
-    let mut x = parent(index, size)?;
-    while x != r {
-        d.push(x);
-        x = parent(x, size)?;
-    }
-    Ok(d)
-}
-
-// Ordered from leaf to root
-// Includes leaf and root
-pub(crate) fn dirpath_long(
-    index: NodeIndex,
+/// Direct path from a leaf node to the root.
+/// Does not include the leaf node but includes the root.
+pub(crate) fn leaf_direct_path(
+    leaf_index: LeafIndex,
     size: LeafIndex,
 ) -> Result<Vec<NodeIndex>, TreeMathError> {
+    let node_index = NodeIndex::from(leaf_index);
     let r = root(size);
-    if index == r {
-        return Ok(vec![r]);
-    }
-
-    let mut x = index;
-    let mut d = vec![index];
-    while x != r {
-        x = parent(x, size)?;
-        d.push(x);
-    }
-    Ok(d)
-}
-
-// Ordered from leaf to root
-// Includes root but not leaf
-pub(crate) fn direct_path_root(
-    index: NodeIndex,
-    size: LeafIndex,
-) -> Result<Vec<NodeIndex>, TreeMathError> {
-    let r = root(size);
-    if index == r {
+    if node_index == r {
         return Ok(vec![r]);
     }
 
     let mut d = vec![];
-    let mut x = index;
+    let mut x = node_index;
     while x != r {
         x = parent(x, size)?;
         d.push(x);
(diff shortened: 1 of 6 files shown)


src/tree/treemath.rs
```diff
@@ -96,71 +97,72 @@ pub(crate) fn sibling(index: NodeIndex, size: LeafIndex) -> Result<NodeIndex, Tr
...
 }
 
-// Ordered from leaf to root
-// Includes root but not leaf
-pub(crate) fn direct_path_root(
-    index: NodeIndex,
+/// Direct path from a parent node to the root.
+/// Includes the parent node and the root.
+/// Returns an error if the `index` is not a parent node.
+pub(crate) fn parent_direct_path(
```
kkohbrok, line 124: I'm not sure this is a good idea. `direct_path` is well defined in the spec not to include the node it starts from. If we want to include the starting node itself, we should either call the function differently, or do it ad-hoc when we need it.

src/tree/treemath.rs
```diff
@@ -96,71 +97,72 @@ pub(crate) fn sibling(index: NodeIndex, size: LeafIndex) -> Result<NodeIndex, Tr
...
+    node_index: NodeIndex,
     size: LeafIndex,
 ) -> Result<Vec<NodeIndex>, TreeMathError> {
```
kkohbrok, line 127: While you're here, you can remove the `Result` and add an `unwrap` at the `parent` call. It's never going to throw an error, because we're never calling it with the root node index.

src/tree/treemath.rs
```diff
@@ -96,71 +97,72 @@ pub(crate) fn sibling(index: NodeIndex, size: LeafIndex) -> Result<NodeIndex, Tr
...
     }
     Ok(d)
 }
 
-// Ordered from leaf to root
-pub(crate) fn copath(index: NodeIndex, size: LeafIndex) -> Result<Vec<NodeIndex>, TreeMathError> {
-    if index == root(size) {
+/// Copath of a leaf.
+/// Ordered from leaf to root.
+pub(crate) fn copath(
```
kkohbrok, line 147: According to spec, the copath is defined for nodes rather than leaves. But it might be ok here, because we're never using it with intermediate nodes.

kkohbrok (commented):
I like the consolidation of `direct_path` functions, but I'm not sure it's a good idea to have a function called `direct_path` that is not actually a direct path according to spec.

src/tree/treemath.rs
franziskuskiefer: I don't really care about the spec here but we should be consistent in the naming and semantics.

src/tree/treemath.rs
franziskuskiefer: It can throw an error if called with a node that's not a parent.

src/tree/treemath.rs
```diff
@@ -96,71 +97,72 @@ pub(crate) fn sibling(index: NodeIndex, size: LeafIndex) -> Result<NodeIndex, Tr
...
+    leaf_index: LeafIndex,
+    size: LeafIndex,
+) -> Result<Vec<NodeIndex>, TreeMathError> {
+    let node_index = NodeIndex::from(leaf_index);
```
franziskuskiefer, line 151: Do we need to check `!node_index.is_parent()`?

franziskuskiefer (changes requested):
Generally looks good to me with two nits.

src/tree/treemath.rs
kkohbrok: True. I guess that observation is valid for the generic implementation of `direct_path` as opposed to the specific leaf- and parent variants.

src/tree/treemath.rs
raphaelrobert: Since we convert it from a leaf index, it cannot be a parent node by design

src/tree/treemath.rs
raphaelrobert: I thought we agreed (and I think it was also your suggestion) that it is ok to deviate from the spec when it makes sense to improve the quality of the implementation. I believe this is one of those instances.
If we do more refactoring around the binary tree, we might have to change this anyway, but for now it's the most generic way and it gives additional type safety.

src/tree/treemath.rs
raphaelrobert: Exactly, I'd have to remove the check for whether it's really a parent node, which I'd rather not do given how the function is supposed to be used.

src/tree/treemath.rs
raphaelrobert: According to the example algorithm in the spec yes, but in a wider sense the `copath` function in the spec is only ever used when starting from a leaf, that's what I wanted to enforce here.

kkohbrok (approved):
Fair enough. I can see how it improves the implementation. Let's get this in, then.

src/tree/treemath.rs
franziskuskiefer: Can we change the name then or something? The actual issue is still here.

franziskuskiefer (changes requested):

src/tree/treemath.rs
raphaelrobert: I'm not sure I follow what the issue is. Why is the name a problem?

For context: Currently the implementation uses direct path in two instances: when starting with a leaf and when starting with a parent node. I named the two variants of direct path to reflect the use case.

franziskuskiefer: Right, the issue is that `parent_direct_path` has a different result than `leaf_direct_path` because it includes the starting node. Having so similar names this behaviour is unexpected. So *something* should change.

raphaelrobert: Would it help if I renamed it to `partial_direct_path()`?

For context: The risk of confusion is low, because the two functions have different signatures as well.

franziskuskiefer: I'm not worried about confusion but about assumptions on what the function does. The easiest would be to always either include the starting node or not.

raphaelrobert: I think what the functions do is the most generic we can get and the documentation is also accurate, which leaves us with the name:

I suggested `partial_direct_path()`, `inner_direct_path()` could also work. If those are not good enough I think it's best if you make a suggestion so that we can move forward.

franziskuskiefer: I'm not excited about the names but don't have a better idea. I'll approve and let you decide :)

franziskuskiefer (approved):

PR #322, opened 2021-01-28
Add ctest_panic macro

kkohbrok:
Adds a macro that allows indicating that we expect a test to fail when using the ciphersuite proc macros.

I built this for the interop test scenario PR, but ended up not using it. It might be useful for the future, though.

### DIFF
--- a/test_macros/src/lib.rs
+++ b/test_macros/src/lib.rs
@@ -24,12 +24,25 @@ impl Parse for TestInput {
 #[proc_macro]
 pub fn ctest(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as TestInput);
-    impl_ciphersuite_tests(input).unwrap()
+    impl_ciphersuite_tests(input, false).unwrap()
 }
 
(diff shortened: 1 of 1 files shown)


test_macros/src/lib.rs
```diff
@@ -24,12 +24,25 @@ impl Parse for TestInput {
...
 }
 
-fn impl_ciphersuite_tests(input: TestInput) -> Result<TokenStream> {
+#[proc_macro]
+pub fn ctest_panic(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as TestInput);
+    impl_ciphersuite_tests(input, true).unwrap()
+}
+
+fn impl_ciphersuite_tests(input: TestInput, should_panic: bool) -> Result<TokenStream> {
```
franziskuskiefer, line 36: You can pass in a `proc_macro::TokenStream` here and the `quote`s as arguments to safe you the conditional.

franziskuskiefer (commented):
thanks. one suggestion to make it more concise.

franziskuskiefer (approved):

PR #324, opened 2021-01-28
KeyPackage constructor check for mismatch

### LINKED ISSUE #294: [Bug] KeyPackage constructor doesn't check if Ciphersuite and SignatureScheme (of the Credential) match
I think for robustness, the constructor of `KeyPackage` (or maybe `KeyPackageBundle`) should check if the two match and return an error if they don't.

raphaelrobert:
Fixes #294 and re-enables the key package unit tests that seem to have been de-activated for a while.


kkohbrok (approved):
Thanks for taking care of this bug! Looks good to me.

franziskuskiefer (approved):

Issue #186, opened 2020-11-03
Implement GroupContext extensions


franziskuskiefer:
The `GroupContext` is missing the `extensions` field. The corresponding handling of extensions within a group is also missing.

cc #32

kkohbrok:
This seems to be solved, as `GroupContext` now has an `extensions` field in `main`. Regarding the handling: To my knowledge, there's currently no extension that goes into the `GroupContext` (see https://github.com/mlswg/mls-protocol/pull/464 and https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#section-16.2).

franziskuskiefer:
Well I guess I don't understand what `extensions` in the `GroupContext` are supposed to be and why they exist in the first place. But we currently hard code it to an empty vector. So if there were extensions, we'd have to implement them (there are also TODOs in the code).

PR #325, opened 2021-01-29
Make composite types 2^32 and make 0 invalid for enums

### LINKED ISSUE #61: Make composite types 2^32 and make 0 invalid for enums (#387)
https://github.com/mlswg/mls-protocol/pull/387

raphaelrobert:
Fixes #61.

Note that this does not yet address #186, in fact I had to deactivate some more code and I mentioned this in TODOs in several places.

### DIFF
--- a/src/ciphersuite/mod.rs
+++ b/src/ciphersuite/mod.rs
@@ -698,7 +698,7 @@ impl SignatureKeypair {
 
     /// Verify a `Signature` on the `payload` byte slice with the key pair's
     /// public key.
-    pub fn verify(&self, signature: &Signature, payload: &[u8]) -> bool {
+    pub fn verify(&self, signature: &Signature, payload: &[u8]) -> Result<(), SignatureError> {
         self.public_key.verify(signature, payload)
     }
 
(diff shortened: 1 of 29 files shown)


kkohbrok:
I've only skimmed over the changes, but this PR doesn't just make composite types 2^32. It also changes a few other things. For example, a `signature_scheme` function is added to `impl Credential` and new things seem to be included in the `GroupContext`. If you're doing more than just fix #61, it would be nice if you could give a rough summary of what this PR is doing so I/we know what to look out for.

raphaelrobert:
> I've only skimmed over the changes, but this PR doesn't just make composite types 2^32. It also changes a few other things. For example, a `signature_scheme` function is added to `impl Credential` and new things seem to be included in the `GroupContext`. If you're doing more than just fix #61, it would be nice if you could give a rough summary of what this PR is doing so I/we know what to look out for.

Yeah, sorry, looks like some bits from #324 made it into here. That should disappear once #324 is merged.

kkohbrok (approved):
I'm still not sure why `GroupContext` needs a new field to address #61, but it probably needs to be done anyway. Other than that, it looks good!

raphaelrobert:
> I'm still not sure why `GroupContext` needs a new field to address #61, but it probably needs to be done anyway. Other than that, it looks good!

Because it was missing the `extensions` field altogether. #61 was about changing how that field is encoded.

kkohbrok:
> > I'm still not sure why `GroupContext` needs a new field to address #61, but it probably needs to be done anyway. Other than that, it looks good!
> 
> Because it was missing the `extensions` field altogether. #61 was about changing how that field is encoded.

Ah, sorry, I had overlooked that. That makes total sense, then.

Issue #206, opened 2020-11-11
Add group config


franziskuskiefer:
In addition to the global config (#85) we also need a per-group configuration that sets the group configuration.

* default lifetime value
* default extensions

kkohbrok:
I think the default lifetime value for KeyPackages would be group independent, as each party publishes their KeyPackages before any group is involved.

kkohbrok:
We have had group configs for a while now. Closing.

PR #327, opened 2021-01-29
Export PublicGroupState

### LINKED ISSUE #192: External Commits
https://github.com/mlswg/mls-protocol/pull/406

- [ ]  #542

raphaelrobert:
This PR exposes the `PublicGroupState` for a given epoch.

This is in preparation for External Commits.

### DIFF
--- a/src/group/mls_group/mod.rs
+++ b/src/group/mls_group/mod.rs
@@ -9,7 +9,7 @@ mod test_mls_group;
 use crate::ciphersuite::*;
 use crate::codec::*;
 use crate::config::Config;
-use crate::credentials::CredentialBundle;
+use crate::credentials::{CredentialBundle, CredentialError};
 use crate::framing::*;
 use crate::group::*;
 use crate::key_packages::*;
@@ -376,6 +376,28 @@ impl MlsGroup {
     pub fn group_id(&self) -> &GroupId {
         &self.group_context.group_id
     }
+
+    /// Get thr groups extensions.
+    /// Right now this is limited to the ratchet tree extension, which is built
+    /// on the fly when calling this function.
+    pub fn extensions(&self) -> Vec<Box<dyn Extension>> {
+        let extensions: Vec<Box<dyn Extension>> = if self.add_ratchet_tree_extension {
+            vec![Box::new(RatchetTreeExtension::new(
+                self.tree().public_key_tree_copy(),
+            ))]
+        } else {
+            Vec::new()
+        };
+        extensions
+    }
+
+    /// Export the `PublicGroupState`
+    pub fn export_public_group_state(
+        &self,
+        credential_bundle: &CredentialBundle,
+    ) -> Result<PublicGroupState, CredentialError> {
+        PublicGroupState::new(self, credential_bundle)
+    }
 }
 
 // Private and crate functions
@@ -390,6 +412,11 @@ impl MlsGroup {
 
     pub(crate) fn secret_tree_mut(&self) -> RefMut<SecretTree> {
         self.secret_tree.borrow_mut()
+    }
+
+    /// Current interim transcript hash of the group
+    pub(crate) fn interim_transcript_hash(&self) -> &[u8] {
+        &self.interim_transcript_hash
     }
 
     #[cfg(all(test, feature = "test-vectors"))]
--- a/src/group/mls_group/mod.rs
+++ b/src/group/mls_group/mod.rs
@@ -377,8 +377,8 @@ impl MlsGroup {
         &self.group_context.group_id
     }
 
-    /// Get thr groups extensions.
-    /// Right now this is limited to the ratchet tree extension, which is built
+    /// Get the groups extensions.
+    /// Right now this is limited to the ratchet tree extension which is built
     /// on the fly when calling this function.
     pub fn extensions(&self) -> Vec<Box<dyn Extension>> {
         let extensions: Vec<Box<dyn Extension>> = if self.add_ratchet_tree_extension {
--- a/src/key_packages/mod.rs
+++ b/src/key_packages/mod.rs
@@ -6,10 +6,7 @@ use crate::ciphersuite::*;
 use crate::codec::*;
 use crate::config::{Config, ProtocolVersion};
 use crate::credentials::*;
-use crate::extensions::{
-    CapabilitiesExtension, Extension, ExtensionError, ExtensionStruct, ExtensionType,
-    LifetimeExtension, ParentHashExtension,
-};
+use crate::extensions::*;
 
 use serde::{
     de::{self, MapAccess, SeqAccess, Visitor},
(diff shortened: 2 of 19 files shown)


src/group/mls_group/mod.rs
```diff
@@ -376,6 +376,28 @@ impl MlsGroup {
     pub fn group_id(&self) -> &GroupId {
         &self.group_context.group_id
     }
+
+    /// Get the groups extensions.
+    /// Right now this is limited to the ratchet tree extension which is built
+    /// on the fly when calling this function.
+    pub fn extensions(&self) -> Vec<Box<dyn Extension>> {
+        let extensions: Vec<Box<dyn Extension>> = if self.add_ratchet_tree_extension {
```
kkohbrok, line 384: If we have `self.add_ratchet_tree_extension`, shouldn't the extension already be present in somewhere in the group state? I mean as opposed to having to create it ad-hoc. After all, it would also have to be included in the `GroupContext`, which as far as I can see it isn't right now.

src/messages/mod.rs
```diff
@@ -318,3 +316,127 @@ impl GroupSecrets {
...
+    pub(crate) external_pub: &'a HPKEPublicKey,
+}
+
+impl<'a> PublicGroupStateTBS<'a> {
+    /// Signs the `PublicGroupStateTBS` with a `CredentialBundle`.
+    fn sign(&self, credential_bundle: &CredentialBundle) -> Result<Signature, CredentialError> {
+        let payload = self
+            .encode_detached()
+            .map_err(CredentialError::CodecError)?;
+        // Unwrapping here is safe, because signing should not cause an error with a
```
kkohbrok, line 436: You're not actually unwrapping, though.

kkohbrok (changes requested):
Looks good to me with one small nit! I'm pretty sure interop is not going to work until we fix #32, after which we would also not have to create the `ratchet_tree` extension here in such an ad-hoc way. But I guess in the end it doesn't matter what we do first.

src/group/mls_group/mod.rs
raphaelrobert: Yes, it's not supported yet because of #32. We would have to refactor that part obviously, so that we don't generate the extension more often than once per epoch.

raphaelrobert:
> Looks good to me with one small nit! I'm pretty sure interop is not going to work until we fix #32, after which we would also not have to create the `ratchet_tree` extension here in such an ad-hoc way. But I guess in the end it doesn't matter what we do first.

Interop should still work now, as long as no group extensions (really only the ratchet tree extension) are present.

src/key_packages/mod.rs
```diff
@@ -6,10 +6,7 @@ use crate::ciphersuite::*;
 use crate::codec::*;
 use crate::config::{Config, ProtocolVersion};
 use crate::credentials::*;
-use crate::extensions::{
-    CapabilitiesExtension, Extension, ExtensionError, ExtensionStruct, ExtensionType,
-    LifetimeExtension, ParentHashExtension,
-};
+use crate::extensions::*;
```
franziskuskiefer, line 9: 😬 why? Please expand again.

src/messages/codec.rs
```diff
@@ -288,3 +282,53 @@ impl Codec for ReInitProposal {
...
+}
+
+impl<'a> Codec for PublicGroupStateTBS<'a> {
+    fn encode(&self, buffer: &mut Vec<u8>) -> Result<(), CodecError> {
+        self.group_id.encode(buffer)?;
+        self.epoch.encode(buffer)?;
+        encode_vec(VecSize::VecU8, buffer, &self.tree_hash)?;
+        encode_vec(VecSize::VecU8, buffer, &self.interim_transcript_hash)?;
+        encode_extensions(&self.extensions, buffer)?;
+        encode_extensions(&self.extensions, buffer)?;
```
franziskuskiefer, line 330: We need this only once.

src/messages/mod.rs
```diff
@@ -318,3 +316,126 @@ impl GroupSecrets {
...
+///     opaque interim_transcript_hash<0..255>;
+///     Extension extensions<0..2^32-1>;
+///     HPKEPublicKey external_pub;
+///     uint32 signer_index;
+///     opaque signature<0..2^16-1>;
+/// } PublicGroupState;
+/// ```
+#[derive(PartialEq, Debug)]
+pub struct PublicGroupState {
+    pub ciphersuite: CiphersuiteName,
```
franziskuskiefer, line 337: No public struct members please.

src/messages/mod.rs
```diff
@@ -318,3 +316,126 @@ impl GroupSecrets {
...
+
+impl PublicGroupState {
+    /// Creates a new `PublicGroupState` struct from the current internal state
+    /// of the group and signs it.
+    pub(crate) fn new(
+        mls_group: &MlsGroup,
+        credential_bundle: &CredentialBundle,
+    ) -> Result<Self, CredentialError> {
+        let ciphersuite = mls_group.ciphersuite();
+        let (_external_priv, external_pub) = mls_group
```
franziskuskiefer, line 356: This should go into the key schedule but I'll do this in my PR that reworks it.

franziskuskiefer (changes requested):
Generally looks good with a couple nits.

For the future please file separate PRs for unrelated changes.

src/messages/mod.rs
raphaelrobert: I thought about that as well, but decided to expose it as a function of `MlsGroup`. The reason is that it does a keypair derivation, which is a fairly expensive operation and it's only needed when External Commits are supported and not otherwise.
Ideally we could have this as part of the configuration (see #206).

raphaelrobert:
> Generally looks good with a couple nits.
> 
> For the future please file separate PRs for unrelated changes.

The only unrelated changes were comment formatting (which happens automatically, still only on nightly unfortunately) and clippy warnings. Happy to discuss whether we should always do those as a separate PR, for now when they don't involve substantial code refactoring I just fix them. Recently there seem to have been a more than average number of changes in clippy rules, so that these warnings keep popping up in older code.

franziskuskiefer (approved):
Ok let's get this in.

> The only unrelated changes were comment formatting (which happens automatically, still only on nightly unfortunately) and clippy warnings. Happy to discuss whether we should always do those as a separate PR, for now when they don't involve substantial code refactoring I just fix them. Recently there seem to have been a more than average number of changes in clippy rules, so that these warnings keep popping up in older code.

Unrelated changes must always go into separat PRs to avoid clutter and have a useful git history. I'l accept it here but not in future PRs.

raphaelrobert:
@kkohbrok anything else you'd like changed?

kkohbrok (approved):
Nope, looks good to me. I just forgot to approve you had responded to my earlier comments. Sorry about that!

PR #328, opened 2021-02-01
Fix codecov test coverage reporting

kkohbrok:
This PR moves the `ignore` section from `.github/workflows/coverage.yml` to a new file `.github/codecov.yml`.

According to the [documentation](https://docs.codecov.io/docs/codecov-yaml), this should fix codecov test coverage testing.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,3 @@
+ignore:
+  - "tests/utils/managed_utils/*"
+  - "tests/utils/mls_utils/*"
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,3 +1,0 @@
-ignore:
-  - "tests/utils/managed_utils/*"
-  - "tests/utils/mls_utils/*"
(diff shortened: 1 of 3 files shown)


raphaelrobert (dismissed):
LGTM, let's see if it works

.github/codecov.yml
```diff
@@ -0,0 +1,3 @@
+ignore:
```
franziskuskiefer, line 1: This should go into the toplevel `codecov.yml`

franziskuskiefer (changes requested):

kkohbrok:
We should probably wait with this one, as it appears to trigger an error within `cargo tarpaulin`. It's likely the same as this one: https://github.com/xd009642/tarpaulin/issues/655

franziskuskiefer:
The issue got fixed in nightly cargo. But now we have a linking issue 🙄

kkohbrok:
I think this would be the corresponding tarpaulin issue https://github.com/xd009642/tarpaulin/issues/517. I assume you switched to nightly, because it includes https://github.com/rust-lang/cargo/pull/9229? In the issue it seems the hope is that it also fixes the linker issue. But at least we're not the only ones having this problem.

Maybe it would be worth waiting for it to reach stable?

franziskuskiefer:
This was merged in #376

PR #330, opened 2021-02-02
key schedule reorg

franziskuskiefer:
In preparation for the key schedule test vectors I re-organized the
key schedule and moved the logic into the actual key schedule module
instead of the group.

There are no changes to functionality PR, only moving code around.

This PR also removes the test-vector feature as it is no longer needed.

### DIFF
--- a/src/group/mls_group/create_commit.rs
+++ b/src/group/mls_group/create_commit.rs
@@ -106,7 +106,7 @@ impl MlsGroup {
         let joiner_secret = JoinerSecret::new(
             ciphersuite,
             provisional_tree.commit_secret(),
-            &self.init_secret,
+            self.epoch_secrets().init_secret().ok_or(GroupError::InitSecretNotFound)?,
         );
 
         // Create group secrets for later use, so we can afterwards consume the
@@ -121,7 +121,7 @@ impl MlsGroup {
         let mut key_schedule = KeySchedule::init(ciphersuite, joiner_secret, None);
         let welcome_secret = key_schedule.welcome()?;
         key_schedule.add_context(&provisional_group_context)?;
-        let provisional_epoch_secrets = key_schedule.epoch_secrets()?;
+        let provisional_epoch_secrets = key_schedule.epoch_secrets(false)?;
 
         // Calculate the confirmation tag
         let confirmation_tag = ConfirmationTag::new(
@@ -220,14 +220,13 @@ impl PlaintextSecret {
         joiner_secret: &JoinerSecret,
         invited_members: Vec<(NodeIndex, AddProposal)>,
         provisional_tree: &RatchetTree,
-    ) -> Result<Vec<Self>, CodecError> {
+    ) -> Result<Vec<Self>, GroupError> {
         // Get a Vector containing the node indices of the direct path to the
         // root from our own leaf.
         let dirpath = treemath::leaf_direct_path(
             provisional_tree.own_node_index(),
             provisional_tree.leaf_count(),
-        )
-        .unwrap();
+        )?;
 
         let mut plaintext_secrets = vec![];
         for (index, add_proposal) in invited_members {
--- a/src/group/mls_group/create_commit.rs
+++ b/src/group/mls_group/create_commit.rs
@@ -40,62 +40,34 @@ impl MlsGroup {
         if apply_proposals_values.self_removed {
             return Err(CreateCommitError::CannotRemoveSelf.into());
         }
-        // Determine if Commit needs path field
-        let path_required =
-            apply_proposals_values.path_required || contains_own_updates || force_self_update;
 
-        let (commit_secret, path, path_secrets_option, kpb_option) = if path_required {
-            // If path is needed, compute path values
-            let (commit_secret, path, path_secrets, key_package_bundle) = provisional_tree
-                .refresh_private_tree(
+        let (path, kpb_option) =
+            if apply_proposals_values.path_required || contains_own_updates || force_self_update {
+                // If path is needed, compute path values
+                let (path, key_package_bundle) = provisional_tree.refresh_private_tree(
                     credential_bundle,
                     &self.group_context.serialized(),
                     apply_proposals_values.exclusion_list(),
                 );
-            (
-                Some(commit_secret),
-                Some(path),
-                Some(path_secrets),
-                Some(key_package_bundle),
-            )
-        } else {
-            // If path is not needed, return empty commit secret
-            (None, None, None, None)
-        };
+                (Some(path), Some(key_package_bundle))
+            } else {
+                // If path is not needed, return empty commit secret
+                (None, None)
+            };
+
         // Create commit message
         let commit = Commit {
             proposals: proposal_reference_list,
             path,
         };
+
         // Create provisional group state
         let mut provisional_epoch = self.group_context.epoch;
         provisional_epoch.increment();
-        // We clone the init secret here, as the `joiner_secret` is only for the
-        // provisional group state.
-        let joiner_secret = JoinerSecret::from_commit_and_init_secret(
-            ciphersuite,
-            commit_secret,
-            &self.init_secret,
-        );
-        // Create group secrets for later use, so we can afterwards consume the
-        // `joiner_secret`.
-        let plaintext_secrets = joiner_secret.group_secrets(
-            apply_proposals_values.invitation_list,
-            &provisional_tree,
-            path_secrets_option,
-        )?;
-        // TODO #141: Implement PSK
-        let intermediate_secret = IntermediateSecret::new(ciphersuite, joiner_secret, None);
-        let welcome_secret = WelcomeSecret::new(ciphersuite, &intermediate_secret);
-
-        // Derive the welcome key material before consuming the `MemberSecret`
-        // immediately afterwards.
-        let (welcome_key, welcome_nonce) = welcome_secret.derive_welcome_key_nonce(ciphersuite);
 
         // Build MLSPlaintext
         let content = MLSPlaintextContentType::Commit(commit);
         let sender = Sender::member(sender_index);
-
         let mut mls_plaintext = MLSPlaintext {
             group_id: self.context().group_id.clone(),
             epoch: self.context().epoch,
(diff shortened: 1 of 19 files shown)


src/schedule/mod.rs
```diff
@@ -43,26 +43,96 @@
 //! welcome_secret) is represented by its own struct to ensure that the keys are
 //! not confused with one-another and/or that the schedule is not derived
 //! out-of-order.
+//!
+//! ## The real key schedules
+//! The spec isn't great here. The key schedule is actually
```
raphaelrobert, line 48: These kind of comments are not very helpful for others. 

From what I gather, the difference between the two key schedules is only the beginning, namely:

```
                  init_secret_[n-1]
                         |                       
                         V                         
    commit_secret -> KDF.Extract                         
                         |
                         V                         
                   DeriveSecret(., "joiner")
                         |
                         V
```

I think it would be much clearer to explain the discrepancy between the spec and the implementation.

raphaelrobert (changes requested):
I see some functional changes in this PR, in the sense that previously the type safety guaranteed a certain order of execution which has now been replaced with an internal state. In general this is more error prone, but I think it's tolerable in this case. 
I also understand the need for the refactoring, because previously it was impossible to extract data for test vectors.
Mostly nits, with some questions about the actual key schedule part.

src/group/mls_group/new_from_welcome.rs
```diff
@@ -32,13 +31,27 @@ impl MlsGroup {
...
+        let group_secrets = GroupSecrets::decode_detached(&group_secrets_bytes)?;
+        let joiner_secret = group_secrets.joiner_secret;
+        // TODO #141: Implement PSK
+        let mut key_schedule = KeySchedule::init(ciphersuite, joiner_secret, None);
+        let (welcome_key, welcome_nonce) = key_schedule
+            .welcome()?
+            .derive_welcome_key_nonce(ciphersuite);
+
+        let group_info_bytes =
+            match welcome_key.aead_open(welcome.encrypted_group_info(), &[], &welcome_nonce) {
```
kkohbrok, line 49: I think `ok_or(...)` would be a bit nicer here.

src/schedule/mod.rs
kkohbrok: I'm curious why you think the extract-expand of the commit and init secrets into the joiner secret is not actually part of the key schedule. Can you elaborate on that?

kkohbrok (commented):
This should work. I'd still prefer a type safe variant without the need for runtime errors, but this might be the more rusty way. If found [this](https://hoverbear.org/blog/rust-state-machine-pattern/) post, which contains an approach that which might be a compromise that uses type safety, but does a better job of hiding the logic from the outside.

src/group/mls_group/create_commit.rs
```diff
@@ -215,3 +202,67 @@ impl MlsGroup {
...
+        invited_members: Vec<(NodeIndex, AddProposal)>,
+        provisional_tree: &RatchetTree,
+    ) -> Result<Vec<Self>, CodecError> {
+        // Get a Vector containing the node indices of the direct path to the
+        // root from our own leaf.
+        let dirpath = treemath::leaf_direct_path(
+            provisional_tree.own_node_index().into(),
+            provisional_tree.leaf_count(),
+        )
+        .expect("create_commit_internal: TreeMath error when computing direct path.");
```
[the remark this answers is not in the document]
franziskuskiefer: Sure I'll fix this. There's more here we might want to do. But this code is only moved and I don't want to touch too much of it in this PR.

src/group/mls_group/new_from_welcome.rs
franziskuskiefer: Sure, will do (`map_err` it is actually). But same comment as above. This code is only moved around and I don't want to touch it too much in this PR.

src/schedule/mod.rs
franziskuskiefer: Well there are both flows you everyone can just compare to see the differences. Also, I'd expect this to change in the spec before its done. And then we should update the comment here.
I updated the description here to explain a little more.

src/schedule/mod.rs
```diff
@@ -92,95 +162,126 @@ impl JoinerSecret {
...
+        }
+
+        Ok(InitSecret::new(
+            self.ciphersuite,
+            self.epoch_secret.as_ref().unwrap(),
+        ))
+    }
+
+    /// Derive the epoch secrets.
+    pub(crate) fn epoch_secrets(&mut self) -> Result<EpochSecrets, KeyScheduleError> {
```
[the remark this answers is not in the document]
franziskuskiefer: There is no order to these secrets. There are only two different functions because we don't always need the `init_secret`. If we want to get the additional guarantee that we can't derive secrets multiple times, we can do so with adding more complexity here. I added the additional state.
But note that first deriving the `EpochSecrets` and then the `InitSecrets` is the right order if the state machine should terminate. We don't always derive the `init_secret`, which would leave the state machine in an open state.
The other option to solve this is integrating the `init_secret` into the `EpochSecrets` struct and only deriving it when requested.

franziskuskiefer (commented):
Thanks for the comments. I'll elaborate a little on the need for this and the issues with the key schedule in the spec.

First, the refactoring does not in any change the type safety guarantees of the secrets. This all works the same way as before. The state machine only has additional checks.
We could certainly put more effort into making the state machine nicer. But it's so small I don't think it's worth the effort.

I started the refactoring in order to implement test vectors. But this is needed in any case to slowly move OpenMLS towards somewhat more maintainable code.

The key schedule as described in the spec isn't really one key schedule. The `joiner_secret` is an intermediate value *and* an output value. This must never be the case within a key schedule. The actual key schedule is therefore only the second half starting with the `joiner_secret`, which indeed is what happens when starting a group from a welcome message.

src/schedule/mod.rs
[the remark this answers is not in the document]
kkohbrok: An alternative could be modelling the key schedule as a state machine with a loop, representing the chain that it is in the spec. That would mean it permanently lives in the group state, but that could be a viable option.

src/schedule/mod.rs
```diff
@@ -225,15 +366,15 @@ impl WelcomeSecret {
...
     secret: Secret,
 }
 
 impl EpochSecret {
     /// Derive an `EpochSecret` from a `JoinerSecret`
-    pub(crate) fn new(
+    fn new(
         ciphersuite: &Ciphersuite,
-        intermediate_secret: IntermediateSecret,
+        intermediate_secret: &IntermediateSecret,
```
kkohbrok, line 377: Can't we consume the intermediate secret here? I believe that should be possible if we use `take()` in `add_context`.

src/group/mls_group/new_from_welcome.rs
kkohbrok: I thought you use `map_err` if you want to use the existing error to compute the new one. `ok_or` just ignores the existing error by default and just outputs the one you specify.

kkohbrok (changes requested):
> The key schedule as described in the spec isn't really one key schedule. The `joiner_secret` is an intermediate value _and_ an output value. This must never be the case within a key schedule. The actual key schedule is therefore only the second half starting with the `joiner_secret`, which indeed is what happens when starting a group from a welcome message.

I see what you mean. It would certainly be nice not to have to "output" the `joiner_secret`, but I think some such step is unfortunately required to allow new members to join the key schedule chain. Do you have an idea to make the structure nicer? You could also view it a bit differently, such that the joiner secret of the next epoch is the "end" of the key schedule as opposed to `init_secret_[n]`.

src/schedule/mod.rs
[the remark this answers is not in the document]
raphaelrobert: From all secrets, the init secret is the most sensitive one, because if it leaks it allows an attacker to break into a group permanently. That's why I am concerned about it being derived more than once. 
You have a point about the fact, that we don't always derive the init secret though. I personally don't care to much about the state machine terminating, being able to derive the init secret only once is more important to me. If we want the state machine to always terminate I guess we would need two different functions:
1. Derive the epoch secrets and terminate
2. Derive the epoch secrets, then the init secret and then terminate.

src/group/mls_group/create_commit.rs
```diff
@@ -221,10 +224,10 @@ impl PlaintextSecret {
         // Get a Vector containing the node indices of the direct path to the
         // root from our own leaf.
         let dirpath = treemath::leaf_direct_path(
-            provisional_tree.own_node_index().into(),
+            provisional_tree.own_node_index(),
             provisional_tree.leaf_count(),
         )
-        .expect("create_commit_internal: TreeMath error when computing direct path.");
+        .unwrap();
```
raphaelrobert, line 230: This is still missing the comment on why the `unwrap()` is safe. I agree it's not part of the PR per-se, but let's finish this correctly.

src/tree/mod.rs
```diff
@@ -460,9 +462,9 @@ impl RatchetTree {
         key_package_bundle.set_key_package(key_package.clone());
 
         (
-            self.private_tree.commit_secret(),
+            // self.private_tree.commit_secret(),
```
raphaelrobert, line 465: comments

raphaelrobert (changes requested):
Thanks for the extensive explanation, the key schedule looks good now.

Just a few nits otherwise.

src/schedule/mod.rs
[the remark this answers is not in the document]
franziskuskiefer: Ok, I don't actually think this is super important. Deleting the key material is more important. But the cleanest option then is to do it all at the same time (the `init_secret` optionally). I'll do that.

kkohbrok (approved):
Looks good to me. At some point, we will want to keep some of the epoch secrets around, while other (e.g. the `init_secret`) should be deleted for FS. Do we have an issue for that already?

franziskuskiefer:
> Looks good to me. At some point, we will want to keep some of the epoch secrets around, while other (e.g. the `init_secret`) should be deleted for FS. Do we have an issue for that already?

Yeah we certainly need to look at some FS aspects at some point.... There are some issues for some of it already. But I don't think there's one for the epoch secrets.

raphaelrobert (approved):
Thanks, LGTM!

I agree about the FS concerns, I think we have #335 for that.

PR #331, opened 2021-02-02
Fix remover bug

### LINKED ISSUE #329: [Bug] Wrong remover in callback function of ManagedGroup
**Describe the bug**
`ManagedGroup` incorrectly reports the sender of a Commit as the remover, when proposals by reference are used.

**To Reproduce**
 - Alice issues a `RemoveProposal` to remove Bob
 - Charlie commits that proposal

**Expected behavior**
The callback function should report "Alice removed Bob"

**Actual behavior**
The callback function reports "Charlie removed Bob"

**Screenshots or debug log**
N/A

all

**Additional context**
N/A

raphaelrobert:
This PR fixes #329 by passing the correct sender to the event dispatcher and additionally fixes the wrong order of leaver and remover in the callback.

A small step in code change, a huge leap in testing.

### DIFF
--- a/src/group/managed_group/mod.rs
+++ b/src/group/managed_group/mod.rs
@@ -456,7 +456,7 @@ impl<'a> ManagedGroup<'a> {
                             self.send_events(
                                 self.ciphersuite(),
                                 &commit.proposals,
-                                &plaintext.sender.sender,
+                                plaintext.sender.sender,
                                 &indexed_members,
                             );
                             // We don't need the pending proposals and key package bundles any
@@ -470,7 +470,7 @@ impl<'a> ManagedGroup<'a> {
                                 self.send_events(
                                     self.ciphersuite(),
                                     &commit.proposals,
-                                    &plaintext.sender.sender,
+                                    plaintext.sender.sender,
                                     &indexed_members,
                                 );
                                 // The group is no longer active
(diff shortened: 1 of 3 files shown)


kkohbrok (dismissed):
Looks good to me! Thanks for the fix!

src/group/managed_group/mod.rs
```diff
@@ -898,7 +898,7 @@ impl<'a> ManagedGroup<'a> {
                     if let Some(queued_proposal) = pending_proposals_queue.get(proposal_reference) {
                         self.send_proposal_event(
                             queued_proposal.proposal(),
-                            sender,
+                            &queued_proposal.sender().to_leaf_index(),
```
franziskuskiefer, line 901: This shouldn't borrow. But feel free to do in a follow up.

franziskuskiefer (approved):
lgtm with one nit.
We definitely need more tests.

src/group/managed_group/mod.rs
raphaelrobert: Agreed, it's a bit inconsistent. I can make `.to_leaf_index()` return a `&LeafIndex` and that would be my preferred way. This will lead to some more code changes elsewhere. Let me know if you're fine with that.

Alternatively I could just change the function signature of `send_proposal_event()` to accept a `LeafIndex` and copy the value. Less preferred, but also less code changes.

raphaelrobert:
> lgtm with one nit.
> We definitely need more tests.

I filed #332.

src/group/managed_group/mod.rs
franziskuskiefer: Well, `LeafIndex` really is a `u32`, which should be passed by value. If the compiler actually uses a reference here, it's probably more expensive than passing the value. So I'd recommend doing the second option.

franziskuskiefer (approved):

PR #334, opened 2021-02-04
Replace the CredentialBundle reference in ManagedGroup with a reference to a simple KeyStore

kkohbrok:
This PR introduces a simple key store implementation to OpenMLS and replaces the reference to a `CredentialBundle` in a `ManagedGroup` with a reference to a `KeyStore`.

The `KeyStore` is currently a very simple design that stores `CredentialBundle`s and `KeyPackageBundle`s. It provides functions to generate fresh instances of both, returning the `Credential` or `KeyPackageBundle` respectively, as well as the retrieval of `CredentialBundle`s given the `Credential`'s `SignaturePublicKey` and of `KeyPackageBundle`s given the `KeyPackage`'s hash.

In `ManagedGroup`, the `CredentialBundle` reference is only replaces in a very shallow way, i.e. as soon as a function requires the `CredentialBundle`, the group looks up its own credential in the public tree via its `own_node_index` and uses the `SignaturePublicKey` of that credential to look up the `CredentialBundle` in the `KeyStore`. In the process, I also removed the derivation of `PartialEq` for `ManagedGroup`. We can still to something very similar by comparing public tree and exporter key. Alternatively, we could implement `PartialEq` manually.

Since the key package (and thus the credential) in the own leaf in the public tree gets blanked or replaced when oneself is removed, `self.credential()` now can return an error, which gets triggered, if it is used after one was removed from the group. This caused changes primarily in callback functions and event listeners in test functions.

One thing I didn't quite know what to do with is the `PoisonError`s. Ideally, only the `KeyStore` would take write locks, however, the getter for `CredentialBundle`s has to return an `RwLock` (or at least that's the only way I could make it work, the Problem is the same as with the original double-`RefCell` approach). This means that outside functions could potentially change `CredentialBundle`s, which is bad from a poison-control standpoint, but also more generally from a design standpoint. @franziskuskiefer any idea how to work around this?


src/key_store/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+impl KeyStore {
+    /// Retrieve a `KeyPackageBundle` from the key store given the hash of the
+    /// corresponding `KeyPackage`. Returns an error if no `KeyPackageBundle`
+    /// can be found corresponding to the given `KeyPackage` hash. TODO: This is
+    /// not in use yet, because the groups are not yet refactored to use the key
+    /// store for KeyPackageBundles.
+    pub(crate) fn _key_package_bundle(
+        &self,
+        kp_hash: &[u8],
+    ) -> Result<&KeyPackageBundle, KeyStoreError> {
```
raphaelrobert, line 32: I'm thinking this should return the value and not just a reference, otherwise we might keep around key package bundles even though they are no longer needed (which is bad for FS).

src/key_store/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    /// Retrieve a `CredentialBundle` from the key store given the
+    /// `SignaturePublicKey` of the corresponding `Credential`. Returns an error
+    /// if no `CredentialBundle` can be found corresponding to the given
+    /// `SignaturePublicKey`.
+    /// TODO: This is currently public, because the groups are not yet
+    /// refactored to use the key store for KeyPackageBundles and thus in tests
+    /// we need to access the credential bundle to create KeyPackageBundles
+    /// ad-hoc.
+    pub fn credential_bundle(
```
raphaelrobert, line 46: Going forward it might be enough to expose the `.sign()` function instead of the whole bundle, but no objections for now.

src/key_store/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+        let kpb = KeyPackageBundle::new(ciphersuites, credential_bundle, extensions)?;
+        let kp = kpb.key_package().clone();
+        let kp_hash = kp.hash();
+        self.key_packages.insert(kp_hash, kpb);
+        Ok(kp)
+    }
+
+    /// Generate a fresh `CredentialBundle` with the given parameters and store
+    /// it in the `KeyStore`. Returns the corresponding `Credential`.
+    pub fn fresh_credential(
```
raphaelrobert, line 76: For both functions I would use `generate_` as a prefix instead of `fresh_`.

src/key_store/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+        &mut self,
+        identity: Vec<u8>,
+        credential_type: CredentialType,
+        signature_scheme: SignatureScheme,
+    ) -> Result<Credential, KeyStoreError> {
+        let cb = CredentialBundle::new(identity, credential_type, signature_scheme)?;
+        let credential = cb.credential().clone();
```
raphaelrobert, line 83: Do we really need to clone here and return the value (same for key packages)? Most likely we will just serialise it shortly after, which works with a reference as well.

src/key_store/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+
+pub use errors::KeyStoreError;
+
+#[derive(Debug, Default)]
+/// The `KeyStore` contains private key material of `Credentials` and
+/// `KeyPackage`s.
+pub struct KeyStore {
+    // Map from signature public keys to credential bundles
+    credentials: HashMap<SignaturePublicKey, CredentialBundle>,
+    key_packages: HashMap<Vec<u8>, KeyPackageBundle>,
```
raphaelrobert, line 20: Do the `HashMap`s really need a value for the key or could this work with a reference as well?

raphaelrobert (changes requested):
Thanks for the PR! Looks good in general.

My take on the question of what the key store should store: It should only store the "init keys" used for new groups. The key packages that are generated for updates in a group should not leave the group (there is no use for them outside of it). The ManagedGroup also knows best how to handle their lifecycle.

src/key_store/mod.rs
kkohbrok: Done! Note, that that's not the case for most of the tests, where I now have to clone, but you're right in that it should save us a clone in most non-test cases.

src/key_store/mod.rs
kkohbrok: Done!

src/key_store/mod.rs
kkohbrok: Done!

kkohbrok (commented):
Thanks for the review! I've implemented the requested changes and also prefixed the names of the "getter" functions wit `get_`. It's a bit unintuitive that the `KeyPackageBundle` getter removes the KPB from the store while the `CredentialBundle` getter doesn't. I'd like to reflect that in the names, but I'm not sure what's the convention here. Maybe `fetch_` or `remove_` as prefix instead of `get_`?

tests/test_managed_group.rs
```diff
@@ -167,19 +172,43 @@ fn managed_group_operations() {
...
-                ciphersuite.signature_scheme(),
-            )
-            .unwrap();
-            let bob_credential_bundle = CredentialBundle::new(
-                "Bob".into(),
-                CredentialType::Basic,
-                ciphersuite.signature_scheme(),
-            )
-            .unwrap();
+            let mut key_store = KeyStore::default();
```
franziskuskiefer, line 175: Not using `Cell` or something in the `KeyStore` of course requires you to have a mutable store. It's not used a lot but it's of course something that later needs refactoring. (Might be something to change from the beginning to safe us some refactoring.)

tests/utils/managed_utils/default_callbacks.rs
```diff
@@ -7,21 +7,29 @@ fn member_added(
     sender: &Credential,
     added_member: &Credential,
 ) {
+    let own_identity = match managed_group.credential() {
```
franziskuskiefer, line 10: duplicated code?

src/group/managed_group/mod.rs
```diff
@@ -53,10 +56,9 @@ use ser::*;
 /// Changes to the group state are dispatched as events through callback
 /// functions (see [`ManagedGroupCallbacks`]).
 #[derive(Debug)]
-#[cfg_attr(test, derive(PartialEq))]
 pub struct ManagedGroup<'a> {
-    // CredentialBundle used to sign messages
-    credential_bundle: &'a CredentialBundle,
+    // Reference to the KeyStore to obtain private key material.
+    key_store: &'a KeyStore,
```
franziskuskiefer, line 61: This implies that the `KeyStore` is owned by the client (however that looks like) and there's only one key store? This should also be clarified in the `KeyStore` module docs.

franziskuskiefer (changes requested):
Thanks for starting this! A couple thoughts.

src/key_store/mod.rs
raphaelrobert: No, I just meant for the key. The value has to be a value of course, but the key doesn't get stored in a `HashMap` it just gets hashed and it shouldn't be a lifetime issue.

kkohbrok: I see what you mean, but I'm not sure it's worth it. This would mean that the hashes of the key packages we store have to live as long as the KeyStore object. When we generate a KPB, the hash of the key package has to be computed locally. It's not part of the key package and will then go out of scope, which causes a lifetime issue. It might be possible, but I honestly just don't really see a good way to do it.

src/key_store/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+};
+
+pub mod errors;
+
+pub use errors::KeyStoreError;
+
+#[derive(Debug, Default)]
+/// The `KeyStore` contains private key material of `Credentials` and
+/// `KeyPackage`s.
+pub struct KeyStore {
```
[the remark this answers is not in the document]
kkohbrok: I created #337 to keep track of the development of the key store. I already put some considerations into the issue. Feel free to add anything you think should be added, but we can also keep discussing it in meetings. 
Doing it in individual PRs allows me to get feedback early and should keep the PRs more easily revieweable, so I'd prefer doing it bit-by-bit.

src/key_store/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+impl KeyStore {
+    /// Retrieve a `KeyPackageBundle` from the key store given the hash of the
+    /// corresponding `KeyPackage`. This removes the `KeyPackageBundle` from the
+    /// store. Returns an error if no `KeyPackageBundle` can be found
+    /// corresponding to the given `KeyPackage` hash. TODO: This is not in use
+    /// yet, because the groups are not yet refactored to use the key store for
+    /// KeyPackageBundles.
+    pub(crate) fn _get_key_package_bundle(
```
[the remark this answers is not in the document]
kkohbrok: This came from https://github.com/openmls/openmls/pull/334#discussion_r570307673. The idea is that the key package is used in the creation of a group, after which it lives on in the group state. This will change eventually with the evolution of the key store (see above or #337).

src/key_store/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// `SignaturePublicKey` of the corresponding `Credential`. Returns an error
+    /// if no `CredentialBundle` can be found corresponding to the given
+    /// `SignaturePublicKey`. TODO: This is currently public, because the groups
+    /// are not yet refactored to use the key store for KeyPackageBundles and
+    /// thus in tests we need to access the credential bundle to create
+    /// KeyPackageBundles ad-hoc.
+    pub fn get_credential_bundle(
+        &self,
+        cred_pk: &SignaturePublicKey,
+    ) -> Result<&CredentialBundle, KeyStoreError> {
```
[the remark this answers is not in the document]
kkohbrok: Right. The reason is that we expect CredentialBundles to be used across groups, while KeyPackageBundles are currently consumed when creating a new group. I expect this to change in the future, when KeyPackageBundles don't live in their groups anymore. Until then, this at least enables the creation of multiple groups with the same CredentialBundle.

tests/test_managed_group.rs
kkohbrok: Right. I'm in the process of finding a good solution for this. The goal of the next PR is to allow access via non-mutable references, such that we can actually use the `KeyPackageBundle` storage functionality.

src/key_store/mod.rs
raphaelrobert: I'm not sure the lifetime issues you describe would occur. I think `HashMap` will not take ownership here.

src/group/managed_group/mod.rs
kkohbrok: Added a sentence for clarification in the KeyStore docs.

src/key_store/mod.rs
```diff
@@ -0,0 +1,88 @@
+use std::collections::HashMap;
```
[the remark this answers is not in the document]
kkohbrok: Done!

src/key_store/mod.rs
kkohbrok: Should we pair up for this? I've tried and I keep running into issues.

kkohbrok (commented):
Thanks for the review! I've tried to answer/address all the comments. I've also added a few simple tests for generation/loading of KPBs and CBs.

tests/utils/managed_utils/default_callbacks.rs
kkohbrok: Done!

src/ciphersuite/mod.rs
```diff
@@ -708,6 +709,16 @@ impl PartialEq for SignaturePublicKey {
     }
 }
 
+impl Eq for SignaturePublicKey {
+    fn assert_receiver_is_total_eq(&self) {}
+}
```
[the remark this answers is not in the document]
kkohbrok: All of the traits are now derived. Note, that this means that for two `SignaturePublicKey` instances to be equal, both the public key _and_ their `SignatureScheme` have to be equal.

src/key_store/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+    extensions::Extension,
+    key_packages::{KeyPackage, KeyPackageBundle},
+};
+
+pub mod errors;
+
+pub use errors::KeyStoreError;
+
+#[derive(Debug, Default)]
+/// The `KeyStore` contains private key material of `Credentials` and
```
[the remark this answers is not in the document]
kkohbrok: Fixed the documentation to reflect the fact that the whole KPB/CB is stored, not only the private key material.

src/key_store/mod.rs
```diff
@@ -1,3 +1,39 @@
...
+//!
+//! // Generate credential bundles
+//! let alice_credential = key_store
+//!     .generate_credential(
+//!         "Alice".into(),
+//!         CredentialType::Basic,
+//!         SignatureScheme::ED25519,
+//!     )
+//!     .unwrap()
+//!     .clone();
```
franziskuskiefer, line 31: Why is the credential cloned after creation?

kkohbrok: If we don't clone it, we have a dangling mutable reference to the KeyStore and can't borrow immutably for the subsequent call to `get_credential_bundle`. I plan to fix this with the next PR.

src/key_store/mod.rs
kkohbrok: It appears that the `HashMap` really needs ownership of the key that you pass in `insert`. And if you pass a reference, it needs to stay alive at least as long as the `HasMap` itself. This seems to have been a design decision. There might be a data structure that doesn't do this, but I'm not sure it would be worth using something non-standard here.

See also: https://github.com/rust-lang/rust/issues/44271

kkohbrok:
Note, that I tried using `identity` to index the `CredentialBundle` instances in the store, but I stumbled across the fact that, at least when using only `BasicCredential`, it's not possible for a single identity to support multiple signature schemes. This either might be what we want in the end, or alternatively we might end up using a different kind of credential for that, but for now I didn't implement that change.

src/group/managed_group/mod.rs
```diff
@@ -635,14 +646,22 @@ impl<'a> ManagedGroup<'a> {
...
+        let tree = self.group.tree();
+        Ok(tree.own_key_package().credential().clone())
     }
 
-    /// Returns own credential
-    pub fn credential(&self) -> &Credential {
-        &self.credential_bundle.credential()
+    /// Tries to obtain the credential bundle corresponding to the group's
+    /// credential from the key store.
+    pub fn credential_bundle(&self) -> Result<&CredentialBundle, ManagedGroupError> {
```
raphaelrobert, line 661: Do we really need this for something?

raphaelrobert (commented):
LGTM in general, just one comment

src/group/managed_group/mod.rs
kkohbrok: It was meant as a convenience function, but I'm removing it with a follow-up PR anyway, since it doesn't play nicely with the `RefCell` approach. So I can remove it already an fetch it "manually" from the key store every time, or I leave it in until the next PR. Alternatively, if this PR doesn't land today, I'm going to close it anyway, so I can continue developing through next week. We can then land everything in one big go the week after next.

src/key_store/mod.rs
franziskuskiefer: Ok but this makes the entire key store pretty much unusable if every reference it returns has to be cloned.

franziskuskiefer (commented):
I don't think this is in a state where it's actually useful such that we shouldn't merge this as is.
I suggested two approaches otr of handling this sort of larger issue. I don't mind which one you're taking but we should follow one. Given that you started like this and have follow-ups lined up it's probably best to keep working on this branch until it has usable key store.
I'm happy to give feedback here (or better in PRs you put on top) when you point me to particular parts but lets wait with the review until we have something that's a little closer to what we need in the end.

src/key_store/mod.rs
kkohbrok: I'm a bit confused. I thought I made clear that this was a limitation of the current key store and that a solution (in the shape of RefCells) comes with the next PR. Note, that the cloning is only necessary for `Credential`s. For `CredentialBundle`s it's fine, because getting them only requires an immutable reference.

I feel like we always get back to this point. But since the `RefCell` solution is in all likelyhood going to be ready until next week, I'll probably end up closing this PR for a version that doesn't require cloning anyway.

kkohbrok:
Updating the Key Store to use RefCells was not that much refactoring after all an it's now included in this PR.

kkohbrok:
I've updated the KeyStore to use `RwLock` instead of `RefCell`, which creates the issue of poison-management (see PR description). We can go with this approach for now, but it's not great. I'd be grateful for ideas on how to improve this design.

src/key_store/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+
+pub use errors::KeyStoreError;
+
+#[derive(Debug, Default)]
+/// The `KeyStore` contains `CredentialBundle`s and `KeyPackageBundle`s and
+/// makes them available via information in the corresponding `Credential` and
+/// `KeyPackage` instances.
+pub struct KeyStore {
+    // Map from signature public keys to credential bundles
+    credential_bundles: RwLock<HashMap<SignaturePublicKey, RwLock<CredentialBundle>>>,
```
franziskuskiefer, line 84: So something that gives you a lot of trouble is the value in the hash map here. Why do you need to have an `RwLock` on it?

src/key_store/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+//! use openmls::prelude::*;
+//!
+//! let key_store = KeyStore::default();
+//!
+//! // Generate a credential bundle.
+//! let alice_credential = key_store
+//!     .generate_credential(
+//!         "Alice".into(),
+//!         CredentialType::Basic,
+//!         SignatureScheme::ED25519,
```
franziskuskiefer, line 31: Would be nice if we could use the `ciphersuite_name` here as well.

src/key_store/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+//! roadmap.
+//!
+//! The current key store enables the storage of `CredentialBundle` instances,
+//! and grants access to `CredentialBundle` references via the
+//! `SignaturePublicKey` of the corresponding `Credential`.
+//!
+//! A `KeyStore` is meant to be used across multiple `ManagedGroup` instances to
+//! allow sharing the same `CredentialBundle`. If this is not desired, multiple
+//! `KeyStore` instances can be used across groups.
+//!
```
franziskuskiefer, line 15: So what I'm missing here is the high level idea of which parts are stored, what's returned as a reference, what's returned as a copy (and why each time).

src/key_store/mod.rs
kkohbrok: Excellent question! It used to be a `RefCell`, because at some point, I think I convinced myself that we needed another layer of interior mutability here. But on further inspection, I think it might not actually be needed. I've removed it and it turns out it's much nicer without it 🙂

Thanks for pointing this out!

src/key_store/mod.rs
kkohbrok: I've extended the documentation to hopefully provide some clarity on this.

src/key_store/mod.rs
kkohbrok: We can and I changed it so that we do.

src/key_store/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+//! `CredentialBundle` instances are only accessible via reference, so that if a
+//! `KeyStore` is used in multiple groups, those groups can also share a
+//! `CredentialBundle`.
+//!
+//! The `KeyStore` also stores "init" `KeyPackageBundle` instances, i.e. the
+//! counterparts to the `KeyPackage` instances that are published so other
+//! parties can use them to add this party to groups. These `KeyPackageBundle`
+//! instances are removed from the `KeyStore` when they are requested via
+//! `take_key_package_bundle`. This is because each `ManagedGroup` owns the
+//! `KeyPackageBundle` in its leaf, so upon creation of the group, it needs to
```
franziskuskiefer, line 20: So this is what we do right now. But I'm not convinced yet that it's something we should be doing. This is why I'd like us to have an upgrade path from here and also describe it. I would expect us to deprecate the `take_` function here and instead introduce a `get_` function that is used when starting the client such that the `ManagedGroup` doesn't need to persist anything here.

src/key_store/mod.rs
franziskuskiefer: Can you also add explanations for `generate_key_package` and `generate_credential`?

src/key_store/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+//! storage of `CredentialBundle` and `KeyPackageBundle` instances. The
+//! development of this module is tracked in #337, which also includes a
+//! roadmap.
+//!
+//! The current key store enables the storage of `CredentialBundle` instances,
+//! and grants access to `CredentialBundle` references via the
+//! `SignaturePublicKey` of the corresponding `Credential`. The
+//! `CredentialBundle` instances are only accessible via reference, so that if a
+//! `KeyStore` is used in multiple groups, those groups can also share a
+//! `CredentialBundle`.
```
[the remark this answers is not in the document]
kkohbrok: Generally, groups will only ever read from a `CredentialBundle`, so `RwLock` should allow parallel access to the same credential without issues. This leads to your question regarding updating/deleting of a credential. I don't think Credentials or CredentialBundles should ever be updated, but only deleted. On an MLS level, the credential is part of the KeyPackage and can thus only be "changed" (i.e. replaced by a fresh one) by doing an Update. As the API stands right now, this requires client activity to intiate it. 

So the lifecycle of a CredentialBundle would look as follows:
- it gets created
- the client decides to use it in one or more groups by creating KeyPackageBundles with that CredentialBundle and updating  the groups with these KeyPackageBundles.
- potentially, the client publishes some KeyPackageBundles created with the CredentialBundle and thus potentially joins new groups with them
- if the client wants to at some point delete the CredentialBundle, it first has to create a new one and replace KeyPackages that use them. This includes KeyPackages the client previously published, but also the ones that are in use in various groups. If the client wants to be totally safe, it has to wait until all KeyPackages that include the CredentialBundle have expired (as others might have downloaded them from the server before the update and still plan to use them).

This raises the question of who keeps track of how/where a CredentialBundle is used and when KeyPackageBundles that use it expire. If at some point the KeyStore is responsible for generation of _all_ of the KeyPackageBundles (not just the "init" ones as is now the case), it could keep track of expiration and use. But that's not the case just yet.

src/key_store/mod.rs
kkohbrok: You're right. That was one of the questions that were left after our last discussion: Who should persist KeyPackages that are allocated to an individual group? Should it be the group (probably instructing the KeyStore to do it) or should it be the KeyStore directly?

Having written the answer to your other question, I now believe that it should be the KeyStore that holds (and persists) all the KeyPackageBundles and not the groups. This would allow it to keep track of expirations and active use of KeyPackageBundles and thus prevent the deletion of CredentialBundles if they are potentially still in use somewhere.

franziskuskiefer: >  I now believe that it should be the KeyStore that holds (and persists) all the KeyPackageBundles and not the groups

Yep, that's where I think we have to end up. But as I said I think we have this upgrade path. But we should also document where we (probably) want to go here.

src/key_store/mod.rs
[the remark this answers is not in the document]
franziskuskiefer: I agree that there should be no modification but only delete and create.
> Generally, groups will only ever read from a CredentialBundle, so RwLock should allow parallel access to the same credential without issues.

Right, but my understanding of the locking you do here is that the entire hashmap gets locked. This way any write operation to the hashmap (in future this might also include any read from something that's permanently stored and then cached in the hashmap) would block a read. Or do I misunderstand the design?

> This raises the question of who keeps track of how/where a CredentialBundle is used and when KeyPackageBundles that use it expire. 

Right, I think this is the main question. I currently expect this to be the client or the key store. We don't have to decide right now but we should make sure the API here is good for any route we might want to take or can be easily upgraded. And we should document all this here.

src/key_store/mod.rs
kkohbrok: I've updated #337 to document this plan.

src/key_store/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+//!
+//! The API of the `KeyStore` allows the generation of `Credential` instances
+//! via `generate_credential`, such that it stores the corresponding
+//! `CredentialBundle`. After storing them, references to `CredentialBundle`
+//! instances can be retrieved via `get_credential_bundle` using the
+//! `SignaturePublicKey` of the corresponding `Credential` as index.
+//!
+//! ## Init `KeyPackageBundle` Instances
+//!
+//! Similarly, the `KeyStore` can generage "init" `KeyPackage` instances via
```
franziskuskiefer, line 24:
```suggestion
//! Similarly, the `KeyStore` can generate "init" `KeyPackage` instances via
```

src/key_store/mod.rs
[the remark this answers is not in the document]
kkohbrok: I've updated #337 to document the plans and also added a short "Future work" section to the end of the module documentation.

src/key_store/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+
+impl KeyStore {
+    /// Retrieve a `KeyPackageBundle` from the key store given the hash of the
+    /// corresponding `KeyPackage`. This removes the `KeyPackageBundle` from the
+    /// store. Returns an error if no `KeyPackageBundle` can be found
+    /// corresponding to the given `KeyPackage` hash.
+    pub fn take_key_package_bundle(&self, kp_hash: &[u8]) -> Option<KeyPackageBundle> {
+        // We unwrap here, because the two functions claiming write locks (this
+        // one and `generate_key_package`) only hold the lock very briefly and
+        // should not panic during that period.
```
franziskuskiefer, line 156: Hm, "should not panic" is very optimistic... We should handle all the potential locks in these functions somehow. Or add it as a todo. But this has huge potential to create bugs that are impossible to debug.

src/key_store/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+#[cfg(test)]
+mod test_key_store;
+
+pub use errors::KeyStoreError;
+
+#[derive(Debug, Default)]
+/// The `KeyStore` contains `CredentialBundle`s and `KeyPackageBundle`s and
+/// makes them available via information in the corresponding `Credential` and
+/// `KeyPackage` instances.
+pub struct KeyStore {
```
franziskuskiefer, line 121: Looks like we currently don't have a way to actually delete anything here? I think we can go without in the first version but should document it. Because this will also be necessary for testing of all the async things it would be nice to have it earlier. But I'm fine without in the first version.

src/key_store/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// it in the `KeyStore`. Returns the corresponding `Credential` or an error
+    /// if the creation of the `CredentialBundle` fails.
+    pub fn generate_credential(
+        &self,
+        identity: Vec<u8>,
+        credential_type: CredentialType,
+        signature_scheme: SignatureScheme,
+    ) -> Result<Credential, KeyStoreError> {
+        let cb = CredentialBundle::new(identity, credential_type, signature_scheme)?;
+        let signature_key = cb.credential().signature_key().clone();
```
franziskuskiefer, line 215: Why all this cloning? Appears to be a little excessive.

franziskuskiefer (commented):
I think we're getting there 🙂

src/key_store/mod.rs
kkohbrok: Yes, I wasn't sure what to write here and I agree that "should not panic" is very optimistic. The way I understand poisoning is that something is considered "poisoned" if a piece of code panics while holding a write lock. This means that as long as we thoroughly check what is going on while we hold a write lock, we should be ok. It's a fine line, but I would argue that it's doable, since the only thing we (currently) do is taking the write lock, inserting or removing a value and then immediately dropping it again. Of course, this is dangerous, because adding new code could break this pattern. That being said, I still believe that if we're careful here, unwrapping should be fine.

src/key_store/mod.rs
kkohbrok: I believe we have to clone here at least once, because right afterwards, the `CredentialBundle` is consumed by `insert`. I've refactored a bit to get rid of a clone. Both `generate_` functions currently return a full `Credential`/`KeyPackage` instance. We could also try returning a reference, but that would require another `Guard` struct and it would keep a read lock on the corresponding `HashMap`. What do you think? I think a clone is ok here.

src/key_store/mod.rs
kkohbrok: Yes, I've not added that yet. I thought I'd add it once we have a way to keep track of `Credential` usage, to make sure that we can safely delete `CredentialBundle` instances. Not sure when that is going to happen, though.

raphaelrobert:
@kkohbrok should we close this in favour of #385?

kkohbrok:
> @kkohbrok should we close this in favour of #385?

Yes, I meant to do that. Thanks for the reminder!

PR #336, opened 2021-02-04
key schedule test vectors

franziskuskiefer:
This PR implements test vector generation and test for the key schedule
https://github.com/mlswg/mls-implementations/blob/master/test-vectors.md#key-schedule


src/schedule/kat_key_schedule.rs
```diff
@@ -0,0 +1,274 @@
...
+    write("test_vectors/kat_key_schedule_openmls-new.json", &tests);
+}
+
+#[test]
+fn run_test_vectors() {
+    let tests: Vec<KeyScheduleTestVector> = read("test_vectors/kat_key_schedule_openmls.json");
+
+    for test_vector in tests {
+        let ciphersuite =
+            CiphersuiteName::try_from(test_vector.cipher_suite).expect("Invalid ciphersuite");
```
kkohbrok, line 183: Wouldn't it make sense to fail gracefully here? I would imagine that in the future, there could be ciphersuites in the test vectors that we don't support. Or is this just for our own test vectors and not for interop?

kkohbrok (approved):
One question, otherwise it looks good to me!

src/schedule/kat_key_schedule.rs
franziskuskiefer: good point. Not here though. This always gets us the ciphersuite (unless its one that's not in the RFC, but we don't care about that case). I'll check for the supported cipher suite a line below. I'll do the same for the encryption test vectors.

kkohbrok (commented):
Thanks! Looks good to me now. Since auto merge is on, I'll leave it to you @raphaelrobert to take a look first and merge it if you like.

src/schedule/kat_key_schedule.rs
raphaelrobert: Good point indeed, we should expect all the ciphersuites that are mentioned in the spec to also appear in the test vectors and we don't support all of them right now (and probably not for a while).

src/schedule/kat_key_schedule.rs
```diff
@@ -0,0 +1,282 @@
...
+    let welcome_secret = key_schedule.welcome().unwrap();
+
+    let confirmed_transcript_hash = randombytes(ciphersuite.hash_length());
+
+    let group_context = GroupContext::new(
+        GroupId::from_slice(group_id),
+        GroupEpoch(epoch),
+        tree_hash.to_vec(),
+        confirmed_transcript_hash.clone(),
+        &[], // Extensions
```
raphaelrobert, line 87: You could add the (one and only) extension here if you wanted. They are not supported in the MLS groups but should work in the key schedule already. You would have to tweak the tree hash however.

raphaelrobert (commented):
LGTM in general, just a question about whether group extensions should be covered as well

src/schedule/kat_key_schedule.rs
franziskuskiefer: It's currently not part of the test vectors. We can extend them and then add them here. But this is the definition right now and it doesn't really change the key schedule. So I'm not sure we want to add anything here.

raphaelrobert (approved):
Let's get this in!

PR #338, opened 2021-02-05
transcript test vectors

franziskuskiefer:
Test vector generation and testing for [Commits and Transcript Hashes](https://github.com/mlswg/mls-implementations/blob/master/test-vectors.md#commits-and-transcript-hashes).


src/group/tests/kat_transcripts.rs
```diff
@@ -0,0 +1,219 @@
...
+                authenticated_data: randombytes(48),
+                content_type: ContentType::Commit,
+                content: MLSPlaintextContentType::Commit(Commit {
+                    proposals: vec![],
+                    path: None,
+                }),
+                signature: Signature::new_empty(),
+                confirmation_tag: None,
+                membership_tag: None,
+            };
```
raphaelrobert, line 79: This probably doesn't matter here, but this Commit message wouldn't be valid because it should either contain a `path` or add proposals.

raphaelrobert (commented):
Looks good in general, but it contains the changes from the previous PR I think

src/group/tests/kat_transcripts.rs
franziskuskiefer: yeah we don't care about validity here

franziskuskiefer:
> Looks good in general, but it contains the changes from the previous PR I think

yeah it was against the another because that was still in review. It's rebased now.

kkohbrok (approved):
Looks good to me!

PR #343, opened 2021-02-08
Initial PSK PR

raphaelrobert:
This PR is an initial PR to tackle PSKs. Since it's already quite big, I'd like to get some feedback on the general strategy before I go further.

It contains:
 - A PSK fetcher callback for `MlsGroup`
 - Support for creating `ExternalPskProposal`s
 - Consuming the PSKs in `MlsGroup` in `new()`, `new_from_welcome()`, `create_commit()` and `apply_commit()`.
 - Exporting resumption secrets in `MlsGroup`
 - A resumption secret store in `ManagedGroup`
 - Initial tests for positive cases (group with PSK as a proposal and in the `Welcome`)
 - Adaptation of the various unit tests
 - Mentions where more code for PSK support is needed
 
It does not contain:
 - Support for PSKs in `ManagedGroup`
 - Support for other types of PSKs


src/group/managed_group/mod.rs
```diff
@@ -459,6 +482,10 @@ impl<'a> ManagedGroup<'a> {
                                 plaintext.sender.sender,
                                 &indexed_members,
                             );
+                            // Extract and store the resumption secret for the current epoch
+                            let resumption_secret = self.group.epoch_secrets().resumption_secret();
+                            self.resumption_secret_store
+                                .add(self.group.context().epoch(), resumption_secret.clone());
```
kkohbrok, line 488: Do we really need to clone here? If we put it in a store, it might be nicer to make it an `Option` in `EpochSecrets` or just just return/derive it separately. I believe then we don't need to derive `clone` at all.

src/schedule/psk.rs
```diff
@@ -32,40 +34,71 @@ impl TryFrom<u8> for PSKType {
...
+            psk: Psk::External(self.external_psk.clone()),
+            psk_nonce: self.nonce.clone(),
+        }
+    }
+    pub fn secret(&self) -> &Secret {
+        &self.secret
+    }
+}
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
 pub struct ReinitPsk {
```
kkohbrok, line 77: Now that they're a bit more relevant, some comments would be nice on each of these structs.

src/tree/mod.rs
```diff
@@ -735,13 +739,30 @@ impl RatchetTree {
...
+        let psks: Vec<PreSharedKeyID> = proposal_queue
+            .filtered_by_type(ProposalType::Presharedkey)
+            .map(|queued_proposal| {
+                // Unwrapping here is safe because we know the proposal type
+                let psk_proposal = queued_proposal.proposal().as_presharedkey().unwrap();
+                psk_proposal.psk
+            })
+            .collect();
+        if psks.is_empty() {
+            println!("No PSKs");
```
kkohbrok, line 752: I believe those should be logging calls as opposed to `println`s.

kkohbrok (changes requested):
Looks very nice already! Just a few comments. Also, we probably need more tests for the `mls_group` PSK functionality. I believe there's currently only one test that makes sure it works, but I think we need more tests for all the failure conditions along the way.

src/messages/proposals.rs
```diff
@@ -133,6 +133,12 @@ impl Proposal {
             _ => None,
         }
     }
+    pub(crate) fn as_presharedkey(&self) -> Option<PreSharedKeyProposal> {
```
franziskuskiefer, line 136: Shouldn't we either take the proposals out of the queue here or (as we do for the others) work with a reference? (That's at least the only call for this I can see.)

src/group/managed_group/mod.rs
```diff
@@ -71,6 +76,8 @@ pub struct ManagedGroup<'a> {
     // The AAD that is used for all outgoing handshake messages. The AAD can be set through
     // `set_aad()`.
     aad: Vec<u8>,
+    // Resumption secret store. This is where the resumption secrets are kept in a rollover list.
+    resumption_secret_store: ResumptionSecretStore,
```
franziskuskiefer, line 80: Should we really keep them here or push them to the key store? Or do you see a need to always have the values around?

franziskuskiefer (changes requested):
Thanks for starting this. I think there are a couple questions open around how to handle the PSKs.

Who's supposed to own the PSKs? Are they supposed to be passed in as reference or owning? This also depends on whether you read it from the key store or pass it in directly.

Running test vectors still skip cases with a PSK. We should enable them. (kat_key_schedule.rs:206)

src/group/mls_group/mod.rs
```diff
@@ -65,6 +66,7 @@ impl MlsGroup {
         ciphersuite_name: CiphersuiteName,
         key_package_bundle: KeyPackageBundle,
         config: GroupConfig,
+        psk_option: Option<PskSecret>,
```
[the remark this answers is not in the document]
raphaelrobert: In this case we need to know the PSK right away, we can't set it later.

src/group/managed_group/mod.rs
raphaelrobert: Since #334 hasn't landed yet, I left it here. We could also so it as apart of #345.

src/group/managed_group/mod.rs
raphaelrobert: I agree that we could take it out here in the context of `ManagedGroup`, but since it's provided by `MlsGroup`, it might be too restrictive there.

src/group/managed_group/resumption.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub(crate) fn new(number_of_secrets: usize) -> Self {
+        Self {
+            number_of_secrets,
+            resumption_secrets: vec![],
+            cursor: 0,
+        }
+    }
+
+    /// Add a new entry to the store.
+    pub(crate) fn add(&mut self, epoch: GroupEpoch, resumption_secret: ResumptionSecret) {
```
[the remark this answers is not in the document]
raphaelrobert: I guess that could work as well, although it feels more brittle. I thought the linear effort for searching in the resumption keys was tolerable, since this is not called very often and we don't expect a huge number of keys.
If you think we need this performance at the expense of robustness (and possibly hidden cost to initialise the `HashMap` whenever we load the state, we can discuss.

src/group/mls_group/create_commit.rs
```diff
@@ -117,10 +118,20 @@ impl MlsGroup {
...
+            &apply_proposals_values.presharedkeys,
         )?;
 
-        // TODO #141: Implement PSK
-        let mut key_schedule = KeySchedule::init(ciphersuite, joiner_secret, None);
+        // Create key schedule
+        let mut key_schedule = KeySchedule::init(
+            ciphersuite,
+            joiner_secret,
+            psk_output(
```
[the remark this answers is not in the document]
raphaelrobert: The reason we have this is because we don't know what the proposals contain beforehand, especially those that we received from other members. We would have to parse them ahead of time to know exactly what PSKs to pass here.

src/group/mls_group/create_commit.rs
```diff
@@ -223,6 +234,7 @@ impl PlaintextSecret {
         joiner_secret: &JoinerSecret,
         invited_members: Vec<(NodeIndex, AddProposal)>,
         provisional_tree: &RatchetTree,
+        presharedkeys: &PreSharedKeys,
```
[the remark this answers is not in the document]
raphaelrobert: I will address this at the spec level, the `Option` doesn't make sense IMHO.

src/group/mls_group/mod.rs
```diff
@@ -196,6 +203,30 @@ impl MlsGroup {
         .map_err(GroupError::CodecError)
     }
 
+    // 11.1.4. PreSharedKey
```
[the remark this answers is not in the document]
raphaelrobert: I stayed with the current scheme, but all of `MlsGroup` needs proper documentation. I filed #346.

src/group/mls_group/new_from_welcome.rs
```diff
@@ -39,8 +40,22 @@ impl MlsGroup {
         )?;
         let group_secrets = GroupSecrets::decode_detached(&group_secrets_bytes)?;
         let joiner_secret = group_secrets.joiner_secret;
-        // TODO #141: Implement PSK
-        let mut key_schedule = KeySchedule::init(ciphersuite, joiner_secret, None);
+
+        // Create key schedule
+        let presharedkeys = PreSharedKeys {
+            psks: match group_secrets.psks {
```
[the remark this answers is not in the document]
raphaelrobert: Fully agree. I implemented what's in the spec literally, and I want to solve this at the spec level. `Option<Vec>` makes no sense.

src/messages/proposals.rs
raphaelrobert: I also agree with that. The problem is that it would require a bigger refactoring of how the proposal queue works, otherwise we immediately run into lifetime issues. I filed #347.

src/schedule/psk.rs
```diff
@@ -32,40 +34,71 @@ impl TryFrom<u8> for PSKType {
...
+
+#[cfg(test)]
+pub struct ExternalPskBundle {
+    secret: Secret,
+    nonce: Vec<u8>,
+    external_psk: ExternalPsk,
+}
+
+#[cfg(test)]
+impl ExternalPskBundle {
```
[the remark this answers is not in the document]
raphaelrobert: fixed

src/schedule/psk.rs
```diff
@@ -126,3 +166,56 @@ impl PreSharedKeys {
...
+    secret: Secret,
+}
+
+impl PskSecret {
+    /// Create a new `PskSecret` from PSK IDs and PSKs
+    pub fn new(ciphersuite: &Ciphersuite, psk_ids: &[PreSharedKeyID], psks: &[Secret]) -> Self {
+        let mut secret = vec![];
+        for (index, psk) in psks.iter().enumerate() {
+            let psk_input = ciphersuite.hkdf_extract(None, psk);
+            let psk_label = PskLabel::new(&psk_ids[index], index as u16, psks.len() as u16)
```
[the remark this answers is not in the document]
raphaelrobert: fixed

tests/test_decryption_key_index.rs
```diff
@@ -17,7 +17,7 @@ ctest!(decryption_key_index_computation {
     let update_policy = UpdatePolicy::default();
     let callbacks = default_callbacks::default_callbacks();
     let managed_group_config =
-        ManagedGroupConfig::new(handshake_message_format, update_policy, 10, callbacks);
+        ManagedGroupConfig::new(handshake_message_format, update_policy, 10, 0, callbacks);
```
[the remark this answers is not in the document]
raphaelrobert: I agree. Any general recommendation? Wrap them in structs?

src/group/managed_group/mod.rs
kkohbrok: In the end it doesn't matter too much, as the resumption secret is actually intended to be potentially used multiple times. I'm ok to leave it as is for now.

src/group/mls_group/mod.rs
```diff
@@ -430,6 +469,13 @@ impl MlsGroup {
     }
 }
 
+// Callback functions
+
+/// This callback functions is used in several places in `MlsGroup`.
```
kkohbrok, line 474:
```suggestion
/// This callback function is used in several places in `MlsGroup`.
```

kkohbrok (approved):
Thanks for addressing the comments. Just one typo, otherwise it looks good!

src/group/mls_group/mod.rs
```diff
@@ -470,6 +470,10 @@ impl MlsGroup {
 }
 
 // Callback functions
+
+/// This callback function is used in several places in `MlsGroup`.
```
franziskuskiefer, line 474: So when we land this now we should add a TODO and an issue number to make clear that this is probably going away again.

src/messages/mod.rs
```diff
@@ -295,10 +295,11 @@ pub(crate) struct GroupSecrets {
 
 impl GroupSecrets {
     /// Create new encoded group secrets.
-    pub(crate) fn new_encoded(
+    pub(crate) fn new_encoded<'a>(
         joiner_secret: &JoinerSecret,
         path_secret: Option<PathSecret>,
-        psks_option: Option<&PreSharedKeys>,
+        //psks_option: Option<&PreSharedKeys>,
```
franziskuskiefer, line 301: remove

franziskuskiefer (changes requested):
A couple nits and you'll need another commit in any case to fix the tests.
Even if not ideal, let's get this in then and file a follow up to get the key store and callback story figured out, and start using better processes after this one.

raphaelrobert:
I fixed the tests. I had to revert one of the changes regarding the review comments. We have to wait until #348 is implemented for that.

src/group/mls_group/create_commit.rs
```diff
@@ -181,6 +181,7 @@ impl MlsGroup {
 
             // Encrypt GroupInfo object
             let (welcome_key, welcome_nonce) = welcome_secret.derive_welcome_key_nonce(ciphersuite);
+            println!("Create Commit: Welcome key: {:?}", welcome_key);
```
franziskuskiefer, line 184: Debug output?

src/messages/mod.rs
```diff
@@ -295,11 +295,10 @@ pub(crate) struct GroupSecrets {
 
 impl GroupSecrets {
     /// Create new encoded group secrets.
-    pub(crate) fn new_encoded<'a>(
+    pub(crate) fn new_encoded(
         joiner_secret: &JoinerSecret,
         path_secret: Option<PathSecret>,
-        //psks_option: Option<&PreSharedKeys>,
-        psks_option: impl Into<Option<&'a PreSharedKeys>> + crate::codec::Codec,
+        psks_option: Option<&PreSharedKeys>,
```
franziskuskiefer, line 301: This change would've been ok.

franziskuskiefer (changes requested):
Good to go when the debug output is removed.

franziskuskiefer (approved):
:shipit:

Issue #205, opened 2020-11-11
Make MAC tag length dependent on the ciphersuite


kkohbrok:
The variable `TAG_BYTES` is currently a hard-coded constant. To support future ciphersuites with variable MAC tag lengths, it would be good to make this a property of the ciphersuite.

franziskuskiefer:
fixed in #378

PR #344, opened 2021-02-08
Make MAC tag length variable

claucece:
This is related to #205 . It is not perfect by any means, but I was thinking it could be an starting point.
Comments welcome @raphaelrobert @franziskuskiefer

### DIFF
--- a/src/ciphersuite/mod.rs
+++ b/src/ciphersuite/mod.rs
@@ -419,7 +419,7 @@ impl Ciphersuite {
         hkdf_extract(self.hmac, salt.value.as_slice(), ikm.value.as_slice())
     }
 
-    /// Get the length of the used hash algorithm.
+    /// Get the length of the used mac algorithm.
     pub(crate) fn mac_length(&self) -> usize {
         // TODO: don't hard-code tag bytes, but use the mac_size func (Issue #205)
         16
--- a/src/ciphersuite/mod.rs
+++ b/src/ciphersuite/mod.rs
@@ -421,6 +421,7 @@ impl Ciphersuite {
 
     /// Get the length of the used hash algorithm.
     pub(crate) fn mac_length(&self) -> usize {
+        // TODO: don't hard-code tag bytes, but use the mac_size func (Issue #205)
         16
     }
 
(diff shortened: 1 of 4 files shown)


kkohbrok (dismissed):
Thanks for the PR!

Looks good to me! At some point, the mac length should probably be exposed via `AeadMode`, but I think `evercrypt` just hardcodes it to 16 at the moment, so this is a good step in the right direction.

src/ciphersuite/mod.rs
```diff
@@ -562,6 +570,7 @@ impl AeadKey {
         AeadKey {
             aead_mode,
             value: aead_key_gen(aead_mode),
+            mac_len: 16,
```
raphaelrobert, line 573: We could pass `&Ciphersuite` as an argument here as well in order to avoid the hard-coding

raphaelrobert (changes requested):
Thanks for the PR! LGTM with one nit

franziskuskiefer:
> Looks good to me! At some point, the mac length should probably be exposed via `AeadMode`, but I think `evercrypt` just hardcodes it to 16 at the moment, so this is a good step in the right direction.

yeah this is currently hard coded. But I'm happy to add something. I want to make a new release soon anyway. File an issue or a PR over there https://www.franziskuskiefer.de/evercrypt-rust

kkohbrok:
> > Looks good to me! At some point, the mac length should probably be exposed via `AeadMode`, but I think `evercrypt` just hardcodes it to 16 at the moment, so this is a good step in the right direction.
> 
> yeah this is currently hard coded. But I'm happy to add something. I want to make a new release soon anyway. File an issue or a PR over there https://github.com/franziskuskiefer/evercrypt-rust/

I've created a PR with a relatively primitive solution that would allow this over in the evercrypt repo.

src/ciphersuite/mod.rs
claucece: Just added it ;)

claucece:
Oh, thank you @kkohbrok should we add that as a new PR once a new release of evercrypt happens?

raphaelrobert:
> Oh, thank you @kkohbrok should we add that as a new PR once a new release of evercrypt happens?

I think @franziskuskiefer was fast for that one: https://github.com/franziskuskiefer/evercrypt-rust/pull/19
But I suppose it requires a new release of evercrypt?

franziskuskiefer:
> But I suppose it requires a new release of evercrypt?

Yeah this needs a new release (or we have to point to the git version in the meantime). I'm waiting for the new HPKE test vectors before I do the new releases. This probably takes a couple more days.

raphaelrobert:
Then we can merge this and do a follow-up PR.

raphaelrobert (dismissed):

raphaelrobert (approved):

kkohbrok (approved):

PR #350, opened 2021-02-22
Enable ratchet tree extensions for MlsGroup

raphaelrobert:
This PR enables the ratchet tree extensions for `MlsGroup` and provides a positive and negative test.

(this PR initially would also add the extension to the group context, which was wrong)


src/group/mls_group/new_from_welcome.rs
```diff
@@ -73,24 +73,27 @@ impl MlsGroup {
...
+                .count()
+                > 1
             {
                 return Err(WelcomeError::DuplicateRatchetTreeExtension);
             }
+            let extension = group_info.extensions_mut().get(i).unwrap();
             match extension.as_ratchet_tree_extension() {
                 Ok(e) => {
-                    let ext = Some(e.clone());
+                    //let ext = Some(e.clone());
```
kkohbrok, line 91: Old code?

src/group/mls_group/new_from_welcome.rs
```diff
@@ -73,24 +73,27 @@ impl MlsGroup {
...
         let ratchet_tree_extension = if let Some(i) = ratchet_tree_ext_index {
-            let extension = group_info.extensions_mut().remove(i);
             // Throw an error if we there is another ratchet tree extension.
             // We have to see if this makes problems later as it's not something
             // required by the spec right now.
             if group_info
                 .extensions()
                 .iter()
-                .any(|e| e.extension_type() == ExtensionType::RatchetTree)
+                .filter(|e| e.extension_type() == ExtensionType::RatchetTree)
```
kkohbrok, line 82: We're iterating twice over the extensions here. I think you can optimize by filtering first, then checking for duplicates and then using the one that's left (if there is one).

src/extensions/test_extensions.rs
```diff
@@ -51,3 +59,152 @@ fn lifetime() {
     std::thread::sleep(std::time::Duration::from_secs(1));
     assert!(!ext.is_valid());
 }
+
+// This tests the ratchet tree extension to deliver the public ratcheting tree
+// in-band
+#[test]
+fn ratchet_tree_extension() {
+    for ciphersuite in Config::supported_ciphersuites() {
```
kkohbrok, line 67: I think this can be replaced by the `ctest` macro.

kkohbrok (changes requested):
Thanks for tackling this! Just a few comments.

franziskuskiefer (changes requested):
What Konrad said.
And please move all the unrelated formatting changes to a separate PR.

src/group/mls_group/create_commit.rs
```diff
@@ -94,14 +94,25 @@ impl MlsGroup {
...
         )?;
 
+        // Calculate tree hash
         let tree_hash = provisional_tree.tree_hash();
-        // TODO #186: Implement extensions
+
+        // Check for group extensions and include them
+        let extensions: Vec<Box<dyn Extension>> = if self.use_ratchet_tree_extension {
+            vec![Box::new(RatchetTreeExtension::new(
+                provisional_tree.public_key_tree_copy(),
```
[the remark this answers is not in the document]
raphaelrobert: I agree that would be a good optimisation for larger groups. Ultimately you need to clone the whole tree of course when you send out a Welcome message, but you shouldn't have to do it when you just apply a Commit.

src/group/mls_group/new_from_welcome.rs
raphaelrobert: Where else are we iterating? I'm not really concerned about performance here, the list of extensions should be < 10 most of the time.

src/group/mls_group/create_commit.rs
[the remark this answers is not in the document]
raphaelrobert: I looked into this again and I don't see an easy way to improve the performance here. The problem is that we need the full payload of the group extensions for the group context, which is just a serialisation of the payload. The byte vector is later used to derive the epoch secret. So in order to get rid of all duplication of the public tree, the epoch secret would have to be generated by iterating over the tree. Looking into that, it turns out that it leads to a number of issues:
 - serde cannot deserialise generic impls, meaning we would have tor write all deserialisation for all extensions by hand
 - we need to introduce lifetimes for all extensions (because this particular one now depends on the lifetime of a tree), which is complicated at best and potentially impossible

Given the above, I'd like to punt on making this more efficient for now. This only introduces an overhead when the ratchet tree extension is actually used, which by design is is incompatible with really large groups anyway.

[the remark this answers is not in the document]
kkohbrok: Thanks for looking into this! I see that it's complicated and I think we can just land this for now. I feel like we've now run into a couple of these issues that we likely want to revisit later to improve performance, so I've created #359 for this issue in particular and #358 to keep track of existing performance bottlenecks.

kkohbrok (approved):
Just the minor issue with the redundant iteration left. Otherwise this looks really good!

franziskuskiefer (commented):
looks like @kkohbrok has this covered

franziskuskiefer (approved):
but looks like I still need to approve...

raphaelrobert:
There's a pending discussion about what should be included in the `GroupContext`.  The consensus seems to be that the ratchet tree extension should not be included, so I need to rewrite this.

raphaelrobert:
I now refactored the PR so that the ratchet tree extension is not added to the group context. Since the group extensions are not clearly defined in the spec, #186 remains open for now.

kkohbrok:
If the RatchetTreeExtension is not a group extension that members agree on, we could in theory handle it on a per-client basis. Right now, it's "inherited" by clients in way. If they receive a Welcome with a RatchetTreeExtension, they will in the future also include a RatchetTreeExtension in any Welcome that they send. Should we maybe move this decision to the ManagedGroupConfig?

src/extensions/ratchet_tree_extension.rs
```diff
@@ -35,7 +35,6 @@ pub struct RatchetTreeExtension {
     tree: Vec<Option<Node>>,
 }
 
-// TODO #186: Implement group extensions
```
franziskuskiefer, line 38: Shall we leave this in then until the issue and https://github.com/mlswg/mls-protocol/pull/464 are resolved?

src/group/mls_group/new_from_welcome.rs
```diff
@@ -63,43 +63,38 @@ impl MlsGroup {
...
+            None
+        } else if ratchet_tree_extensions.len() == 1 {
+            let extension = ratchet_tree_extensions
+                .pop()
+                // Unwrappig here is safe because we know we only have one element
+                .unwrap()
+                .as_ratchet_tree_extension()
+                // Unwrapping here is safe, because we know the extension type already
+                .unwrap()
+                // We clone the nodes here upon extraction, so that we don't have to clone
```
franziskuskiefer, line 87: Not sure I understand why they need cloning. But this is a performance issue that will either become one later and we can fix it then, or it doesn't 🙂

src/group/mls_group/new_from_welcome.rs
```diff
@@ -63,43 +63,38 @@ impl MlsGroup {
...
+                // them later when we build the tree
+                .clone();
+            Some(extension)
+        } else {
+            // Throw an error if we there is more than one ratchet tree extension.
```
franziskuskiefer, line 92:
```suggestion
            // Throw an error if there is more than one ratchet tree extension.
```

franziskuskiefer: Do we have a test for this?

franziskuskiefer (changes requested):
This might be something that's not clearly defined in the spec. But we can still take an implementation decision here that would still be fine with the spec. But I'm ok with leaving it like this and seeing what comes from https://github.com/mlswg/mls-protocol/pull/464

lgtm but we should see if we can test the more than one ratchet tree extension error case.

src/extensions/ratchet_tree_extension.rs
raphaelrobert: Yes, #186 is essentially blocked by that.

src/group/mls_group/new_from_welcome.rs
raphaelrobert: We need an actual copy here for the Welcome message.

raphaelrobert:
> If the RatchetTreeExtension is not a group extension that members agree on, we could in theory handle it on a per-client basis. Right now, it's "inherited" by clients in way. If they receive a Welcome with a RatchetTreeExtension, they will in the future also include a RatchetTreeExtension in any Welcome that they send. Should we maybe move this decision to the ManagedGroupConfig?

This is definitely an implementation decision we take here. We still need to implement it for `ManagedGroup` though.

franziskuskiefer (approved):
lgtm

PR #357, opened 2021-03-01
Expose key schedule test vector generation and checking behind flag

kkohbrok:
This PR exposes the generation and checking of the key schedule test vector behind a feature flag. This is necessary to enable the test vector functionality of the (planned) interop client.

To make the code a bit more usable, I've replaced the asserts in the kat-checking code with Errors.

The goal is to expose all test vectors this way, but I'm starting with the key schedule one to discuss how to best go about this. Once we decide on a way to do this, I'll create a follow-up PR with the rest of the test vectors.

### DIFF
--- a/src/schedule/kat_key_schedule.rs
+++ b/src/schedule/kat_key_schedule.rs
@@ -50,7 +50,7 @@ struct Epoch {
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone, Default)]
-struct KeyScheduleTestVector {
+pub struct KeyScheduleTestVector {
     cipher_suite: u16,
     group_id: String,
     initial_init_secret: String,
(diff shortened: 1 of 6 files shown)


raphaelrobert (commented):
Thanks! LGTM in general. 

Since the KAT vector generation is now not part of the usual tests anymore, do we expect any errors in the generation that we would miss otherwise? Say, an error that would otherwise not manifest itself?

kkohbrok:
> Thanks! LGTM in general.
> 
> Since the KAT vector generation is now not part of the usual tests anymore, do we expect any errors in the generation that we would miss otherwise? Say, an error that would otherwise not manifest itself?

KAT generation and checking should still be part of the tests via `read_`/`write_test_vectors`, so I would expect no change.

src/schedule/kat_key_schedule.rs
```diff
@@ -209,14 +215,16 @@ fn run_test_vectors() {
             }
 
             let joiner_secret = JoinerSecret::new(ciphersuite, &commit_secret, &init_secret);
-            assert_eq!(hex_to_bytes(&epoch.joiner_secret), joiner_secret.as_slice());
+            if hex_to_bytes(&epoch.joiner_secret) != joiner_secret.as_slice() {
+                return Err(KSTestVectorError::JoinerSecretMismatch);
```
franziskuskiefer, line 219: This is a bad idea. This code is supposed to test if things are working and hence should assert at the exact point where something goes wrong. Why do you think you need errors?

src/schedule/kat_key_schedule.rs
```diff
@@ -49,7 +50,7 @@ struct Epoch {
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone, Default)]
-struct KeyScheduleTestVector {
+pub struct KeyScheduleTestVector {
```
franziskuskiefer, line 53: Why do you need this and some of the other functions in here public? This module isn't public anywhere afaik such that this doesn't make any difference.

src/schedule/kat_key_schedule.rs
```diff
@@ -108,76 +109,81 @@ fn generate(
     )
 }
 
+#[cfg(any(feature = "expose-test-vectors", test))]
```
franziskuskiefer, line 112: Since the entire module is only included if `#[cfg(test)]` these feature configs don't make any difference.

franziskuskiefer (changes requested):

kkohbrok:
I didn't touch this PR in a while and have, for the sake of moving a little faster, primarily worked on the `konrad/managed_client` branch during last weeks interop focus. The result is pretty messy, as was to be expected. I'm going to work on figuring out how to convert changes on `konrad/managed_client` into nice PRs, but that might take some time. As part of that process, I hope to update this PR as well in the near future.

src/schedule/kat_key_schedule.rs
kkohbrok: In the context of interop, I'd like to call this function and, if it fails, it would be nice to get some information back, which can then be reported to the test runner. If I have asserts here, I could still try to use `catch_unwind`, but using errors seemed to be the cleaner approach. I'm happy to use another approach here if you have a suggestion.

In case it helps: The calling function should panic right after an error is returned, so not too far from where things go wrong.

src/schedule/kat_key_schedule.rs
kkohbrok: Right, module exposure is something I missed in this PR and have "fixed" in the interop branch. But as I said in another comment, after the interop week is over now, I need to take some time to figure out how to do these things cleanly.

src/schedule/kat_key_schedule.rs
franziskuskiefer: If you really want to return errors here, which might make sense in the case you describe, you should still keep the assert but only in `cfg(test)` to keep the test as it is.

franziskuskiefer:
> I didn't touch this PR in a while and have, for the sake of moving a little faster, primarily worked on the `konrad/managed_client` branch during last weeks interop focus. The result is pretty messy, as was to be expected. I'm going to work on figuring out how to convert changes on `konrad/managed_client` into nice PRs, but that might take some time. As part of that process, I hope to update this PR as well in the near future.

Ok, I'll ignore this PR then for now and wait for an update 🙂

src/schedule/kat_key_schedule.rs
kkohbrok: Aah, that's a great idea! I hadn't thought about that.

kkohbrok:
This PR has gotten stale. A lot of work has been going on in the `interop` branch in the meantime and various changes have already been ported to `main`.

PR #360, opened 2021-03-04
Allow context as input for exported secret

kkohbrok:
This PR changes the API of `ManagedGroup` and `MlsGroup` by allowing `context` as additional input to `export_secret` in accordance with the MLS Spec: https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#section-8.6


raphaelrobert (changes requested):
Thanks for picking it up! LGTM in general, I think it would be good to have a negative test to show that 

```
export(label1, context1, key_length1) != export(label1, context2, key_length1)
```

kkohbrok:
Thanks for the review! Added the requested tests on the level of the managed group.

franziskuskiefer:
@kkohbrok what's the state of this? Does this need review?

kkohbrok:
This arose from interop testing and it's one of the few things I put in a PR to get into main, as it was an easy change. But after merging #370, all that remains is the test that @raphaelrobert asked for. I don't think that's something that's tested yet, but it's also relatively trivial. Feel free to decide if we should merge or close.

franziskuskiefer (approved):
I'm always for more tests. Let's get it in.

kkohbrok:
@raphaelrobert, I think you need to bless this PR first, as you requested changes 🙏🏻

PR #361, opened 2021-03-04
Return events directly in ManagedGroup

raphaelrobert:
After individually running in the same difficulties with @kkohbrok when trying to use event callbacks, we decided to try returning the events directly from `process_messages()` on `ManagedGroup`.

This R does that and also extends the test coverage for events (which now finally possible). In that context one missing event was fixed (Commits that contain an `UpdatePath` now return an update event in addition to the proposals they cover).

If consumers require callbacks for events in the future, they can still implement them. Or alternatively we can provide an event streamer.


src/group/managed_group/mod.rs
```diff
@@ -929,85 +950,67 @@ impl<'a> ManagedGroup<'a> {
...
+                    update_proposal.key_package.credential().clone(),
+                ))
             }
             // Remove proposals
             Proposal::Remove(remove_proposal) => {
                 let removal = Removal::new(
-                    self.credential_bundle.credential(),
-                    sender_credential,
-                    &indexed_members[&LeafIndex::from(remove_proposal.removed)],
+                    self.credential_bundle.credential().clone(),
```
kkohbrok, line 994: There are some additional clones here and a bit below. I assume this was previously done in the callbacks?

kkohbrok (commented):
Thanks for taking care of these changes. This should make it a bit more straight-forward to use the the ManagedGroup API in the ManagedClient. One small question, but otherwise it looks great!

src/group/managed_group/mod.rs
raphaelrobert: As mentioned out-of-band: We have to clone all the data since we can no longer use references to data that is stored inside the group. Now that we return the events directly, we have to assume that they are long-lived and will outlive the ephemeral group state. While this creates some data duplication, I don't see any way around it conceptually.

kkohbrok (approved):

src/group/managed_group/mod.rs
```diff
@@ -530,7 +530,15 @@ impl<'a> ManagedGroup<'a> {
...
-                                panic!("apply_commit_error did not return an ApplyCommitError.");
+                                let error_string =
+                                    "apply_commit() did not return an ApplyCommitError."
+                                        .to_string();
+                                events.push(GroupEvent::Error(ErrorEvent::new(
+                                    ManagedGroupError::LibraryError(ErrorString::from(
+                                        error_string,
+                                    )),
+                                )));
+                                panic!();
```
kkohbrok, line 541: The commit message says "avoid a panic", but this is still panicking. Is that intentional?

src/group/managed_group/mod.rs
```diff
@@ -384,11 +386,17 @@ impl<'a> ManagedGroup<'a> {
     // === Process messages ===
 
     /// Processes any incoming messages from the DS (MLSPlaintext &
-    /// MLSCiphertext) and triggers the corresponding callback functions
-    pub fn process_messages(&mut self, messages: Vec<MLSMessage>) -> Result<(), ManagedGroupError> {
+    /// MLSCiphertext) and triggers the corresponding callback functions.
+    /// Return a list of `GroupEvent` that contain the individual events that
+    /// occured whil processing messages.
```
franziskuskiefer, line 391:
```suggestion
    /// occurred while processing messages.
```

src/group/managed_group/mod.rs
```diff
@@ -384,11 +386,17 @@ impl<'a> ManagedGroup<'a> {
...
+    pub fn process_messages(
+        &mut self,
+        messages: Vec<MLSMessage>,
+    ) -> Result<Vec<GroupEvent>, ManagedGroupError> {
```
franziskuskiefer, line 395: If managed group functions always return something like this, we should have a return type for it, i.e. a `Result` type in here with this signature to force consistency and make the API easier to understand.

src/group/managed_group/events.rs
```diff
@@ -0,0 +1,268 @@
...
+    pub fn aad(&self) -> &[u8] {
+        &self.aad
+    }
+
+    /// Get a reference to the member added event's sender.
+    pub fn sender(&self) -> &Credential {
+        &self.sender
+    }
+
+    /// Get a reference to the member added event's added member.
```
franziskuskiefer, line 48: Not sure I understand these comments ... 😄 Maybe simplify the sentences for all these comments.

src/group/managed_group/events.rs
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// Helper structs
+
+/// This enum lists the 4 different variants of a removal, depending on who the
+/// remover and who the leaver is.
+#[derive(Debug, PartialEq, Clone)]
+pub enum Removal {
+    ///  We previously issued a RemoveProposal for ourselves and this was now
+    /// commited by someone else
```
franziskuskiefer, line 239:
```suggestion
    /// committed by someone else
```

src/group/managed_group/events.rs
```diff
@@ -0,0 +1,268 @@
...
+    WeLeft,
+    /// Another member issued a RemoveProposal for itself that was now committed
+    TheyLeft(Credential),
+    /// Another member issued a RemoveProposal for ourselves that was now
+    /// committed
+    WeWereRemovedBy(Credential),
+    /// Member A issued a RemoveProposal for member B that was now commited
```
franziskuskiefer, line 246:
```suggestion
    /// Member A issued a RemoveProposal for member B that was now committed
```

src/group/managed_group/events.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// Get a reference to the re init event's psk id.
+    pub fn psk_id(&self) -> &ReInitProposal {
+        &self.re_init_proposal
+    }
+}
+
+// Application messages
+
+/// Event that occurs when an application message is received.
+/// `sender` contains the mesagge's sender and `message` contains
```
franziskuskiefer, line 160:
```suggestion
/// `sender` contains the message's sender and `message` contains
```

franziskuskiefer (commented):
Generally lgtm. A couple questions.

So you want to replace all the callbacks with returning events instead? May I ask what the issues where that you had with the callbacks?
I think I generally prefer callbacks but I'm not opposed to returning events instead.

src/group/managed_group/mod.rs
raphaelrobert: This is the only function that returns group events. I can add it if you insist, but the signature looked pretty clean to me.

src/group/managed_group/mod.rs
franziskuskiefer: With this panic it doesn't make sense to have a new event. So I guess this should go...

franziskuskiefer (changes requested):

franziskuskiefer (approved):
thanks!

PR #362, opened 2021-03-05
Generate complete tree math test vectors

bifurcation:
This just extends the tree math KAT so that it generates complete outputs.  Right now, it only generates results for half of the tree.

NB: I was not able to build OpenMLS, so I don't know if this patch actually builds.


bifurcation:
@kkohbrok - It looks like the known-answer test JSON file needs to be regenerated.  Is there an easy way to do that?

kkohbrok (approved):
LGTM! Thanks for making this work!

I've already integrated your changes into the interop branch. Please wait for @franziskuskiefer to take a look before merging this to main.

PR #367, opened 2021-03-11
Extract interop-week progress into separate PRs

kkohbrok:
This is a draft PR to review quick-and-dirty changes made during interop week for inclusion into OpenMLS `main`. This is not intended to ever become a full PR, so feel free to disregard.


src/schedule/psk.rs
```diff
@@ -262,7 +262,8 @@ impl PskSecret {
         }
         let mut secret = vec![];
         for (index, psk) in psks.iter().enumerate() {
-            let psk_input = ciphersuite.hkdf_extract(None, psk);
+            let zero_secret = Secret::from(zero(ciphersuite.hash_length()));
+            let psk_input = ciphersuite.hkdf_extract(&zero_secret, Some(psk));
```
franziskuskiefer, line 266: Probably both secrets in `hkdf_extract` should be `Option` (or none).

franziskuskiefer (commented):
The changes to the library appear to be relatively limited. I suggest to pull out the changes to the library code to get them landed and continue using this branch for interop testing and hot fixes. This allows us to move quickly without loosing track of the important changes. We can then later get the managed client and interop client in.

PR #370, opened 2021-03-19
Interop fixes 1/3

franziskuskiefer:
This is extracted from #367 and contains all the changes to the library.

* fix order of inputs in extraction operations
* fix default value for secrets in general and CommitSecret in particular
* switch which input of extract is an Option type
* fix derivation of Welcome secret
* Fix MLSPlaintext und MLSCiphertext encoding/decoding
* Fix node index in tree context

This PR also adds quite some logging and contains some residual changes from the other PRs that might not make sense on their own.


franziskuskiefer:
> I think the way I exposed the test vectors is not optimal to say the least. Also, clippy has quite a few complaints, but again, most of them related to the test vector exposition.

yeah, the warnings should go away with #371 merged in. (This doesn't even pass tests right now.) The test vectors here are actually not `pub`. This must still happen on the interop branch and we should clean it up a little. This PR only puts the foundation in to do this.

> Other than that, I think the other changes are ok. @franziskuskiefer I believe you suggested removing the `Option` entirely on the `extract` inputs. I have no strong opinion there, but that's certainly something we can do.

The API could be a little indeed. But not sure if it's worth the additional churn right now.

raphaelrobert (approved):
LGTM!

kkohbrok (approved):

PR #371, opened 2021-03-19
Interop fixes 2/3

franziskuskiefer:
This PR contains changes to the test vector code that is needed for the interop client.
While the interop client is not landing just yet there are changes in here that we want.

This PR only changes test code.


kkohbrok (approved):
Thanks for preparing this for merge into `main`. Looks good to me!

PR #372, opened 2021-03-19
update test vectors

franziskuskiefer:


kkohbrok (approved):
I didn't look at the contents of the `.json` files to be honest, but it's probably generally a good idea to update the test vector files 😄

PR #376, opened 2021-03-24
test macro and codecov fix

franziskuskiefer:
This includes the Code from #328 and changes the test proc macro. 
This still requires nightly (when merging I’ll file a follow up to use stable again once it’s released). The linking issue came from a cyclic dependency between the test macro and the main lib. While this is fine for the compiler tarpaulin appears to have issues with it. And since it’s not great in general I rewrote it a little. It’s not as seem less now but therefore more flexible and it fixes the codecov issues.

### DIFF
--- a/tests/utils/mod.rs
+++ b/tests/utils/mod.rs
@@ -2,3 +2,21 @@
 pub mod managed_utils;
 /// A framework to create integration tests of the "raw" mls_group API.
 pub mod mls_utils;
+
+#[allow(unused_macros)]
+macro_rules! ctest_ciphersuites {
+    ($name:ident, test($param_name:ident: $t:ty) $body:block) => {
+        test_macros::ctest!(
+            $name
+            [
+                CiphersuiteName::MLS10_128_DHKEMX25519_AES128GCM_SHA256_Ed25519,
+                CiphersuiteName::MLS10_128_DHKEMP256_AES128GCM_SHA256_P256,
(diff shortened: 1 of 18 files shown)


tests/utils/mod.rs
```diff
@@ -2,3 +2,21 @@
 pub mod managed_utils;
 /// A framework to create integration tests of the "raw" mls_group API.
 pub mod mls_utils;
+
+#[allow(unused_macros)]
+macro_rules! ctest_ciphersuites {
```
kkohbrok, line 7: Just a minor point: Can't we just expose this code under some flag as part of `openmls` instead of duplicating it in `/tests` and `/src/test_util`?

kkohbrok (approved):
Thanks for picking this up. But ... wow, I somehow really didn't miss all those codedov complaints.

I realize that it's necessary, but we might want to think about how to reduce the noise. For example, there seem to be quite a few complaints related to test vector generation.

tests/utils/mod.rs
franziskuskiefer: There's no easy way unfortunately. The integration tests are running against a regular build of the library. So the only "easy" way I see is to require a feature flag to be present for testing, which isn't great.

> I realize that it's necessary, but we might want to think about how to reduce the noise. For example, there seem to be quite a few complaints related to test vector generation.

Yeah, this is getting us a baseline again now that we can look at and see if we want to disable something.

PR #378, opened 2021-03-24
ciphersuite secret cleanup

franziskuskiefer:
After looking at dependencies between the modules I started cleaning up the ciphersuite and key schedule. There are a couple more things that could be done here but this would require having something like a `trait Secret`.
Similar cleanups are needed for the other modules as well.


kkohbrok (approved):
❤️ 

After interop is done and possible from `main`, we should really take some time for more refactoring, cleanup and doc PRs.

PR #381, opened 2021-03-26
Fix parent hash verification after to spec change

### LINKED ISSUE #380: [MLS Spec change] Parent hash verification algorithm fix (#460)
A word was missing from the parent has verification algorithm.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/460

kkohbrok:
This PR fixes #380.

### DIFF
--- a/src/tree/hashes.rs
+++ b/src/tree/hashes.rs
@@ -237,8 +237,8 @@ impl RatchetTree {
         }
         let right = child;
 
-        // "If R is a leaf node, the check fails"
-        if right.is_leaf() {
+        // "If R is a blank leaf node, the check fails"
+        if self.nodes[child].is_blank() && right.is_leaf() {
             return Err(ParentHashError::EndedWithLeafNode);
         }
 


franziskuskiefer (commented):
We should only land this with a test. A unit test would be fine or actually enabling parent hash verification, then this will be tested as well.

src/tree/hashes.rs
```diff
@@ -237,8 +237,8 @@ impl RatchetTree {
         }
         let right = child;
 
-        // "If R is a leaf node, the check fails"
-        if right.is_leaf() {
+        // "If R is a blank leaf node, the check fails"
+        if self.nodes[child].is_blank() && right.is_leaf() {
```
raphaelrobert, line 241: Not that it makes a difference for now, but I'd suggest to use `right` as an index for `self.nodes` for clarity.

raphaelrobert (changes requested):
Thanks for fixing it!

I guess this PR reflects the spec change but we should keep an eye on this until we have a real proof of the algorithm

src/tree/hashes.rs
kkohbrok: I'll fix it when I create tests for this issue. What do you mean by "real proof of the algorithm"? I fixed this, because I ran into cases that wouldn't work without this fix.

kkohbrok:
I'm closing this in favor of #390, which includes this fix.

PR #382, opened 2021-04-01
Make Secret protocol version and cipher suite aware

### LINKED ISSUE #231: Make version number variable in key deriavation label
The `"mls10 "` part of the `KdfLabel` depends on the MLS version and can't be hardcoded.

_Originally posted by @franziskuskiefer in https://github.com/openmls/openmls/pull/225#discussion_r529557576_

This mostly involves pushing `ProtocolVersion` down to places where we derive keys.

franziskuskiefer:
The main purpose of this patch is to make the labeled key derivation
dependent on the cipher suite and MLS version (#231).
This got a little bigger than expected ...

This patch stores the the cipher suite and MLS version of a secret
as part of the struct and enforces that it is only used with compatible
other secrets. It also removes the need to pass the ciphersuite to
many functions.
At the same time this increases the complexity of creating and parsing
secrets and other structs that don't have the necessary information.

This patch also introduces constant time comparison for secrets, in
particular for Mac values, which are now secrets as well.

Fixes #231

### DIFF
--- a/src/ciphersuite/mod.rs
+++ b/src/ciphersuite/mod.rs
@@ -199,6 +199,16 @@ impl KdfLabel {
     }
 }
 
+/// Compare two byte slices in a way that's hopefully not optimised out by the
+/// compiler.
+fn equal_ct(a: &[u8], b: &[u8]) -> bool {
+    let mut diff = 0u8;
+    for (l, r) in a.iter().zip(b.iter()) {
+        diff |= l ^ r;
+    }
+    diff == 0
+}
+
 /// A struct to contain secrets. This is to provide better visibility into where
 /// and how secrets are used and to avoid passing secrets in their raw
 /// representation.
@@ -245,11 +255,7 @@ impl PartialEq for Secret {
             );
             return false;
         }
-        let mut diff = 0u8;
-        for (l, r) in self.value.iter().zip(other.value.iter()) {
-            diff |= l ^ r;
-        }
-        diff == 0
+        equal_ct(&self.value, &other.value)
     }
 }
 
@@ -293,14 +299,6 @@ impl Secret {
             mls_version,
             ciphersuite,
         }
-    }
-
-    /// HMAC-Hash(salt, IKM). For all supported ciphersuites this is the same
-    /// HMAC that is also used in HKDF.
-    /// Compute the HMAC on `self` with key `ikm`.
-    pub(crate) fn mac(&self, ikm: &Secret) -> Self {
-        // FIXME: this should be on ikm and the other input shouldn't be a secret.
-        self.hkdf_extract(ikm)
     }
 
     /// HKDF extract where `self` is `salt`.
@@ -393,13 +391,8 @@ impl Secret {
     }
 
     /// Returns the inner bytes of a secret
-    pub(crate) fn to_bytes(&self) -> &[u8] {
+    pub(crate) fn as_slice(&self) -> &[u8] {
         &self.value
-    }
-
-    /// Returns the MLS version of the secret
-    pub(crate) fn version(&self) -> ProtocolVersion {
-        self.mls_version
     }
 
     /// Returns the ciphersuite of the secret
@@ -416,6 +409,36 @@ impl From<&[u8]> for Secret {
             value: bytes.to_vec(),
             mls_version: ProtocolVersion::default(),
             ciphersuite: Ciphersuite::default(),
+        }
+    }
+}
+
+/// 9.2 Message framing
+///
+/// struct {
+///     opaque mac_value<0..255>;
+/// } MAC;
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct Mac {
+    pub(crate) mac_value: Vec<u8>,
+}
+
+impl PartialEq for Mac {
+    // Constant time comparison.
+    fn eq(&self, other: &Mac) -> bool {
+        equal_ct(&self.mac_value, &other.mac_value)
+    }
+}
+
+impl Mac {
+    /// HMAC-Hash(salt, IKM). For all supported ciphersuites this is the same
+    /// HMAC that is also used in HKDF.
+    /// Compute the HMAC on `salt` with key `ikm`.
+    pub(crate) fn new(salt: &Secret, ikm: &[u8]) -> Self {
+        Mac {
+            mac_value: salt
+                .hkdf_extract(&Secret::from_slice(ikm, salt.mls_version, salt.ciphersuite))
+                .value,
         }
     }
 }
--- a/src/ciphersuite/mod.rs
+++ b/src/ciphersuite/mod.rs
@@ -266,9 +266,6 @@ impl Secret {
             ciphersuite.name,
             mls_version
         );
-        // if ciphersuite.name == CiphersuiteName::MLS10_128_DHKEMP256_AES128GCM_SHA256_P256 {
-        //     log::trace!("{:?}", backtrace::Backtrace::new());
-        // }
         Secret {
             value: get_random_vec(ciphersuite.hash_length()),
             mls_version,
--- a/src/ciphersuite/mod.rs
+++ b/src/ciphersuite/mod.rs
@@ -26,7 +26,7 @@ mod ser;
 
 use crate::{
     codec::*,
-    config::{Config, ConfigError},
+    config::{Config, ConfigError, ProtocolVersion},
 };
 
 use ciphersuites::*;
@@ -34,6 +34,8 @@ pub(crate) use errors::*;
 
 #[cfg(test)]
 mod test_ciphersuite;
+#[cfg(test)]
+mod test_secrets;
 
 pub(crate) const NONCE_BYTES: usize = 12;
 pub(crate) const REUSE_GUARD_BYTES: usize = 4;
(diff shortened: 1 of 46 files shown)


src/config/mod.rs
```diff
@@ -166,6 +166,7 @@ impl Config {
 pub enum ProtocolVersion {
     Reserved = 0,
     Mls10 = 1,
+    Mls10Draft12 = 255, // pre RFC version
```
raphaelrobert, line 169: We are on draft 11 right now, just in case. I think we need room for ~10 pre-RFC draft versions, so maybe start at 200?

src/group/mls_group/test_mls_group.rs
```diff
@@ -60,7 +61,9 @@ fn test_failed_groupinfo_decryption() {
             let tree_hash = vec![1, 2, 3, 4, 5, 6, 7, 8, 9];
             let confirmed_transcript_hash = vec![1, 1, 1];
             let extensions = Vec::new();
-            let confirmation_tag = ConfirmationTag::from(vec![6, 6, 6]);
+            let confirmation_tag = ConfirmationTag(Mac {
+                mac_value: Secret::random(ciphersuite, None),
+            });
```
raphaelrobert, line 66: 🤘➡️🗑️

src/schedule/mod.rs
```diff
@@ -624,16 +644,24 @@ impl ConfirmationKey {
...
+                confirmed_transcript_hash,
+                self.secret.version(),
+                ciphersuite,
+            )).into()
+            // ciphersuite
+            //     .mac(
+            //         &self.secret,
+            //         &Secret::from(confirmed_transcript_hash.to_vec()),
+            //     )
+            //     .into(),
```
raphaelrobert, line 664: comments

raphaelrobert (changes requested):
Thanks for tackling that! I did a quick pass and flagged the stuff that stood out.

src/ciphersuite/mod.rs
```diff
@@ -201,84 +202,223 @@ impl KdfLabel {
...
+            value: bytes.to_vec(),
+            mls_version,
+            ciphersuite,
+        }
+    }
+
+    /// HMAC-Hash(salt, IKM). For all supported ciphersuites this is the same
+    /// HMAC that is also used in HKDF.
+    /// Compute the HMAC on `self` with key `ikm`.
+    pub(crate) fn mac(&self, ikm: &Secret) -> Self {
```
kkohbrok, line 304: Do we really need any secret to be able to compute `mac`s? I think we have distinct types for keys that we expect to compute macs with.

src/ciphersuite/mod.rs
```diff
@@ -201,84 +202,223 @@ impl KdfLabel {
...
+        // FIXME: this should be on ikm and the other input shouldn't be a secret.
+        self.hkdf_extract(ikm)
```
kkohbrok, line 306: I believe that this is how it was before, but isn't it generally a bit strange that `hkdf_extract` and not `hmac`?

src/ciphersuite/mod.rs
```diff
@@ -201,84 +202,223 @@ impl KdfLabel {
...
             self.value,
             label,
-            ciphersuite.name()
+            self.ciphersuite.name()
         );
-        self.kdf_expand_label(ciphersuite, label, &[], ciphersuite.hash_length())
+        self.kdf_expand_label(label, &[], self.ciphersuite.hash_length())
+    }
+
+    /// Update the ciphersuite and MLS version of this secret.
```
kkohbrok, line 386: I think this is a valuable observation that would be nice to have in the right place. Is there an existing issue to document this in? Alternatively, create a new one?

kkohbrok (commented):
Just a partial review for now. I'll continue after the weekend.

src/ciphersuite/mod.rs
```diff
@@ -201,84 +202,223 @@ impl KdfLabel {
...
+                other.mls_version,
+                other.value.len()
+            );
+            return false;
+        }
+        let mut diff = 0u8;
+        for (l, r) in self.value.iter().zip(other.value.iter()) {
+            diff |= l ^ r;
+        }
+        diff == 0
```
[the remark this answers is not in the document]
franziskuskiefer: They are very slim. This is the standard way of doing this. Happy to adjust if you show me there's an issue ;)

src/ciphersuite/mod.rs
franziskuskiefer: Macs are only wrappers right now and don't live here. The entire Mac handling needs a refactor. But that's not the goal of this PR.

src/ciphersuite/mod.rs
```diff
@@ -201,84 +202,223 @@ impl KdfLabel {
...
+            mls_version,
+            ciphersuite,
+        }
+    }
+
+    /// HMAC-Hash(salt, IKM). For all supported ciphersuites this is the same
+    /// HMAC that is also used in HKDF.
+    /// Compute the HMAC on `self` with key `ikm`.
+    pub(crate) fn mac(&self, ikm: &Secret) -> Self {
+        // FIXME: this should be on ikm and the other input shouldn't be a secret.
```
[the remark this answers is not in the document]
franziskuskiefer: see other comment above. This all needs refactoring.

src/ciphersuite/mod.rs
franziskuskiefer: It's the same thing. This code was only moved around.

src/ciphersuite/mod.rs
franziskuskiefer: See #383

src/config/mod.rs
franziskuskiefer: Yeah I figured by the time we need this we are on 12 ;)
Using an enum for versions is just not a great idea... But let's start at 200.

src/schedule/mod.rs
```diff
@@ -487,28 +490,41 @@ impl EncryptionSecret {
...
 
+// #[cfg(any(feature = "expose-test-vectors", test))]
+// #[doc(hidden)]
+// impl From<&[u8]> for EncryptionSecret {
+//     fn from(bytes: &[u8]) -> Self {
+//         Self {
+//             secret: Secret::from(bytes),
+//         }
+//     }
+// }
```
kkohbrok, line 526: I believe this can be removed.

kkohbrok (changes requested):
Thanks for picking this up (and also thanks for adding a ton more logging everywhere). I think having these guarantees is worth the effort of adding the extra parameters everywhere.

That being said, I'm a bit concerned by the fact that we now have to `config` various structs after decoding. You already introduced a `decode_with_context` function for `MLSPlaintext`. Wouldn't it be worth introducing that for most of the other "configurable" structs as well? Maybe even have a separate `Codec` trait for them?

Other than that, just a few, minor, comments.

raphaelrobert (commented):
LGTM! I'll rely on @kkohbrok's expertise regarding the crypto details and to do the final approval.

src/framing/codec.rs
```diff
@@ -40,6 +42,28 @@ impl Codec for MLSPlaintext {
     }
 }
 
+impl MLSPlaintext {
+    /// Decode an `MLSPlaintext` with ciphersuite and protocol version information.
+    /// This should be used instead of the raw decoding function in order to
+    /// update the `MLSPlaintext` with the missing information.
+    pub fn decode_with_context(
```
franziskuskiefer, line 49: Yeah, this is a spec issue imo. MlsPlaintext isn't well defined if you want to be as defensive as we are here now. But this is not for the library to decide here but for consumers. If there's a consumer that needs a different API we can think about how to make it better. Right now it doesn't look like there is one.

src/framing/codec.rs
```diff
@@ -118,13 +142,22 @@ impl MLSPlaintextContentType {
...
     fn encode(&self, buffer: &mut Vec<u8>) -> Result<(), CodecError> {
-        encode_vec(VecSize::VecU8, buffer, &self.mac_value)?;
+        encode_vec(VecSize::VecU8, buffer, self.mac_value.to_bytes())?;
         Ok(())
     }
 
     fn decode(cursor: &mut Cursor) -> Result<Self, CodecError> {
         let mac_value = decode_vec(VecSize::VecU8, cursor)?;
-        Ok(Self { mac_value })
+        // The secret is instantiated with default values here because we don't
```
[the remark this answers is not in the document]
franziskuskiefer: We need this unless we want to remove the `MlsPlaintext::decode` entirely. We could add a `_with_context` in addition to use in the corresponding decoding functions. I don't think it's worth it (see my general comments), but if you prefer I can add it.

src/framing/plaintext.rs
```diff
@@ -342,9 +347,9 @@ impl MLSPlaintextContentType {
 /// struct {
 ///     opaque mac_value<0..255>;
 /// } MAC;
-#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub(crate) struct Mac {
-    pub(crate) mac_value: Vec<u8>,
+    pub(crate) mac_value: Secret,
```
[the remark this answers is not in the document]
franziskuskiefer: There are two separat reasons. This can be solved differently as well, but this appeared to be the easiest right now.

The `Mac` should have the same ciphersuite/version information attached as a secret. Before the value was always converted to a `Secret` before handling it, that's difficult when this information is missing. The alternative would be to store the necessary information in the `Mac`.

Checking `Mac` validity is a timing sensitive issue in general (it was the source of some TLS attacks for example). So we want to have "constant time" `Mac` comparison. And that's something that's better implemented on the generic secret. The alternative would be to have a generic "constant time" comparison function and use it in both places.

franziskuskiefer (commented):
> That being said, I'm a bit concerned by the fact that we now have to config various structs after decoding. You already introduced a decode_with_context function for MLSPlaintext. Wouldn't it be worth introducing that for most of the other "configurable" structs as well? Maybe even have a separate Codec trait for them?

Right, there are two different ways to handle this. Either decode as we do right now and then update the structs with the additional information; or pass in a context when decoding.
I think both is pretty ugly not really what should happen. The structs should have the necessary information encoded in them. But that's a spec issue.
I went for the minimal effort approach here with both examples of handling this. Ideally we don't need either in the end. But I don't want to optimise the code for a spec issue just yet (while there's a chance the spec still changes).

kkohbrok (approved):
Thanks for implementing the changes. Let's get it in!

Issue #375, opened 2021-03-22
Change clippy CI


franziskuskiefer:
Clippy on CI currently uses `all-features`, which leads to a lot of unused code warnings due to the current interop test setup.
Only enable features here we actually want enabled for clippy.

https://github.com/openmls/openmls/blob/6ef531b38b7a03452c6e78c890f1187ef63abae5/.github/workflows/tests.yml#L46

raphaelrobert:
While clippy still doesn't work, this is no longer a problem.

PR #384, opened 2021-04-06
Work on clippy warnings

raphaelrobert:
This PR tries to get rid of the clippy warnings that were introduced in past PRs.

No other code changes were made, except for the ones strictly necessary to get rid of the warnings.

### DIFF
--- a/cli/src/user.rs
+++ b/cli/src/user.rs
@@ -25,7 +25,7 @@ pub struct Group {
     members: Vec<Vec<u8>>,
     conversation: Conversation,
     mls_group: RefCell<MlsGroup>,
-    pending_proposals: Vec<MLSPlaintext>,
+    pending_proposals: Vec<MlsPlaintext>,
 }
 
 pub struct User {
@@ -97,7 +97,7 @@ impl User {
 
         // Send mls_ciphertext to the group
         let msg = GroupMessage::new(
-            MLSMessage::Ciphertext(mls_ciphertext),
+            MlsMessage::Ciphertext(mls_ciphertext),
             &self.recipients(group),
         );
         log::debug!(" >>> send: {:?}", msg);
@@ -122,9 +122,9 @@ impl User {
                     // approve first ...)
                     self.join_group(welcome)?;
                 }
-                Message::MLSMessage(message) => {
+                Message::MlsMessage(message) => {
                     let msg = match message {
-                        MLSMessage::Ciphertext(ctxt) => {
+                        MlsMessage::Ciphertext(ctxt) => {
                             let mut group = match self.groups.get(&ctxt.group_id.as_slice()) {
                                 Some(g) => g.mls_group.borrow_mut(),
                                 None => {
@@ -146,7 +146,7 @@ impl User {
                                 }
                             }
                         }
-                        MLSMessage::Plaintext(msg) => msg,
+                        MlsMessage::Plaintext(msg) => msg,
                     };
                     let group = match self.groups.get_mut(&msg.group_id().as_slice()) {
                         Some(g) => g,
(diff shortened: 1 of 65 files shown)


kkohbrok (changes requested):
Thanks for taking care of this. The clippy warnings were getting really annoying. 

I was wondering if we wanted to stick with the all-capitalized MLS-prefix just to conform with the struct names of the spec, but I'm fine with either approach.

Do we really need to allow dead code everywhere, though? Can't we just get away with running clippy with `--all-targets` or something similar?

franziskuskiefer:
> I was wondering if we wanted to stick with the all-capitalized MLS-prefix just to conform with the struct names of the spec, but I'm fine with either approach.

This is a change I wanted to do as well. We should stick to Rust naming conventions, not the spec here.

> Do we really need to allow dead code everywhere, though? Can't we just get away with running clippy with `--all-targets` or something similar?

We shouldn't allow dead code. This is what #375 is about. We should instead remove the `--all-features` on CI and replace it with the features we want enabled.

kkohbrok:
> This is a change I wanted to do as well. We should stick to Rust naming conventions, not the spec here.

Is there a particular reason why you prefer the Rust naming convention over the spec? Again, I don't have a strong opinion either way, but it doesn't seem that obvious to me.

franziskuskiefer:
> Is there a particular reason why you prefer the Rust naming convention over the spec? Again, I don't have a strong opinion either way, but it doesn't seem that obvious to me.

RFCs are "rough consensus and running code". So what really counts is the code. The code is the living document that has to be maintained and improved over time. The RFC is "only" documentation for it. So if it comes down to what an implementation should optimise for, it's always the code.

kkohbrok:
> > Is there a particular reason why you prefer the Rust naming convention over the spec? Again, I don't have a strong opinion either way, but it doesn't seem that obvious to me.
> 
> RFCs are "rough consensus and running code". So what really counts is the code. The code is the living document that has to be maintained and improved over time. The RFC is "only" documentation for it. So if it comes down to what an implementation should optimise for, it's always the code.

I was under the impression that the RFC represents a specification and that the code implements that specification. Although I agree that in practice it might sometimes be the way that you describe. If the RFC is documentation for code, then which code is it documenting? Typically there are multiple implementations, but only one RFC. But maybe let's discuss this offline.

raphaelrobert:
Regarding the naming convention: We had a mixed bag already (`MlsGroup`). I really think we should stick to the Rust convention, even if I personally don't find it very beautiful.

Regarding the dead code: Ultimately we should get rid of it, but I haven't found how just yet. The issue the the newly introduced feature flag. By default clippy doesn't look at feature flags, neither in the CI nor in vscode. If you have any pointers how to instruct clippy, I'd be happy.

franziskuskiefer:
> Regarding the dead code: Ultimately we should get rid of it, but I haven't found how just yet. The issue the the newly introduced feature flag. By default clippy doesn't look at feature flags, neither in the CI nor in vscode. If you have any pointers how to instruct clippy, I'd be happy.

As I said, clippy is being run with all features enable. We just shouldn't do that with the features we have right now.
https://github.com/openmls/openmls/blob/a454a5a649f5663ce299c29dedf54938240e68ef/.github/workflows/tests.yml#L42-L46

raphaelrobert:
> As I said, clippy is being run with all features enable. We just shouldn't do that with the features we have right now.
> 
> https://github.com/openmls/openmls/blob/a454a5a649f5663ce299c29dedf54938240e68ef/.github/workflows/tests.yml#L42-L46

Removing that flag disables clippy on all features, which is not what I want to achieve. This might be ok for the CI for now, but not for editors.
The real problem is why clippy thinks the code is dead, even though it's clearly used in on the `expose-test-vectors` feature.

raphaelrobert:
I fixed the dead code part and did some renaming inside of comments (sorry for blowing up the size of the PR because of that).

kkohbrok (approved):

PR #385, opened 2021-04-07
Change the Managed Group to accept a KeyStore ref instead of keeping a CredentialBundle ref

kkohbrok:
This PR replaces #334, as it goes a step further and doesn't keep the KeyStore reference in the ManagedGroup struct, but instead passes a KeyStore reference into every call into the Managed Group API. This allows us to avoid lifetime issues, as Rust doesn't allow struct-internal references.

One caveat in this PR that might be worth discussing:
* We currently ignore (i.e. `unwrap`) PoisonErrors when accessing KeyPackageBundles and CredentialBundles. The argument being that there is (to my knowledge) no way that the code can panic while it holds a write lock. In particular while holding a write lock, we only interact with the HashMap using `get`, `insert` or `remove`, which should be safe.

The PR is rather large, but should be easy to review, as a lot of the changes are due to the change of API, so function calls will look different, but in most places, there are no changes in terms of functionality.

### DIFF
--- a/src/group/managed_group/mod.rs
+++ b/src/group/managed_group/mod.rs
@@ -757,7 +757,9 @@ impl ManagedGroup {
     }
 
     /// Returns own credential. If the group is inactive, it returns a
-    /// `UseAfterEviction` error.
+    /// `UseAfterEviction` error. This function currently returns a full
+    /// `Credential` rather than just a reference. This issue is tracked in
+    /// issue #387.
     pub fn credential(&self) -> Result<Credential, ManagedGroupError> {
         if !self.is_active() {
             return Err(ManagedGroupError::UseAfterEviction(UseAfterEviction::Error));
--- a/src/group/managed_group/mod.rs
+++ b/src/group/managed_group/mod.rs
@@ -8,11 +8,12 @@ mod ser;
 mod test_managed_group;
 
 use crate::{
-    credentials::{Credential, CredentialBundle},
+    credentials::Credential,
     error::ErrorString,
     framing::*,
     group::*,
     key_packages::{KeyPackage, KeyPackageBundle},
+    key_store::KeyStore,
     messages::{proposals::*, Welcome},
     schedule::ResumptionSecret,
     tree::{index::LeafIndex, node::Node},
@@ -60,10 +61,7 @@ use ser::*;
 /// Changes to the group state are dispatched as events through callback
 /// functions (see [`ManagedGroupCallbacks`]).
 #[derive(Debug)]
-#[cfg_attr(test, derive(PartialEq))]
-pub struct ManagedGroup<'a> {
-    // CredentialBundle used to sign messages
-    credential_bundle: &'a CredentialBundle,
+pub struct ManagedGroup {
     // The group configuration. See `ManagedGroupCongig` for more information.
     managed_group_config: ManagedGroupConfig,
     // the internal `MlsGroup` used for lower level operations. See `MlsGroup` for more
@@ -86,17 +84,23 @@ pub struct ManagedGroup<'a> {
     active: bool,
 }
 
-impl<'a> ManagedGroup<'a> {
+impl ManagedGroup {
     // === Group creation ===
 
-    /// Creates a new group from scratch with only the creator as a member.
+    /// Creates a new group from scratch with only the creator as a member. This
+    /// function removes the `KeyPackageBundle` corresponding to the
+    /// `key_package_hash` from the `key_store`. Throws an error if no
+    /// `KeyPackageBundle` can be found.
     pub fn new(
-        credential_bundle: &'a CredentialBundle,
+        key_store: &KeyStore,
         managed_group_config: &ManagedGroupConfig,
         group_id: GroupId,
-        key_package_bundle: KeyPackageBundle,
+        key_package_hash: &[u8],
     ) -> Result<Self, ManagedGroupError> {
         // TODO #141
+        let key_package_bundle = key_store
+            .take_key_package_bundle(key_package_hash)
+            .ok_or(ManagedGroupError::NoMatchingKeyPackageBundle)?;
         let group = MlsGroup::new(
             &group_id.as_slice(),
             key_package_bundle.key_package().ciphersuite_name(),
@@ -109,7 +113,6 @@ impl<'a> ManagedGroup<'a> {
             ResumptionSecretStore::new(managed_group_config.number_of_resumption_secrets);
 
         let managed_group = ManagedGroup {
-            credential_bundle,
             managed_group_config: managed_group_config.clone(),
             group,
             pending_proposals: vec![],
@@ -127,20 +130,22 @@ impl<'a> ManagedGroup<'a> {
 
     /// Creates a new group from a `Welcome` message
     pub fn new_from_welcome(
-        credential_bundle: &'a CredentialBundle,
+        key_store: &KeyStore,
         managed_group_config: &ManagedGroupConfig,
         welcome: Welcome,
         ratchet_tree: Option<Vec<Option<Node>>>,
-        key_package_bundle: KeyPackageBundle,
-    ) -> Result<Self, GroupError> {
+    ) -> Result<Self, ManagedGroupError> {
+        let resumption_secret_store =
+            ResumptionSecretStore::new(managed_group_config.number_of_resumption_secrets);
+        let key_package_bundle = welcome
+            .secrets()
+            .iter()
+            .find_map(|egs| key_store.take_key_package_bundle(&egs.key_package_hash))
+            .ok_or(ManagedGroupError::NoMatchingKeyPackageBundle)?;
         // TODO #141
         let group = MlsGroup::new_from_welcome(welcome, ratchet_tree, key_package_bundle, None)?;
 
-        let resumption_secret_store =
-            ResumptionSecretStore::new(managed_group_config.number_of_resumption_secrets);
-
         let managed_group = ManagedGroup {
-            credential_bundle,
             managed_group_config: managed_group_config.clone(),
             group,
             pending_proposals: vec![],
@@ -167,6 +172,7 @@ impl<'a> ManagedGroup<'a> {
     /// [`Welcome`](crate::prelude::Welcome) message.
     pub fn add_members(
         &mut self,
+        key_store: &KeyStore,
         key_packages: &[KeyPackage],
     ) -> Result<(Vec<MLSMessage>, Welcome), ManagedGroupError> {
         if !self.active {
@@ -194,16 +200,22 @@ impl<'a> ManagedGroup<'a> {
             .iter()
             .collect::<Vec<&MLSPlaintext>>();
 
+        let credential = self.credential()?;
+        let credential_bundle = key_store
+            .get_credential_bundle(credential.signature_key())
+            .ok_or(ManagedGroupError::NoMatchingCredentialBundle)?;
+
         // Create Commit over all proposals
         // TODO #141
         let (commit, welcome_option, kpb_option) = self.group.create_commit(
             &self.aad,
-            &self.credential_bundle,
+            &credential_bundle,
             proposals_by_reference,
             proposals_by_value,
             false,
             None,
         )?;
+
         let welcome = match welcome_option {
             Some(welcome) => welcome,
             None => {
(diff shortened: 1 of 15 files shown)


src/group/managed_group/mod.rs
```diff
@@ -168,6 +173,7 @@ impl<'a> ManagedGroup<'a> {
     /// [`Welcome`](crate::prelude::Welcome) message.
     pub fn add_members(
         &mut self,
+        key_store: &KeyStore,
         key_packages: &[KeyPackage],
```
franziskuskiefer, line 177: Shouldn't we pass in the key package hashes here now?

src/key_store/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+//!
+//! In contrast to the functions providing access to `CredentialBundle`
+//! instances, the function to retrieve `KeyPackageBundle` instances deletes
+//! them from the `KeyStore`. This is because each `ManagedGroup` currently owns
+//! the `KeyPackageBundle` in its leaf, so upon creation of the group, it needs
+//! to consume a `KeyPackageBundle` instance. Note, that in contrast to
+//! `CredentialBundle` instances, `KeyPackageBundle` instances should not be
+//! used across groups.
+//!
+//! This design is temporary and only until the `ManagedGroup` is refactored to
```
franziskuskiefer, line 39: Can you add an issue number here?

src/key_store/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+        // period.
+        let mut kpbs = self.init_key_package_bundles.write().unwrap();
+        kpbs.insert(kp.hash(), kpb);
+        Ok(kp)
+    }
+
+    /// Generate a fresh `CredentialBundle` with the given parameters and store
+    /// it in the `KeyStore`. Returns the corresponding `Credential` or an error
+    /// if the creation of the `CredentialBundle` fails.
+    pub fn generate_credential(
```
franziskuskiefer, line 208: It is called `generate_credential` but returns a `CredentialBundle` that's confusing.

src/key_store/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+//! `SignaturePublicKey` of the corresponding `Credential` as index.
+//!
+//! ## Init `KeyPackageBundle` Instances
+//!
+//! Similarly, the `KeyStore` can generate "init" `KeyPackage` instances via
+//! `generate_key_package` and store the corresponding `KeyPackageBundle`. Init
+//! `KeyPackage` instances are meant to be published so other parties can use
+//! them to add the publishing party to groups.
+//!
+//! ### `KeyPackageBundle` Ownership
```
franziskuskiefer, line 29: This makes me expect a "`CredentialBundle` Ownership" section.

src/key_store/mod.rs
franziskuskiefer: Also, why only return the credential?

src/key_store/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+            index: signature_public_key,
+        })
+    }
+
+    /// Generate a fresh `KeyPackageBundle` with the given parameters, store it
+    /// in the `KeyStore` and return the corresponding `KeyPackage`. Throws an
+    /// error if no `CredentialBundle` can be found in the `KeyStore`
+    /// corresponding to the given `Credential` or if an error occurred during
+    /// the creation of the `KeyPackageBundle`.
+    pub fn generate_key_package(
```
franziskuskiefer, line 185: This generates a `KeyPackageBundle` not a `key_package`.

src/group/managed_group/mod.rs
```diff
@@ -699,14 +756,14 @@ impl<'a> ManagedGroup<'a> {
...
-    pub fn set_credential_bundle(&mut self, credential_bundle: &'a CredentialBundle) {
-        self.credential_bundle = credential_bundle;
-    }
-
-    /// Returns own credential
-    pub fn credential(&self) -> &Credential {
-        &self.credential_bundle.credential()
+    /// Returns own credential. If the group is inactive, it returns a
+    /// `UseAfterEviction` error.
+    pub fn credential(&self) -> Result<Credential, ManagedGroupError> {
```
franziskuskiefer, line 761: Why do you need this to return a copy?

src/group/managed_group/mod.rs
```diff
@@ -725,17 +782,23 @@ impl<'a> ManagedGroup<'a> {
...
     pub fn self_update(
         &mut self,
+        key_store: &KeyStore,
         key_package_bundle_option: Option<KeyPackageBundle>,
     ) -> Result<(Vec<MLSMessage>, Option<Welcome>), ManagedGroupError> {
         if !self.active {
             return Err(ManagedGroupError::UseAfterEviction(UseAfterEviction::Error));
         }
 
+        let credential = self.credential()?;
```
franziskuskiefer, line 792: I see this block a lot in here. The code should be factored out somehow to avoid having so many copies of this code.

src/lib.rs
```diff
@@ -39,6 +39,7 @@ mod extensions;
 pub mod framing;
 pub mod group;
 mod key_packages;
+mod key_store;
```
franziskuskiefer, line 42: `pub`?

tests/test_decryption_key_index.rs
```diff
@@ -20,7 +20,6 @@ ctest_ciphersuites!(decryption_key_index_computation, test(param: CiphersuiteNam
         ManagedGroupConfig::new(handshake_message_format, update_policy, 10, 0, callbacks);
     let number_of_clients = 20;
     let setup = ManagedTestSetup::new(managed_group_config, number_of_clients);
-    setup.create_clients();
         // Create a basic group with more than 4 members to create a tree with intermediate nodes.
```
franziskuskiefer, line 23: Can you fix the indentation here?

franziskuskiefer (changes requested):
Thanks! I think we're getting there this time 🙂 A couple comments, mostly around the API and some cleanup.

src/key_store/mod.rs
raphaelrobert: I think this is fine though, because the bundle is now only stored within the KeyStore. From an outside perspective, we only ever deal with key packages. The return type is consistent with the function name and the caller doesn't need to know what's happening internally.

src/key_store/test_key_store.rs
```diff
@@ -0,0 +1,79 @@
+use crate::config::*;
+use crate::{ciphersuite::*, credentials::CredentialBundle};
+use std::convert::TryFrom;
+
+use crate::credentials::CredentialType::Basic;
+use crate::key_packages::KeyPackageBundle;
+
+use super::{KeyStore, KeyStoreError};
+
+ctest_ciphersuites!(key_storage, test(param: CiphersuiteName) {
```
raphaelrobert, line 10: A brief description of the test would be good

raphaelrobert (changes requested):
Thanks for tackling this again! Definitely looks good, and we can always change things later if we need them.

kkohbrok (commented):
Thanks for the reviews! I tried to answer to all comments that I didn't immediately address in the code.

src/group/managed_group/mod.rs
```diff
@@ -86,17 +84,23 @@ pub struct ManagedGroup<'a> {
...
 
-impl<'a> ManagedGroup<'a> {
+impl ManagedGroup {
     // === Group creation ===
 
-    /// Creates a new group from scratch with only the creator as a member.
+    /// Creates a new group from scratch with only the creator as a member. This
+    /// function removes the `KeyPackageBundle` corresponding to the
+    /// `key_package_hash` from the `key_store`. Throws an error if no
+    /// `KeyPackageBundle` can be found.
```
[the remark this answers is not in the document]
raphaelrobert: You are right and I'm wrong.

src/group/managed_group/mod.rs
```diff
@@ -86,17 +84,23 @@ pub struct ManagedGroup<'a> {
...
     pub fn new(
-        credential_bundle: &'a CredentialBundle,
+        key_store: &KeyStore,
         managed_group_config: &ManagedGroupConfig,
         group_id: GroupId,
-        key_package_bundle: KeyPackageBundle,
+        key_package_hash: &[u8],
```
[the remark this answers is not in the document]
raphaelrobert: True,, let's then keep it as is.

src/group/managed_group/mod.rs
```diff
@@ -195,16 +201,22 @@ impl<'a> ManagedGroup<'a> {
             .iter()
             .collect::<Vec<&MLSPlaintext>>();
 
+        let credential = self.credential()?;
+        let credential_bundle = key_store
+            .get_credential_bundle(credential.signature_key())
+            .ok_or(ManagedGroupError::NoMatchingCredentialBundle)?;
+
```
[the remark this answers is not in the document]
raphaelrobert: It might also just work, let's worry about it when we get there.

src/group/managed_group/mod.rs
```diff
@@ -330,17 +355,24 @@ impl<'a> ManagedGroup<'a> {
...
     ) -> Result<Vec<MLSMessage>, ManagedGroupError> {
         if !self.active {
             return Err(ManagedGroupError::UseAfterEviction(UseAfterEviction::Error));
         }
+
+        let credential = self.credential()?;
+        let credential_bundle = key_store
+            .get_credential_bundle(credential.signature_key())
+            .ok_or(ManagedGroupError::NoMatchingCredentialBundle)?;
+
```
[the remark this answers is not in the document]
raphaelrobert: Since @franziskuskiefer asks the same question below, I will close this and we can continue the discussion there.

src/group/managed_group/mod.rs
raphaelrobert: I don't understand why this would be a problem for a helper function, but I haven't tried it myself. I was thinking about something along the lines of:

```
fn fetch_credential_bundle(credential: &Credential, key_store: &'a KeyStore) -> Result<CbGuard, ManagedGroupError> { 
    ... 
}
```

raphaelrobert (approved):
Except for the code duplication isuue raised by @franziskuskiefer, all other comments were addressed, so I will approve this now from my side.

src/key_store/mod.rs
```diff
@@ -0,0 +1,224 @@
...
+//!
+//! # Example
+//!
+//! A simple example for the generation and the retrieval of a
+//! `CredentialBundle` and a `KeyPackageBundle`.
+//!
+//! ```
+//! use openmls::prelude::*;
+//!
+//! let key_store = KeyStore::default();
```
raphaelrobert, line 54: I think there is a `mut` missing here

src/group/managed_group/mod.rs
franziskuskiefer: Ok, for the sake of getting this landed and as you say this should chnage in future, please add a ToDo and issue number.

src/group/managed_group/mod.rs
franziskuskiefer: Right, I don't think you can clean this up without changing `self.credential()` to return a reference.

src/key_store/mod.rs
franziskuskiefer: That should be a generates, not returns in the first comment.
My issue is not with returning a credential, but you don't generate a credential but a credential bundle. That's unexpected.

src/key_store/mod.rs
franziskuskiefer: It's in line with the other functions. But it's still not doing what it says, i.e. it's not generating only a key package.

src/key_store/mod.rs
kkohbrok: I see. So `generate_credential_bundle` (and `_key_package_bundle`) as a replacement? The documentation should then explain why it only returns a `Credential` (or `KeyPackage` respectively).

franziskuskiefer: Both isn't really great. I don't have a good suggestion for how to resolve this. But `generate_credential_bundle` is at least doing what it says. You could also have a separate generate and get function, i.e. don't return anything. But really this is some sort of `generate_key_pair`, not a `generate_random_point`, so this has to be made clear.
But since the API will change anyway I'm also fine with an explicit follow up issue.

franziskuskiefer (approved):
Let's get this in 🚢 and follow up on the issues to improve later.

PR #388, opened 2021-04-19
Fix error description visibility

raphaelrobert:
Error descriptions could not be accessed from outside the crate.


kkohbrok (approved):

franziskuskiefer:
@raphaelrobert this is not what you want. I suggest to back out this change. This overrides the deprecated `description` function of the `Error` trait now. There should be no need for this and this might make problems when interacting with other structs implementing the `Error` trait.

raphaelrobert:
Any suggestions? Because without that change the error structs are not usable outside of the crate.

Issue #387, opened 2021-04-19
[Feature request] Avoid copying the credential returned by the Managed API `credential` function


kkohbrok:
**Is your feature request related to a problem? Please describe.**
Once the `KeyStore` is merged, the `credential` function of the managed api will copy the credential every time it is called. This was done to avoid lifetime issues, as the actual credential lives in the tree, which in turn lives in a `RefCell`.

**Describe the solution you'd like**
There should be a guard structure not unlike the `CbGuard` used in the `KeyStore` for a similar purpose, which gets returned and which keeps the borrow of the tree alive. The guard structure then dereferences to the `Credential`.

**Describe alternatives you've considered**
We could change the way we store the tree such that we can return an actual reference.

**Additional context**
Since we're planning a rewrite of the TreeKEM component, it might make sense to wait until that is done and see if this issue persists.

PR #390, opened 2021-04-20
API changes for TreeKEM Test Vector integration

### LINKED ISSUE #380: [MLS Spec change] Parent hash verification algorithm fix (#460)

kkohbrok:
This PR changes the default behaviour for the `add_members` function in the Managed API such that it always forces a "self update", i.e. that the commit includes a `path`. 

This breaks various existing tests, so I had to pull in the fixes I made for the test vectors as well. This means there's no super well defined scope anymore for this PR, but I tried to include the individual fixes one commit at time.

One of the fixes in this PR fixes #380.

This PR additionally extends the Managed API slightly with a few functions that are only exposed for tests or interop testing.

### DIFF
--- a/src/group/managed_group/mod.rs
+++ b/src/group/managed_group/mod.rs
@@ -22,6 +22,9 @@ use crate::{
 use std::collections::HashMap;
 use std::io::{Error, Read, Write};
 
+#[cfg(any(feature = "expose-test-vectors", test))]
+use std::cell::Ref;
+
 pub use callbacks::*;
 pub use config::*;
 pub use errors::{
@@ -928,8 +931,8 @@ impl ManagedGroup {
     }
 
     #[cfg(any(feature = "expose-test-vectors", test))]
-    pub fn export_path_secrets(&self) -> Vec<Secret> {
-        self.group.tree().private_tree().path_secrets().to_vec()
+    pub fn export_path_secrets(&self) -> Ref<[Secret]> {
+        Ref::map(self.group.tree(), |tree| tree.private_tree().path_secrets())
     }
 
     #[cfg(any(feature = "expose-test-vectors", test))]
--- a/src/group/managed_group/mod.rs
+++ b/src/group/managed_group/mod.rs
@@ -168,7 +168,8 @@ impl ManagedGroup {
     ///
     /// New members are added by providing a `KeyPackage` for each member.
     ///
-    /// The `include_path` flag controls if a Path is included in the Commit.
+    /// This operation results in a Commit with a `path`, i.e. it includes an
+    /// update of the committer's leaf `KeyPackage`.
     ///
     /// If successful, it returns a `Vec` of
     /// [`MLSMessage`](crate::prelude::MLSMessage) and a
@@ -177,7 +178,6 @@ impl ManagedGroup {
         &mut self,
         key_store: &KeyStore,
         key_packages: &[KeyPackage],
-        include_path: bool,
     ) -> Result<(Vec<MlsMessage>, Welcome), ManagedGroupError> {
         if !self.active {
             return Err(ManagedGroupError::UseAfterEviction(UseAfterEviction::Error));
@@ -216,7 +216,7 @@ impl ManagedGroup {
             &credential_bundle,
             proposals_by_reference,
             proposals_by_value,
-            include_path,
+            true,
             None,
         )?;
 
--- a/src/group/managed_group/mod.rs
+++ b/src/group/managed_group/mod.rs
@@ -168,6 +168,8 @@ impl ManagedGroup {
     ///
     /// New members are added by providing a `KeyPackage` for each member.
     ///
+    /// The `include_path` flag controls if a Path is included in the Commit.
+    ///
     /// If successful, it returns a `Vec` of
     /// [`MLSMessage`](crate::prelude::MLSMessage) and a
     /// [`Welcome`](crate::prelude::Welcome) message.
--- a/src/group/managed_group/mod.rs
+++ b/src/group/managed_group/mod.rs
@@ -175,6 +175,7 @@ impl ManagedGroup {
         &mut self,
         key_store: &KeyStore,
         key_packages: &[KeyPackage],
+        include_path: bool,
     ) -> Result<(Vec<MlsMessage>, Welcome), ManagedGroupError> {
         if !self.active {
             return Err(ManagedGroupError::UseAfterEviction(UseAfterEviction::Error));
@@ -213,7 +214,7 @@ impl ManagedGroup {
             &credential_bundle,
             proposals_by_reference,
             proposals_by_value,
-            false,
+            include_path,
             None,
         )?;
 
@@ -922,6 +923,26 @@ impl ManagedGroup {
     /// Export the Ratchet Tree
     pub fn export_ratchet_tree(&self) -> Vec<Option<Node>> {
         self.group.tree().public_key_tree_copy()
+    }
+
+    #[cfg(any(feature = "expose-test-vectors", test))]
+    pub fn export_path_secrets(&self) -> Vec<Secret> {
+        self.group.tree().private_tree().path_secrets().to_vec()
+    }
+
+    #[cfg(any(feature = "expose-test-vectors", test))]
+    pub fn export_group_context(&self) -> &GroupContext {
+        self.group.context()
+    }
+
+    #[cfg(any(feature = "expose-test-vectors", test))]
+    pub fn tree_hash(&self) -> Vec<u8> {
+        self.group.tree().tree_hash()
+    }
+
+    #[cfg(any(feature = "expose-test-vectors", test))]
+    pub fn print_tree(&self, message: &str) {
+        _print_tree(&self.group.tree(), message)
     }
 }
 
--- a/src/tree/hashes.rs
+++ b/src/tree/hashes.rs
@@ -123,19 +123,17 @@ impl RatchetTree {
     pub(crate) fn original_child_resolution(&self, index: NodeIndex) -> Vec<&HPKEPublicKey> {
         // Build the exclusion list that consists of the unmerged leaves of the parent
         // node
-        let mut unmerged_nodes = vec![];
+        let mut unmerged_leaves = vec![];
         // If the current index is not the root, we collect the unmerged leaves of the
         // parent
         if let Ok(parent_index) = treemath::parent(index, self.leaf_count()) {
             // Check if the parent node is not blank
             if let Some(parent_node) = &self.nodes[parent_index].node {
-                for &index in &parent_node.unmerged_leaves {
-                    unmerged_nodes.push(index);
-                }
+                unmerged_leaves.extend_from_slice(&parent_node.unmerged_leaves);
             }
         };
         // Convert the exclusion list to a HashSet for faster searching
-        let exclusion_list: HashSet<&LeafIndex> = unmerged_nodes.iter().collect();
+        let exclusion_list: HashSet<&LeafIndex> = unmerged_leaves.iter().collect();
 
         // Compute the resolution for the index with the exclusion list
         let resolution = self.resolve(index, &exclusion_list);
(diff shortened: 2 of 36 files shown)


src/group/managed_group/mod.rs
```diff
@@ -923,6 +926,26 @@ impl ManagedGroup {
     pub fn export_ratchet_tree(&self) -> Vec<Option<Node>> {
         self.group.tree().public_key_tree_copy()
     }
+
+    #[cfg(any(feature = "expose-test-vectors", test))]
+    pub fn export_path_secrets(&self) -> Vec<Secret> {
+        self.group.tree().private_tree().path_secrets().to_vec()
```
franziskuskiefer, line 932: Should we return the slice here? (not sure how it's used but a reference appears to be fine for the group context below)

franziskuskiefer (dismissed):
lgtm with one nit

src/group/managed_group/mod.rs
kkohbrok: I think we run into the same problem as with the `credential` in the recent KeyStore PR, as the tree lives in a `RefCell`. See #387.

raphaelrobert:
I'm not sure about the `include_path` parameter. I left this out originally, because there should be an `UpdatePolicy` that decides whether or not an update is required. Consumers should not be bothered with that decision taking. We haven't implemented that yet and I understand you might need it for interop.

If you really need it urgently for interop, we should get this in as a temporary workaround with a ticket to fix it later.
If you have some more time, you could maybe think of some basic update policies? One policy simply could be aggressive and always include an update, that way it would work for the interop client as well.

kkohbrok:
I think that decision depends on the circumstances and thus either requires complex policies that consider at least group state and the inputs to the add, or a direct way to include a path. The consumer would have to make decisions in both cases. Either by defining the policies in a very detailed way (e.g. "I want to include a path whenever I add more than 10 people", or "I want to include a path whenever the group is sparse already", or "I want to include a path when I add people to a new group", etc) or by taking control over how they query the API.

On a different note, I think it would be a good idea security-wise to have Add include a path by default. We can then add a policy later, that allows the consumer to decide to sometimes not include a path.

I can also work around it if we want to postpone the discussion.

kkohbrok:
I've updated the description to reflect the new direction this PR has taken after some offline discussion. Please check it out before reviewing 🙂

src/group/managed_group/mod.rs
franziskuskiefer: Not really, you just have to call the `to_vec` on the outside if needed.

src/tree/hashes.rs
```diff
@@ -123,19 +123,19 @@ impl RatchetTree {
...
-        let mut unmerged_leaves = vec![];
+        let mut unmerged_nodes = vec![];
         // If the current index is not the root, we collect the unmerged leaves of the
         // parent
         if let Ok(parent_index) = treemath::parent(index, self.leaf_count()) {
             // Check if the parent node is not blank
             if let Some(parent_node) = &self.nodes[parent_index].node {
                 for index in &parent_node.unmerged_leaves {
-                    unmerged_leaves.push(NodeIndex::from(*index as usize));
+                    unmerged_nodes.push(NodeIndex::from(*index));
```
franziskuskiefer, line 133: something like `for &index in ...` looks nicer.

src/tree/mod.rs
```diff
@@ -195,7 +195,7 @@
                 node.unwrap()
                     .unmerged_leaves()
                     .iter()
-                    .map(|n| NodeIndex::from(*n)),
+                    .map(|n| NodeIndex::from(LeafIndex::from(*n))),
```
franziskuskiefer, line 198: `|&n|` instead of `*n`.
Also, the borrow for `node` above is unnecessary.

src/tree/mod.rs
```diff
@@ -591,6 +598,19 @@
         Ok(())
     }
 
+    /// Add a node for the provided key package the tree on the right side.
+    /// Note, that this function will not fill blank leaves.
+    fn add_node(&mut self, key_package: &KeyPackage) -> (NodeIndex, Credential) {
+        if !self.nodes.is_empty() {
+            self.nodes.push(Node::new_blank_parent_node());
+        }
+        self.nodes.push(Node::new_leaf(Some((key_package).clone())));
```
franziskuskiefer, line 607: This `clone` shouldn't be necessary. You have to change `add_nodes` not to take a reference. But then I don't see a need for cloning.

src/tree/mod.rs
```diff
@@ -605,37 +625,37 @@
         // Note that zip makes it so only the first free_leaves().len() nodes are taken.
         let free_leaves = self.free_leaves();
         let free_leaves_len = free_leaves.len();
-        for (new_kp, leaf_index) in new_kps.iter().zip(free_leaves) {
-            self.nodes[leaf_index] = Node::new_leaf(Some((*new_kp).clone()));
+        for (&new_kp, leaf_index) in new_kps.iter().zip(free_leaves) {
+            self.nodes[leaf_index] = Node::new_leaf(Some((new_kp).clone()));
```
franziskuskiefer, line 629: Same here with the `clone`.

franziskuskiefer (changes requested):
Generally looks good to me with a couple nits where I think we can do better. But that's more general cleanup that we should do while we're here 🙂

src/tree/mod.rs
```diff
@@ -605,37 +625,37 @@ impl RatchetTree {
...
                 .expect("add_nodes: Error when computing direct path.");
             for d in dirpath.iter() {
                 if !self.nodes[d].is_blank() {
                     let node = &self.nodes[d];
-                    let index = d.as_u32();
                     // TODO handle error
                     let mut parent_node = node.node.clone().unwrap();
-                    if !parent_node.unmerged_leaves().contains(&index) {
-                        parent_node.add_unmerged_leaf(index);
+                    if !parent_node.unmerged_leaves().contains(&leaf_index.into()) {
```
raphaelrobert, line 651: Clippy warning

src/tree/mod.rs
```diff
@@ -605,37 +625,37 @@ impl RatchetTree {
...
+                        parent_node.add_unmerged_leaf(leaf_index.into());
```
raphaelrobert, line 652: Clippy warning

src/tree/tests/test_resolution.rs
```diff
@@ -89,7 +89,7 @@ fn test_original_child_resolution() {
         // Resolution for root left child
         const LEFT_CHILD_RESOLUTION: &[usize] = &[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14];
         // Arbitrary unmerged leaves for root
-        const ROOT_UNMERGED_LEAVES: &[u32] = &[4, 6, 8];
+        const ROOT_UNMERGED_LEAVES: &[u32] = &[2, 3, 4];
```
raphaelrobert, line 92: 3 should be an invalid `NodeIndex` if you want to index a leaf

raphaelrobert (changes requested):
Thanks for the PR! Looks good in general, I commented on a few Clippy warnings.

src/group/managed_group/mod.rs
kkohbrok: Then I don't understand how to do it. I can't just omit the `to_vec()` and change the return type to `&[Secret]`. If I do that, `cargo` complains that I cannot return a value referencing a temporary value.

src/tree/mod.rs
kkohbrok: Well, the KeyPackage references get down here all the way from the proposals. I could try to refactor how proposals are handled, but I gave that a go and found that it's not straight-forward. Would you prefer if I have the caller do the cloning?

src/tree/mod.rs
kkohbrok: Same as above, without a refactor of the proposal handling, I believe that we have to clone somewhere.

src/tree/mod.rs
```diff
@@ -195,7 +195,7 @@ impl RatchetTree {
                 node.unwrap()
                     .unmerged_leaves()
                     .iter()
-                    .map(|n| NodeIndex::from(*n)),
+                    .map(|n| NodeIndex::from(LeafIndex::from(*n))),
```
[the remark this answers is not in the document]
kkohbrok: This became clear during refactoring. The resolution is not the unmerged leaves of that node, but that node, as well as its unmerged leaves. And the output should be `Vec<NodeIndex>`, so we need to indeed transform to `NodeIndex` here.

src/tree/tests/test_resolution.rs
kkohbrok: But `unmerged_leaves` are now  `Vec<LeafIndex>`, so 2, 3 and 4 are now the indices corresponding to the earlier 4, 6 and 8.

kkohbrok:
Thanks for the reviews! I've tried to address all the comments. In particular, I refactored the `resolve` function to take an exclusion list made of `LeafIndex`, which led to quite a few simplifications. It bloats the PR a bit, but I believe it's cleaner and less error prone now.

src/tree/hashes.rs
```diff
@@ -123,19 +123,19 @@ impl RatchetTree {
     pub(crate) fn original_child_resolution(&self, index: NodeIndex) -> Vec<&HPKEPublicKey> {
         // Build the exclusion list that consists of the unmerged leaves of the parent
         // node
-        let mut unmerged_leaves = vec![];
+        let mut unmerged_nodes = vec![];
```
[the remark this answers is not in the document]
raphaelrobert: I would then still call it `unmerged_leaves`, no?

src/tree/mod.rs
raphaelrobert: I guess the only way to get rid of the `clone()` would be to consume the `KeyPackage`. Not sure if that works for the caller though.

src/tree/tests/test_resolution.rs
raphaelrobert: Right, that makes sense now.

raphaelrobert (changes requested):
Almost there, just a nit from my side.

src/group/managed_group/mod.rs
franziskuskiefer: Something like this would work. But also not too important...
```rust
pub fn export_path_secrets(&self) -> Ref<[Secret]> {
    Ref::map(self.group.tree(), |tree| tree.private_tree().path_secrets())
}
```

src/tree/mod.rs
franziskuskiefer: I don't see a reason why it shouldn't work. `add_nodes` gets called with a list of key package references. It should be a list of key packages instead, i.e. the `add_proposals.iter()` there should actually take out the elements instead of referencing. This would also avoid any potential bug of re-using key packages because of wrong indices.
But this would actually require to refactor the way the proposals are applied... so let's leave it like this for now as we need the key package in two places right now.

franziskuskiefer (approved):
lgtm. Feel free to change the `export_path_secrets` or keep as is. I don't have a strong opinion here as it's not compiled outside of tests.

src/tree/hashes.rs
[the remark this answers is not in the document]
kkohbrok: 🤦🏻‍♂️ totally forgot about that. Thanks!

src/tree/mod.rs
kkohbrok: That's what I meant. It probably makes sense to do, but this PR already has some serious scope creep 😄

src/group/managed_group/mod.rs
kkohbrok: Ah, I had totally forgotten about `Ref::map()`. I could probably use that for the `credential` case, too...

raphaelrobert (dismissed):

raphaelrobert (approved):

PR #391, opened 2021-04-20
Store encoded key_pacakge

franziskuskiefer:
Encoding of key packages is a performance bottleneck.
In order to speed up things like tree hashes this PR
serializes the key package on creation and every update
and stores the serialized value in the key package itself.

### DIFF
--- a/src/key_packages/codec.rs
+++ b/src/key_packages/codec.rs
@@ -4,7 +4,7 @@ use crate::key_packages::*;
 
 impl Codec for KeyPackage {
     fn encode(&self, buffer: &mut Vec<u8>) -> Result<(), CodecError> {
-        buffer.append(&mut self.unsigned_payload()?);
+        buffer.extend_from_slice(&self.encoded);
         self.signature.encode(buffer)?;
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


src/key_packages/codec.rs
```diff
@@ -16,14 +16,16 @@ impl Codec for KeyPackage {
...
+        let mut kp = KeyPackage {
             protocol_version,
             ciphersuite: Config::ciphersuite(cipher_suite_name)?,
             hpke_init_key,
             credential,
             extensions,
             signature,
+            encoded: Vec::new(),
         };
+        kp.encoded = kp.unsigned_payload()?;
```
kkohbrok, line 28: It seems a bit odd to decode a key package and then immediately encode it again. But I assume this is a quirk our decoding mechanism?

kkohbrok (commented):
Nice optimization for both signatures and hashes! One question, otherwise it looks good to me.

src/key_packages/codec.rs
franziskuskiefer: I didn't want to touch the `Cursor` to copy the encoded payload... There will be more versions of this in future that make all this nicer...

kkohbrok: I took the liberty of creating issue #393. Looks good to me otherwise!

kkohbrok (approved):

PR #392, opened 2021-04-20
Slight refactoring of path secrets in preparation of TreeKEM kat merge

kkohbrok:
This is the second PR in a series of smaller PRs to bring the TreeKEM test vector to `main`. It contains slightly modified work from @franziskuskiefer's `franziskus/kat-tree-kem` branch.

This PR contains a single change to the API of the tree (how path secrets are exposed) and otherwise does some simple refactoring.

I apologize for the messy commit history of this branch! I hope it's small enough to understand what's going on.


raphaelrobert (approved):
This looks like a reasonable change to me. My only worry is that the code coverage decreases a bit, but I'm unsure of the reason. I'll approve this, but maybe you could check if there are low hanging fruits in terms of testing.

kkohbrok:
> This looks like a reasonable change to me. My only worry is that the code coverage decreases a bit, but I'm unsure of the reason. I'll approve this, but maybe you could check if there are low hanging fruits in terms of testing.

We are generally lacking tests for the tree functionalities. To properly test them, I'd want to have some sort of test framework to generate meaningful test scenarios. However, I think that's too much effort right now given that we want to re-write the tree anyway.

At the end of this PR-chain, we should have the inclusion of the TreeKEM test vector, which (using the managed testing framework as an integration test) will provide improved coverage.

PR #394, opened 2021-04-21
use tls-codec for tree hash structs

franziskuskiefer:
This PR starts using the tls-codec for OpenMLS TLS encoding, starting
with the tree hash structs. Note that this is a mix now, but the tree
hash inputs are pretty self contained and therefore a good starting
point. Some of the encoding is a little more manual than before. This
should be improved later on when more structs use the tls-codec crate.

The tls-codec crate now offers a way to pre-allocate memory for
serializing. The `TlsSize` trait must be implemented for any struct
that want to implement `Serialize` and provides a `serialized_len`
method to get the length of the struct before serializing.

This reduces the tree_hash time by about 30%. There's a lot more to
do here, but this is an easy start to improve performance.

### DIFF
--- a/src/ciphersuite/codec.rs
+++ b/src/ciphersuite/codec.rs
@@ -51,7 +51,7 @@ impl Codec for Signature {
 impl tls_codec::TlsSize for Signature {
     #[inline]
     fn serialized_len(&self) -> usize {
-        2 + self.value.len()
+        VecSize::VecU16.len_len() + self.value.len()
     }
 }
 
--- a/src/ciphersuite/codec.rs
+++ b/src/ciphersuite/codec.rs
@@ -48,6 +48,13 @@ impl Codec for Signature {
     }
 }
 
+impl tls_codec::TlsSize for Signature {
+    #[inline]
+    fn serialized_len(&self) -> usize {
+        2 + self.value.len()
+    }
+}
+
 impl Codec for HPKEPublicKey {
     fn encode(&self, buffer: &mut Vec<u8>) -> Result<(), CodecError> {
         encode_vec(VecSize::VecU16, buffer, self.as_slice())?;
--- a/src/tree/codec.rs
+++ b/src/tree/codec.rs
@@ -1,3 +1,5 @@
+use tls_codec::TlsSize;
+
 use crate::tree::{node::*, secret_tree::*, *};
 use std::convert::TryFrom;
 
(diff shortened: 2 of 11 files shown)


src/ciphersuite/codec.rs
```diff
@@ -48,6 +48,13 @@ impl Codec for Signature {
     }
 }
 
+impl tls_codec::TlsSize for Signature {
+    #[inline]
+    fn serialized_len(&self) -> usize {
+        2 + self.value.len()
```
raphaelrobert, line 54: It would be good to have a comment when const values are used, so that we know what to change in case the const changes some day.

src/tree/codec.rs
```diff
@@ -54,6 +56,17 @@ impl Codec for ParentNode {
     }
 }
 
+impl tls_codec::TlsSize for ParentNode {
+    #[inline]
+    fn serialized_len(&self) -> usize {
+        self.public_key.serialized_len()
+            + 4
```
raphaelrobert, line 63: const

src/tree/codec.rs
```diff
@@ -54,6 +56,17 @@ impl Codec for ParentNode {
...
+            + self.unmerged_leaves.len() * 4
+            + 1
```
raphaelrobert, line 65: const

raphaelrobert (commented):
Pre-allocating the memory is definitely a good start! I noticed that the code gets a lot uglier though. If that's the price we have to pay in the end, so be it. But specifically for the memory pre-allocation, couldn't we re-use `Cursor` to then serialise everything?

franziskuskiefer:
> But specifically for the memory pre-allocation, couldn't we re-use `Cursor` to then serialise everything?

I want to get rid of the `Cursor` eventually. The change here isn't always optimal. But we have to keep working what we have unless we want to rewrite everything in one go, which would be a huge PR.

What exactly do you think got worse here? I don't see much changes except for the manual vector encoding (which is about two lines instead of one now, which isn't too bad).

src/ciphersuite/codec.rs
franziskuskiefer: Not sure I understand.

kkohbrok: I think what @raphaelrobert means is that we have a `2` here without an explanation of why it's `2` and not some other constant. (And the same for the comments below.)

raphaelrobert (approved):

kkohbrok (commented):
Looks good to me with the exception of the magic numbers in the code (see comment). It would be great if those could be replaced by named constants.

franziskuskiefer:
I made a small change to get the lengths from the `VecSize` because you all don't like constants....

kkohbrok (approved):
Great, thanks!

PR #397, opened 2021-04-23
Refactor Managed API to accept and return individual messages

kkohbrok:
This PR refactors the functions in the Managed API to accept individual messages (`process_message`) and return individual messages (`propose_...`, etc).

This necessitates some refactoring of the events returned and in particular allows us to return errors if messages are invalid as opposed to just putting error events into the queue.

### DIFF
--- a/src/group/managed_group/errors.rs
+++ b/src/group/managed_group/errors.rs
@@ -5,7 +5,8 @@
 
 use crate::codec::CodecError;
 use crate::config::ConfigError;
-use crate::error::{ErrorPayload, ErrorString};
+use crate::error::ErrorString;
+use crate::framing::MlsCiphertextError;
 use crate::group::{ApplyCommitError, CreateCommitError, ExporterError, GroupError};
 use crate::key_store::KeyStoreError;
 
@@ -36,6 +37,7 @@ implement_error! {
             EmptyInput(EmptyInputError) =
                 "Empty input. Additional detail is provided.",
             KeyStoreError(KeyStoreError) = "See [`KeyStoreError`](`crate::key_store::KeyStoreError`) for details",
+            InvalidMessage(InvalidMessageError) = "The message could not be processed.",
         }
     }
 }
@@ -64,12 +66,14 @@ implement_error! {
         Simple {
             MembershipTagMismatch =
                 "A Proposal with an invalid membership tag was received.",
+            InvalidProposal =
+                "The given proposal is invalid.",
+            CommitWithInvalidProposals =
+                "A commit contained an invalid proposal.",
         }
         Complex {
-            InvalidCiphertext(ErrorPayload) =
+            InvalidCiphertext(MlsCiphertextError) =
                 "An invalid ciphertext was provided. The error returns the associated data of the ciphertext.",
-            CommitWithInvalidProposals(ErrorString) =
-                "A commit contained an invalid proposal. Additional detail is provided.",
             CommitError(ApplyCommitError) =
                 "See [`ApplyCommitError`](`crate::group::ApplyCommitError`) for details",
             GroupError(GroupError) =
--- a/src/group/managed_group/mod.rs
+++ b/src/group/managed_group/mod.rs
@@ -175,7 +175,7 @@ impl ManagedGroup {
         &mut self,
         key_store: &KeyStore,
         key_packages: &[KeyPackage],
-    ) -> Result<(Vec<MlsMessage>, Welcome), ManagedGroupError> {
+    ) -> Result<(MlsMessage, Welcome), ManagedGroupError> {
         if !self.active {
             return Err(ManagedGroupError::UseAfterEviction(UseAfterEviction::Error));
         }
@@ -233,7 +233,7 @@ impl ManagedGroup {
 
         // Convert MlsPlaintext messages to MLSMessage and encrypt them if required by
         // the configuration
-        let mls_messages = self.plaintext_to_mls_messages(vec![commit])?;
+        let mls_messages = self.plaintext_to_mls_message(commit)?;
 
         // Since the state of the group was changed, call the auto-save function
         self.auto_save();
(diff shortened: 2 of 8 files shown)


src/group/managed_group/mod.rs
```diff
@@ -802,67 +755,60 @@ impl ManagedGroup {
...
+                    true, /* force_self_update */
+                    None,
+                )?
             }
         };
+
+        // Take the new KeyPackageBundle and save it for later
+        let kpb = kpb_option.ok_or(ManagedGroupError::LibraryError(
+            "We didn't get a key package for a full commit on self update.".into(),
+        ))?;
```
franziskuskiefer, line 792: What clippy says.

src/group/managed_group/mod.rs
```diff
@@ -956,24 +902,20 @@ impl ManagedGroup {
...
-                }
-            };
-            out.push(msg);
-        }
-        Ok(out)
+        plaintext: MlsPlaintext,
+    ) -> Result<MlsMessage, ManagedGroupError> {
+        let msg = match self.configuration().handshake_message_format {
+            HandshakeMessageFormat::Plaintext => MlsMessage::Plaintext(plaintext),
+            HandshakeMessageFormat::Ciphertext => {
```
franziskuskiefer, line 911: Codecov suggests that we never use this with `Ciphertext`. Can you check if this is really the case?

tests/utils/managed_utils/client.rs
```diff
@@ -175,12 +173,16 @@ impl Client {
         let action_results = match action_type {
             ActionType::Commit => {
                 let (messages, welcome) = group.add_members(&self.key_store, key_packages)?;
-                (messages, Some(welcome))
+                (vec![messages], Some(welcome))
+            }
+            ActionType::Proposal => {
+                let mut messages = Vec::new();
+                for key_package in key_packages {
```
franziskuskiefer, line 180: This looks like it should be a `for_each()` with `collect()`.

tests/utils/managed_utils/client.rs
```diff
@@ -201,11 +203,19 @@ impl Client {
...
-                None,
-            ),
+            ActionType::Commit => {
+                let (message, welcome_option) =
+                    group.remove_members(&self.key_store, target_indices)?;
+                (vec![message], welcome_option)
+            }
+            ActionType::Proposal => {
+                let mut messages = Vec::new();
+                for &target_index in target_indices {
```
franziskuskiefer, line 213: Same here.

src/group/managed_group/mod.rs
```diff
@@ -802,67 +755,60 @@ impl ManagedGroup {
...
-            &credential_bundle,
-            &messages_to_commit,
-            &[],
-            true, /* force_self_update */
-            None,
-        )?;
-
-        // Add the Commit message to the other pending messages
-        plaintext_messages.push(commit);
+        let messages_to_commit: Vec<&MlsPlaintext> = self.pending_proposals.iter().collect();
```
franziskuskiefer, line 759: Having to collect something here sounds like the APIs are bad. But that's probably nothing to solve in this PR.

src/group/managed_group/mod.rs
```diff
@@ -437,170 +418,145 @@ impl ManagedGroup {
...
-                    }
+        // Check the type of message we received
+        let (plaintext, aad_option) = match message {
+            // If it is a ciphertext we decrypt it and return the plaintext message
+            MlsMessage::Ciphertext(ciphertext) => {
+                let aad = ciphertext.authenticated_data.clone();
+                (
+                    self.group
+                        .decrypt(&ciphertext)
+                        .map_err(|e| InvalidMessageError::InvalidCiphertext(e))?,
```
franziskuskiefer, line 437: What clippy says.

src/group/managed_group/mod.rs
```diff
@@ -437,170 +418,145 @@ impl ManagedGroup {
...
-                            events.push(GroupEvent::InvalidMessage(InvalidMessageEvent::new(
-                                InvalidMessageError::InvalidCiphertext(aad.into()),
-                            )));
-                            // Since we cannot decrypt the MlsCiphertext to a MlsPlaintext we move
-                            // to the next message
-                            continue;
-                        }
-                    }
+        // Check the type of message we received
+        let (plaintext, aad_option) = match message {
```
franziskuskiefer, line 430: This entire function needs to be refactored... maybe file a follow up.

franziskuskiefer (changes requested):
Generally looks good to me. Mostly nits and some follow ups.

tests/utils/managed_utils/client.rs
kkohbrok: Good idea! I'm not sure how I should handle the potential error thrown by `propose_add_member` in the `for_each` closure, though. Same in the other code below.

kkohbrok (commented):
Thanks for the review! I've addressed what I could and otherwise replied to the comments.

src/group/managed_group/mod.rs
franziskuskiefer: Do you want to add a test (or parameterise some tests to use ciphertexts)? If not, please file a follow up.

src/group/managed_group/errors.rs
```diff
@@ -36,6 +37,7 @@ implement_error! {
             EmptyInput(EmptyInputError) =
                 "Empty input. Additional detail is provided.",
             KeyStoreError(KeyStoreError) = "See [`KeyStoreError`](`crate::key_store::KeyStoreError`) for details",
+            InvalidMessage(InvalidMessageError) = "The message could not be processed.",
```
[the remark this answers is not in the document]
franziskuskiefer: Can you file a follow up then for a negative test case?

tests/utils/managed_utils/client.rs
franziskuskiefer: Ah, I didn't see the error there. Then let's leave it like this for now.

franziskuskiefer (approved):
lgtm, with the remaining comment addressed in some way.

src/group/managed_group/errors.rs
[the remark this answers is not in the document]
kkohbrok: I've included a link to this conversation in #133.

src/group/managed_group/mod.rs
kkohbrok: I've changed one of the managed group tests to use ciphertexts. That should increase the coverage.

franziskuskiefer (approved):
:shipit:

PR #398, opened 2021-04-29
check that tree math input is valid

franziskuskiefer:
I ran into some infinite loops while testing because tree math didn't check that inputs are valid. I added a couple checks that should cover most cases.

### DIFF
--- a/src/tree/treemath.rs
+++ b/src/tree/treemath.rs
@@ -6,6 +6,9 @@ implement_error! {
         LeafHasNoChildren = "Leaf nodes don't have children.",
         RootHasNoParent = "Root nodes don't have parents.",
         NotAParentNode = "Node index was not a parent node.",
+        LeafNotInTree = "The leaf index is larger than the tree size.",
+        NodeNotInTree = "The node index is larger than the tree size.",
+        InvalidInput = "The provided input is invalid for tree math.",
     }
 }
 
@@ -69,6 +72,7 @@ pub(crate) fn right(index: NodeIndex, size: LeafIndex) -> Result<NodeIndex, Tree
     Ok(NodeIndex::from(r))
 }
 
+// The parent here might be beyond the right edge of the tree.
 pub(crate) fn parent_step(x: usize) -> usize {
     let k = level(NodeIndex::from(x));
     let b = (x >> (k + 1)) & 0x01;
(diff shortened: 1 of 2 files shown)


src/tree/treemath.rs
```diff
@@ -69,40 +72,73 @@ pub(crate) fn right(index: NodeIndex, size: LeafIndex) -> Result<NodeIndex, Tree
...
+fn unsafe_parent(index: NodeIndex, size: LeafIndex) -> Result<NodeIndex, TreeMathError> {
     let x = index.as_usize();
     let n = size.as_usize();
     if index == root(size) {
         return Err(TreeMathError::RootHasNoParent);
     }
     let mut p = parent_step(x);
-    while p >= node_width(n) {
-        p = parent_step(p)
+    let node_width = node_width(n);
```
kkohbrok, line 96: I'm just curious: Why the explicit `node_width` variable here?

src/tree/treemath.rs
```diff
@@ -69,40 +72,73 @@ pub(crate) fn right(index: NodeIndex, size: LeafIndex) -> Result<NodeIndex, Tree
...
+    while p >= node_width {
+        let new_p = parent_step(p);
+        if new_p == p {
```
kkohbrok, line 99: Can `parent_step` of an index be the same as the index itself? I thought we should be checking here if the output is outside of the tree.

src/tree/treemath.rs
```diff
@@ -69,40 +72,73 @@ pub(crate) fn right(index: NodeIndex, size: LeafIndex) -> Result<NodeIndex, Tree
...
 }
 
 pub(crate) fn parent(index: NodeIndex, size: LeafIndex) -> Result<NodeIndex, TreeMathError> {
+    node_in_tree(index, size)?;
+    unsafe_parent(index, size)
+}
+
+// This function is only safe to use if index <= size.
+// If this is not checked before calling the function, `parent` should be used.
+fn unsafe_parent(index: NodeIndex, size: LeafIndex) -> Result<NodeIndex, TreeMathError> {
```
kkohbrok, line 89: If this function returns `Result` anyway, why not just have it return an error if the input index is outside of the tree?

kkohbrok (commented):
Thanks for picking this up! Just a few questions, otherwise it should be good to go.

src/tree/treemath.rs
franziskuskiefer: I guess it's a debug artefact that I forgot to remove...

src/tree/treemath.rs
franziskuskiefer: This is a pure "performance improvement". It's avoiding the check when we know that it was done before. (I don't think we'd notice the performance difference...)

src/tree/treemath.rs
franziskuskiefer: Well, this is a purely opportunistic check. If this ends up being true, we'd end up in an endless loop. It's not an actual correctness check.

kkohbrok (dismissed):
Let's get this in!

src/tree/treemath.rs
```diff
@@ -69,40 +72,72 @@ pub(crate) fn right(index: NodeIndex, size: LeafIndex) -> Result<NodeIndex, Tree
...
+    let p = unsafe_parent(index, size)?;
     match index.cmp(&p) {
         Ordering::Less => right(p, size),
         Ordering::Greater => left(p),
         Ordering::Equal => left(p),
     }
 }
 
+#[inline(always)]
+fn node_in_tree(node_index: NodeIndex, size: LeafIndex) -> Result<(), TreeMathError> {
```
raphaelrobert, line 117: This function should have a dedicated test.

src/tree/treemath.rs
```diff
@@ -69,40 +72,72 @@ pub(crate) fn right(index: NodeIndex, size: LeafIndex) -> Result<NodeIndex, Tree
...
+    if node_index.as_usize() >= node_width(size.as_usize()) {
+        Err(TreeMathError::NodeNotInTree)
+    } else {
+        Ok(())
+    }
+}
+
+#[inline(always)]
+fn leaf_in_tree(leaf_index: LeafIndex, size: LeafIndex) -> Result<(), TreeMathError> {
```
raphaelrobert, line 126: This function should have a dedicated test.

raphaelrobert (changes requested):
Thanks! LGTM in general, I just flagged two missing tests that I think should be there.

raphaelrobert (commented):
Thanks for adding the tests! I think codecov is unhappy because you used `.unwrap()` instead of `.expect()`.

raphaelrobert (approved):

PR #399, opened 2021-05-03
Fix bug in path_secrets function and include TreeKEM test vector

kkohbrok:
The final PR in a sequence of PRs to integrate message and TreeKEM test vectors. As with the previous PRs, this is mostly @franziskuskiefer's code that I'm organizing into PRs.

This PR notably includes a fix to the `path_secret` function, which now checks if nodes are blank before assuming that we have a given path secret.


src/ciphersuite/mod.rs
```diff
@@ -393,7 +393,7 @@ impl Secret {
     }
 
     /// Returns the inner bytes of a secret
-    pub(crate) fn as_slice(&self) -> &[u8] {
+    pub fn as_slice(&self) -> &[u8] {
```
franziskuskiefer, line 396: Do we really want this to be public?

src/extensions/mod.rs
```diff
@@ -15,7 +15,7 @@ pub use errors::*;
 pub use key_package_id_extension::KeyIdExtension;
 pub use life_time_extension::LifetimeExtension;
 pub(crate) use parent_hash_extension::ParentHashExtension;
-pub(crate) use ratchet_tree_extension::RatchetTreeExtension;
+pub use ratchet_tree_extension::RatchetTreeExtension;
```
franziskuskiefer, line 18: Why should this be public?

src/extensions/ratchet_tree_extension.rs
```diff
@@ -37,7 +37,7 @@ pub struct RatchetTreeExtension {
 
 impl RatchetTreeExtension {
     /// Build a new extension from a vector of `Node`s.
-    pub(crate) fn new(tree: Vec<Option<Node>>) -> Self {
+    pub fn new(tree: Vec<Option<Node>>) -> Self {
```
franziskuskiefer, line 40: pub?

src/key_packages/mod.rs
```diff
@@ -455,6 +455,28 @@ impl KeyPackageBundle {
...
+            .into_keys();
+        KeyPackageBundle {
+            key_package: key_package.clone(),
+            private_key,
+            leaf_secret: leaf_secret.clone(),
+        }
+    }
+
+    #[cfg(any(feature = "expose-test-vectors", test))]
+    pub fn get_leaf_secret(&self) -> Secret {
```
franziskuskiefer, line 477:
```suggestion
    pub fn leaf_secret(&self) -> Secret {
```

Why not return a reference?

src/key_packages/mod.rs
```diff
@@ -455,6 +455,28 @@ impl KeyPackageBundle {
             leaf_secret,
         }
     }
+
+    #[cfg(any(feature = "expose-test-vectors", test))]
+    pub fn from_key_package_and_leaf_secret(
```
franziskuskiefer, line 460: This looks way to complex for something that's a test only function. Either we should actually use this function somewhere in the code or we shouldn't need it for thest.

src/tree/mod.rs
```diff
@@ -790,7 +789,15 @@ impl RatchetTree {
...
         // root from our own leaf.
-        if let Ok(dirpath) = treemath::leaf_direct_path(self.own_node_index(), self.leaf_count()) {
+        if let Ok(mut dirpath) =
+            treemath::leaf_direct_path(self.own_node_index(), self.leaf_count())
+        {
+            // Filter out blanks for which we don't have path secrets
+            let dirpath: Vec<NodeIndex> = dirpath
+                .drain(..)
+                .filter(|&index| !self.nodes[index].is_blank())
+                .collect();
```
franziskuskiefer, line 799: The next call on `dirpath` is an `iter()`. Then you don't need to collect here.

src/tree/tests/kat_tree_kem.rs
```diff
@@ -0,0 +1,252 @@
...
+//! for more description on the test vectors.
+//!
+//! The test vector describes a tree of `n` leaves adds a new leaf with
+//! `my_key_package` and `my_path_secret` (common ancestor of `add_sender` and
+//! `my_key_package`).
+//! Then an update, sent by `update_sender` with `update_path` is processed, which
+//! is processed by the newly added leaf as well.
+//!
+//! Some more points
+//! * An empty group context is used.
```
franziskuskiefer, line 13: Please update the description here. At least this part isn't correct anymore.

src/tree/tests/kat_tree_kem.rs
```diff
@@ -0,0 +1,252 @@
...
+//! * update path with empty exclusion list.
+
+#[cfg(any(feature = "expose-test-vectors", test))]
```
franziskuskiefer, line 16: I'd expect this file only to be compiled for tests or test vectors. So why is this needed? (Same for the other `cfg` below.)

src/tree/tests/kat_tree_kem.rs
```diff
@@ -0,0 +1,252 @@
...
+    if tree_after.verify_parent_hashes().is_err() {
+        if cfg!(test) {
+            panic!("Parent hash mismatch in the 'after' tree.");
+        }
+        return Err(TreeKemTestVectorError::AfterParentHashMismatch);
+    }
+
+    // Initialize private portion of the RatchetTree
+    let add_sender = test_vector.add_sender;
+    println!(
```
franziskuskiefer, line 129: Can you convert all the `println`s here to logging?

src/tree/tests/kat_tree_kem.rs
```diff
@@ -0,0 +1,252 @@
...
+    let root_secret_after_add: &PathSecret = &Secret::from_slice(
+        &hex_to_bytes(&test_vector.root_secret_after_add),
+        ProtocolVersion::default(),
+        ciphersuite,
+    )
+    .into();
+
+    if root_secret_after_add
+        != tree_before
+            //.private_tree()
```
franziskuskiefer, line 168: commented code

tests/test_tree_kem_kat.rs
```diff
@@ -0,0 +1,278 @@
...
+//! for more description on the test vectors.
+//!
+//! The test vector describes a tree of `n` leaves adds a new leaf with
+//! `my_key_package` and `my_path_secret` (common ancestor of `add_sender` and
+//! `my_key_package`).
+//! Then an update, sent by `update_sender` with `update_path` is processed, which
+//! is processed by the newly added leaf as well.
+//!
+//! Some more points
+//! * An empty group context is used.
```
franziskuskiefer, line 13: Duplicated (outdated) description.

tests/test_tree_kem_kat.rs
```diff
@@ -0,0 +1,278 @@
...
+
+use openmls::{prelude::*, test_util::*};
+
+mod utils;
+use utils::managed_utils::*;
+
+use serde::{self, Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct TreeKemTestVector {
```
franziskuskiefer, line 24: This file contains a lot of duplicated code. We should find a way to share code that's used in both, generation and test.

tests/test_tree_kem_kat.rs
```diff
@@ -0,0 +1,278 @@
...
+    if n_leaves <= 1 {
+        panic!("test vector can only be generated with two or more members")
+    }
+    // Set up a group with `n_leaves` members.
+    let handshake_message_format = HandshakeMessageFormat::Plaintext;
+    let update_policy = UpdatePolicy::default();
+    let callbacks = ManagedGroupCallbacks::default();
+    let managed_group_config =
+        ManagedGroupConfig::new(handshake_message_format, update_policy, 0, 0, callbacks);
+    let setup = ManagedTestSetup::new(managed_group_config, n_leaves as usize + 10);
```
franziskuskiefer, line 75: Why `+ 10`?

tests/test_tree_kem_kat.rs
```diff
@@ -0,0 +1,278 @@
...
+        .unwrap();
+
+    let mut groups = setup.groups.borrow_mut();
+    let group = groups.get_mut(&group_id).unwrap();
+    let remover_id = group.random_group_member();
+    let mut target_id = group.random_group_member();
+    while remover_id == target_id {
+        target_id = group.random_group_member();
+    }
+    //let remover_index = thread_rng().gen_range(0..n_leaves);
```
franziskuskiefer, line 99: Commented code.

tests/test_tree_kem_kat.rs
```diff
@@ -0,0 +1,278 @@
...
+    println!("remover id: {:?}", remover_id);
```
franziskuskiefer, line 100: Please use logging instead of println.

tests/test_tree_kem_kat.rs
```diff
@@ -0,0 +1,278 @@
...
+    // We then have the same client who removed the target add a fresh member.
+    let adder_id = remover_id;
+    println!("adder id: {:?}", adder_id);
+    let (adder_index, _) = group
+        .members
+        .iter()
+        .find(|(_, id)| id == &adder_id)
+        .unwrap()
+        .clone();
+    let addees = setup.random_new_members_for_group(group, 1).unwrap();
```
franziskuskiefer, line 132: Instead of using this as `addees[0]` every time you could actually get out the one addee in there.

tests/test_tree_kem_kat.rs
```diff
@@ -0,0 +1,278 @@
...
+                }
+            }
+            return false;
+        })
+        .unwrap()
+        .as_ref()
+        .unwrap()
+        .key_package()
+        .unwrap()
+        .clone();
```
franziskuskiefer, line 244: Why do you need to clone it?

tests/test_tree_kem_kat.rs
```diff
@@ -0,0 +1,278 @@
...
+            if let Some(node) = node_option {
+                if let Some(key_package) = node.key_package() {
+                    if key_package.credential().identity() == &addees[0] {
+                        return true;
+                    }
+                }
+            }
+            return false;
+        })
+        .unwrap()
```
franziskuskiefer, line 239: There are a lot of unwraps here... not sure if they are all necessary.

franziskuskiefer (changes requested):

raphaelrobert (dismissed):
The fix looks good, I'll leave the review of the test vector related parts to @franziskuskiefer.

tests/test_tree_kem_kat.rs
kkohbrok: Good point. I've made the values in the "internal" test vector definition public such that we can use it here.

tests/test_tree_kem_kat.rs
kkohbrok: It's just lots of option types to dig through until I reach my own key package. It's going to be fewer after the tree rewrite ;-)

kkohbrok (commented):
Thanks for the review and sorry for the obvious stuff in the test vector files. I really should have looked over those.

src/key_packages/mod.rs
franziskuskiefer: The problem is not the complexity of this code but that you are not testing any code that's actually being used. This should rather share code with `new_from_leaf_secret`, which is generating a `KeyPackageBundle` from a `leaf_secret` already.

src/tree/mod.rs
```diff
@@ -790,9 +789,16 @@ impl RatchetTree {
     pub(crate) fn path_secret(&self, index: NodeIndex) -> Option<&PathSecret> {
         // Get a Vector containing the node indices of the direct path to the
         // root from our own leaf.
-        if let Ok(dirpath) = treemath::leaf_direct_path(self.own_node_index(), self.leaf_count()) {
+        if let Ok(mut dirpath) =
+            treemath::leaf_direct_path(self.own_node_index(), self.leaf_count())
+        {
+            // Filter out blanks for which we don't have path secrets
+            let mut dirpath = dirpath
```
franziskuskiefer, line 796: No need for the `let`.

franziskuskiefer (changes requested):
Two more nits.

src/key_packages/mod.rs
kkohbrok: Ah, I see what you mean now. I think I tried that, but then I would need the `CredentialBundle` that was used to sign the `KeyPackage` to begin with and that's not part of the test vector. The way that the test vector is structured doesn't really match any of the code we use in practice. As a compromise I've restructured the code such that we now at least test the function that is used when refreshing an existing key package.

src/tree/mod.rs
kkohbrok: I don't see a way to just remove it. I change `dirpath` from a vector to an iterator, so I think I have to use a `let` here. I've changed the variables to make that clearer. Let me know if I'm missing something here.

src/key_packages/mod.rs
```diff
@@ -300,20 +300,29 @@ impl KeyPackageBundle {
...
-
-        // Repackage everything as a KeyPackageBundle
         let mut new_key_package = key_package.clone();
+        // Set the public key and re-set the unsigned payload.
         new_key_package.set_hpke_init_key(public_key);
-        KeyPackageBundle::new_from_values(new_key_package, private_key, leaf_secret)
+        KeyPackageBundle {
+            key_package: new_key_package,
+            private_key,
+            leaf_secret: leaf_secret.clone(),
```
franziskuskiefer, line 324: What clippy says.

franziskuskiefer (changes requested):
Thanks, just one clippy warning to fix

franziskuskiefer (approved):

PR #400, opened 2021-05-04
Add messages test vector

kkohbrok:
This PR adds the message test vector, as well missing codec implementations for the `Capabilities` and `Lifetime` extensions. It also adds as a couple of (cfg-gated) functions required for the test vector generation.

(This PR is labeled XL, because it contains the actual messages test vector.)


src/group/tests/kat_messages.rs
```diff
@@ -0,0 +1,512 @@
...
+
+    let psk_id = PreSharedKeyId::new(
+        PskType::External,
+        Psk::External(ExternalPsk::new(get_random_vec(ciphersuite.hash_length()))),
+        get_random_vec(ciphersuite.hash_length()),
+    );
+
+    let psk_proposal = PreSharedKeyProposal { psk: psk_id };
+    let reinit_proposal = ReInitProposal {
+        group_id: group_id.clone(),
```
raphaelrobert, line 110: clippy warning

raphaelrobert (changes requested):
Thanks! It would be good if you could a a test for the encoding/decoding of the two extensions.

raphaelrobert (dismissed):

src/extensions/capabilities_extension.rs
```diff
@@ -143,3 +143,27 @@ impl Extension for CapabilitiesExtension {
...
+        encode_vec(VecSize::VecU8, &mut buffer, &self.ciphersuites)?;
+        encode_vec(VecSize::VecU8, &mut buffer, &self.extensions)?;
+        Ok(())
+    }
+
+    fn decode(cursor: &mut Cursor) -> Result<Self, crate::codec::CodecError> {
+        let version_numbers: Vec<u8> = decode_vec(VecSize::VecU8, cursor)?;
+        let mut versions = Vec::new();
+        for &version_number in version_numbers.iter() {
+            versions.push(ProtocolVersion::try_from(version_number)?)
```
franziskuskiefer, line 159: Why don't you use the `decode` function of `ProtocolVersion`?

src/group/tests/kat_messages.rs
```diff
@@ -0,0 +1,512 @@
+use crate::{
```
franziskuskiefer, line 1: Can you add a description to what's happening here and where it's coming from?

src/group/tests/kat_messages.rs
```diff
@@ -0,0 +1,512 @@
...
+use evercrypt::prelude::get_random_vec;
+
+use serde::{self, Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, Clone, Default)]
+pub struct MessagesTestVector {
+    key_package: String,  // serialized KeyPackage,
+    capabilities: String, // serialized Capabilities,
+    lifetime: String,     // serialized {uint64 not_before; uint64 not_after;},
+    ratchet_tree: String, /* serialized optional<Node> ratchet_tree<1..2^32-1>; */
```
franziskuskiefer, line 19: Why the different commenting styles?

src/group/tests/kat_messages.rs
```diff
@@ -0,0 +1,512 @@
...
+
+    commit: String, /* serialized Commit */
+
+    mls_plaintext_application: String, /* serialized MLSPlaintext(ApplicationData) */
+    mls_plaintext_proposal: String,    /* serialized MLSPlaintext(Proposal(*)) */
+    mls_plaintext_commit: String,      /* serialized MLSPlaintext(Commit) */
+    mls_ciphertext: String,            /* serialized MLSCiphertext */
+}
+
+#[cfg(any(feature = "expose-test-vectors", test))]
```
franziskuskiefer, line 43: Shouldn't the entire module guarded with this and hence any further cfg guards be unnecessary in here?

src/group/tests/kat_messages.rs
```diff
@@ -0,0 +1,512 @@
...
+    let group_info = GroupInfo::new(
+        group_id.clone(),
+        GroupEpoch(0),
+        get_random_vec(ciphersuite.hash_length()),
+        get_random_vec(ciphersuite.hash_length()),
+        vec![Box::new(RatchetTreeExtension::new(ratchet_tree.clone()))],
+        ConfirmationTag(Mac {
+            mac_value: get_random_vec(ciphersuite.hash_length()),
+        }),
+        LeafIndex::from(0u32),
```
franziskuskiefer, line 82: Should this be random as well?

src/group/tests/kat_messages.rs
```diff
@@ -0,0 +1,512 @@
...
+
+    let psk_proposal = PreSharedKeyProposal { psk: psk_id };
+    let reinit_proposal = ReInitProposal {
+        group_id,
+        version: ProtocolVersion::Mls10,
+        ciphersuite: ciphersuite_name,
+        extensions: vec![Box::new(RatchetTreeExtension::new(ratchet_tree.clone()))],
+    };
+    // We don't support external init proposals yet.
+    let external_init_proposal = vec![0u8];
```
franziskuskiefer, line 116: We should leave the fields we don't support empty instead of adding values. This could be interpreted as a wrong value.

src/group/tests/kat_messages.rs
```diff
@@ -0,0 +1,512 @@
...
+    } else {
+        panic!("Wrong content of MLS plaintext");
+    };
+    let welcome = welcome_option.unwrap();
+    let mls_ciphertext_application = group
+        .create_application_message(b"aad", b"msg", &credential_bundle, random_u8() as usize)
+        .unwrap();
+    let mls_plaintext_application = group.decrypt(&mls_ciphertext_application).unwrap();
+
+    let mls_ciphertext = group
```
franziskuskiefer, line 150: Should we randomly pick which plaintext to encrypt? Or encrypt all of them? If it increases coverage, we should probably encrypt all of them.

franziskuskiefer (changes requested):
Generally looks good! A couple nits and questions

src/group/tests/kat_messages.rs
kkohbrok: Can't remember. I've aligned it now.

src/group/tests/kat_messages.rs
kkohbrok: As far as I can see, there's only one field for MLSCiphertext that I can populate. Also, since the ciphertext is just decoded and encoded (not decrypted), the coverage should remain mostly the same, as the only thing that really changes is the content type. I've randomized the encryption target, which should cover all the encoding and decoding of any content type.

kkohbrok (commented):
Thanks for the review! I think I've addressed all the comments. Let me know if I've missed anything.

franziskuskiefer (approved):
:shipit:

PR #403, opened 2021-05-11
Refactor GroupContext and expose extensions to Mls- and managed group APIs

### LINKED ISSUE #186: Implement GroupContext extensions

kkohbrok:
This PR fixes #186 by allowing the group creator to set extensions when creating a group. 

When sending a Welcome message, the GroupInfo is created from the GroupContext with the exception of the RatchetTreeExtensions, which is included iff the flag in the MlsGroup is set. Similarly, when creating a new group from a Welcome message, a GroupContext is created from the GroupInfo, where a potential RatchetTreeExtension is filtered out.

Both the MlsGroup, as well as the ManagedGroup APIs are extended to allow outsiders to access extensions of a group.

Finally, instead of cloning extensions when creating a GroupContext, the group creator will have to pass in a copy of any extension they want to have in the group. I made this change, because this might help us avoid a clone if the extension has to be instantiated for all new groups anyway, where previously, extensions would always be cloned.


src/group/mls_group/mod.rs
```diff
@@ -73,20 +73,19 @@ impl MlsGroup {
         config: GroupConfig,
         psk_option: impl Into<Option<PskSecret>>,
         version: impl Into<Option<ProtocolVersion>>,
+        extensions: Vec<Box<dyn Extension>>,
```
franziskuskiefer, line 76: Is there a reason this has to be consumed? If not, use a slice.

src/group/group_context.rs
```diff
@@ -7,26 +12,65 @@
         epoch: GroupEpoch,
         tree_hash: Vec<u8>,
         confirmed_transcript_hash: Vec<u8>,
-        extensions: &[Box<dyn Extension>],
+        extensions: Vec<Box<dyn Extension>>,
```
franziskuskiefer, line 15: Slice is more generic unless there's a reason to consume the vector.

src/group/group_context.rs
```diff
@@ -7,26 +12,65 @@
...
             serialized: vec![],
         };
         let serialized = group_context.encode_detached()?;
         group_context.serialized = serialized.to_vec();
         Ok(group_context)
     }
+
+    /// Create the group context of a new epoch from the one of the previous
+    /// epoch.
+    pub(crate) fn from_previous_group_context(
```
franziskuskiefer, line 32: `from_group_context`?
Also it's not entirely clear how this is a `from`, but requiring additional arguments.

src/group/group_context.rs
```diff
@@ -7,26 +12,65 @@
...
+        GroupContext::new(
+            group_context.group_id.clone(),
+            epoch,
+            tree_hash,
+            confirmed_transcript_hash,
+            group_context.extensions.clone(),
+        )
+    }
+
+    pub(crate) fn into_group_info(
```
franziskuskiefer, line 48: Please add a comment what this is doing.

src/group/group_context.rs
```diff
@@ -7,26 +12,65 @@
...
+        mut self,
+        option_ratchet_tree_extension: Option<RatchetTreeExtension>,
```
franziskuskiefer, line 50: `impl Into`?

src/group/mls_group/create_commit.rs
```diff
@@ -97,16 +95,11 @@ impl MlsGroup {
...
 
-        // TODO #186: Implement extensions
-        let extensions: Vec<Box<dyn Extension>> = Vec::new();
-
         // Calculate group context
-        let provisional_group_context = GroupContext::new(
-            self.group_context.group_id.clone(),
-            provisional_epoch,
-            tree_hash.clone(),
+        let provisional_group_context = GroupContext::from_previous_group_context(
```
franziskuskiefer, line 99: Looking at how this is used here. Would it make sense to have a `ProvusionalGroupContext` struct instead that only stores the necessary changes and/or references?

src/group/managed_group/mod.rs
```diff
@@ -102,6 +102,7 @@
         managed_group_config: &ManagedGroupConfig,
         group_id: GroupId,
         key_package_hash: &[u8],
+        extensions: Vec<Box<dyn Extension>>,
```
franziskuskiefer, line 105: Maybe this should get a builder pattern at some point. It's getting wuite a lot of arguments now and some, like the `extensions` here, might not be needed in many cases (at least in all the tests right now).

franziskuskiefer (changes requested):
Generally looks good with a couple nits.


Also note that I suspect that this will create some merge conflicts with #405. Make sure you coordinate the two.

PR #404, opened 2021-05-11
info is empty and aad is the group context

franziskuskiefer:
The info and aad parameters were swapped ([spec for reference](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hpke-08.txt#section-5.1.1)).

Thanks for Théophile for finding this.


franziskuskiefer:
> I'm confused. As I read the spec, it should be first info, then aad. But then your changes move the `group_context` to the `aad` field?

Order doesn't matter, it's different everywhere. The spec defines `SetupBaseS(node_public_key, "")` for `SetupBaseS(pkR, info)` and `Seal(group_context, path_secret)` for `ContextS.Seal(aad, pt)`. So the `info` is empty and the `group_context` is the `aad`. This is currently the other/wrong way.

kkohbrok (approved):

raphaelrobert (approved):

PR #405, opened 2021-05-12
Refactor group info extensions for MlsGroup

raphaelrobert:
This PR prepares `MlsGroup` for custom group info extensions.
Details:

 - I got rid of `GroupConfig`, because it was only used for the ratchet tree extension
 - `MlsGroup` still has an internal state `bool` to track whether to use the ratchet tree extension. The reason for that is that it is not possible to treat the ratchet tree extension like any other group info extension. When a `Commit` is created, the corresponding `Welcome` message needs the ratchet tree of the provisional state. Since the provisional state is ephemeral (for now), it is not possible to generate the ratchet tree extension from outside the group. Long story short, this part hasn't changed. The alternative would be to not store the state in the group and instead pass a `bool` for every `create_commit` call, but this might get harder to track.
  - `create_commit` now accepts a `Vec` of group info extensions that are to be included in the `Welcome` message. Conversely, `new_from_welcome` now also return all group info extensions, except for the ratchet tree extension.
  - A new trait for group info extensions is introduced: `GroupInfoExtension`. The idea behind it is that it can be used by consumer code easily by only implementing `Codec` and attributing an extension number.
  - There is a test with a custom extension that implements `GroupInfoExtension`.
  - None of this affects the `ManagedGroup` so far, that will follow in another PR.

### DIFF
--- a/cli/src/user.rs
+++ b/cli/src/user.rs
@@ -229,14 +229,13 @@ impl User {
         let mut group_aad = group_id.to_vec();
         group_aad.extend(b" AAD");
         let kpb = self.identity.borrow_mut().update();
-        let config = GroupConfig::default();
         let mls_group = MlsGroup::new(
             group_id,
             CIPHERSUITE,
             kpb,
-            config,
-            None, /* Initial PSK */
-            None, /* MLS version */
+            false, /* use ratchet tree extension */
+            None,  /* Initial PSK */
+            None,  /* MLS version */
         )
         .unwrap();
         let group = Group {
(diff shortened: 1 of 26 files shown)


franziskuskiefer:
What's the plan for this one now? Shall I review or do we cancel reviews and make it a draft until we know how to proceed?

raphaelrobert:
> What's the state of this @raphaelrobert ?

Now that the spec changed regarding extensions and their inclusion in the group context we should not continue this PR. There is some code that can be re-used for future PRs tackling the spec changes, so I set it to draft for now. Will close when the new PRs are ready.

PR #379, opened 2021-03-25
interop fixes

franziskuskiefer:
This pulls out some changes from the `interop` branch.

* finished encryption and transcript test vectors
* framing changes from #373
* updated and mlspp test vectors

We should maybe wait with landing until https://github.com/mlswg/mls-protocol/pull/466 got merged.

PR #407, opened 2021-05-18
Minor fixes to message formats and mlspp test vectors

### LINKED ISSUE #373: [MLS Spec change] message format changes (#466)
PR #466 to the MLS spec contains some changes to the MLS message format. Relevant for OpenMLS are:
* `confirmation_tag` in MLSPlaintextCommitAuthData is now `optional`
* `extension_data` in `Extension` is now up to `2^32-1` long

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/466/

kkohbrok:
This PR brings over a few changes from the interop branch. In particular, it adds mlspp test vectors, which get checked along with OpenMLS' own test vectors and adds a missing chunk of encryption test vector checking. Additionally, it fixes #373. It's essentially an updated version of #379 by @franziskuskiefer plus extending the size of the `extension_data` vector as per spec change.


src/group/mls_group/apply_commit.rs
```diff
@@ -145,11 +145,7 @@ impl MlsGroup {
         key_schedule.add_context(&provisional_group_context)?;
         let provisional_epoch_secrets = key_schedule.epoch_secrets(true)?;
 
-        let mls_plaintext_commit_auth_data =
-            match MlsPlaintextCommitAuthData::try_from(mls_plaintext) {
-                Ok(mpcad) => mpcad,
-                Err(_) => return Err(ApplyCommitError::ConfirmationTagMissing),
-            };
+        let mls_plaintext_commit_auth_data = MlsPlaintextCommitAuthData::from(mls_plaintext);
```
franziskuskiefer, line 148: I think this might have been why I kept it a `try_from`. We probably still want to fail here if the tag is missing in the `mls_plaintext` even if we can now just put the `None` in the `MlsPlaintextCommitAuthData` as well.

franziskuskiefer (changes requested):
lgtm except that I think I'd keep the `try_from` to catch missing tags. Maybe we should add a test case for this as well to make sure we fail properly. I don't think we excercise this right now and I'm not sure we handle it right.

src/group/mls_group/apply_commit.rs
kkohbrok: Good catch! We should definitely fail when there's no ConfirmationTag. I'll revert it to `try_from` and add a failure condition for when no tag is present. Also, a test, of course.

franziskuskiefer (approved):
Yay, more tests. Thanks!

src/group/tests/kat_transcripts.rs
```diff
@@ -190,13 +203,23 @@ pub fn run_test_vector(test_vector: TranscriptTestVector) -> Result<(), Transcri
         return Err(TranscriptTestVectorError::MembershipTagVerificationError);
     }
 
-    let my_confirmation_tag = confirmation_key.tag(&confirmed_transcript_hash_before);
+    //let my_confirmation_tag = confirmation_key.tag(&confirmed_transcript_hash_before);
+    let confirmed_transcript_hash_after =
+        hex_to_bytes(&test_vector.confirmed_transcript_hash_after);
```
raphaelrobert, line 208: This seems to trigger a clippy warning above (L181), because `confirmed_transcript_hash_before` doesn't need to be cloned anymore.

raphaelrobert (changes requested):
Thanks, LGTM! (just one nit)

raphaelrobert (approved):

PR #409, opened 2021-05-18
Fix message processing in the managed testing framework

kkohbrok:
This PR fixes a bug, where the managed testing framework would try to make a client process a message after it was already removed.

It also allows it to detect error events and throw an error if one occurs.


kkohbrok:
Closing in favor of #410.

PR #410, opened 2021-05-19
Fix message processing in the managed testing framework

kkohbrok:
This PR fixes a bug, where the managed testing framework would try to make a client process a message after it was already removed.

It also allows it to detect error events and throw an error if one occurs.


tests/utils/managed_utils/client.rs
```diff
@@ -104,7 +104,20 @@ impl Client {
         let group_state = group_states
             .get_mut(&group_id)
             .ok_or(ClientError::NoMatchingGroup)?;
-        group_state.process_message(message.clone())?;
+        // Prevent feeding further messages to client after it was removed
+        // by one of the messages.
+        if !group_state.is_active() {
+            return Ok(());
```
raphaelrobert, line 110: Don't you want to return an error message when the group is no longer active?

franziskuskiefer (dismissed):

tests/utils/managed_utils/client.rs
kkohbrok: Come to think about it, this check should no longer be necessary. After removing it, the group will throw an error when it's inactive and we're calling `process_message`. And you're right, that's how it should be.

raphaelrobert (approved):

PR #414, opened 2021-05-21
Add ratchet tree extension to managed group

raphaelrobert:
This PR adds the ratchet tree extension to the  `ManagedGroup` config and adds an integration test.

This is likely to be rewritten once we tackle extension for good, but for now it just aligns `ManagedGroup` with `MlsGroup` in that regard.

### DIFF
--- a/src/group/managed_group/config.rs
+++ b/src/group/managed_group/config.rs
@@ -22,6 +22,8 @@ pub struct ManagedGroupConfig {
     pub(crate) padding_size: usize,
     /// Number of resumtion secrets to keep
     pub(crate) number_of_resumption_secrets: usize,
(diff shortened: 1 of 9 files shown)


kkohbrok (approved):
Thanks for integrating this! That will enable us to simplify the managed group testing framework as well, where we currently pass around the tree manually.

franziskuskiefer (commented):
Can you file a follow-up to add a builder pattern or some other constructor for the `GroupConfig`? This is getting ugly.

raphaelrobert:
> Can you file a follow-up to add a builder pattern or some other constructor for the `GroupConfig`? This is getting ugly.

Good point! #415

franziskuskiefer (approved):

PR #418, opened 2021-06-02
signature refactor

franziskuskiefer:
In preparation for the new key store I cleaned up the way signing and verification with credentials works.
With this PR the `signable` module is the only place where `sign` and `verify` on credentials is called. This will make it easier to move around credentials and in particular the underlying keys.

To achieve this signed structs needed a little refactor. Notably, `KeyPackage`/`KeyPackageBundle` can't be modified anymore but modifying requires to convert them to unsigned representations. Signing gets you back the `KeyPackage`/`KeyPackageBundle`.

Another notable and somewhat unrelated change in this PR is that comparing `KeyPackage` now ignores the signature. This is necessary because the key package might be signed twice such that a different signature is attached even though the key package is the same. This is probably the better way to define `PartialEq` for key packages and necessary for KAT tests that re-sign key packages.


src/ciphersuite/signable.rs
```diff
@@ -1,29 +1,47 @@
...
+    {
+        let payload = self.unsigned_payload()?;
+        let signature = credential_bundle
+            .sign(&payload)
+            .map_err(|_| CredentialError::SignatureError)?;
+        Ok(Self::SignedOutput::from_payload(self, signature))
     }
+}
+
+pub trait Verifiable {
```
kkohbrok, line 35: Maybe add some documentation on the idea behind this trait? You mention later on that Signable and Verifiable shouldn't be implemented by the same struct. Maybe that would be worth mentioning here.

src/framing/plaintext.rs
```diff
@@ -179,7 +185,7 @@ impl MlsPlaintext {
         credential: &Credential,
     ) -> Result<(), VerificationError> {
         let tbs_payload = MlsPlaintextTbs::new_from(&self, Some(serialized_context));
-        tbs_payload.verify(credential, &self.signature)
+        tbs_payload.verify(credential).map_err(|e| e.into())
```
kkohbrok, line 188: Nitpick: Maybe make this into `...?;` followed by `Ok()` just to be consistent in style? Not that we're really being consistent anywhere else. Feel free to ignore, I guess...

src/key_packages/test_key_packages.rs
```diff
@@ -70,24 +70,21 @@ fn key_package_id_extension() {
...
         assert!(kpb.key_package().verify().is_ok());
+        let mut kpb = kpb.unsigned();
 
         // Add an ID to the key package.
         let id = [1, 2, 3, 4];
-        kpb.key_package_mut()
-            .add_extension(Box::new(KeyIdExtension::new(&id)));
-
-        // This is invalid now.
-        assert!(kpb.key_package().verify().is_err());
```
kkohbrok, line 87: Nice, this isn't even possible anymore 😄

src/tree/mod.rs
```diff
@@ -413,72 +413,70 @@ impl RatchetTree {
...
         // Generate new keypair
         let own_index = self.own_node_index();
+        let own_key_package = self.own_key_package();
 
         // Replace the init key in the current KeyPackage
-        let mut key_package_bundle =
-            KeyPackageBundle::from_rekeyed_key_package(self.own_key_package());
+        let key_package_bundle_unsigned =
+            KeyPackageBundlePayload::from_rekeyed_key_package(own_key_package);
+        // FIXME: THIS IS UNNECESSARY
```
kkohbrok, line 424: Is this because we sign the KPB again a few lines below? I guess this is due to a restriction that exists because of the split types for KP and KPB and would require some more refactoring to go away. Is that correct? If so, we should probably file an issue.

kkohbrok (changes requested):
Thanks for making these changes and great use of the type checker to get some extra guarantees! Just a few nits from my side.

kkohbrok (dismissed):
Thanks for addressing the comments! Looks good to me now.

src/framing/errors.rs
```diff
@@ -9,10 +9,17 @@ use crate::tree::secret_tree::SecretTreeError;
...
+        Simple {
+            NotAnApplicationMessage = "The MlsPlaintext message is not an application message.",
+            UnknownSender = "Sender is not part of the group",
+            InvalidSignature = "The MlsPlaintext signature is invalid",
+            InvalidMembershipTag = "The MlsPlaintext membership tag is invalid",
+        }
+        Complex {
+            CodecError(CodecError) = "Codec error",
+            CredentialError(CredentialError) = "See [`CredentialError`](`crate::credentials::CredentialError`) for details.",
+            VerificationError(VerificationError) = "See [`VerificationError`](`VerificationError`) for details.",
```
raphaelrobert, line 21: This new error type seems untested

raphaelrobert (changes requested):
Thanks for the PR! The new logic looks rather straight forward, just a couple of nits inline.

kkohbrok (approved):

raphaelrobert (approved):

PR #422, opened 2021-06-08
Enforce MlsPlaintext signature verification

franziskuskiefer:
This PR enforces that MlsPlaintext signatures are always verified.
It is now impossible to construct an MlsPlaintext without signature.
When creating an MlsPlaintext appropriate functions must be used.
When receiving an MlsPlaintext a VerifiableMlsPlaintext must be created
instead, which can be converted to an MlsPlaintext by calling verify.

There are a number of follow-up issues that should ad the same safety
guarantees to
* membership tags
* confirmation tags
* key packages
* group info
* public group state

This should then also allow us to get rid of the verify_no_out function
in Verifiable.

This PR also adds tests for MlsPlaintext signature, membership tag, and confirmation tag verification.
Note that it also fixes a bug where the group state is modified when verifying an invalid confirmation tag. This is only a hot fix and should get fixed properly with the tree rewrite.


src/framing/test_framing.rs
```diff
@@ -399,3 +371,148 @@ fn confirmation_tag_presence() {
...
+    assert_eq!(decoded_commit.encode_detached().unwrap(), original_encoded_commit);
+
+    // Remove membership tag.
+    let good_membership_tag = commit.membership_tag.clone();
+    commit.membership_tag = None;
+    let membership_error = commit.verify_membership(
+        group_alice.context().serialized(),
+        group_alice.epoch_secrets().membership_key())
+        .err()
+        .expect("Membership verification should have returned an error");
```
kkohbrok, line 445: Nitpick: Any particular reason not to use `expect_err` here? Same below.

src/group/mls_group/apply_commit.rs
```diff
@@ -36,14 +36,16 @@ impl MlsGroup {
             ciphersuite,
             &commit.proposals,
             proposals_by_reference,
-            mls_plaintext.sender,
+            mls_plaintext.sender().clone(),
```
kkohbrok, line 39: What clippy says.

src/group/mls_group/create_commit.rs
```diff
@@ -67,23 +67,16 @@ impl MlsGroup {
...
-            authenticated_data: aad.to_vec(),
-            content_type: ContentType::from(&content),
-            content,
-            signature: Signature::new_empty(),
-            confirmation_tag: None,
-            membership_tag: None,
-        };
-
-        // Add signature and membership tag to the MlsPlaintext
+        // let content = MlsPlaintextContentType::Commit(commit);
```
kkohbrok, line 70: Remove obsolete code?

src/group/mls_group/apply_commit.rs
```diff
@@ -36,14 +36,16 @@ impl MlsGroup {
...
         ) {
             Ok(proposal_queue) => proposal_queue,
             Err(_) => return Err(ApplyCommitError::MissingProposal),
         };
 
         // Create provisional tree and apply proposals
         let mut provisional_tree = self.tree.borrow_mut();
+        // FIXME: this is a copy of the nodes in the tree to reset the original state.
```
kkohbrok, line 47: I know that this should be fixed with the tree rewrite. But maybe file an issue for this just to be sure?

src/group/mls_group/test_mls_group.rs
```diff
@@ -286,28 +286,29 @@ fn test_update_path() {
             path: Some(broken_path),
         };
 
-        let broken_commit_content = MlsPlaintextContentType::Commit(broken_commit);
+        // let broken_commit_content = MlsPlaintextContentType::Commit(broken_commit);
```
kkohbrok, line 289: Remove obsolete code.

src/group/tests/kat_transcripts.rs
```diff
@@ -36,6 +36,7 @@ pub struct TranscriptTestVector {
     tree_hash_before: String,
     confirmed_transcript_hash_before: String,
     interim_transcript_hash_before: String,
+    credential: String,
```
kkohbrok, line 39: I assume this is included so that the signature on the commit can be checked as well? Then we should definitely file an issue to re-include the mlspp test vectors and probably also to change the test vector definitions.

src/group/tests/kat_transcripts.rs
```diff
@@ -275,16 +279,17 @@ fn read_test_vectors() {
...
-    let tv_files = [
-        "test_vectors/mlspp/mlspp_transcript_1.json",
-        "test_vectors/mlspp/mlspp_transcript_2.json",
-        "test_vectors/mlspp/mlspp_transcript_3.json",
-    ];
-    for &tv_file in tv_files.iter() {
-        let tv: TranscriptTestVector = read(tv_file);
-        run_test_vector(tv).expect("Error while checking key schedule test vector.");
-    }
+    // FIXME: change test vector spec
```
kkohbrok, line 282: As mentioned above, we should probably file an issue for this.

src/messages/proposals.rs
```diff
@@ -172,16 +172,16 @@ impl<'a> QueuedProposal<'a> {
...
+        let proposal = match mls_plaintext.content() {
             MlsPlaintextContentType::Proposal(p) => p,
             _ => return Err(QueuedProposalError::WrongContentType),
         };
         let proposal_reference = ProposalReference::from_proposal(ciphersuite, &proposal);
         Ok(Self {
             proposal,
             proposal_reference,
-            sender: mls_plaintext.sender,
+            sender: mls_plaintext.sender().clone(),
```
kkohbrok, line 184: What clippy says.

src/tree/mod.rs
```diff
@@ -49,6 +49,7 @@ pub struct RatchetTree {
     /// All nodes in the tree.
     /// Note that these only hold public values.
     /// Private values are stored in the `private_tree`.
+    /// FIXME: this must not be public
```
kkohbrok, line 52: This will also be fixed in the tree rewrite. I just created #423, so maybe we can just mention that in place of the existing FIXMEs.

kkohbrok (changes requested):
Thanks for this overhaul of the way we do signed payloads and general cleanup effort. Again a very cool change to leverage the type system for extra safety! A few notes, but nothing substantial.

src/framing/ciphertext.rs
```diff
@@ -179,12 +182,14 @@ impl MlsCiphertext {
...
-        log::trace!(
+        log_content!(
+            trace,
             "  Successfully decoded MlsPlaintext with: {:x?}",
             mls_ciphertext_content.content
         );
+
         // Return the MlsPlaintext
-        Ok(MlsPlaintext {
+        let plaintext = MlsPlaintext {
```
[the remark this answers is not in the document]
franziskuskiefer: So right now it is impossible to do this outside of the framing module. It is still nice to do this for unit tests and in some other cases.
If we want to disallow it here we have to move the `plaintext` and `ciphertext` modules further apart (the struct fields are `pub(super)`). But you are right we should do this.
I locked this down further and made the `MlsPlaintext` members all private.

src/ciphersuite/signable.rs
```diff
@@ -45,18 +52,31 @@ pub trait Signable: Sized {
...
 /// `unsigned_payload` function.
-pub trait Verifiable {
+pub trait Verifiable: Sized {
     /// Return the unsigned, serialized payload that should be verified.
     fn unsigned_payload(&self) -> Result<Vec<u8>, CodecError>;
 
     /// A reference to the signature to be verified.
     fn signature(&self) -> &Signature;
 
+    /// Verifies the payload against the given `credential` and `signature`.
```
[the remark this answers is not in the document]
franziskuskiefer: Good point. updated the comment.

src/framing/plaintext.rs
```diff
@@ -30,145 +53,159 @@ use std::convert::TryFrom;
...
+    pub(super) confirmation_tag: Option<ConfirmationTag>,
+    pub(super) membership_tag: Option<MembershipTag>,
 }
 
 impl MlsPlaintext {
-    /// This constructor builds a new `MlsPlaintext` from the parameters
-    /// provided. It is only used internally.
-    pub(crate) fn new_from_member(
+    /// Convenience function for creating an `MlsPlaintext`.
+    #[inline]
```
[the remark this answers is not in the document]
kkohbrok: Thanks for the pointer!

kkohbrok (approved):
Looks good to me! Thanks for addressing my comments and questions!

openmls/src/framing/plaintext.rs
```diff
@@ -30,145 +53,200 @@ use std::convert::TryFrom;
...
+            context,
+        )?;
+        mls_plaintext.set_membership_tag(context.serialized(), membership_key)?;
         Ok(mls_plaintext)
     }
 
     /// This constructor builds an `MlsPlaintext` containing a Proposal.
     /// The sender type is always `SenderType::Member`.
-    pub fn new_from_proposal_member(
+    pub fn new_proposal(
```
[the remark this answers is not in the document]
raphaelrobert: For context: external proposals won't have the membership tag.

openmls/src/codec/mod.rs
```diff
@@ -69,28 +69,50 @@ impl<'a> Cursor {
...
     fn decode_detached(bytes: &[u8]) -> Result<Self, CodecError> {
         let cursor = &mut Cursor::new(bytes);
         Self::decode(cursor)
     }
 }
 
-impl Codec for u8 {
+pub trait Codec: Encode + Decode {}
+
+macro_rules! implement_codec {
```
raphaelrobert, line 93: Is the macro really needed here? Can't we solve this with only using traits?

raphaelrobert (changes requested):
Great PR, thanks! Just a few nits inline. The PR also contains quite a few unrelated changes (but I don't really mind).

raphaelrobert (approved):

openmls/src/codec/mod.rs
franziskuskiefer: This is a common macro use to safe writing out some code manually. As discussed I'll leave it in because it will go away soon anyway with the use of `tls_codec`.

openmls/src/framing/plaintext.rs
```diff
@@ -30,145 +53,200 @@ use std::convert::TryFrom;
...
+        self.membership_tag = Some(tag);
+    }
 }
 
 impl MlsPlaintext {
-    /// This constructor builds a new `MlsPlaintext` from the parameters
-    /// provided. It is only used internally.
-    pub(crate) fn new_from_member(
+    /// Convenience function for creating an `MlsPlaintext`.
+    #[inline]
```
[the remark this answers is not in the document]
franziskuskiefer: This isn't forcing anything (you can't actually force the compiler to inline). It's just a hint that it would make sense to inline this because of the way it's used.

openmls/src/framing/plaintext.rs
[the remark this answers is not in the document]
franziskuskiefer: Right, as discussed, I deleted code that wasn't used or tested. As for the API naming, I consider this the default case and the external proposal will get a separate name.

PR #431, opened 2021-06-21
Make naming in MlsGroup API consistent

kkohbrok:
In an effort to make the naming around the MlsGroup API a little more consistent and for general cleanup around the API, this PR does the following 
* rename `GroupConfig` to `MlsGroupConfig`
* rename `GroupError` to `MlsGroupError`
* make all MlsGroup API functions return an `MlsGroupError` rather than an `MlsCiphertextError` or `MlsPlaintextError`.


raphaelrobert (approved):
Thanks, LGTM!

openmls/src/group/mod.rs
```diff
@@ -80,22 +80,22 @@ impl GroupContext {
 
 /// Configuration for an MLS group.
 #[derive(Clone, Copy, Debug)]
-pub struct GroupConfig {
+pub struct MlsGroupConfig {
```
franziskuskiefer, line 83: I'm generally not keen on using the `Mls` prefix everywhere. But at least it's consistent this way..

franziskuskiefer (approved):

openmls/src/group/mod.rs
kkohbrok: I'm not attached to it, but if we want to change it, we should do it soon, before even more things depend on that API. I'll merge for now, but again, I'm not opposed to a different naming scheme.

PR #432, opened 2021-06-24
Move managed test utils into their own crate

kkohbrok:
This PR moves the testing framework for the managed API into its own crate. This is in preparation of merging the `interop_v2` into main.

Concretely, this PR moves the relevant files into a new crate outside of OpenMLS and does some minor renaming in the crate documentation.


franziskuskiefer (approved):
lgtm

PR #434, opened 2021-07-01
Generate TreeKEM kat externally

kkohbrok:
This PR introduces a new module, which is only exposed for kat generation and which exposes the necessary API for that purpose. It also moves the gneration and checking of the TreeKEM kat to an external crate.


kkohbrok:
This PR is obsolete, as we can now generate the TreeKEM kat internally (as opposed to as part of the integration tests).

PR #435, opened 2021-07-02
Refactor KeyPackageBundles to store path instead of leaf secrets

kkohbrok:
This PR refactors KeyPackageBundles to store the path secret derived from the leaf secret instead of the leaf secret itself.

It also refactors the constructors for the private tree, which now rely on a KeyPackageBundle, as well as the test vector generation specific function of the KeyStore, which allows us to obtain the leaf secret which we still need to generate treekem test vectors.

TODO: 
Evaluate if we can make the `PathSecret` in the KeyPackageBundle optional so that we can remove it for processing later. Alternatively derive `Clone` for PathSecret so that it can be stored in the PrivateTree independent from the KeyPackageBundle.

### DIFF
--- a/openmls/src/ciphersuite/mod.rs
+++ b/openmls/src/ciphersuite/mod.rs
@@ -403,11 +403,6 @@ impl Secret {
     pub(crate) fn ciphersuite(&self) -> &'static Ciphersuite {
         self.ciphersuite
     }
-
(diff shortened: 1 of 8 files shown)


kkohbrok:
This PR has gone stale and I think the underlying problem was solved in another way. Closing for now.

PR #439, opened 2021-07-09
Pull testing framework into OpenMLS crate

kkohbrok:
This PR pulls the testing framework from the `test_utils` crate into the `test_utils` module inside the OpenMLS crate. It also moves the tree kem test vector generation to the same place where tree kem test vectors are tested. This will allow us to move the interop client to main and start doing interop from the main branch


openmls/src/lib.rs
```diff
@@ -31,7 +31,7 @@ mod codec;
 
 #[cfg(any(feature = "expose-test-vectors", test))]
 #[macro_use]
-pub mod test_util;
+pub mod test_utils;
```
franziskuskiefer, line 34: Can you change the feature to something like `test-utils`?

openmls/src/test_utils/mod.rs
```diff
@@ -7,6 +7,8 @@ use std::{
     io::{BufReader, Write},
 };
 
+pub mod test_framework;
+
 pub fn write(file_name: &str, obj: impl Serialize) {
```
franziskuskiefer, line 12: A  little unrelated, but can you check that all these functions in here need to be public?

franziskuskiefer (changes requested):
lgtm with the small nit

openmls/src/lib.rs
kkohbrok: Done.

openmls/src/test_utils/mod.rs
kkohbrok: Done. All of these functions are crate-public now.

openmls/Cargo.toml
```diff
@@ -27,6 +27,7 @@ itertools = {version = "0.10", optional = true}
 default = ["rust-crypto"]
 rust-crypto = ["evercrypt/rust-crypto-aes"]
 expose-test-vectors = ["itertools"]
+test_utils = []
```
franziskuskiefer, line 30: There's no clear guideline on this yet ([see for example this discussion](https://github.com/rust-lang/api-guidelines/discussions/101)). But all other features are kebab case. So we should probably stick to it

openmls/src/lib.rs
```diff
@@ -29,7 +29,7 @@ pub mod error;
 #[macro_use]
 mod codec;
 
-#[cfg(any(feature = "expose-test-vectors", test))]
+#[cfg(any(feature = "test-utils", feature = "expose-test-vectors", test))]
```
franziskuskiefer, line 32: you even use kebab case here 😉

franziskuskiefer (changes requested):

openmls/Cargo.toml
kkohbrok: Right you are. That was really just a typo.

openmls/Cargo.toml
```diff
@@ -27,6 +27,7 @@ itertools = {version = "0.10", optional = true}
 default = ["rust-crypto"]
 rust-crypto = ["evercrypt/rust-crypto-aes"]
 expose-test-vectors = ["itertools"]
```
franziskuskiefer, line 29: Do you want to keep both feature flags? I thought `test-utils` could expose the test vectors as well.

kkohbrok: I thought that was what you were suggesting. I have no preference either way. Then I would propose we only keep the `expose-test-vector` for now or at least until we need to expose only the test_utils.

kkohbrok: Same as above, only that I actually changed the name everywhere to `test-utils`, so there's no more `expose-test-vectors`.

franziskuskiefer (approved):

raphaelrobert (approved):
Looks straight forward and relying on @franziskuskiefer for the nits.

PR #443, opened 2021-07-12
Fix possible circumvention of verification type structure

### LINKED ISSUE #437: [Bug] MlsPlaintext can be built from VerifiableMlsPlaintext without verification
**Describe the bug**
The `from_verifiable` constructor of `MlsPlaintext` is not hidden and can be used to create an `MlsPlaintext` from a `VerifiableMlsPlaintext` without actually verifying the authenticity.

**Expected behavior**
The user should be forced to actually call `verify` on a `VerifiableMlsPlaintext` to create an `MlsPlaintext` object.

kkohbrok:
Fixes #437.


franziskuskiefer (commented):
Thanks! This certainly works.
But did you think about the other direction we talked about? (Having an `into_verified` instead of this `from_verifiable`.)

kkohbrok:
> Thanks! This certainly works.
> But did you think about the other direction we talked about? (Having an `into_verified` instead of this `from_verifiable`.)

Yes, I thought about this, but I don't see how this wouldn't allow the same "attack". My take was that we have three parties involved:
- the "creator" of the trait
- the implementer of the trait
- and the user of the trait implementation

The goal is for the first two to "cooperate" in such a way that the user of the trait implementation _has_ to use `verify` to create the `VerifiedStruct`. Since the creator of the trait doesn't know how to construct the object, there has to be some constructor implemented by the implementer of the trait. I didn't quite see how it matters if it's `from_verifiable` or `into_verified`. I think we can have either, but we have to have some sort of "access control" mechanism as the one I have implemented regardless of what way the constructor goes. Or did I miss something there?

franziskuskiefer (approved):
Yeah I think I agree that the other options aren't much better... Let's do this.
Please also get these changes on the interop_v2 branch. This is a breaking change for the branch.

kkohbrok:
> Yeah I think I agree that the other options aren't much better... Let's do this.
> Please also get these changes on the interop_v2 branch. This is a breaking change for the branch.

Good point. After this is merged, my next task will be bringing the interop_v2 branch to `main`, so I"ll have to deal with this anyway.

raphaelrobert (approved):

PR #444, opened 2021-07-13
Enable Managed API to process VerifiableMlsPlaintext

kkohbrok:
This PR enables the Managed API to process VerfiableMlsPlaintext objects, which is required, as it's not possible (and not meant to be) to deserialize `MlsPlaintext` directly. This PR creates a new enum `MlsMessageIn` which contains either an unverified plaintext or a ciphertext. The ManageAPI then decrypts the ciphertext (decryption already deals with verification of the contained ciphertext) or verifies the plaintext (this step was missing) and continues processing.

This is a temporary fix until verification is dealt with properly as per #133.


kkohbrok:
> Is this really specific to inbound messages? It sounds like you want to introduce a distinction between inbound and outbound messages. This might make sense, I just want to know if that's really the intention and also what the motivation is.

Since @franziskuskiefer made the verification of signatures type-safe, you can't de-serialize a payload directly into an `MlsPlaintext`, but instead you have to de-serialize into a `VerifiableMlsPlaintext`, which you can then verify. As of now, the verification takes place behind the Managed API (although this might change in the future), so the input messages have to be assumed to be unverified, thus the change. Meanwhile the output messages don't have to be verified (except in the context of the test framework, where we don't serialize and de-serialize betwen clients), since they originate with the locally running client (which we trust).

franziskuskiefer (changes requested):
I'm not sure I fully understand this PR. While something like this certainly is necessary the code here should have either broken or be unusable. Since neither has been the case I suspect that it's simply not used (properly)? Then the question would be why the code is even there. It then either needs tests and usage examples or should be removed.

> This is a temporary fix until verification is dealt with properly as per #133.

Depending on the answer of my questions above we might not need a temporary "fix". (Also, if this is fixing something, there should have been something broken and we need corresponding tests.)

kkohbrok:
> I'm not sure I fully understand this PR. While something like this certainly is necessary the code here should have either broken or be unusable. Since neither has been the case I suspect that it's simply not used (properly)? Then the question would be why the code is even there. It then either needs tests and usage examples or should be removed.

My understanding is that we don't have a real consumer of the Managed API that serializes and de-serializes messages. Otherwise this problem would have arisen when we removed the ability to de-serialize bytes into MlsPlaintext.

We have the testing framework, which uses the Managed API, but that doesn't serialize and de-serialize and the cli uses the MlsGroup API, which only takes MlsPlaintexts as input. The difference here is that Managed API wants to allow both ciphertexts and plaintexts as inputs, which means that either all verification has to performed by the API or only ciphertexts are verified by the API, while plaintexts have to be verified by the consumer. The core difference here being that the Managed API deals with decryption, while the MlsGroup API makes the consumer do that.

To conclude: We have several options:
- split verification to outside for plaintexts and inside for ciphertexts, which would allow us to keep the `MlsMessage` enum as-is.
- only take plaintexts as input and enable the consumer to deal with decryption themselves, which in turn enables them to verify both plaintexts _and_ (after decryption) ciphertexts

I'm not a huge fan of the first option, because it seems inconsistent and, after all, the goal of the Managed API is to deal with this sort of stuff for the consumer as much as possible.

franziskuskiefer:
> My understanding is that we don't have a real consumer of the Managed API that serializes and de-serializes messages.

Exactly. That's why it must get a consumer. Only then we can talk about the right way of "fixing" this.

kkohbrok:
> > My understanding is that we don't have a real consumer of the Managed API that serializes and de-serializes messages.
> 
> Exactly. That's why it must get a consumer. Only then we can talk about the right way of "fixing" this.

The consumer would be the interop client, and trying to merge the interop client into main is exactly why I stumbled across this problem and filed this PR 😄

Should I make this PR part of the PR that merges the interop client into main?

Alternatively, I can start serializing and de-serializing in the testing framework.

franziskuskiefer:
> Alternatively, I can start serializing and de-serializing in the testing framework.

This would certainly be nicer, because it tests more. Having it configurable would be even nicer to be able to switch it off for debugging. But it's not really necessary.

kkohbrok:
I essentially copied some of the code of `DsMlsMessage` from `ds-lib` to the test framework, because it seems like we need this `MessageType` in both places. The use is slightly different in both places and I don't see this flag as part of OpenMLS, so I copied the relevant parts of the code for now.

openmls/src/framing/plaintext.rs
```diff
@@ -335,6 +335,18 @@ impl MlsPlaintext {
     pub fn authenticated_data(&self) -> &[u8] {
         self.authenticated_data.as_slice()
     }
+
+    #[cfg(test)]
+    pub(crate) fn tamper_with_signature(&mut self) {
```
franziskuskiefer, line 340: maybe something like invalidate?

```suggestion
    pub(crate) fn invalidate_signature(&mut self) {
```

franziskuskiefer: If this exists here, `signature_mut` shouldn't be necessary.

openmls/src/group/managed_group/test_managed_group.rs
```diff
@@ -268,3 +271,85 @@ ctest_ciphersuites!(export_secret, test(ciphersuite_name: CiphersuiteName) {
...
+        .expect("error creating self update");
+
+    drop(client);
+    drop(clients);
+
+    // Tamper with the message such that signature verification fails
+    let mut msg_invalid_signature = mls_message.clone();
+    match &mut msg_invalid_signature {
+        MlsMessage::Plaintext(pt) => pt.tamper_with_signature(),
+        MlsMessage::Ciphertext(_) => panic!("This should be a plaintext!"),
```
franziskuskiefer, line 321: You don't get a codecov warning if you use `if let` instead of the `match`.

openmls/src/test_utils/test_framework/messages.rs
```diff
@@ -0,0 +1,63 @@
+//! This module contains code to convert messages between the
+//! `MlsMessage`/`MlsMessageIn` formats as used by the Managed API, which the
+//! Clients are built on and the `DsMlsMessage`, which includes the message type
```
franziskuskiefer, line 3: c&p ? `DsMlsMessage`

openmls/src/test_utils/test_framework/messages.rs
```diff
@@ -0,0 +1,63 @@
...
+#[repr(u8)]
+pub enum MessageType {
+    /// An MlsCiphertext message.
+    MlsCiphertext = 0,
+
+    /// An MlsPlaintext message.
+    MlsPlaintext = 1,
+}
+
+impl<'a> tls_codec::Size for MlsMessageIn<'a> {
```
franziskuskiefer, line 24: Unrelated but if you need something to do and want to poke around the tls codec code; This pattern of implementing (de)serialization of enums could be put into the library and would safe us quite some code.

openmls/src/test_utils/test_framework/mod.rs
```diff
@@ -168,6 +174,15 @@ impl ManagedTestSetup {
...
     /// client by `get_fresh_key_package`.
     pub fn deliver_welcome(&self, welcome: Welcome, group: &Group) -> Result<(), SetupError> {
+        // Serialize and de-serialize the Welcome if the bit was set.
+        if self.use_codec {
+            let serialized_welcome = welcome
+                .tls_serialize_detached()
+                .map_err(|e| ClientError::TlsCodecError(e))?;
+            let deserialized_welcome = Welcome::tls_deserialize(&mut serialized_welcome.as_slice())
+                .map_err(|e| ClientError::TlsCodecError(e))?;
+            assert_eq!(deserialized_welcome, welcome);
```
franziskuskiefer, line 184: Like in the other case you could use the `deserialized_welcome` going forward instead of comparing the two. But both is fine.

openmls/src/test_utils/test_framework/mod.rs
```diff
@@ -81,6 +83,9 @@ pub struct ManagedTestSetup {
     // This maps key package hashes to client ids.
     pub waiting_for_welcome: RefCell<HashMap<Vec<u8>, Vec<u8>>>,
     pub default_mgc: ManagedGroupConfig,
+    /// Flag to indicate if messages should be serialized and de-serialized as
+    /// part of message distribution
+    pub use_codec: bool,
```
franziskuskiefer, line 88: Can you make this an enum so that we don't have to use `true` and `false` when setting up the config?

franziskuskiefer (changes requested):
thanks, now with tests 🙂 
But there are a couple things we can do even better

kkohbrok (commented):
Thanks for the review! I believe I've addressed all comments.

kkohbrok:
The last failing test on this one should go away after merging #451. (Or just re-running the test, but that would be cheating!)

openmls/src/group/managed_group/mod.rs
```diff
@@ -1086,7 +1099,27 @@ impl ManagedGroup {
     }
 }
 
-/// Unified message type
+/// Unified message type for input to the managed API
+#[derive(Debug, Clone)]
+pub enum MlsMessageIn<'a> {
```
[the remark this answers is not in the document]
franziskuskiefer: I see you made it `MlsMessageOut`, that's fine with me

openmls/src/group/managed_group/mod.rs
```diff
@@ -1086,50 +1105,82 @@ impl ManagedGroup {
...
         }
     }
 
     /// Returns `true` if this is a handshake message and `false` otherwise.
     pub fn is_handshake_message(&self) -> bool {
         match self {
-            MlsMessage::Ciphertext(m) => m.is_handshake_message(),
-            MlsMessage::Plaintext(m) => m.is_handshake_message(),
+            MlsMessageOut::Ciphertext(m) => m.is_handshake_message(),
+            MlsMessageOut::Plaintext(m) => m.is_handshake_message(),
```
franziskuskiefer, line 1171: So this is never called with a plaintext according to codecov? If so, we should add a test case.

openmls/src/group/managed_group/test_managed_group.rs
```diff
@@ -268,3 +273,88 @@ ctest_ciphersuites!(export_secret, test(ciphersuite_name: CiphersuiteName) {
...
+        .members
+        .iter()
+        .find(|(index, _)| index == &0)
+        .unwrap()
+        .clone();
+
+    let clients = setup.clients.borrow();
+    let client = clients.get(client_id).unwrap().borrow();
+
+    let (mls_message, _welcome_option) = client
```
franziskuskiefer, line 316: I'm not saying it's nicer. But something like this would allow to avoid the `drop`s below.
```suggestion
    let (mls_message, _welcome_option) = setup.clients.borrow().get(client_id).unwrap().borrow()
```

openmls/src/group/managed_group/test_managed_group.rs
```diff
@@ -268,3 +273,88 @@ ctest_ciphersuites!(export_secret, test(ciphersuite_name: CiphersuiteName) {
...
+    );
+
+    // Tamper with the message such that sender lookup fails
+    let mut msg_invalid_sender = mls_message.clone();
+    match &mut msg_invalid_sender {
+        MlsMessageOut::Plaintext(pt) => pt.set_sender(Sender {
+            sender_type: pt.sender().sender_type,
+            sender: LeafIndex::from(group.members.len() + 1),
+        }),
+        MlsMessageOut::Ciphertext(_) => panic!("This should be a plaintext!"),
```
franziskuskiefer, line 347: `if let` would allow you to get rid of the codecov warning. Otherwise we should make sure we actually end up in this match arm.

franziskuskiefer (approved):
lgtm with some nits that you don't have to address.

raphaelrobert (approved):
Thanks for the PR! LGTM!

PR #445, opened 2021-07-13
Truncate tree after members were removed

### LINKED ISSUE #420: [MLS Spec change]  Truncate tree size on removal (#461)
Trim the tree after removal of a group member.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/461

kkohbrok:
Fixes #420.


openmls/src/tree/mod.rs
```diff
@@ -776,6 +776,11 @@ impl RatchetTree {
         // Determine if Commit needs a path field
         let path_required = has_updates || has_removes || !has_adds;
 
+        // If members were removed, truncate the tree.
+        if has_removes {
+            self.trim_tree()
```
franziskuskiefer, line 781: This needs a test.

franziskuskiefer (changes requested):

openmls/src/tree/mod.rs
kkohbrok: Ugh. I really should have thought of that. Fixing.

kkohbrok: I really blame codecov. It was so green I didn't even think of adding any tests 😅

openmls/src/tree/tests/mod.rs
```diff
@@ -20,6 +20,8 @@ mod test_resolution;
 #[cfg(test)]
 mod test_secret_tree;
 #[cfg(test)]
+mod test_tree_truncation;
+#[cfg(test)]
```
franziskuskiefer, line 24: ugh, these `cfg` need to get cleaned up. This should get split into a `test-utils` and `test` module. Please file a follow up

openmls/src/tree/tests/test_tree_truncation.rs
```diff
@@ -0,0 +1,46 @@
...
+        .clone();
+
+    // Remove the rightmost 2 members in the tree
+    setup
+        .remove_clients_by_index(ActionType::Commit, group, &remover_id, &[6, 7])
+        .expect("error while removing members from group");
+
+    // Test if the tree was truncated. The tree's size should be ((number of
+    // members) * 2) - 1, i.e. 11 with 6 members.
+    assert_eq!(group.public_tree.len(), 11)
```
franziskuskiefer, line 45: We also need to test the other case where no truncation happens, for example removing only one member in this scenario.

franziskuskiefer (changes requested):
Thanks for the test! Can I get another one? 🙂

openmls/src/tree/tests/test_tree_truncation.rs
kkohbrok: Wouldn't this have to be a test for the `trim` function we use as part of the `remove` process? As a consumer of the `trim` function, I rely on it to do what it says.

Don't get me wrong, I'm happy to implement another test, either for the `remove` operation, or for the `trim` function. I'm just trying to understand what's the right approach here.

kkohbrok: It doesn't hurt to have tests for _both_, of course, but shouldn't the general approach be that I don't have to test for correct behaviour of all underlying functions I'm using?

franziskuskiefer: The code you add is a
```
if remove {
 trim
}
```
So the tests to cover this must cover
* `if remove` is `true`
  * `trim` has an effect
  * `trim` has no effect
* `if remove` is `false`

That `trim` itself isn't actually tested is another issue. But that's why it's even more important to test properly when it's used.

kkohbrok: Ah, then I misunderstood. I thought you wanted to test if trim has an effect if we remove one member, but there's nothing to trim. That sounds sensible.

kkohbrok: Although I'm not sure how to test this actually, because trim will really only have an effect if someone is removed at the right-most leaf of the tree. Consequently, in all other cases calling trim will have no effect, so we can't effectively test whether trim is run or not. We could of course remove the `if` statement as a consequence, but then again the spec says to only trim after removal 🤷🏻

franziskuskiefer: > in all other cases calling trim will have no effect

there you have your options for test cases

openmls/src/tree/tests_and_kats/unit_tests/test_tree_truncation.rs
```diff
@@ -0,0 +1,146 @@
...
+        // We build a leaf node from the key packages
+        let leaf_node = Node {
+            node_type: NodeType::Leaf,
+            key_package: Some(key_package_bundle.key_package().clone()),
+            node: None,
+        };
+        key_package_bundles.push(key_package_bundle);
+        nodes.push(Some(leaf_node));
+        // We skip the last parent node (trees should always end with a leaf node)
+        if i != 9 {
```
franziskuskiefer, line 36: It's only a test. But this is dangerous. If the range of the loop is change, this breaks.

franziskuskiefer: Instead of hard-coding values for the tree size etc. It would be better to have it all variable and then also run this for a bunch of tree sizes. It's probably not necessary to cover all of `trim_tree` but it makes sure the test doesn't easily break.

openmls/src/tree/tests_and_kats/unit_tests/test_tree_truncation.rs
```diff
@@ -0,0 +1,146 @@
...
+        .expect("error while removing members from group");
+
+    // Test if the tree was truncated. The tree's size should be ((number of
+    // members) * 2) - 1, i.e. 11 with 6 members.
+    assert_eq!(group.public_tree.len(), 11)
+}
+
+#[test]
+fn test_truncation_after_update() {
+    // Set up a group with 8 members.
```
franziskuskiefer, line 109: The setup code is all the same. Please factor it out and reuse it.

franziskuskiefer (changes requested):
Thanks. A couple nits in the test code.

franziskuskiefer (approved):
:shipit:

PR #449, opened 2021-07-14
Re-organize tests and kats

### LINKED ISSUE #448: Better organize test-utils vs. test cfg flags in test modules
Currently, many modules have a test module that has a test-or-test-utils cfg flag. In turn, that module needs to qualify all of its files as either one or the other. A solution could be to have two modules, one for test-utils and one for test.

kkohbrok:
Fixes #448.


openmls/src/ciphersuite/tests/mod.rs
```diff
@@ -0,0 +1,4 @@
+//! Unit tests for the ciphersuites.
```
franziskuskiefer, line 1: To make it easier to find files I generally prefer not to use too many `mod.rs`. What do you think about putting this in `ciphersuite/tests.rs` instead? (Similarly for all the others.)

I didn't touch any of this sort of re-org yet, but while you're at it we can at least change the ones we touch.

franziskuskiefer (changes requested):
Thanks, looking much better already.
One thought inline. But if you strongly prefer this layout I can live with it.

openmls/src/ciphersuite/tests/mod.rs
kkohbrok: I had thought about this before, my editor helpfully puts the directory behind the `mod.rs`, so it didn't impact me too much. But sure, let's start here and change the file names as we go from now on.

kkohbrok: Although the `mod.rs` approach does have the added benefit of making the relevant "mod" file easily identifiable in a given directory. I guess I don't have a strong opinion either way.

franziskuskiefer (approved):
:shipit:

raphaelrobert (approved):

PR #451, opened 2021-08-05
Fix application of exclusion list in the computation of resolution

kkohbrok:
Previously, we didn't take the exclusion list into account when adding leaf nodes to the resolution based on full parent nodes. This PR fixes this issue and adds a test for that scenario.

Thanks @raphaelrobert for helping me find and squash this bug!


openmls/src/tree/tests_and_kats/unit_tests/test_resolution.rs
```diff
@@ -178,3 +182,61 @@ fn test_original_child_resolution() {
...
+    let setup = ManagedTestSetup::new(managed_group_config.clone(), number_of_clients);
+
+    let group_id = setup.create_group(Ciphersuite::default()).unwrap();
+
+    let mut groups = setup.groups.borrow_mut();
+    let group = groups.get_mut(&group_id).unwrap();
+
+    let (_, group_creator_id) = group.members.first().unwrap().clone();
+
+    // We add 62 members so that we have a group of 63.
```
raphaelrobert, line 202: How do we get to 62? The number of clients is 16 in line 192.

kkohbrok: Uh, leftover from before I realized I could reproduce the bug with a smaller group. I'll fix.

raphaelrobert (approved):

openmls/src/tree/tests_and_kats/unit_tests/test_resolution.rs
```diff
@@ -178,3 +182,66 @@ fn test_original_child_resolution() {
...
+/// parent hashes
+#[test]
+fn test_exclusion_for_parent_nodes() {
+    // Create a large tree members.
+    let managed_group_config = ManagedGroupConfig::test_default();
+
+    // We need 16 clients, such that we can create a group with 16 members. 16
+    // members means that we have two layers between the root and the leaves.
+    let number_of_clients = 16;
+    let setup = ManagedTestSetup::new(managed_group_config.clone(), number_of_clients);
```
franziskuskiefer, line 196: Why do you need to clone this?

openmls/src/tree/tests_and_kats/unit_tests/test_resolution.rs
```diff
@@ -178,3 +182,66 @@ fn test_original_child_resolution() {
         );
     }
 }
+
+/// Test the `original_child_resolution` function that is used to calculate
+/// parent hashes
+#[test]
+fn test_exclusion_for_parent_nodes() {
```
franziskuskiefer, line 189: I don't a test in here. This is only doing tree operations. It either needs a test or a description of the test that is happening somewhere inside the test framework if this is the case.

franziskuskiefer (changes requested):
lgtm but the test needs to test something

openmls/src/tree/tests_and_kats/unit_tests/test_resolution.rs
kkohbrok: Good point. I added more explanations and changed the last `unwrap` to `expect`, which is where it will fail if the parent hash computation is incorrect.

franziskuskiefer (approved):

raphaelrobert (approved):

PR #452, opened 2021-08-05
DER encode ECDSA signatures

### LINKED ISSUE #450: [Bug] Use DER encoding for ECDSA signatures
ECDSA signatures should be DER encoded.

Also see https://github.com/mlswg/mls-protocol/pull/475

kkohbrok:
This PR fixes #450 by encoding ECDSA signatures as specified in https://github.com/mlswg/mls-protocol/pull/475.

It also adds some test and refactors the SignatureError such that codec errors are taken into account.


raphaelrobert (changes requested):
LGTM in general, just some nits on the comments

openmls/src/ciphersuite/mod.rs
```diff
@@ -538,6 +546,67 @@ impl Signature {
...
+    /// throw a `der::Error`.
+    pub(crate) fn der_encode(raw_signature: &[u8]) -> Result<Self, der::Error> {
+        // We DER encode the ECDSA signature as per spec, assuming that
+        // `sign` returns two concatenated values (r||s).
+        let (r, s) = raw_signature.split_at(SCALAR_SIZE);
+        let r_encoded = UIntBytes::new(r)?;
+        let s_encoded = UIntBytes::new(s)?;
+        let mut bytes = [0; 2 * SCALAR_SIZE + MAX_DER_OVERHEAD];
+        let mut encoder = der::Encoder::new(&mut bytes);
+        encoder.message(&[&r_encoded, &s_encoded])?;
```
franziskuskiefer, line 602: So you don't even parse the signature? Or is the DER crate checking the OID etc. for you?

franziskuskiefer (changes requested):
I'll have to do a proper review but some thoughts.
I'm not sure I understand why you even pull in the DER crate. The ASN.1 structure for P256 signatures is pretty simple and probably doesn't need more code than you wrote here. Unless we want to do more x509 ;)

openmls/src/ciphersuite/mod.rs
```diff
@@ -538,6 +546,67 @@ impl Signature {
...
+            let decoded_scalar = UIntBytes::decode(decoder)?;
+            let decoded_scalar_len: usize = decoded_scalar.len().try_into()?;
+            // The verification algorithm expects the scalars to be 32 bytes
+            // long, buffered with zeroes.
+            let mut padded_scalar = vec![0u8; SCALAR_SIZE - decoded_scalar_len];
+            padded_scalar.extend_from_slice(decoded_scalar.as_bytes());
+            Ok(padded_scalar)
+        };
+
+        let mut out = Vec::with_capacity(2 * SCALAR_SIZE + MAX_DER_OVERHEAD);
```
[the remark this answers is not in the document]
kkohbrok: Ugh. Good point.

openmls/src/ciphersuite/mod.rs
kkohbrok: Maybe I don't understand what's meant to go on under the hood when DER encoding an ECDSA signature, but I was under the impression that all it really is is encoding the scalars to `UIntBytes` and add the relevant DER-specific bits in the beginning as explained in the comment above `MAX_DER_OVERHEAD`. I was not aware that I needed to do any checking or parsing of the given scalars. In particular, I was mostly following what is going on [here](https://docs.rs/ecdsa/0.12.4/src/ecdsa/der.rs.html#98) since I assumed that it DER encoding didn't carry any additional security implications with it. Was I wrong in assuming that?

kkohbrok (commented):
Thanks for your thoughts and I'm looking forward to a more thorough review! This is my first encounter with DER encoding and I was mostly trying to follow the approach taken in the `ecdsa` crate [here](https://docs.rs/ecdsa/0.12.4/src/ecdsa/der.rs.html), which seemed just fine to me (and it seems to work in terms of interop with MLSpp). Please let me know if I was too naive in this approach. I've heard a lot about woes with these encoding formats in the past, so I'm also genuinely curious on how one goes about this stuff in more general terms.

We could of course just "manually" encode the signature. My thinking was that using the `der` crate would be a good way to go, for one because I was lazy and also because we might want to DER-encode other things in the future.

kkohbrok:
In an oob discussion, @franziskuskiefer notes that we should be as strict as possible when decoding the signature in order not to rely on the crypto backend to handle invalid inputs gracefully. Also, he recommends not to use the `der` crate, but instead implement the specific decoding routine that we need for our purposes. @kkohbrok will take care of this and change the PR accordingly.

openmls/src/ciphersuite/mod.rs
```diff
@@ -546,63 +549,150 @@ impl Signature {
...
-// - 1 byte length tag
-// - 1 byte to indicate if the integer is signed
-const MAX_DER_OVERHEAD: usize = 9;
+// DER encoding SEQUENCE tag.
+const SEQUENCE_TAG: u8 = 0x30;
 
 impl Signature {
     /// This function takes a DER encoded ECDSA signature and decodes it to the
     /// bytes representing the concatenated scalars. If the decoding fails, it
     /// will throw a `der::Error`.
```
franziskuskiefer, line 566: outdated?

openmls/src/ciphersuite/mod.rs
```diff
@@ -546,63 +549,150 @@ impl Signature {
...
+                };
+                // Since we just read that byte, we decrease the length by 1.
+                scalar_length -= 1;
+            };
+
+            let mut scalar = vec![0; scalar_length];
+            buffer.read_exact(&mut scalar)?;
+
+            // The scalar should have the right length.
+            if scalar.len() != scalar_length {
```
franziskuskiefer, line 612: Isn't this guaranteed by `read_exact`?

openmls/src/ciphersuite/mod.rs
```diff
@@ -546,63 +549,150 @@ impl Signature {
...
+                return Err(SignatureError::DecodingError);
+            };
+
             // The verification algorithm expects the scalars to be 32 bytes
             // long, buffered with zeroes.
-            let mut padded_scalar = vec![0u8; SCALAR_SIZE - decoded_scalar_len];
-            padded_scalar.extend_from_slice(decoded_scalar.as_bytes());
+            let mut padded_scalar = vec![0u8; P256_SCALAR_LENGTH - scalar_length];
+            padded_scalar.extend_from_slice(&scalar);
```
franziskuskiefer, line 619: You can avoid copying things around by reading into the `padded_scalar` directly. But then you have to check the length again. You can use `append` to make this a move rather than a copy, which would be fine I guess.

openmls/src/ciphersuite/mod.rs
```diff
@@ -546,63 +549,150 @@ impl Signature {
...
+            return Err(SignatureError::DecodingError);
+        }
+
+        // The remaining bytes should be equal to the encoded length.
+        if signature_bytes.len() != length {
+            return Err(SignatureError::DecodingError);
+        }
+
+        let r = decode_scalar(&mut signature_bytes)?;
+        let s = decode_scalar(signature_bytes)?;
```
franziskuskiefer, line 653: Why is this not `&mut`?

openmls/src/ciphersuite/mod.rs
```diff
@@ -546,63 +549,150 @@ impl Signature {
...
+
+        let mut out = Vec::with_capacity(2 * P256_SCALAR_LENGTH);
         out.extend_from_slice(&r);
```
franziskuskiefer, line 656: This needs to clone. You should move `r` and `s`.

franziskuskiefer (commented):
drive-by

openmls/src/ciphersuite/mod.rs
kkohbrok: Honestly, I'm not sure what is happening under the hood here exactly. I know that the first one has to be `mut` for `decode_scalar` to move the read pointer forward. For the second call I'm not sure.

kkohbrok: I now made it `mut` so that we can afterwards check that indeed everything has been read from the `signature_bytes` reader.

openmls/src/ciphersuite/mod.rs
```diff
@@ -538,6 +549,203 @@ impl Signature {
...
+            let mut msb = scalar
+                .read_u8()
+                .map_err(|_| SignatureError::EncodingError)?;
+            while msb == 0x00 {
+                msb = scalar
+                    .read_u8()
+                    .map_err(|_| SignatureError::EncodingError)?;
+            }
+
+            // If the most significant bit is 1, we have to prepend 0x00 to indicate
```
franziskuskiefer, line 680: So we first remove all `0x00` and then add one again? You could peek the next byte. But this might end up being more complex so i guess this is fine.

openmls/src/ciphersuite/mod.rs
```diff
@@ -538,6 +549,203 @@ impl Signature {
...
+            // The encoded scalar needs to start with integer tag.
+            encoded_scalar
+                .write_u8(INTEGER_TAG)
+                .map_err(|_| SignatureError::EncodingError)?;
+            // This is safe, as we know that r is at most 33 bytes long.
+            encoded_scalar
+                .write_u8(scalar_uint.len() as u8)
+                .map_err(|_| SignatureError::EncodingError)?;
+            encoded_scalar
+                .write_all(&scalar_uint)
```
franziskuskiefer, line 705: hm, so you write this twice? Why don't you write the `scalar` directly  into `encoded_scalar`?

openmls/src/ciphersuite/mod.rs
```diff
@@ -538,6 +549,203 @@ impl Signature {
...
+
+        // Check overall length
+        if raw_signature.len() != 2 * P256_SCALAR_LENGTH {
+            return Err(SignatureError::EncodingError);
+        }
+
+        // We DER encode the ECDSA signature as per spec, assuming that
+        // `sign` returns two concatenated values (r||s).
+        let r = raw_signature
+            .get(..P256_SCALAR_LENGTH)
```
franziskuskiefer, line 719: You check the length of `raw_signature` before, so you can just do `raw_signature[..P256_SCALAR_LENGTH]` (or don't do the length check).

openmls/src/ciphersuite/mod.rs
```diff
@@ -538,6 +549,203 @@ impl Signature {
...
+        let total_length = r_encoded.len() + s_encoded.len();
+        let mut encoded_signature: Vec<u8> = Vec::with_capacity(2 + 2 * (P256_SCALAR_LENGTH + 3));
+        encoded_signature
+            .write_u8(SEQUENCE_TAG)
+            .map_err(|_| SignatureError::EncodingError)?;
+        encoded_signature
+            .write_u8(total_length as u8)
+            .map_err(|_| SignatureError::EncodingError)?;
+        encoded_signature
+            .write_all(&r_encoded)
```
franziskuskiefer, line 737: So now you write the `r_encoded` again. This is the 3rd time I think. You can just pass the writer to `encode_scalar` to write it directly.

franziskuskiefer (changes requested):
Mostly nits now. Generally looks good

openmls/src/ciphersuite/mod.rs
```diff
@@ -538,6 +549,203 @@ impl Signature {
...
+    /// bytes representing the concatenated scalars. If the decoding fails, it
+    /// will throw a `SignatureError`.
+    pub(crate) fn der_decode(&self) -> Result<Vec<u8>, SignatureError> {
+        // A small function to DER decode a single scalar.
+        fn decode_scalar<R: Read>(mut buffer: R) -> Result<Vec<u8>, SignatureError> {
+            // Check header bytes of encoded scalar.
+
+            // 1 byte INTEGER tag should be 0x02
+            let integer_tag = buffer
+                .read_u8()
```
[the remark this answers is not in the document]
franziskuskiefer: I think we can drop byteorder now?

openmls/src/ciphersuite/mod.rs
```diff
@@ -538,6 +549,193 @@ impl Signature {
...
+        // at most 1 byte to indicating that the integer is unsigned.
+        let length = signature_bytes
+            .read_u8()
+            .map_err(|_| SignatureError::DecodingError)? as usize;
+        if length > 2 * (P256_SCALAR_LENGTH + 3) {
+            return Err(SignatureError::DecodingError);
+        }
+
+        // The remaining bytes should be equal to the encoded length.
+        if signature_bytes.len() != length {
```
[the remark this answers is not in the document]
franziskuskiefer: Does 👍🏻 mean you want do add some? 🙂

openmls/src/ciphersuite/mod.rs
franziskuskiefer: Well, it's a function that can panic. In that sense, yes. But that's what you get when using array access. I'm fine with the `get` here. But you check the length twice because you already know `raw_signature.len() == 2 * P256_SCALAR_LENGTH`. If you want to avoid the potential for panics here I suggest just to drop the `raw_signature.len() != 2 * P256_SCALAR_LENGTH` check.

openmls/src/ciphersuite/mod.rs
```diff
@@ -538,6 +548,249 @@ impl Signature {
...
+
+        encode_scalar(&r, &mut encoded_signature)?;
+        encode_scalar(&s, &mut encoded_signature)?;
+
+        // Now we can determine the length of the scalars by taking the overall
+        // length and removing 1 for the Sequence tag and 1 for the length
+        // field. The conversion to u8 is safe, because we know that each of the
+        // scalars is at most 33 bytes long.
+        *encoded_signature
+            .get_mut(1)
```
franziskuskiefer, line 785: Ah I see how you do this now. Another option to do this is to first get the length and write it directly. I suspect that this would be faster (because you can allocate the memory you need at once). But I'm fine with leaving it like this for now.

franziskuskiefer (changes requested):
Thanks! Almost there.

openmls/src/ciphersuite/mod.rs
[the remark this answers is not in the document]
kkohbrok: Done. Sorry about that.

kkohbrok (commented):
I've now changed it such that I determine the length beforehand and encode it directly.

franziskuskiefer (approved):
Let's get this in. Looks good.
I filed #486 as follow up to make sure we can test this.

PR #465, opened 2021-09-27
Native Rust Crypto Provider

franziskuskiefer:
This PR adds a new cryptography backend using native Rust crypto implementations.
The new backend is the default.
To use evercrypt, default features have to be disabled and the `evercrypt-backend` feature has to be enabled.

The basic approach here is that there is one feature for the Rust crypto backend and one for the evercrypt backend. However, features are additive. So the other feature has to be explicitly disabled.
The `ciphersuite` module has a new `crypto` sub-module that branches off into the `evercrypt_provider` and the `rust_crypto_provider`, each implementing the piping to call the respective crypto backends. No other part of OpenMLS must call either backend.

This PR also adds new cross compilation targets with ios, arm64 Android, wasm, and 32-bit Linux.

The way randomness that is generated in OpenMLS itself (and not in the crypto libraries or hpke) has changed. The cipher suite holds a ChaCha20 RNG object now. We want to make the general handling of randomness better later. But the changes in this PR abstract away the PRNG and implements the infrastructure for the global state.

Note that this PR also contains some changes to the KATs that had bug in encoding some secrets, which failed with the new backend.


franziskuskiefer:
The fuzzer fails right now because of https://github.com/rust-fuzz/cargo-fuzz/issues/276

openmls/src/ciphersuite/mod.rs
```diff
@@ -301,6 +359,9 @@ impl PartialEq for Secret {
 impl Secret {
     /// Randomly sample a fresh `Secret`.
     /// This default random initialiser uses the default Secret length of `hash_length`.
+    ///
+    /// **NOTE: This has to wait until it can acquire the lock to get randomness!**
+    /// TODO: This panics if another thread holding the rng panics.
```
raphaelrobert, line 364: Can you file an issue for this so we don't forget?

openmls/src/ciphersuite/mod.rs
```diff
@@ -569,9 +636,10 @@ pub struct Ciphersuite {
     name: CiphersuiteName,
     signature_scheme: SignatureScheme,
     hpke: Hpke,
-    aead: AeadMode,
-    hash: DigestMode,
-    hmac: HmacMode,
+    // The rng really must not be accessed concurrently
+    rng: Mutex<ChaCha20Rng>,
```
raphaelrobert, line 640: Having the state here is probably the right place. I'm however wondering about the following: in the past we used to instantiate new `Ciphersuite` objects from a `CiphersuiteName`. This was not an issue, because there was no state but this changes now. Did you already check if we run into trouble because of that?

raphaelrobert (commented):
Thanks for the PR! I did a quick pass and left some comments. I need to check more thoroughly for a full review. I saw that `hmac` was substituted with `hash` in some places but didn't check why that it is. Could you quickly explain why?

openmls/src/ciphersuite/mod.rs
franziskuskiefer: I filed #467 to generally address the handling of randomness

openmls/src/ciphersuite/mod.rs
franziskuskiefer: This will just take the same `Ciphersuite` object and hence the same randomness pool. Which is fine from a security perspective but of course adds a single mutex for everything using this cipher suite.
This is something to consider in #467 and #468

franziskuskiefer:
> I saw that `hmac` was substituted with `hash` in some places but didn't check why that it is. Could you quickly explain why?

HMAC is defined by the hash algorithm it is using. In some places this is called an hmac type, in others the hash algorithm is used to set up the hmac. I decided to take the hash type here because it's more general.

.github/workflows/tests.yml
```diff
@@ -38,16 +38,48 @@ jobs:
           rustup target install aarch64-apple-ios
       - name: Tests
         run: cargo test --verbose
+      - name: Build w/ Evercrypt
+        if: matrix.os != 'macos-latest'
```
raphaelrobert, line 42: We are we leaving out chips for evercrypt? Isn't this working now?

.github/workflows/tests.yml
```diff
@@ -38,16 +38,48 @@ jobs:
...
+            deps: sudo apt update && sudo apt install gcc-multilib
+          # wasm
+          - target: wasm32-unknown-unknown
+            rust: stable
+          # ARM64 Android
+          - target: aarch64-linux-android
+            rust: stable
+          # ARM64 iOS
+          - target: aarch64-apple-ios
+            rust: stable
```
raphaelrobert, line 71: We will need the iOS simulator and Android emulator here at some point

README.md
```diff
@@ -23,7 +23,34 @@ A WIP Rust implementation of [Messaging Layer Security](https://github.com/mlswg
 
 ### Dependencies
 
-OpenMLS relies on [EverCrypt](https://github.com/project-everest/hacl-star/tree/master/providers/evercrypt), a high-performance, cross-platform, formally verified modern cryptographic provider through [EverCrypt Rust bindings](https://crates.io/crates/evercrypt).
+#### Cryptography
+
+OpenMLS does not implement its own cryptographic primitives.
+Instead, it relies on existing implementations of the cryptographic primitives used.
+There are two different cryptography backends supported right now.
```
raphaelrobert, line 30: Nitpicking: passive voice

README.md
```diff
@@ -23,7 +23,34 @@ A WIP Rust implementation of [Messaging Layer Security](https://github.com/mlswg
...
+
+##### Rust Crypto
+
+This is the default cryptographic backend, using [hkdf], [sha2], [p256], [p384], [x25519-dalek-ng], [ed25519-dalek] [chacha20poly1305], [aes-gcm].
+
+
+##### Evercrypt
+
+[EverCrypt] provides a formally verified cryptographic backend through the through [EverCrypt Rust bindings] to OpenMLS.
```
raphaelrobert, line 39: "through the through"

README.md
```diff
@@ -23,7 +23,34 @@ A WIP Rust implementation of [Messaging Layer Security](https://github.com/mlswg
...
+The default features have to disabled and the `evercrypt-backend` feature has to be enabled.
```
raphaelrobert, line 40: "have to be"/ "must be"

openmls/src/ciphersuite/crypto/rust_crypto_provider.rs
```diff
@@ -0,0 +1,214 @@
+//! # Evercrypt Crypto Provider
```
raphaelrobert, line 1: Rust Crypto

openmls/src/ciphersuite/crypto/rust_crypto_provider.rs
```diff
@@ -0,0 +1,214 @@
...
+//!
+//! Use evercrypt for all crypto operations.
```
raphaelrobert, line 3: Same

openmls/src/ciphersuite/crypto/rust_crypto_provider.rs
```diff
@@ -0,0 +1,214 @@
...
+
+// XXX: ed25519-dalek depends on an old version of rand.
+//      https://github.com/dalek-cryptography/ed25519-dalek/issues/162
+// extern crate ed25519_dalek;
+extern crate rand_07;
```
raphaelrobert, line 8: Isn't this pre 2018 edition syntax? Or is this needed for `ed25519_dalek` to work?

openmls/src/ciphersuite/crypto/rust_crypto_provider.rs
```diff
@@ -0,0 +1,214 @@
...
+        _ => Err(CryptoError::UnsupportedSignatureScheme),
+    }
+}
+
+pub(crate) fn sign(alg: SignatureScheme, data: &[u8], key: &[u8]) -> Result<Vec<u8>, CryptoError> {
+    match alg {
+        SignatureScheme::ECDSA_SECP256R1_SHA256 => {
+            let k = SigningKey::from_bytes(key).map_err(|_| CryptoError::CryptoLibraryError)?;
+            let signature = k.sign(data);
+            Ok(signature.to_der().to_bytes().into())
```
raphaelrobert, line 204: Does this make #452 specific to evercrypt?

openmls/src/ciphersuite/mod.rs
```diff
@@ -138,33 +142,87 @@ impl From<CiphersuiteName> for SignatureScheme {
...
+    Aes128Gcm = 0x0001,
+
+    /// AES GCM 256
+    Aes256Gcm = 0x0002,
+
+    /// ChaCha20 Poly1305
+    ChaCha20Poly1305 = 0x0003,
+
+    /// HPKE Export-only
+    HpkeExport = 0xFFFF,
```
raphaelrobert, line 159: Is this an HPKE thing? Do we need it?

openmls/src/ciphersuite/mod.rs
```diff
@@ -496,7 +559,7 @@ impl Mac {
 #[derive(Debug)]
 #[cfg_attr(test, derive(PartialEq))]
 pub struct AeadKey {
-    aead_mode: AeadMode,
+    aead_mode: AeadType,
```
raphaelrobert, line 562: Do we still need `mac_len` now?

openmls/src/ciphersuite/crypto/evercrypt_provider.rs
```diff
@@ -0,0 +1,187 @@
...
+    Ok(match hash_type {
+        HashType::Sha1 => HmacMode::Sha1,
+        HashType::Sha2_256 => HmacMode::Sha256,
+        HashType::Sha2_384 => HmacMode::Sha384,
+        HashType::Sha2_512 => HmacMode::Sha512,
+        _ => return Err(CryptoError::UnsupportedKdf),
+    })
+}
+
+pub(crate) fn support(signature_scheme: SignatureScheme) -> Result<(), CryptoError> {
```
raphaelrobert, line 79: Nitpicking, but `support` sounds a bit strange. Maybe `supports`, `has_support`, `check_support`?

raphaelrobert (changes requested):
I did a more thorough review and have a few remarks:

 - RNG state is temporary, needs to be moved with new API (as discussed offline)
 - Bad test coverage; it needs more testing, only test MLS KAT vectors for now (can be a follow-up PR)
 - Shouldn't there be a trait for crypto provider? It's not strictly needed, but might simplify the harmonization. Alternatively existing traits from RustCrypto could be used for that if they are suitable.

openmls/src/ciphersuite/crypto/evercrypt_provider.rs
franziskuskiefer: This function will go away anyway, but sure I can call it `support`

openmls/src/ciphersuite/crypto/rust_crypto_provider.rs
franziskuskiefer:
```suggestion
//! # Native Rust Crypto Provider
```

openmls/src/ciphersuite/crypto/rust_crypto_provider.rs
franziskuskiefer:
```suggestion
//! Use native Rust crypto for all operations.
```

openmls/src/ciphersuite/crypto/rust_crypto_provider.rs
franziskuskiefer: yeah, dalek requires an old version and this is the "nicest" way to get it in

openmls/src/ciphersuite/crypto/rust_crypto_provider.rs
franziskuskiefer: Yes, the code from  #452 will move into the `evercrypt_provider.rs`.

openmls/src/ciphersuite/mod.rs
franziskuskiefer: hm, probably not

README.md
franziskuskiefer: "have to be" would be wrong. But happy to do "must"

```suggestion
The default features have to disabled and the `evercrypt-backend` feature must be enabled.
```

README.md
franziskuskiefer:
```suggestion
[EverCrypt] provides a formally verified cryptographic backend through the [EverCrypt Rust bindings] to OpenMLS.
```

README.md
franziskuskiefer: so what?

.github/workflows/tests.yml
franziskuskiefer: Evercrypt doesn't support AES-GCM on old chips. And GH actions mac machines are too old.

franziskuskiefer (commented):
>  Bad test coverage; it needs more testing, only test MLS KAT vectors for now (can be a follow-up PR)

It's as bad as it has always been 😉 Since this PR doesn't introduce new logic there's no point in new tests. But yes, we have to increase test coverage.

> Shouldn't there be a trait for crypto provider? It's not strictly needed, but might simplify the harmonization. Alternatively existing traits from RustCrypto could be used for that if they are suitable.

RustCrypto traits don't really help here. But yes, there will be a trait in a follow up. But it's not necessary the way it's used here. Only when we pass in a crypto object we'll need one.

openmls/src/utils.rs
```diff
@@ -1,15 +1,11 @@
 use crate::extensions::*;
 use crate::tree::{index::*, node::*, *};
 
-use evercrypt::prelude::*;
+// === The folowing functions aren't necessarily cryptographically secure!
```
kkohbrok, line 4: Which should be ok, because they're behind the test-utils or test config flags.

kkohbrok (changes requested):
Thanks for taking care of this! Just a few nits from my side.

README.md
```diff
@@ -23,7 +23,34 @@ A WIP Rust implementation of [Messaging Layer Security](https://github.com/mlswg
...
+```
+
+Note that using this provider restricts OpenMLS usage to the following platforms right now
+* Linux x64
+* Linux Arm64
+* MacOS x64
+* Windows x64
+
+⚠️ Note that using this feature tests don't work out of the box.
+You need to change the `openmls` dev-dependency in the `Cargo.toml` to use the evercrypt backend as well.
```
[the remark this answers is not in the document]
franziskuskiefer: Right, that's not possible right now because the tests require the openmls with test-utils. But the general way this is handled will change again with the crypto object and I think we should be able to run with both backends then.

openmls/src/ciphersuite/crypto.rs
```diff
@@ -0,0 +1,29 @@
...
+#[cfg(all(feature = "evercrypt-backend", not(feature = "rust-crypto")))]
+pub(crate) use evercrypt_provider::*;
+#[cfg(all(feature = "rust-crypto", not(feature = "evercrypt-backend")))]
+mod rust_crypto_provider;
+#[cfg(all(feature = "rust-crypto", not(feature = "evercrypt-backend")))]
+pub(crate) use rust_crypto_provider::*;
+
+#[cfg(test)]
+pub(crate) fn aead_key_gen(alg: super::AeadType) -> Vec<u8> {
+    use rand::{rngs::OsRng, RngCore};
```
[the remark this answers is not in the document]
franziskuskiefer: Not really. This is all pretty ad-hoc. I just wanted to have something that works. And this is test code so I don't really care

openmls/src/ciphersuite/crypto/evercrypt_provider.rs
```diff
@@ -0,0 +1,187 @@
...
+    match alg {
+        AeadType::Aes128Gcm => Ok(AeadMode::Aes128Gcm),
+        AeadType::Aes256Gcm => Ok(AeadMode::Aes256Gcm),
+        AeadType::ChaCha20Poly1305 => Ok(AeadMode::Chacha20Poly1305),
+        AeadType::HpkeExport => Err(CryptoError::UnsupportedAeadAlgorithm),
+    }
+}
+
+#[inline(always)]
+fn kdf_from_hash(hash_type: HashType) -> Result<HmacMode, CryptoError> {
```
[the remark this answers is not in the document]
franziskuskiefer: sure, it's never used for hmac, but we can call it that :)

openmls/src/ciphersuite/crypto/evercrypt_provider.rs
```diff
@@ -0,0 +1,187 @@
...
+    if SignatureMode::try_from(signature_scheme).is_err() {
+        Err(CryptoError::UnsupportedSignatureScheme)
+    } else if DigestMode::try_from(signature_scheme).is_err() {
+        Err(CryptoError::UnsupportedSignatureScheme)
+    } else {
+        Ok(())
+    }
+}
+
+pub(crate) fn hkdf_extract(
```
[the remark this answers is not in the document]
franziskuskiefer: sure, this will become irrelevant with traits, but I can copy paste the comments here.

openmls/src/ciphersuite/crypto/rust_crypto_provider.rs
```diff
@@ -0,0 +1,214 @@
+//! # Native Rust Crypto Provider
+//!
+//! Use native Rust crypto for all operations.
+
+// XXX: ed25519-dalek depends on an old version of rand.
```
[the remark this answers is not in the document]
franziskuskiefer: There's an issue linked below. There's nothing we can do here.

openmls/src/utils.rs
franziskuskiefer: right, they are probably fine. But I just want to make sure everyone knows that we no one made sure that these are safe.

openmls/src/ciphersuite/mod.rs
```diff
@@ -366,20 +428,21 @@ impl Secret {
 
         Self {
             value: hkdf_extract(
-                self.ciphersuite.hmac,
+                self.ciphersuite.hash,
                 self.value.as_slice(),
                 ikm.value.as_slice(),
-            ),
+            )
+            .unwrap(),
```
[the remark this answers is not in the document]
franziskuskiefer: Yeah, this again should go away with the next PR. But I added a comment with explanation.

openmls/src/ciphersuite/crypto/rust_crypto_provider.rs
franziskuskiefer: Actually... I'm not sure why this is still in here. I probably forgot about it 🤔

raphaelrobert (approved):
Let's get this in!

kkohbrok (approved):

PR #470, opened 2021-09-30
Kat tree kem fix

franziskuskiefer:
* 5 year life time for key packages
* fix decoding error (no tls serialization on secrets)


kkohbrok (approved):

raphaelrobert (approved):
Thanks for the fix!
I'm still a bit puzzled about the fact that this didn't turn out to be a problem in #460, but let's get this in now.

Issue #520, opened 2021-10-29
Better tests for Evercrypt backend


franziskuskiefer:

franziskuskiefer:
Maybe a good initial set of tests would be to run test vectors with both crypto backends.

PR #521, opened 2021-10-29
Evercrypt provider

franziskuskiefer:
This PR adds back the possibility to use evercrypt as crypto backend after the changes in #500.

Note that the provider is hardly used and more and better tests are needed #520.


openmls/tests/test_managed_group.rs
```diff
@@ -84,6 +84,33 @@ fn auto_save(managed_group: &ManagedGroup) {
         .expect("Could not write group state to file");
 }
 
+// The following enables the OpenMlsEvercrypt provider on machines that support
+// it. This is a basic check to ensure that the provider generally works.
+// TODO: #520 - Better tests for Evercrypt backend
+#[cfg(all(
+    target_arch = "x86_64",
+    not(target_os = "macos"),
```
raphaelrobert, line 92: Any reason we exclude `macOS` currently?

openmls/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "openmls"
 version = "0.3.0"
 authors = ["OpenMLS Authors"]
-edition = "2018"
+edition = "2021"
```
raphaelrobert, line 5: Good to move it to 2021, but did you check everything that could be changed?

raphaelrobert (approved):
Looks good generally, just a couple of questions.

openmls/tests/test_managed_group.rs
franziskuskiefer: The Github CI doesn't support AES-NI on macOS. This should change with the new runners in about a month.

openmls/Cargo.toml
franziskuskiefer: Nope, I wanted the new resolver and this seemed like the easiest way. And nothing complained.
Note that the hpke version used here requires 2021 already. Nothing broke and cargo fix was happy as well.
I don't have any specific items we should change. But we should make use of `#![doc = include_str!("../Readme.md")]` for documentation.

raphaelrobert: Maybe worthwhile creating a ticket to check if we should change anything.

kkohbrok (dismissed):
Looks good! Maybe a good initial set of tests would be to run test vectors with both crypto backends.

franziskuskiefer:
Just a re-base with manual conflict resolution.

kkohbrok (approved):

PR #522, opened 2021-11-02
Validation & inspectable proposals

raphaelrobert:
This PR is the first in a series to improve semantic validation and to introduce a better API for authorization.
While many things are still missing, the size of the changes grows large quickly and reviews become more difficult.

 - Introduces new parsing functions in `ManagedGroup` according to the [wiki](https://github.com/openmls/openmls/wiki/Message-validation)
 - Introduces a ProposalStore to store proposals within an epoch
 - Separates the proposal queue handling between creating Commits and staging Commits
 - new (temporary) function in `tree/mod.rs`: `apply_staged_proposals`. It is identical to `apply_proposals` except for the function signature. This will disappear in a follow-up PR.
 - The `Verfifiable` trait is extended with a verification function that accepts signature public keys instead of just a credential (just credentials won't work with all AS scenarios).
 
Follow-ups:
 - Get rid of the current callback mechanism in `ManagedGroup`
 - More testing if we agree with the structure
 - Align `create_commit()` with the changes, so that we can reuse the `ProposalStore`


openmls/src/group/mls_group/proposals.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+impl StagedProposal {
+    /// Creates a new [StagedProposal] from an [MlsPlaintext]
+    pub fn from_mls_plaintext(
+        ciphersuite: &Ciphersuite,
+        backend: &impl OpenMlsCryptoProvider,
+        mls_plaintext: MlsPlaintext,
+    ) -> Result<Self, StagedProposalError> {
+        debug_assert!(mls_plaintext.content_type() == &ContentType::Proposal);
```
kkohbrok, line 74: I assume this assert is removed in the end?

openmls/src/group/mls_group/proposals.rs
```diff
@@ -0,0 +1,208 @@
...
+            proposal_reference,
+            sender,
+            proposal_or_ref_type: ProposalOrRefType::Proposal,
+        })
+    }
+    /// Returns the `Proposal` as a reference
+    pub(crate) fn proposal(&self) -> &Proposal {
+        &self.proposal
+    }
+    /// Returns the `ProposalReference` as a reference
```
kkohbrok, line 106: The function doesn't return a reference, though.

openmls/src/group/mls_group/staged_commit.rs
```diff
@@ -58,11 +59,14 @@ impl MlsGroup {
...
-        let apply_proposals_values =
-            match provisional_tree.apply_proposals(backend, proposal_queue, own_key_packages) {
-                Ok(res) => res,
-                Err(_) => return Err(StageCommitError::OwnKeyNotFound.into()),
-            };
+        let apply_proposals_values = match provisional_tree.apply_staged_proposals(
+            backend,
+            &proposal_queue,
+            own_key_packages,
+        ) {
```
kkohbrok, line 66: I know this is how it previously was, but can't we use `map_err` here?

openmls/src/group/managed_group/validation.rs
```diff
@@ -0,0 +1,244 @@
...
+            // If it is a ciphertext we decrypt it and return the plaintext message.
+            // Attempting to decrypt it will also check the bounds of the epoch.
+            InboundMessage::Ciphertext(ciphertext) => {
+                let aad = ciphertext.authenticated_data.clone().into_vec();
+
+                let plaintext = self.group.decrypt(&ciphertext, backend)?;
+
+                (plaintext, aad)
+            }
+            // If it is a plaintext message we return it with an empty AAD
```
kkohbrok, line 50: Wouldn't `Option` be more suitable for the AAD here? Otherwise we can't distinguish between an empty AAD and a non-existent one.

openmls/src/framing/plaintext.rs
```diff
@@ -268,6 +268,11 @@ impl MlsPlaintext {
         self.sender.to_leaf_index()
     }
 
+    /// Returns `true` if the `membership_tag` fiel is present and `false otherwise.
```
kkohbrok, line 271:
```suggestion
    /// Returns `true` if the `membership_tag` field is present and `false` otherwise.
```

openmls/src/group/managed_group/validation.rs
```diff
@@ -0,0 +1,244 @@
...
+                // TODO #106: Membership tag is not expected for external senders
+                membership_tag_required = true;
+                (plaintext, vec![])
+            }
+        };
+
+        // Check that the sender is a valid member of the tree
+        // The sender index must be within the tree and the corresponding leaf node must not be blank
+        let sender = plaintext.sender().to_leaf_index();
+        if sender > self.group.tree().leaf_count() || self.group.tree().nodes[sender].is_blank() {
```
kkohbrok, line 63: What about external inits or preconfigured senders?

openmls/src/group/managed_group/validation.rs
```diff
@@ -0,0 +1,244 @@
...
+            .unwrap()
+            .credential()
+            .clone();
+
+        // Get the serialized group context for further verification
+        let serialized_context = self.group.context().tls_serialize_detached().map_err(|_| {
+            ManagedGroupError::LibraryError("Could not serialize group context".into())
+        })?;
+
+        // Verifiy the membership tag if needed
```
kkohbrok, line 83:
```suggestion
        // Verify the membership tag if needed
```

openmls/src/group/managed_group/validation.rs
```diff
@@ -0,0 +1,244 @@
...
+        */
+        self.group.merge_commit(staged_commit);
+        self.proposal_store.empty();
+        Ok(())
+    }
+}
+
+/// Inbound message from the DS.
+pub enum InboundMessage {
+    Plaintext(MlsPlaintext),
```
kkohbrok, line 207: Shouldn't this be a `VerifiablePlaintext`?

openmls/src/group/managed_group/validation.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// This processing function does most of the semantic verifications.
+    /// It returns a [ProcessedMessage] enum.
+    pub fn process_unverified_message(
+        &mut self,
+        message: UnverifiedMessage,
+        signature_key: Option<SignaturePublicKey>,
+        backend: &impl OpenMlsCryptoProvider,
+    ) -> Result<ProcessedMessage, ManagedGroupError> {
+        /*
+         - Signature verification, either with leaf key or optional parameter
```
kkohbrok, line 124: If I read this function correctly, some of these things are not actually checked (yet), are they?

openmls/src/group/managed_group/validation.rs
```diff
@@ -0,0 +1,244 @@
...
+                InvalidMessageError::UnknownSender,
+            ));
+        }
+
+        // Extract the credential
+        // Unwrapping here is safe, because we know the leaf node exists and is not blank
+        let credential = self.group.tree().nodes[sender]
+            .key_package
+            .as_ref()
+            .unwrap()
```
kkohbrok, line 74: Remove unwrap with something that throws an error?

kkohbrok (commented):
Some comments, but generally, this looks really good!

openmls/src/group/managed_group/validation.rs
kkohbrok: Let's do this in a follow up, when we add these functionalities.

openmls/src/group/managed_group/validation.rs
```diff
@@ -0,0 +1,223 @@
...
+    /// Returns an [UnverifiedMessage] that can be inspected and later processed in
+    /// [self::process_unverified_message()].
+    pub fn parse_message(
+        &mut self,
+        message: InboundMessage,
+        backend: &impl OpenMlsCryptoProvider,
+    ) -> Result<UnverifiedMessage, ManagedGroupError> {
+        /*
+        High level checks:
+         - epoch must be within bounds
```
kkohbrok, line 18: Created #527 to track that these checks are actually performed.

franziskuskiefer (dismissed):

franziskuskiefer (approved):

Issue #528, opened 2021-11-04
Make sure that validation checks are carried out in `process_unverified_message`


kkohbrok:

raphaelrobert:
With validation for incoming MLS messages being completed, we can close this.

Issue #525, opened 2021-11-04
Test message validation


kkohbrok:
With the new validation mechanism added in #522 in place, we should be able to better test individual validation steps. Writing these test should essentially yield a negative test vector, which we can use to test against other implementations.

PR #545, opened 2021-11-14
Introduce type system for validation

raphaelrobert:
This PR is a follow-up PR towards implementing validation. It does not yet contain tests, which will be tackled with #525, #527, and #528.

 - Use strict types for the different stages of message validation
 - This is based on `VerifiableMlsPlaintext` but supersedes its purpose somewhat. We can debate whether `VerifiableMlsPlaintext` is still needed in the future.
 - Move MlsMessageIn & MlsMessageOut to `framing` and add serialization/deserialization for these types
 - Said serialization now looks at the first byte of the serialized data to determine the wire_format, instead of prepending a duplicate
 - Improvement: it membership tags are now also verified for Commit messages (probably more will be discovered with better testing)


kkohbrok:
Just did a rough pass for now. One thing I would suggest from a structural standpoint is to put more of the logic that is now in functions like `parse_message` in functions that are part of the implementation of the individual datatypes. So that one can call ```
let decrypted_message = DecryptedMessage::from_inbound_message(message, ...)?;
```
On the one hand, that requires passing potentially unneeded information (e.g. decryption keys even if the message is a plaintext), but on the other hand, we get rid of some of the complexity. What do you think?

**Comment by raphaelrobert** [comment]
> Just did a rough pass for now. One thing I would suggest from a structural standpoint is to put more of the logic that is now in functions like `parse_message` in functions that are part of the implementation of the individual datatypes. So that one can call
> 
> ```
> let decrypted_message = DecryptedMessage::from_inbound_message(message, ...)?;
> ```
> 
> On the one hand, that requires passing potentially unneeded information (e.g. decryption keys even if the message is a plaintext), but on the other hand, we get rid of some of the complexity. What do you think?

I think I can move some of the checks into the new types. Regarding the substantial differences in the arguments, I would prefer to have two separate functions to create a `DecryptedMessage`.

**Comment by kkohbrok** [comment]
Fair enough. The match statements that error out in case you call the function with the wrong enum always feel very ugly to me, but I understand that the unnecessary arguments are also not pretty.

**Comment by franziskuskiefer** [comment]
> The match statements that error out in case you call the function with the wrong enum always feel very ugly to me

These should probably be `if let ...` instead of `match`.

File: openmls/src/framing/validation.rs [hunk 1]
```diff
@@ -0,0 +1,266 @@
+//! # Validation steps for incoming messages
+//! ```text
+//! MlsMessageIn (exposes: wire format, group, epoch)
+//! |
+//! V
+//! parse_message()
+//! |
+//! V
+//! DecryptedMessage
```
kkohbrok, line 9: This state is not really visible to the outside, though, is it? Or at least it's not the return type of `parse_message`.

openmls/src/framing/validation.rs
```diff
@@ -0,0 +1,266 @@
...
+
+impl UnverifiedMemberMessage {
+    pub(crate) fn into_verified(
+        self,
+        backend: &impl OpenMlsCryptoProvider,
+        signature_key: Option<&SignaturePublicKey>,
+    ) -> Result<VerifiedMemberMessage, ValidationError> {
+        // If a signature key is provided it will be used,
+        // otherwise we take the key from the credential
+        match signature_key {
```
kkohbrok, line 198: Sorry to nitpick here, but all this matching seems a bit much. Maybe something like this?
```
        let verified_member_message = if let Some(signature_public_key) = signature_key {
            self.plaintext
                .verify_with_key(backend, signature_public_key)
        } else {
            self.plaintext.verify(backend, &self.credential)
        }
        .map(|plaintext| VerifiedMemberMessage { plaintext })?;
        Ok(verified_member_message)
```

openmls/src/group/managed_group/validation.rs
```diff
@@ -126,47 +108,78 @@ impl ManagedGroup {
...
+                let verified_member_message =
+                    member_message.into_verified(backend, signature_key)?;
+
+                Ok(match verified_member_message.plaintext().content() {
+                    MlsPlaintextContentType::Application(application_message) => {
+                        ProcessedMessage::ApplicationMessage(
+                            application_message.as_slice().to_vec(),
+                        )
+                    }
+                    MlsPlaintextContentType::Proposal(_proposal) => {
```
kkohbrok, line 133: Would it be worth revisiting the constructors for StagedProposal and `stage_commit` to take the content instead of the whole plaintext?

openmls/src/group/managed_group/validation.rs
```diff
@@ -126,47 +108,78 @@ impl ManagedGroup {
...
+                })
             }
-            MlsPlaintextContentType::Proposal(_proposal) => {
-                ProcessedMessage::ProposalMessage(Box::new(
-                    StagedProposal::from_mls_plaintext(self.ciphersuite(), backend, plaintext)
-                        .unwrap(),
-                ))
+            UnverifiedContextMessage::External(external_message) => {
+                // Signature verification
+                match signature_key {
```
kkohbrok, line 157: Nitpicking again: Even though it makes returning the error slightly less explicit: I think I'd use an `ok_or` here. Or maybe an `if let`.

kkohbrok (changes requested):
A few nitpicks, but otherwise looks good!

openmls/src/framing/validation.rs
```diff
@@ -0,0 +1,266 @@
...
+        {
+            return Err(ValidationError::MissingConfirmationTag);
+        }
+        // Check that application messages are always encrypted
+        if plaintext.content_type() == ContentType::Application {
+            if plaintext.wire_format() != WireFormat::MlsCiphertext {
+                return Err(ValidationError::UnencryptedApplicationMessage);
+            } else if !plaintext.sender().is_member() {
+                // This should not happen because the sender of an MlsCiphertext should always be a member
+                return Err(ValidationError::LibraryError);
```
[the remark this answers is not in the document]
raphaelrobert: Decrypting an `MlsCiphertext` would fail if it wasn't encrypted by a member, i.e. we wouldn't be able to fetch the right key material for decryption.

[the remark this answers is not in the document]
kkohbrok: I see, thanks!

openmls/src/framing/validation.rs
```diff
@@ -0,0 +1,266 @@
...
+use crate::ciphersuite::signable::Verifiable;
+
+use super::*;
+
+pub struct DecryptedMessage {
+    plaintext: VerifiableMlsPlaintext,
+}
+
+impl DecryptedMessage {
+    pub(crate) fn from_inbound_plaintext(
```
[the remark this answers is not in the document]
kkohbrok: Thanks for adding docs on the structs! The functions still don't have documentation. The functions aren't hugely complex, but having at least a brief sentence what they do would be nice. Especially for the `pub` functions.

openmls/src/framing/validation.rs
```diff
@@ -139,6 +142,10 @@ impl UnverifiedMessage {
     }
 }
 
+/// Contains an [VerifiableMlsPlaintext] and a [Credential] if it is a member message.
```
kkohbrok, line 145: `Contains an [VerifiableMlsPlaintext]` -> `Contains a [VerifiableMlsPlaintext]`

openmls/src/framing/validation.rs
```diff
@@ -40,6 +40,9 @@ use crate::ciphersuite::signable::Verifiable;
 
 use super::*;
 
+/// Contains a [VerifiableMlsPlaintext]. Can be built either from a plaintext or ciphertext.
+/// In the latter case, it attempts to decrypt the ciphertext.
+/// Checks the presence of the membership tag and confirmation tag.
```
kkohbrok, line 45: Some of this feels like it should be documentation for the functions that are part of this struct's implementation. For example, `... it attempts to decrypt the ciphertext`. What is it? Same for the last sentence.

kkohbrok (changes requested):
Almost there! Just a few more comments on the docs.

raphaelrobert:
> Almost there! Just a few more comments on the docs.

The plan is to add more explicit documentation about the different checks once the tests are there.

kkohbrok (approved):
Perfect, thanks!

PR #546, opened 2021-11-15
Fixes for clippy errors and warnings

raphaelrobert:
Clippy was blocked by an error (also on the CI) which is fixed now and I also fixed several warnings.

This PR only contains the clippy suggestions.


kkohbrok (approved):
Thanks for making clippy happy 📎!

PR #557, opened 2021-11-18
Create CredentialBundle from existing keys

raphaelrobert:
This PR adds a function to `CredentialBundle` to create it from existing keys. This feature was requested by OpenMLS users.


kkohbrok (approved):
Looks good! It might be worth in the future to tag public/private keys with signature schemes in a similar way that we tag Secrets with MLS versions and Ciphersuites.

PR #559, opened 2021-11-18
Import existing key material for CredentialBundle

### LINKED ISSUE #461: [Feature request] Support importing external `private_key` for `CredentialBundle`

Currently there is no way to pass an external `private_key` during creating of an `CredentialBundle` instance which does not allow for protocols where the private key is also used for features outside of MLS.

An additional constructor in `CredentialBundle` which allows me to pass on a `private_key` via `SignatureKeypair` as a method argument.

**Additional context**

https://github.com/openmls/openmls/blob/main/openmls/src/credentials/mod.rs#L159

raphaelrobert:
This is a follow-up to #557. This PR adds functions to import existing keys for `CredentialBundle`.
Fixes #461.


adzialocha:
I think the problem is this line here: https://github.com/openmls/openmls/blob/30fbec303bb228bd280714eed03e531899d475b5/openmls/src/ciphersuite/mod.rs#L43 :+1:

adzialocha:
Also related to this: How do I create a new `SignaturePrivateKey` instance from an external secret key? `SignaturePublicKey` provides a way (`::new(bytes, signature_scheme)`) which is convenient.

franziskuskiefer:
This PR needs

- [ ] a feature flag as proposed in #596 
- [ ] fix the prelude export (I recommend removing it from there. It's nothing that is commonly used.)

> How do I create a new `SignaturePrivateKey` instance from an external secret key?

@adzialocha why do you need to create a `SignaturePrivateKey`? `SignatureKeypair::from_bytes` as introduced in this PR allows you to get a key pair from raw bytes. There shouldn't be the need to create only the private key.

adzialocha:
> why do you need to create a SignaturePrivateKey? SignatureKeypair::from_bytes as introduced in this PR allows you to get a key pair from raw bytes. There shouldn't be the need to create only the private key.

Yes, this can be considered as resolved now, at the point of my question it was not possible. So, alles gut 🖖

And thank you!

openmls/src/ciphersuite/signature.rs
```diff
@@ -137,6 +149,10 @@ impl SignaturePublicKey {
             .map_err(|_| CryptoError::InvalidSignature)
     }
 
+    /// Get the signature scheme of the keypair
```
kkohbrok, line 152:
```suggestion
    /// Get the signature scheme of the public key
```

openmls/src/ciphersuite/signature.rs
```diff
@@ -159,4 +175,9 @@ impl SignaturePrivateKey {
             Err(_) => Err(CryptoError::CryptoLibraryError),
         }
     }
+
+    /// Get the signature scheme of the keypair
```
kkohbrok, line 179:
```suggestion
    /// Get the signature scheme of the private key
```

kkohbrok (commented):
Looks good! Just two nits.

franziskuskiefer (approved):

PR #560, opened 2021-11-18
Start validation tests

raphaelrobert:
This PR starts the validation testing. The tests are named after the semantic validation checks as outlined in the [wiki](https://github.com/openmls/openmls/wiki/Message-validation). This works towards #133.

This PR only contains 2 tests so far. If we agree on the structure I'll follow up with more tests.


kkohbrok (approved):
Looks good to me!

PR #564, opened 2021-11-22
Implement & document more validation checks

raphaelrobert:
This PR implements new validation checks and documents existing ones according to the [wiki](https://github.com/openmls/openmls/wiki/Message-validation).

The following checks are now covered:

*Semantic validation of message framing*

    ValSem1 Wire format
    ValSem2 Group id
    ValSem3 Epoch
    ValSem4 Sender: Member: check the member exists
    ValSem5 Application messages must use ciphertext
    ValSem6 Ciphertext: decryption needs to work
    ValSem7 Membership tag presence
    ValSem8 Membership tag verification
    ValSem9 Confirmation tag presence
    ValSem10 Signature verification

*Semantic validation of proposals covered by a Commit*

    ValSem100 Add Proposal: Identity in proposals must be unique among proposals
    ValSem101 Add Proposal: Signature public key in proposals must be unique among proposals
    ValSem102 Add Proposal: HPKE init key in proposals must be unique among proposals
    ValSem103 Add Proposal: Identity in proposals must be unique among existing group members
    ValSem104 Add Proposal: Signature public key in proposals must be unique among existing group members
    ValSem105 Add Proposal: HPKE init key in proposals must be unique among existing group members
    ValSem106 Add Proposal: required capabilities
    ValSem107 Remove Proposal: Removed member must be unique among proposals
    ValSem108 Remove Proposal: Removed member must be an existing group member
    ValSem109 Update Proposal: Identity must be unchanged between existing member and new proposal
    ValSem110 Update Proposal: Signature public key must be unique among existing members
    ValSem111 Update Proposal: HPKE init key must be unique among existing members

*Commit message validation*

    ValSem200 Commit must not cover inline self Remove proposal
    ValSem201 Path must be present, if Commit contains Removes or Updates
    ValSem202 Path must be the right length
    ValSem203 Path secrets must decrypt correctly
    ValSem204 Public keys from Path must be verified and match the private keys from the direct path
    ValSem205 Confirmation tag must be successfully verified

Note that not all validation checks are covered by tests yet. This is blocked by #560 and will be done in a subsequent PR.

### DIFF
--- a/openmls/src/group/errors.rs
+++ b/openmls/src/group/errors.rs
@@ -18,7 +18,7 @@ implement_error! {
         Simple {
             InitSecretNotFound =
                 "Missing init secret when creating commit.",
-            NoSignatureKey = "No signature was found.",
+            NoSignatureKey = "No signature key was found.",
         }
         Complex {
             MlsCiphertextError(MlsCiphertextError) =
--- a/openmls/src/group/errors.rs
+++ b/openmls/src/group/errors.rs
@@ -6,6 +6,7 @@
 use crate::ciphersuite::CryptoError;
 use crate::config::ConfigError;
 use crate::credentials::CredentialError;
+use crate::framing::errors::ValidationError;
 use crate::framing::errors::{MlsCiphertextError, MlsPlaintextError, VerificationError};
 use crate::messages::errors::ProposalQueueError;
 use crate::schedule::errors::{KeyScheduleError, PskSecretError};
@@ -17,6 +18,7 @@ implement_error! {
         Simple {
             InitSecretNotFound =
                 "Missing init secret when creating commit.",
+            NoSignatureKey = "No signature was found.",
         }
         Complex {
             MlsCiphertextError(MlsCiphertextError) =
@@ -49,6 +51,8 @@ implement_error! {
                 "See [`CredentialError`](crate::credentials::CredentialError) for details.",
             TreeError(TreeError) =
                 "See [`TreeError`](crate::tree::TreeError) for details.",
+            ValidationError(ValidationError) =
+                "See [`ValidationError`](crate::framing::ValidationError) for details.",
             FramingValidationError(FramingValidationError) =
                 "See [`FramingValidationError`](crate::group::FramingValidationError) for details.",
             ProposalValidationError(ProposalValidationError) =
@@ -216,7 +220,7 @@ implement_error! {
     pub enum FramingValidationError {
         WrongGroupId = "Message group ID differs from the group's group ID.",
         WrongEpoch = "Message epoch differs from the group's epoch.",
-        UnknownSender = "The sender is not a member of the group.",
+        UnknownMember = "The sender could not be matched to a member of the group.",
         UnencryptedApplicationMessage = "Application messages must always be encrypted.",
         NonMemberApplicationMessage = "An application message was sent from an external sender.",
         MissingMembershipTag = "Membership tag is missing.",
@@ -226,7 +230,7 @@ implement_error! {
 
 implement_error! {
     pub enum ProposalValidationError {
-        UnknownSender = "The sender is not a member of the group.",
+        UnknownMember = "The sender could not be matched to a member of the group.",
         DuplicateIdentityAddProposal = "Found two add proposals with the same identity.",
         DuplicateSignatureKeyAddProposal = "Found two add proposals with the same signature key.",
         DuplicatePublicKeyAddProposal = "Found two add proposals with the same HPKE public key.",
@@ -236,6 +240,7 @@ implement_error! {
         UpdateProposalIdentityMismatch = "The identity of the update proposal did not match the existing identity.",
         ExistingSignatureKeyUpdateProposal = "Signature key of the update proposal already existed in tree.",
         ExistingPublicKeyUpdateProposal = "HPKE public key of the update proposal already existed in tree.",
-        UnknownMemberRemoval = "The remove proposal refrenced a non-existing member.",
+        DuplicateMemberRemoval = "Duplicate remove proposals for the same member.",
+        UnknownMemberRemoval = "The remove proposal referenced a non-existing member.",
     }
 }
--- a/openmls/src/group/errors.rs
+++ b/openmls/src/group/errors.rs
@@ -49,6 +49,10 @@ implement_error! {
                 "See [`CredentialError`](crate::credentials::CredentialError) for details.",
             TreeError(TreeError) =
                 "See [`TreeError`](crate::tree::TreeError) for details.",
+            FramingValidationError(FramingValidationError) =
+                "See [`FramingValidationError`](crate::group::FramingValidationError) for details.",
+            ProposalValidationError(ProposalValidationError) =
+                "See [`ProposalValidationError`](crate::group::ProposalValidationError) for details.",
         }
     }
 }
@@ -189,6 +193,7 @@ implement_error! {
     pub enum StagedProposalQueueError {
         Simple {
             ProposalNotFound = "Not all proposals in the Commit were found locally.",
+            SelfRemoval = "The sender of a Commit tried to remove themselves.",
         }
         Complex {
             NotAProposal(StagedProposalError) = "The given MLS Plaintext was not a Proposal.",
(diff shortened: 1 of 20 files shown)


openmls/src/group/errors.rs
```diff
@@ -206,3 +211,31 @@ implement_error! {
         }
     }
 }
+
+implement_error! {
+    pub enum FramingValidationError {
+        WrongGroupId = "Message group ID differs from the group's group ID.",
+        WrongEpoch = "Message epoch differs from the group's epoch.",
+        UnknownSender = "The sender is not a member of the group.",
```
kkohbrok, line 219: If I understand the code correctly, this is thrown either if the advertised sender is outside of the tree or if it has a public key that can't be found in the tree. It would be nice to address these cases separately here, since one will go away with preconfigured senders or external init and one will always be a straight-up error.

kkohbrok: Same for the error below, I think.

openmls/src/group/mls_group/proposals.rs
```diff
@@ -195,6 +217,106 @@ impl StagedProposalQueue {
     }
 }
 
+pub struct StagedAddProposal<'a> {
+    add_proposal: &'a AddProposal,
+    sender: &'a Sender,
+}
+
+impl<'a> StagedAddProposal<'a> {
+    pub(crate) fn try_from_staged_proposal(staged_proposal: &'a StagedProposal) -> Option<Self> {
```
kkohbrok, line 226: Doesn't `try` imply that it will result in an error upon failure? Seems a bit odd that this would return `Option`.

kkohbrok (changes requested):
I only made a relatively superficial pass for now in that I didn't actually verify that the code does the check that it is annotated with. I think it's a good idea to index the individual checks, but it would be great if we could find an indexing scheme that somehow conveys the nature of the check as well. For example we could just write out the complete entry in the list of checks written out in the PR description, i.e. instead of having the following code comment:
```
// ValSem100
```
We would have his one:
```
// ValSem100 Add Proposal: Identity in proposals must be unique among proposals
```
This would make reviewing easier now and in the future.

One more nit besides the inline comments: It would be great if you could add short documentation strings to every public or crate-public function, even if most of them are just getters or setters.

raphaelrobert:
I addressed the issues we discussed and filed #568 for the documentation. I feel pretty strongly about keeping the documentation concise at this point, otherwise it will be a hassle to update it whenever the validation stuff is touched.

openmls/src/group/errors.rs
```diff
@@ -17,6 +18,7 @@ implement_error! {
         Simple {
             InitSecretNotFound =
                 "Missing init secret when creating commit.",
+            NoSignatureKey = "No signature was found.",
```
kkohbrok, line 21:
```suggestion
            NoSignatureKey = "No signature key was found.",
```

openmls/src/group/mls_group/validation.rs
```diff
@@ -0,0 +1,244 @@
...
+    ///  - ValSem9
+    pub fn validate_plaintext(
+        &self,
+        plaintext: &VerifiableMlsPlaintext,
+    ) -> Result<(), MlsGroupError> {
+        // ValSem4
+        let sender = plaintext.sender();
+        if sender.is_member() {
+            let sender_index = sender.to_leaf_index();
+            if sender_index > self.tree().leaf_count() || self.tree().nodes[sender_index].is_blank()
```
kkohbrok, line 41:
```suggestion
            if sender_index >= self.tree().leaf_count() || self.tree().nodes[sender_index].is_blank()
```

kkohbrok, line 41: In the PR description, ValSem4 is described with "checking that the member exists". This brings us back to the discussion of what a member is. Here, we're checking if the sender's claimed leaf index is within the bounds of the tree. Does that mean that "the member exists"? This is a somewhat philosophical question, but I think what I'm trying to say is that there could be a better label for this check 😄

openmls/src/group/mls_group/validation.rs
```diff
@@ -0,0 +1,244 @@
...
+            {
+                return Err(FramingValidationError::UnknownMember.into());
+            }
+        }
+
+        // ValSem5
```
kkohbrok, line 47: Here, we check more than what's advertised by ValSem5. Should the second test get its own label? It's related, but not quite the same, I think. No strong opinion, though.

openmls/src/group/mls_group/validation.rs
```diff
@@ -0,0 +1,244 @@
...
+        if plaintext.content_type() == ContentType::Application {
+            if plaintext.wire_format() != WireFormat::MlsCiphertext {
+                return Err(FramingValidationError::UnencryptedApplicationMessage.into());
+            } else if !plaintext.sender().is_member() {
+                return Err(FramingValidationError::NonMemberApplicationMessage.into());
+            }
+        }
+
+        // ValSem7
+        if plaintext.sender().is_member()
```
kkohbrok, line 57: The label just says "Membership tag presence". Maybe it would be worthwhile adding the conditions under which a membership tag actually has to be present? This seems to be done for other labels. Same for the confirmation tag presence test below.

openmls/src/group/mls_group/validation.rs
```diff
@@ -0,0 +1,244 @@
...
+    // === Proposals ===
+
+    /// Validate Add proposals. This function implements the following checks:
+    ///  - ValSem100
+    ///  - ValSem101
+    ///  - ValSem102
+    ///  - ValSem103
+    ///  - ValSem104
+    ///  - ValSem105
+    ///  - ValSem106
```
kkohbrok, line 82:
```suggestion
    ///  - TODO: ValSem106
```

openmls/src/group/mls_group/validation.rs
```diff
@@ -0,0 +1,244 @@
...
+                return Err(ProposalValidationError::ExistingPublicKeyAddProposal.into());
+            }
+        }
+        // TODO #538: ValSem106: Check the required capabilities of the add proposals
+        Ok(())
+    }
+
+    /// Validate Remove proposals. This function implements the following checks:
+    ///  - ValSem107
+    ///  - ValSem109
```
kkohbrok, line 149:
```suggestion
    ///  - ValSem108
```

openmls/src/group/mls_group/validation.rs
```diff
@@ -0,0 +1,244 @@
...
+
+        let mut removes_set = HashSet::new();
+        let tree = &self.tree();
+
+        let index_set: HashSet<NodeIndex> =
+            HashSet::from_iter(tree.indexed_key_packages().map(|(index, _kp)| index));
+
+        for remove_proposal in remove_proposals {
+            let removed = remove_proposal.remove_proposal().removed();
+            // ValSem107
```
kkohbrok, line 164: I believe the labels are switched here.

openmls/src/group/mls_group/process.rs
```diff
@@ -0,0 +1,175 @@
...
+            credential,
+        ))
+    }
+
+    /// This processing function does most of the semantic verifications.
+    /// It returns a [ProcessedMessage] enum.
+    /// Checks the following semantic validation:
+    ///  - ValSem8
+    ///  - ValSem10
+    ///  - ValSem100
```
kkohbrok, line 81: In the function body, I could only find references to ValSem8 and ValSem10. Did I miss something?

openmls/src/group/mls_group/staged_commit.rs
```diff
@@ -255,20 +293,16 @@ pub struct StagedCommit {
 }
 
 impl StagedCommit {
-    pub fn adds(&self) -> impl Iterator<Item = &StagedProposal> {
-        self.staged_proposal_queue
-            .filtered_by_type(ProposalType::Add)
+    pub fn add_proposals(&self) -> impl Iterator<Item = StagedAddProposal> {
```
kkohbrok, line 296: Could you add one-liner docs here as well?

kkohbrok (changes requested):
Thanks for integrating my previous comments! Some more remarks on the labels and some of the checks, then we can get this in once the CI is green.

openmls/src/group/mls_group/validation.rs
raphaelrobert: I changed the label in the wiki.

openmls/src/group/mls_group/validation.rs
raphaelrobert: I don't understand what you mean. I added a comment to make the check more clear.

openmls/src/group/mls_group/validation.rs
raphaelrobert: I added some more comments.

openmls/src/group/mls_group/process.rs
raphaelrobert: That should be fixed now.

openmls/src/group/mls_group/validation.rs
kkohbrok: I was just wondering if we should formally introduce the additional check that we make here: That application messages must originate from a group member.

openmls/src/group/mls_group/validation.rs
kkohbrok: I meant changing the actual label attached to ValSem7. Since the current official index of checks with their associated labels is the PR description, my suggestion was to change it there. I.e. "Membership tag presence" -> "Membership tag presence for plaintext messages from group members". Generally, the style of label is not uniform, but maybe we should deal with the labels when we commit them to the rust book or similar.

kkohbrok (approved):
Looks good! I commented on the two remaining issues, but we can get this in either way. Thanks!

PR #570, opened 2021-11-23
Diff mechanics for the binary tree

kkohbrok:
This PR adds the diff functionality to the binary tree.

There are still a lot of warnings regarding unused code. I expect those to disappear with following PRs when we integrate treesync into the rest of the code.

Open questions:
- Is the file/folder structure ok? Do we need the type alias indirection? 
- Is there a way to make the tests more programmatic, i.e. test for multiple tree sizes generically? How to generate known-good answers?
- Should there be an additional struct layer that enforces tree/diff invariants with more complex functions on top of it? E.g. to enforce, for example, that `add_node` and `remove_node` are used instead of accessing the diff directly, etc.


openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,602 @@
...
+        } else {
+            right(subtree_root_node_index, self.size())?
+        };
+
+        NodeReference::try_from_node_index(self, copath_node_index)
+    }
+
+    /// Returns a vector of `NodeReference`s, where the first reference is to
+    /// the root of the shared subtree of the two given leaf indices followed by
+    /// references to the nodes in its direct path. Returns an error if either
```
raphaelrobert, line 289: Maybe worth clarifying that 'it' refers to the common ancestor and not to either leaf index.

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,602 @@
...
+            return Err(ABinaryTreeDiffError::TreeTooLarge);
+        }
+        // Check that we're extending the tree by at most one.
+        if node_index > self.size() {
+            return Err(ABinaryTreeDiffError::ExtendingOutOfBounds);
+        }
+        self.diff.insert(node_index, node);
+        // Finally, check if the new node increases the size of the diff.
+        if node_index == self.size() {
+            self.size = node_index + 1
```
raphaelrobert, line 513: I guess this is where the size check should go instead.

raphaelrobert (commented):
I did a first pass and left some questions

openmls/src/binary_tree/array_representation.rs
```diff
@@ -1,258 +1,10 @@
...
-}
-
-#[cfg(test)]
-impl PartialEq for ABinaryTree<u32> {
-    fn eq(&self, other: &Self) -> bool {
-        self.nodes == other.nodes
-    }
-}
+#[allow(dead_code)]
+/// FIXME: There's some dead code in treemath that will be used in treesync in
```
franziskuskiefer, line 8: What do we do about this FIXME?
If it is supposed to stay in, it needs an issue number and the dead code in treeemath should be guarded to avoid warnings.

openmls/src/binary_tree/array_representation.rs
```diff
@@ -1,258 +1,10 @@
-//! This module contains an implementation of a binary tree based on an array
-//! representation.
+#[cfg(any(feature = "test-utils", test))]
```
franziskuskiefer, line 1: What's this module doing?
Please add some doc comments about it.

openmls/src/binary_tree/array_representation/tree.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// Return the number of nodes in the tree.
+    pub(in crate::binary_tree) fn size(&self) -> NodeIndex {
+        let len = self.nodes.len();
+        debug_assert!(len <= u32::MAX as usize);
+        self.nodes.len() as u32
+    }
+
+    /// Return the number of leaves in the tree.
+    pub(crate) fn leaf_count(&self) -> TreeSize {
+        (self.size() + 1) / 2
```
franziskuskiefer, line 82: This will overflow for `self.size() == usize::MAX`

openmls/src/binary_tree/array_representation/tree.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /// Creates and returns an empty `AbDiff`.
+    pub(crate) fn empty_diff(&self) -> AbDiff<'_, T> {
+        self.into()
+    }
+
+    /// Merges the changes applied to the `StagedAbDiff` into the tree. This
+    /// function should not fail and only returns a `Result`, because it might
+    /// throw a [LibraryError](ABinaryTreeError::LibraryError).
```
franziskuskiefer, line 114: Please use something like 

```suggestion
/// throw a [`LibraryError`].
```

wherever possible to keep the comments more readable in code.

openmls/src/binary_tree/array_representation/tree.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub(crate) fn merge_diff(&mut self, diff: StagedAbDiff<T>) -> Result<(), ABinaryTreeError> {
+        // The diff size should fit into a 32 bit usize.
+        let diff_size = usize::try_from(diff.size()).map_err(|_| ABinaryTreeError::LibraryError)?;
+        // If the size of the diff is smaller than the tree, truncate the tree
+        // to the size of the diff.
```
franziskuskiefer, line 119: Why?

openmls/src/binary_tree/array_representation/tree.rs
```diff
@@ -0,0 +1,165 @@
...
+                // something has gone wrong.
+                node_index if node_index > self.size() => {
+                    return Err(ABinaryTreeError::LibraryError)
+                }
+                // If the node_index points to somewhere within the size of the
+                // tree, do a swap-remove.
+                node_index => {
+                    // Perform swap-remove.
+                    self.nodes.push(diff_node);
+                    self.nodes.swap_remove(
```
franziskuskiefer, line 137: You want to update the value at `node_index`? Why not just do that?
```rust
self.nodes[node_index] = diff_node;
```
You reallocate memory here each time because you push something to the vector. That's pretty expensive (and complicated) for such a simple operation.

openmls/src/binary_tree/array_representation/tree.rs
```diff
@@ -0,0 +1,165 @@
...
+                        usize::try_from(node_index).map_err(|_| ABinaryTreeError::LibraryError)?,
+                    );
+                }
+            }
+        }
+        Ok(())
+    }
+
+    /// Export the nodes of the tree in the array representation.
+    pub(crate) fn export_nodes(&self) -> Vec<T> {
```
franziskuskiefer, line 149: Do we really need this? I'd prefer having a function that returns a slice and any caller can copy it if needed.

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,602 @@
+//! This module provides the `AbDiff` and `StagedAbDiff` structs that allow
```
franziskuskiefer, line 1: Can you add headings to these files in general?

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,602 @@
...
+    fn from(diff: AbDiff<'a, T>) -> Self {
+        StagedAbDiff {
+            diff: diff.diff,
+            size: diff.size,
+        }
+    }
+}
+
+impl<T: Clone + Debug> StagedAbDiff<T> {
+    pub(super) fn diff(self) -> BTreeMap<NodeIndex, T> {
```
franziskuskiefer, line 48: Can these two functions get comments? :)

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,602 @@
...
+/// be used to access the node at that position or to navigate to other,
+/// neighbouring nodes via the `sibling`, `left_child` and `right_child`
+/// functions of the `AbDiff`.
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct NodeReference {
+    node_index: NodeIndex,
+}
+
+impl NodeReference {
+    /// Creates a new `NodeReference` to a node at the given index. Returns an
```
franziskuskiefer, line 67: Because we've been talking about comment guidelines.
The "Return" part of a comment should always be in a new line (probably separated by a new line) to make it easier to find.

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,602 @@
...
+    /// error if the given index is outside the bounds of the diff.
+    pub(super) fn try_from_node_index<T: Clone + Debug>(
```
franziskuskiefer, line 69: So this is the only function a `NodeReference` has? Do we really need it then?

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,602 @@
...
+    }
+
+    // Helper functions for node addition and removal
+
+    /// Add a node to the diff at the given index. This function can be used
+    /// both to extend the diff (by choosing a `node_index` that is equal to the
+    /// size of the diff) and to place a node at the given index such that any
+    /// previous node in the tree at the same position is replaced upon merging
+    /// the diff. Returns an error if adding the node would increase the size of
+    /// the diff beyond `NodeIndex::max_value()` or if the given node index is
```
franziskuskiefer, line 499: Again, please no `max_value` 😊

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,602 @@
...
+    /// larger than the current size of the diff.
+    fn add_node(&mut self, node_index: NodeIndex, node: T) -> Result<(), ABinaryTreeDiffError> {
+        // Prevent the tree from becoming too large.
+        if self.size() == NodeIndex::max_value() {
+            return Err(ABinaryTreeDiffError::TreeTooLarge);
+        }
+        // Check that we're extending the tree by at most one.
```
franziskuskiefer, line 506: I think doing extending and replacing in the same function is error-prone. How about doing to separate functions?

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,602 @@
...
+        let node_index = to_node_index(leaf_index);
+        self.add_node(node_index, new_leaf)?;
+        Ok(())
+    }
+
+    /// Obtain a `NodeReference` to the leaf with the given `LeafIndex`. Returns
+    /// an error if the given leaf index does not correspond to a leaf in the
+    /// diff.
+    pub(crate) fn leaf(
+        &'a self,
```
franziskuskiefer, line 146: Why do you need the lifetime here? It's not used anywhere/

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,602 @@
...
+        } else if leaf_index_1 >= self.leaf_count() || leaf_index_2 >= self.leaf_count() {
+            return Err(ABinaryTreeDiffError::OutOfBounds);
+        }
+        let subtree_root_node_index =
+            lowest_common_ancestor(to_node_index(leaf_index_1), to_node_index(leaf_index_2));
+        let leaf_index_1_direct_path = direct_path(to_node_index(leaf_index_1), self.size())?;
+
+        leaf_index_1_direct_path
+            .iter()
+            .position(|&direct_path_node_index| direct_path_node_index == subtree_root_node_index)
```
franziskuskiefer, line 252: This is a nice additional check but I think it can also get expensive.
How about we only do it for debug builds?

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,602 @@
...
+    /// Returns `NodeReference` to the copath node of the `leaf_index_1` that is
+    /// in the direct path of `leaf_index_2`. Returns an error if both leaf
+    /// indices are identical or if one of the leaf indices does not correspond
+    /// to a leaf in the diff.
+    pub(crate) fn subtree_root_copath_node(
+        &'a self,
+        leaf_index_1: LeafIndex,
+        leaf_index_2: LeafIndex,
+    ) -> Result<NodeReference, ABinaryTreeDiffError> {
+        if leaf_index_1 == leaf_index_2 {
```
franziskuskiefer, line 266: These checks are duplicate code. Can you pull them out into a separate check function?

franziskuskiefer (changes requested):
Only a partial review so far. But there are some comments already. I'll finish it later today.

kkohbrok (commented):
Thanks for all the comments! I believe I've addressed or at least replied to all of them.

openmls/src/binary_tree/array_representation.rs
kkohbrok: I've removed it here and I've also removed the `allow` one module level higher. We've got lots of warnings now, but I'll deal with them the binary tree and treesync are integrated.

openmls/src/binary_tree/array_representation/tree.rs
```diff
@@ -0,0 +1,176 @@
...
+    /// from left to right. This function should not fail and only returns a
+    /// [`Result`], because it might throw a
+    /// [`LibraryError`](ABinaryTreeError::LibraryError).
+    pub(crate) fn leaves(&self) -> Result<Vec<&T>, ABinaryTreeError> {
+        let mut leaf_references = Vec::new();
+        for leaf_index in 0..self.leaf_count()? {
+            let node_index = usize::try_from(to_node_index(leaf_index))
+                // The tree size and thus the leaf count should fit into usize on 32
+                // bit architectures.
+                .map_err(|_| ABinaryTreeError::LibraryError)?;
```
franziskuskiefer, line 107: Because you where complaining about having to use for loops.
You can always `collect::<Result<Vec<_>, _>>` and them handle the error later.

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,602 @@
...
+        }
+        Ok(())
+    }
+
+    /// Removes a node from the right edge of the diff, thus decreasing the size
+    /// of the diff by one. Throws an error if this would make the diff too
+    /// small (i.e. < 1 node).
+    fn remove_node(&mut self) -> Result<(), ABinaryTreeDiffError> {
+        // First make sure that the tree isn't getting too small.
+        if self.size() <= 1 {
```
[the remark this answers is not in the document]
franziskuskiefer: Well, it's pretty easy to get an empty diff. There's a function called `empty_diff` that creates one.

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,634 @@
...
+        &mut self,
+        parent_node: T,
+        new_leaf: T,
+    ) -> Result<LeafIndex, ABinaryTreeDiffError> {
+        // Prevent the tree from becoming too large.
+        if self.size() >= NodeIndex::MAX - 1 {
+            return Err(ABinaryTreeDiffError::TreeTooLarge);
+        }
+        let original_size = self.size();
+        self.diff.insert(original_size, parent_node);
```
franziskuskiefer, line 128: `insert` returns a value if there has been one before. This shouldn't happen here but would be a good check for a `debug_assert`.
The same is true for the other operations on the diff.

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,634 @@
...
+    ///
+    /// Returns an error if the given leaf index is larger than the leaf count
+    /// of the diff.
+    pub(crate) fn replace_leaf(
+        &mut self,
+        leaf_index: LeafIndex,
+        new_leaf: T,
+    ) -> Result<(), ABinaryTreeDiffError> {
+        let node_index = to_node_index(leaf_index);
+        self.replace_node(node_index, new_leaf)?;
```
franziskuskiefer, line 155: No `?` and `Ok(())` needed.

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,634 @@
...
+    }
+
+    /// Returns the size of the diff.
+    pub(in crate::binary_tree) fn size(&self) -> NodeIndex {
+        self.size
+    }
+
+    /// Returns the leaf count of the diff.
+    pub(crate) fn leaf_count(&self) -> LeafIndex {
+        (self.size() + 1) / 2
```
franziskuskiefer, line 362: We need to check at least with a `debug_assert` that this doesn't overflow (in case you're certain that it shouldn't).

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,634 @@
...
+
+    // Functions around individual [`NodeReference`]s
+    ///////////////////////////////////////////////
+
+    /// Returns a reference to the node pointed to by the [`NodeReference`].
+    /// Returns an Error if the [`NodeReference`] points to a node outside of the
+    /// bounds of the tree. This can happen, for example, if the node was
+    /// removed while shrinking the diff after the creation of the
+    /// [`NodeReference`].
+    pub(crate) fn try_deref(&self, node_ref: NodeReference) -> Result<&T, ABinaryTreeDiffError> {
```
franziskuskiefer, line 373: Would it make sense to implement the `Deref`/`DerefMut` traits here instead?

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -0,0 +1,634 @@
...
+        leaf_index_1: LeafIndex,
+        leaf_index_2: LeafIndex,
+    ) -> Result<(), ABinaryTreeDiffError> {
+        let node_index_1 = to_node_index(leaf_index_1);
+        let node_index_2 = to_node_index(leaf_index_2);
+
+        self.out_of_bounds(node_index_1)?;
+        self.out_of_bounds(node_index_2)?;
+
+        if leaf_index_1 == leaf_index_2 {
```
franziskuskiefer, line 569: Shouldn't this check come first?

franziskuskiefer (changes requested):
Thanks for the changes. I think it's almost there.

> To be honest, I had not really considered 16bit systems. I could try to make it 16bit compatible. Is that something I should spend time on?

We shouldn't try to be 16bit compatible. But we have to be clear about it and add a check somewhere.

raphaelrobert:
Re 16bit compatibility: I think we should rule that out completely and say so in the README.

openmls/src/binary_tree/array_representation/diff.rs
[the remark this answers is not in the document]
raphaelrobert: So it looks like empty diffs can at least exist temporarily. Maybe they could also exist in scenarios where a Commit only references a PSK proposal or some proprietary proposals (neither of which mandate an update path in the Commit).

kkohbrok (commented):
Thanks again for the comments. I've introduced a dedicated `ArchitectureError` and added a sentence to the readme stating that we don't support 16 bit architectures.

README.md
```diff
@@ -38,6 +38,8 @@ The Github CI also builds (but doesn't test) the following rust targets.
 - aarch64-apple-ios-sim
 - wasm32-unknown-unknown
 
+OpenMLS does not support 16 bit platforms.
```
franziskuskiefer, line 41: What about 8?
You want to say that we support 32 an 64

openmls/src/binary_tree/array_representation/diff.rs
franziskuskiefer: 😄 maybe just `NodeId`?

openmls/src/binary_tree/array_representation/diff.rs
franziskuskiefer: No, just nit picking. Let’s leaf it like this.

openmls/src/binary_tree/array_representation/diff.rs
franziskuskiefer: Ok I see. Yes you'll need keep the `&self` in scope. I think it's the function name again 😉 
How about just `node`?

openmls/src/binary_tree/array_representation/diff.rs
franziskuskiefer: If you move it all the way up, you don't need to call functions if there's an error. The earlier you can return, the better.

raphaelrobert (approved):
Let's get this in, I think it's good enough for now!

franziskuskiefer (approved):
thanks

PR #571, opened 2021-11-23
Continue validation tests

### LINKED ISSUE #527: Make sure that validation checks are carried out in `parse_message`

raphaelrobert:
This PR continues the validation tests that were started in #560.
Fixes #527.

Note that the tests are all carried out on the `ManagedGroup`, which now directly passes it on t the corresponding function `MlsGroup`, so that both are being tested.


openmls/src/group/tests/test_validation.rs
```diff
@@ -213,3 +224,371 @@ fn test_valsem2() {
...
+        .expect("Could not add member.");
+
+    let serialized_message = message
+        .tls_serialize_detached()
+        .expect("Could not serialize message.");
+
+    let mut plaintext = VerifiableMlsPlaintext::tls_deserialize(&mut serialized_message.as_slice())
+        .expect("Could not deserialize message.");
+
+    plaintext.set_epoch(GroupEpoch(100));
```
kkohbrok, line 251: Since the Epoch can be both too high and too low, it would be nice to have one test case for each.

kkohbrok (changes requested):
Nice to see the validation in action 🙂. So far, I can only see negative tests. We kind of do positive tests implicitly through the scenarios and the test vectors, but it would be nice to have an explicit positive test alongside each negative test here as well.

raphaelrobert:
> Nice to see the validation in action 🙂. So far, I can only see negative tests. We kind of do positive tests implicitly through the scenarios and the test vectors, but it would be nice to have an explicit positive test alongside each negative test here as well.

Thanks for the review! I'm usually all for positive tests as well. In these tests however I take the same Commit message most of the time and I manipulate it. The positive test would be that the Commit message can be parsed and processed. I can add that, but I wonder how beneficial that is.

kkohbrok:
I'd prefer adding positive tests, just to be methodical. Can't you clone the commit before manipulation and process it after processing the manipulated message has failed? That would to the trick and I would imagine that that's relatively low-effort.

kkohbrok (approved):
Thanks a bunch! Let's get it in 🙂

PR #576, opened 2021-11-25
Remove callbacks in ManagedGroup

### LINKED ISSUE #524: Remove `callback` mechanism in MangedGroup API
This is a follow-up after the changes of #522, which make the callback functionality unnecessary.

raphaelrobert:
⚠️ WARNING ⚠️: This PR introduces important breaking changes in how incoming messages are processed in `ManagedGroup`.

Fixes #524.

This PR removes the existing callbacks for events and auto-save. The vents are replaced by the new message type `ProcessedMessage` that is inspectable by applications.
The auto-save callback is replaced by a flag that indicates whether the internal group state has changed and that can be inspected by calling `.state_changed()`.

Note that the new API is more explicit and therefore more verbose, which increases the size of the tests.


openmls/src/messages/proposals.rs
```diff
@@ -183,7 +183,7 @@ pub struct AddProposal {
 
 impl AddProposal {
     /// Get a reference to the key package in the proposal.
-    pub(crate) fn key_package(&self) -> &KeyPackage {
+    pub fn key_package(&self) -> &KeyPackage {
```
kkohbrok, line 186: Do I understand correctly, that those functions need to be public now for inspection by the consumer?

openmls/tests/test_managed_group.rs
```diff
@@ -258,43 +243,76 @@ fn managed_group_operations() {
...
-
-            // Check that we received the correct event
-            match events.last().expect("Expected an event to be returned") {
-                GroupEvent::ApplicationMessage(application_message_event) => {
-                    assert_eq!(application_message_event.sender(), &alice_credential);
-                    assert_eq!(application_message_event.message(), message_alice);
-                }
-                _ => unreachable!("Expected an ApplicationMessage event"),
+
+            let unverified_message = bob_group
```
kkohbrok, line 247: Would it be worth to introduce a combined parse-process function? Even if just for testing purposes?

kkohbrok (commented):
Looks good already with a few questions.

openmls/src/messages/proposals.rs
raphaelrobert: Yes, consumers can now inspect this through the `StagedCommit` they might receive when processing messages.

openmls/tests/test_managed_group.rs
raphaelrobert: The thought had crossed my mind. It really only affects the tests, because I don't think we should expose that to consumers (they should be forced to inspect processed messages). I don't have a strong opinion about it though, if you think it's worth it, I'll put it in.

kkohbrok: No worries. It's effectively already in the test framework, so we should be fine. And should it lead to extensive code duplication in the future, we can still add it then.

openmls/src/group/managed_group/processing.rs
```diff
@@ -1,164 +1,66 @@
...
+                &self.own_kpbs,
+                backend,
+            )
+            .map_err(|e| e.into())
+    }
+
+    /// Stores a standalone proposal in the internal [ProposalStore]
+    pub fn store_pending_proposal(&mut self, proposal: StagedProposal) {
+        // Store the proposal in in the internal ProposalStore
+        self.proposal_store.add(proposal);
```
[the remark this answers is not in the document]
raphaelrobert: Storing proposals will always change the state.

Persisting the state is somewhat expensive, depending on the size of the group. Consumers will have to come up with a tradeoff of when they want to do it. Once the library is panic-free, consumers should be able to do bulk processing of messages and then only persist the modified state at the end.

openmls/src/framing/validation.rs
```diff
@@ -315,8 +315,38 @@ impl VerifiedExternalMessage {
...
 /// [StagedCommit] and [StagedProposal] can be inspected for authorization purposes.
+#[derive(Debug)]
 pub enum ProcessedMessage {
-    ApplicationMessage(Vec<u8>),
+    ApplicationMessage(ApplicationMessage),
     ProposalMessage(Box<StagedProposal>),
     StagedCommitMessage(Box<StagedCommit>),
 }
+
+/// Application message received through a [ProcessedMessage].
```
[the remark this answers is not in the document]
raphaelrobert: According to [this](https://doc.rust-lang.org/rustdoc/linking-to-items-by-name.html), backticks will be stripped anyway which is why I stopped adding them. This will obviously need a pass to harmonise things in the end.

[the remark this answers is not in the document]
franziskuskiefer: Ah that got stabilised. Never mind me.

openmls/src/group/managed_group/mod.rs
```diff
@@ -236,91 +239,19 @@ impl ManagedGroup {
...
     }
 
     /// Group framing parameters
     fn framing_parameters(&self) -> FramingParameters {
         FramingParameters::new(&self.aad, self.managed_group_config.wire_format)
     }
 }
+
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum InnerState {
```
franziskuskiefer, line 254: This needs a description, esp. when public.

openmls/src/messages/proposals.rs
```diff
@@ -208,12 +208,12 @@ impl AddProposal {
     Debug, PartialEq, Clone, Serialize, Deserialize, TlsDeserialize, TlsSerialize, TlsSize,
 )]
 pub struct UpdateProposal {
-    pub(crate) key_package: KeyPackage,
+    pub key_package: KeyPackage,
```
franziskuskiefer, line 211: Why does the member need to be public? There's the `key_package` function below to get it.

franziskuskiefer (approved):
lgtm

kkohbrok (approved):

kkohbrok (approved):

PR #587, opened 2021-11-28
Correct decryption methods for AEAD

adzialocha:
The given decryption implementations for AEAD in `rust_crypto` for the cases `ChaCha20Poly1305` and `Aes256Gcm` did not use the regarding decryption but encryption methods of `aes-gcm`. I assume this bug was introduced by copy / paste, this PR should fix it :+1:

### DIFF
--- a/rust_crypto/src/lib.rs
+++ b/rust_crypto/src/lib.rs
@@ -184,14 +184,14 @@ impl OpenMlsCrypto for RustCrypto {
             AeadType::Aes256Gcm => {
                 let aes =
                     Aes256Gcm::new_from_slice(key).map_err(|_| CryptoError::CryptoLibraryError)?;
-                aes.encrypt(nonce.into(), Payload { msg: ct_tag, aad })
+                aes.decrypt(nonce.into(), Payload { msg: ct_tag, aad })
                     .map(|r| r.as_slice().into())
                     .map_err(|_| CryptoError::AeadDecryptionError)
             }
             AeadType::ChaCha20Poly1305 => {
                 let aes = ChaCha20Poly1305::new_from_slice(key)
                     .map_err(|_| CryptoError::CryptoLibraryError)?;
-                aes.encrypt(nonce.into(), Payload { msg: ct_tag, aad })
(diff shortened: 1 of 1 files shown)


rust_crypto/src/lib.rs
```diff
@@ -184,14 +184,14 @@ impl OpenMlsCrypto for RustCrypto {
...
-                aes.encrypt(nonce.into(), Payload { msg: ct_tag, aad })
+                aes.decrypt(nonce.into(), Payload { msg: ct_tag, aad })
                     .map(|r| r.as_slice().into())
                     .map_err(|_| CryptoError::AeadDecryptionError)
             }
             AeadType::ChaCha20Poly1305 => {
                 let aes = ChaCha20Poly1305::new_from_slice(key)
                     .map_err(|_| CryptoError::CryptoLibraryError)?;
-                aes.encrypt(nonce.into(), Payload { msg: ct_tag, aad })
+                aes.decrypt(nonce.into(), Payload { msg: ct_tag, aad })
```
franziskuskiefer, line 194: The variable name should get changed to something like `chacha_poly`.
```suggestion
                chacha_poly.decrypt(nonce.into(), Payload { msg: ct_tag, aad })
```

franziskuskiefer (changes requested):
Thanks!
Indeed, looks like a copy paste issue. Interesting that it didn't blow up.
One more thing to update though.

raphaelrobert:
Nice catch! I would also have suspected that this would have been caught by tests earlier (at least for ChaCha20Poly1305, AESGCM256 is not used by MLS). This means we need better testing, I filed #588.

adzialocha:
I've changed the variable name to `chacha_poly` @franziskuskiefer  :+1:

> Nice catch! I would also have suspected that this would have been caught by tests earlier.

Yes, I was surprised myself, it started troubleshooting by reading the `aes-gcm` code and specification since the "decrypted" plaintext was suddenly 32 bytes larger and I was sure I didn't understand something about AES :sweat_smile:

franziskuskiefer:
Clippy and codecov choked for some reason, I'll merge manually.

PR #590, opened 2021-11-30
fixing rustdoc warnings

franziskuskiefer:


raphaelrobert (approved):
Thanks, LGTM! (just one nit)

franziskuskiefer:
> just one nit

do you have that anywhere @raphaelrobert ? 🙂

openmls/src/ciphersuite/mod.rs
```diff
@@ -35,7 +35,7 @@ mod signature;
 
 pub(crate) use aead::*;
 pub use ciphersuites::*;
-pub(crate) use errors::*;
+pub use errors::*;
```
raphaelrobert, line 38: This will be removed in #592 anyway.

raphaelrobert (approved):
Heh, looks like GH swallowed my comment from before.

franziskuskiefer:
Ok, I'll leave the change than as is when it's going to be removed in a follow-up anyway.

Issue #591, opened 2021-12-01
Refactor ciphersuite to remove unwraps


raphaelrobert:
A few `unwrap`s are left in the ciphersuite module that need to be removed by refactoring the ciphersuite.

Issue #438, opened 2021-07-09
Remove all unwraps


franziskuskiefer:
There are still a lot of unwraps in the code. This issue tracks removing all of them.

cc #261

Issue #428, opened 2021-06-16
[Bug] Potential panic when unwrapping in SignedStruct impl of KeyPackage


kkohbrok:
**Describe the bug**
The result of a `from_payload` call is unwrapped even though it could potentially yield an encoding error.

**Expected behavior**
Either the unwrap should be explained (i.e. why it's ok to unwrap here) or the error case should be handled.

**Actual behavior**
The return value is simply unwrapped without associated comment.

PR #592, opened 2021-12-01
Remove most unwraps from library

### LINKED ISSUE #261: Define strategy about how to surface library errors
- Define what the exact rules are to allow `.unwrap()`
 - See where trickling up library errors makes sense

raphaelrobert:
This PR works towards tackling #438. It removes most of the `unwrap`s in the library (see exceptions below).

Fixes #261.

Most `unwrap`s could be removed by propagating `CryptoError` or by refactoring the code. Only very few `LibraryError`s were introduced. Potentially the library errors could be removed completely with more refactoring if necessary.

Exceptions:
 - The `ciphersuite` module still contains `unwrap`s that can only be removed with a bigger refactoring. I filed #591 for that.
 - The `tree` module was only modified to propagate `CryptoError`, the other `unwrap`s were not yet removed since this will be rewritten with the tree rewrite anyway.
 - The `unwrap`s in tests were not yet removed, this will happen in a follow-up PR.
 - #428 is not yet addressed.
 
After this PR we should not introduce new `unwrap`s in the library code.


openmls/src/schedule/mod.rs
```diff
@@ -361,61 +366,76 @@ impl KeySchedule {
             return Err(KeyScheduleError::InvalidState(ErrorState::NotInit));
         }
 
-        Ok(WelcomeSecret::new(
-            backend,
-            self.intermediate_secret.as_ref().unwrap(),
-        ))
+        let intermediate_secret = match &self.intermediate_secret {
```
kkohbrok, line 369: Can't we use `.as_ref().ok_or(KeyScheduleError::LibraryError)?` here and where we use the same pattern below? Not a strong opinion, but I somehow find these `match` patterns with `Some(x) => x` unsatisfactory.

openmls/src/tree/hashes.rs
```diff
@@ -205,14 +210,16 @@ impl RatchetTree {
...
-                ParentHashInput::new(tree, index, former_index_sibling, parent_hash.as_slice())
-                    // FIXME: remove unwrap.
-                    // It is ok to use `unwrap()` here, since we can be sure the node is not blank
-                    .unwrap()
-                    .hash(tree.ciphersuite, backend)
+                Ok(
+                    ParentHashInput::new(tree, index, former_index_sibling, parent_hash.as_slice())
+                        // FIXME: remove unwrap.
+                        // It is ok to use `unwrap()` here, since we can be sure the node is not blank
+                        .unwrap()
```
kkohbrok, line 217: Is there a particular reason we're not propagating the error here?

openmls/src/tree/sender_ratchet.rs
```diff
@@ -46,21 +53,33 @@ impl SenderRatchet {
         if generation <= self.generation {
             let window_index =
                 (self.past_secrets.len() as u32 - (self.generation - generation) - 1) as usize;
-            let secret = self.past_secrets.get(window_index).unwrap().clone();
-            let ratchet_secrets = self.derive_key_nonce(ciphersuite, backend, &secret, generation);
+            let secret = match self.past_secrets.get(window_index) {
```
kkohbrok, line 56: Same as in `schedule`: Any reason we're not using `ok_or(...)?` here?

kkohbrok (changes requested):
Thanks for doing this rather tedious work! Quite a few functions now return `Result`, but I didn't see a lot of changes in the documentation. We have an issue for improvement of documentation for the APIs, but nothing for the internal functions. I think it would make sense to either change the docs as part of this PR (although it's already rather large) or at least create a follow-up issue to improve documentation in general.

openmls/src/schedule/mod.rs
raphaelrobert: I guess it's a style question, but I personally find `match` statements easy to read and understand. I changed it though.

openmls/src/tree/hashes.rs
raphaelrobert: I really only touched the code in `/tree` where necessary. If you think this won't get overwritten by the tree rewrite I can always take this further.

franziskuskiefer:
> although it's already rather large

The PR is a little big indeed 😬  Maybe we can split out the changes in tests (`unwrap` -> `expect`) and otherwise try to do one module per PR or something?

raphaelrobert (commented):
Thanks for the review!

Regarding the documentation: I think there is a good introduction in #590 now: https://github.com/openmls/openmls/blob/8a370ac5af541eeea5627c0abdfedc6e873b6e21/openmls/src/lib.rs#L11
Let me know if you think we should be more explicit for now.

raphaelrobert:
> > although it's already rather large
> 
> The PR is a little big indeed 😬 Maybe we can split out the changes in tests (`unwrap` -> `expect`) and otherwise try to do one module per PR or something?

Sorry this PR turned out to be so big, but I tried to keep it as tight as possible.
As mentioned, I have not touched the tests yet (that's another 650 `unwrap`s to remove 😄). I have also not touched the parts that will need much more refactoring (like `ciphersuite` and one `unwrap` in `key_packages`). I tried splitting it up by module, but bubbling up these errors makes everything very interconnected. Any `unwrap` that I removed led to dozens of changes across the whole codebase. Trying to do it by module now would mean a lot more work for me and probably more code for you to review.

kkohbrok (dismissed):
You have a point with the expressiveness of the match statements. Again, not a strong opinion, but especially for `LibraryError` instances, my preferences are still with `ok_or(...)?`.

PR #595, opened 2021-12-02
Introduce the rstest framework

### LINKED ISSUE #520: Better tests for Evercrypt backend

raphaelrobert:
This PR introduces the [rstest framework](https://crates.io/crates/rstest) to get better testing with all ciphersuites and crypto providers. It is applied to all unit & integration tests.

This is a step towards fixing #520 and fixes #520.

Some details:
 - There is one configuration for both unit & integration tests (there were two before)
 - In addition to the ciphersuites, the backends can now also be specified, which means evercrypt will get tested much more often
 - The `ctest_ciphersuites` macro has been removed, as well as the `test_macros` crate
 - All tests now make use of one of the following templates, depending on their needs: `ciphersuites`, `backends` or `ciphersuites_and_backends`
 - The KAT vectors are written using RustCrypto but are read with all available backends
 - The tree unit tests are not changed given the pending tree rewrite
 
The code changes in this PR are as follows:
 - The new templates are applied to every test (this only changes the function declaration)
 - The templates are defined in `src/test_utils/mod.rs` (that's the most important file to review)
 - A few variables are renamed for consistency
 - The bulk of the changed lines are just new indentation and `rstfmt` kicking in
 - No other changes were made

Issues:
 - The names of the tests when printed in the console don't contain the ciphersuite anymore
 - The clients from the test setup cannot make use of this because they each require their own key store, which mandates a larger rewrite
 - rust analyzer seems to report an issue when applying the templates. It is however a false positive that does not affect the compilation. Fix for VSC in `setting.json`: `"rust-analyzer.diagnostics.disabled": ["unresolved-macro-call"]`
 - rstest is still young and a bit brittle. The issues above might go away in the future.

Issues found: 
 - #600


franziskuskiefer (commented):
Looks like something we should use 👍🏻

kkohbrok (approved):
Cool!

openmls/Cargo.toml
```diff
@@ -27,6 +27,8 @@ getrandom = { version = "0.2", features = ["js"] }
 itertools = { version = "0.10", optional = true }
 openmls_rust_crypto = { version = "0.1", path = "../openmls_rust_crypto", optional = true }
 evercrypt_backend = { version = "0.1", path = "../evercrypt_backend", optional = true }
+rstest = "^0.11"
```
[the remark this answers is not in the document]
kkohbrok: We might even get away with having them only as a dev dependency, as I think they're not really used in any of the `test-utils` functions.

openmls/src/test_utils/mod.rs
```diff
@@ -57,27 +67,144 @@ pub(crate) fn hex_to_bytes(hex: &str) -> Vec<u8> {
...
+            "Using backend: {}, ciphersuite: {}",
+            type_of($backend),
+            $ciphersuite.name()
         );
     };
 }
+
+// === Define backend per platform ===
+
+// For now we only use Evercrypt on specific platforms and only if the feature was enabled
```
kkohbrok, line 114: Why only if the feature was enabled?

kkohbrok (changes requested):
From the PR description: `This is a step towards fixing #520 and fixes #520.` Is this a typo?

I reviewed a few changed tests files and it looks like a good approach to me. Beyond that, I'll trust that the tests didn't change. Due to GHs marking of what's changed and what hasn't it's almost impossible to review.

One comment on the dependencies, otherwise this looks good to go.

openmls/src/test_utils/mod.rs
```diff
@@ -57,27 +67,144 @@ pub(crate) fn hex_to_bytes(hex: &str) -> Vec<u8> {
...
     match hex {
         Some(s) => hex_to_bytes(&s),
         None => vec![],
     }
 }
 
+// === Print debug information about ciphersuites & backends ===
+
+#[allow(unused_macros)]
+macro_rules! debug_show_ciphersuite {
```
franziskuskiefer, line 80: Why are these macros needed but unused?

openmls/src/test_utils/mod.rs
```diff
@@ -57,27 +67,144 @@ pub(crate) fn hex_to_bytes(hex: &str) -> Vec<u8> {
...
+
+#[cfg(any(
+    not(target_arch = "x86_64"),
+    target_os = "macos",
+    target_family = "wasm",
+    not(feature = "evercrypt")
+))]
+#[template]
+#[rstest(backend,
+    case(&OpenMlsRustCrypto::default()),
```
franziskuskiefer, line 137: Can you name all the cases please? :) Otherwise it's impossible to read or filter the tests.
Something like the suggestion here for example (and similar for all the other cases).
```suggestion
    case::rust_crypto(&OpenMlsRustCrypto::default()),
```

franziskuskiefer (commented):
What @kkohbrok said and naming the cases.
Otherwise this looks good to me.

openmls/src/test_utils/mod.rs
raphaelrobert: Because otherwise it simply isn't available

kkohbrok (approved):
Thanks for addressing the comments! 🚢  it!

PR #597, opened 2021-12-02
Replace 'unwrap' by 'expect' in tests

raphaelrobert:
This PR replaces every `unwrap()` by `expect("An unexpected error occurred.")` in all unit & integration tests. It does nothing more than that.
The goal is to spot remaining `unwrap`s more easily and to increase the codecov test coverage.

Sorry for the size of the PR, but it should be easy to review.


kkohbrok (approved):
Looks good, thanks!

Issue #541, opened 2021-11-11
[MLS Spec change] Use smaller, fixed-size hash-based identifiers


raphaelrobert:
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/510

510 super-seeds https://github.com/mlswg/mls-protocol/pull/491

franziskuskiefer:
Blocked by #401, in particular definition of public facing and private functions.

PR #601, opened 2021-12-03
Introduce HashReference

franziskuskiefer:
This is the first PR for #541. It only introduces `HashReference` without using it.


openmls/src/ciphersuite/hash_ref.rs
```diff
@@ -0,0 +1,68 @@
...
+pub struct HashReference {
+    value: Value,
+}
+
+/// A reference to a key package.
+/// This value uniquely identifies a key package.
+pub type KeyPackageRef = HashReference;
+
+/// A reference to a proposal.
+/// THis value uniquely identifies a proposal.
```
raphaelrobert, line 44:
```suggestion
/// This value uniquely identifies a proposal.
```

openmls/src/ciphersuite/hash_ref.rs
```diff
@@ -0,0 +1,68 @@
...
+
+/// A reference to an MLS object computed as an HKDF of the value.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, TlsDeserialize, TlsSerialize, TlsSize)]
+pub struct HashReference {
+    value: Value,
+}
+
+/// A reference to a key package.
+/// This value uniquely identifies a key package.
+pub type KeyPackageRef = HashReference;
```
raphaelrobert, line 41: I'm wondering if we shouldn't go for stronger type safety here. With type aliases, the compiler won't warn us when we confuse the two.

raphaelrobert (changes requested):
Thanks for starting it! Since you're at it, can you add the corresponding `TODO #541` at the relevant places (key packages and proposals)?

kkohbrok (dismissed):
Looks good!

openmls/src/ciphersuite/hash_ref.rs
franziskuskiefer: Using structs is safer but also requires us to write more code. I don't have strong feelings either way but preferred the easier solution here. I don't see that type confusion here would have any security relevance (it will just work or break) hence the type aliases.

kkohbrok (approved):

raphaelrobert (approved):

PR #602, opened 2021-12-03
New Rust version, new clippy warnings

### LINKED ISSUE #585: [Feature request] Convert MlsMessageOut to MlsMessageIn
There are cases where the DS needs to be able to convert `MlsMessageOut` to `MlsMessageIn` without serializing/deserializing.

raphaelrobert:
This PR fixes new clippy warning introduced in v1.57. 
It also fixes #585, because it just meant lifting the test-only visibility of the conversion.

The main change is that `MlsMessageIn/Out` now use `Box`, which makes sense, since they are likely to be used in a `Vec` anyway.

### DIFF
--- a/openmls/src/framing/codec.rs
+++ b/openmls/src/framing/codec.rs
@@ -281,11 +281,11 @@ impl tls_codec::Deserialize for MlsMessageIn {
                 match wire_format {
                     WireFormat::MlsPlaintext => {
                         let plaintext = VerifiableMlsPlaintext::tls_deserialize(&mut chain)?;
-                        Ok(MlsMessageIn::Plaintext(plaintext))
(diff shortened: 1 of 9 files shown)


kkohbrok (approved):
Thanks for picking those up!

kkohbrok:
I think I was a bit premature on this one. Two questions: Why the `Box` in the first place? Also, we briefly discussed this before and I still don't think we should consider requirements of the DS in OpenMLS code. I think a cleaner approach would be to extract something like an `mls_types` or `mls_messages` crate that both OpenMLS and any rust DS implementation can use.

kkohbrok:
I filed #603 to track this idea.

PR #605, opened 2021-12-06
TreeSync integration

kkohbrok:
This PR removes large parts of the old `tree` module (keeping mostly `secrettree` and its dependencies) and instead integrates TreeSync/TreeKEM into the code.

In particular, it introduces the `treesync` module, which contains the `TreeSync` struct, as well as submodules for the diff functionality, TreeKEM, i.e. path encryption and decryption, as well as some submodules for node structs.

This PR touches a lot of the tests, since many tests directly access the tree of a given group, mostly to export nodes.

There is still `treemath` stub in the old `tree` module that contains the functions required for `secret_tree`.

TODOs for follow-up PRs:
- write helper code/framework to generate TreeKEM test vectors
- write tests
- get rid of one of the apply_proposals functions
- make cli, ds and interop client work

Note that this PR is not aimed at `main`.

Topics for discussion
 * Which invariants should the tree enforce?
   * validity/consistency of unmerged leaves in all parent nodes
     * all indices point to non-blank leaves in the subtree of the given node
     * unmerged leaves in a node are consistent with the unmerged leaves in its parent
   * uniqueness of public keys in the tree
   * validity of all parent hashes in the tree
   * tree size is below u32::MAX
   * our own leaf is not blank
   * our own leaf index points to inside of the tree
   * we have a private key for our own leaf
   * our direct path has private keys according to the unmerged leaves in the nodes
   * there are private keys only in our leaf and our direct path
   * the tree hash is non-empty
 * should `export_nodes` return a vector of nodes or node references?
   * in most existing cases, we clone the nodes anyway after exporting (e.g. to create a RatchetTreeExtension)
 * do we need both conditionals in the the ValSem4 check? The second check should imply the first one.


kkohbrok:
Closing in favor of #611.

PR #608, opened 2021-12-07
Move MlsGroup behind a feature flag

### LINKED ISSUE #215: Make MlsGroup private
The `MlsGroup` API should be considered hazmat and should only be used by experienced consumers that want to implement a lot of the basic MLS logic that's not part of the standard themselves. The managed group should be the entry point for most users of the library as exposes a safe API.

raphaelrobert:
⚠️ WARNING ⚠️: This PR introduces breaking changes by renaming `ManagedGroup` to `MlsGroup`

Fixes #215.
 
This PR does the following:
 - Renames `MlsGroup` to `CoreGroup` (just `Group` was too ambiguous in my mind)
 - Renames `ManagedGroup` to `MlsGroup`
 - Makes the `CoreGroup` private and exposes it behind the `coregroup` feature flag with its own `prelude.rs`
 - Attempts to do a bit of decluttering in the main `prelude.rs` by pushing some of it to the `CoreGroup`
 - Remove dead/unused code around `CoreGroup`. Note that some functions now seem unused because External Commits are not fully implemented yet.


franziskuskiefer (changes requested):
Can you please split out the mechanical changes such as renaming to make this reviewable?

raphaelrobert:
I tried massaging `git` in different ways, but it just won't pick up a good part of the file & directory renaming. Even doing it in smaller chunks didn't work. 
As a workaround, I've created #610. It has the same exact content as this PR, except the files & directories are not yet renamed. Obviously, it won't compile because of that, but this way `git` picks up the changes like it should and reviewing it should be easier.

After the review, we can either merge this PR (by backporting potential changes), or do the renaming in the final step in #610. You can let me know which one you prefer.

*Guideline for the review*

Files with interesting changes:
 - `openmls/src/group/mls_group/mod.rs`
 - `openmls/src/group/mls_group/prelude.rs`
 
Other changes:
- WireFormat was moved into `framing/`
- MlsGroupConfig was moved into `src/group/mls_group/mod.rs`
- Lots of renaming

raphaelrobert:
Closing this PR in favor of #605.

Issue #423, opened 2021-06-09
[Meta] Tree Rewrite


kkohbrok:
The whole implementation of the tree-based logic should be rewritten along the lines of the drafted API in the wiki. This is a meta issue to act as target for issues in the code that should go away as soon as the tree is rewritten.

- [x] #424 
- [x] #224
- [x] #514 
- [x] #233 

Issues that arise after merging treesync functionality into `main`:
- [x] #626
- [x] make cli, ds and interop client work

Follow-up issues to increase TreeSync robustness are tracked in #708.

kkohbrok:
Closing this, as the rewrite of TreeSync/TreeKEM is complete. Future robustness improvements are tracked in #708.

Issue #424, opened 2021-06-10
Ensure apply commit operates on provisional tree


franziskuskiefer:
I know that this should be fixed with the tree rewrite. But maybe file an issue for this just to be sure?

_Originally posted by @kkohbrok in https://github.com/openmls/openmls/pull/422#r648294251_

Issue #224, opened 2020-11-20
Improve FS in PrivateTree


kkohbrok:
We currently store the `CommitSecret` and the `PathSecrets` in the Private tree, although we don't have to and we destroy any chance to get in-epoch FS. It should be refactored, such that both values are outputs that get consumed by other functions (e.g. the CommitSecret should get consumed by the derivation of the joiner secret.

PR #611, opened 2021-12-08
TreeSync integration

kkohbrok:
This PR removes large parts of the old `tree` module (keeping mostly `secrettree` and its dependencies) and instead integrates TreeSync/TreeKEM into the code.

In particular, it introduces the `treesync` module, which contains the `TreeSync` struct, as well as submodules for the diff functionality, TreeKEM, i.e. path encryption and decryption, as well as some submodules for node structs.

This PR touches a lot of the tests, since many tests directly access the tree of a given group, mostly to export nodes.

There is still `treemath` stub in the old `tree` module that contains the functions required for `secret_tree`.

TODOs for follow-up PRs:
- write helper code/framework to generate TreeKEM test vectors
- write tests
- get rid of one of the apply_proposals functions
- make cli, ds and interop client work

Note that this PR is not aimed at `main`.

Topics for discussion
 * Which invariants should the tree enforce?
   * validity/consistency of unmerged leaves in all parent nodes
     * all indices point to non-blank leaves in the subtree of the given node
     * unmerged leaves in a node are consistent with the unmerged leaves in its parent
   * uniqueness of public keys in the tree
   * validity of all parent hashes in the tree
   * tree size is below u32::MAX
   * our own leaf is not blank
   * our own leaf index points to inside of the tree
   * we have a private key for our own leaf
   * our direct path has private keys according to the unmerged leaves in the nodes
   * there are private keys only in our leaf and our direct path
   * the tree hash is non-empty
 * should `export_nodes` return a vector of nodes or node references?
   * in most existing cases, we clone the nodes anyway after exporting (e.g. to create a RatchetTreeExtension)
 * do we need both conditionals in the the ValSem4 check? The second check should imply the first one.

See #423 for the meta-issue on the tree rewrite.


openmls/src/group/mls_group/staged_commit.rs
```diff
@@ -243,52 +241,35 @@ impl MlsGroup {
...
-
         // If there is a key package from the Commit's update path, add it to the proposal queue
         // TODO #424: This won't be necessary anymore, we can just apply the proposals first
         // and add a new fake Update proposal to the queue after that
         if let Some(key_package) = path_key_package {
             let proposal = Proposal::Update(UpdateProposal { key_package });
-            let staged_proposal =
-                StagedProposal::from_proposal_and_sender(ciphersuite, backend, proposal, sender)
-                    .map_err(|_| MlsGroupError::LibraryError)?;
+            let staged_proposal = StagedProposal::from_proposal_and_sender(
```
raphaelrobert, line 255: #424 should be addressed with the tree rewrite as well.

openmls/src/group/mls_group/staged_commit.rs
```diff
@@ -121,16 +121,15 @@ impl MlsGroup {
         }
 
         // Determine if Commit is own Commit
-        let sender_index = sender.to_leaf_index();
-        let is_own_commit = sender_index == provisional_tree.own_node_index();
+        let sender = mls_plaintext.sender_index();
```
raphaelrobert, line 124: Comment for the lines above: If we were removed from the group, we should also return a diff that can be applied to the group in order to reflect the new membership state.

openmls/src/tree/tests_and_kats/unit_tests/test_secret_tree.rs
```diff
@@ -1,9 +1,11 @@
 use openmls_rust_crypto::OpenMlsRustCrypto;
 use openmls_traits::random::OpenMlsRand;
 
+use crate::prelude::ProtocolVersion;
```
raphaelrobert, line 4: Let's not use prelude anymore for unit tests.

openmls/src/tree/tests_and_kats/unit_tests/test_sender_ratchet.rs
```diff
@@ -3,14 +3,14 @@ use openmls_rust_crypto::OpenMlsRustCrypto;
 use crate::tree::sender_ratchet::SenderRatchet;
 
 use crate::config::Config;
-use crate::prelude::{LeafIndex, Secret};
+use crate::prelude::Secret;
```
raphaelrobert, line 6: Let's not use prelude anymore for unit tests.

openmls/src/group/mls_group/create_commit.rs
```diff
@@ -48,40 +49,59 @@ impl MlsGroup {
...
-        let mut provisional_tree = RatchetTree::new_from_public_tree(&self.tree());
+        let mut diff: TreeSyncDiff = self.tree().empty_diff()?;
 
         // Apply proposals to tree
         let apply_proposals_values =
-            match provisional_tree.apply_proposals(backend, proposal_queue, &[]) {
-                Ok(res) => res,
-                Err(_) => return Err(CreateCommitError::OwnKeyNotFound.into()),
-            };
+            self.apply_proposals(&mut diff, backend, proposal_queue, &[])?;
```
raphaelrobert, line 56: This doesn't have to be part of this PR, but I think it's worthwhile mentioning it now:
We discussed that it would make sense to return a `StagedCommit` as well as the actual commit message. I'm wondering to what extent that would change the strategy here already.

raphaelrobert (changes requested):
Amazing you tackled this! I did a first pass and flagged some things (some of which might need offline discussion).

openmls/src/group/mls_group/create_commit.rs
```diff
@@ -124,18 +143,19 @@ impl MlsGroup {
 
         let joiner_secret = JoinerSecret::new(
             backend,
-            provisional_tree.commit_secret(),
+            commit_secret_option.as_ref(),
```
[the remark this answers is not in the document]
kkohbrok: Good catch! This relates to #224, which we can now fix by changing the signature as you suggest. Previously, the CommitSecret was stored in the PrivateTree. Now it's a return value and we can actually consume it here.

openmls/src/group/mls_group/create_commit.rs
```diff
@@ -124,18 +143,19 @@ impl MlsGroup {
...
 
         // Create group secrets for later use, so we can afterwards consume the
         // `joiner_secret`.
-        let plaintext_secrets = PlaintextSecret::new(
+        let plaintext_secrets = PlaintextSecret::from_plain_update_path(
+            &diff,
             &joiner_secret,
             apply_proposals_values.invitation_list,
-            &provisional_tree,
+            plain_path_option.as_deref(),
```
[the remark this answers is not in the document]
kkohbrok: I can change the signature to use `impl`, but I think I still have to use `as_deref` here, because in the end I want `Option<&'a [PlainUpdatePathNode]>` not `&Option<Vec<PlainUpdatePathNode>>`.

openmls/src/group/mls_group/staged_commit.rs
kkohbrok: Ah, right. That's not what's happening right now, though, is it? So it should be a FIXME, ideally with an issue number? Also, there is a trade-off here, as this would require us to allow a valid TreeSync state, where the `own_leaf_index` potentially points somewhere outside of the tree.

openmls/src/group/mls_group/staged_commit.rs
kkohbrok: With the tree rewrite, path processing implies replacing the current leaf node of the committer with the one in the path. Do I understand correctly, that we can then remove this whole `if let ...`?

kkohbrok (commented):
Thanks for the feedback!

openmls/src/treesync/treekem.rs
```diff
@@ -0,0 +1,314 @@
...
+        path: &[PlainUpdatePathNode],
+        group_context: &[u8],
+        exclusion_list: &HashSet<&LeafIndex>,
+        key_package: &KeyPackage,
+    ) -> Result<UpdatePath, TreeKemError> {
+        let copath_resolutions = self.copath_resolutions(self.own_leaf_index(), exclusion_list)?;
+
+        let mut update_path_nodes = Vec::new();
+        // Encrypt the secrets
+        for (node, resolution) in path.iter().zip(copath_resolutions.iter()) {
```
franziskuskiefer, line 52: Do we know the lengths of the two match?
And you can pre-allocate the vector again.

openmls/src/treesync/treekem.rs
```diff
@@ -0,0 +1,314 @@
...
+            }
+        }
+
+        // Finally, we append the derived path to the part of the update path
+        // below the first node that we have a private key for.
+        let mut path: Vec<ParentNode> = update_path
+            .nodes()
+            .iter()
+            .take(path_position)
+            .map(|update_path_node| update_path_node.public_key().clone().into())
```
franziskuskiefer, line 124: Is it necessary that `update_path` is a reference? If not we could avoid cloning here.

openmls/src/treesync/treekem.rs
```diff
@@ -0,0 +1,314 @@
...
+    fn encrypted_path_secrets(&self, ciphertext_index: usize) -> Option<&HpkeCiphertext> {
+        self.encrypted_path_secrets.get(ciphertext_index)
+    }
+
+    /// Return the `public_key`.
+    fn public_key(&self) -> &HpkePublicKey {
+        &self.public_key
+    }
+
+    /// Flip the last byte of every `encrypte_path_secret` in this node.
```
franziskuskiefer, line 161: encrypted_path_secret

openmls/src/treesync/diff.rs
```diff
@@ -0,0 +1,806 @@
...
+    /// This turns the diff into a staged diff. In the process, the diff
+    /// computes and sets the new tree hash.
+    pub(crate) fn into_staged_diff(
+        mut self,
+        backend: &impl OpenMlsCryptoProvider,
+        ciphersuite: &Ciphersuite,
+    ) -> Result<StagedTreeSyncDiff, TreeSyncDiffError> {
+        let new_tree_hash = self.compute_tree_hash(backend, ciphersuite)?;
+        #[cfg(test)]
+        self.verify_parent_hashes(backend, ciphersuite)
```
franziskuskiefer, line 641: Why the `cfg(test)`? Shouldn't we do this at least for debug builds?

openmls/src/treesync/diff.rs
```diff
@@ -0,0 +1,806 @@
...
+            diff: self.diff.into(),
+            new_tree_hash,
+        })
+    }
+
+    /// Helper function to compute and set the tree hash of the given node and
+    /// all nodes below it in the tree. This function respects cached tree hash
+    /// values. If a cached value is found it is returned without further
+    /// computation of hashes of the node or the nodes below it.
+    fn set_tree_hash(
```
franziskuskiefer, line 653: Something like `compute_tree_hash`? `set_` doesn't sound like computing and the cashing is confusing then.

openmls/src/group/mls_group/create_commit.rs
```diff
@@ -48,40 +61,62 @@ impl MlsGroup {
...
+                    key_package_bundle.key_package(),
+                )?;
+                PathProcessingResult {
+                    commit_secret: Some(commit_secret),
+                    encrypted_path: Some(encrypted_path),
+                    plain_path: Some(plain_path),
+                    key_package_bundle: Some(key_package_bundle),
+                }
+            } else {
+                // If path is not needed, return empty commit secret
```
raphaelrobert, line 112:
```suggestion
                // If path is not needed, return empty commit secret
```
```suggestion
                // If path is not needed, return empty path processing results
```

openmls/src/schedule/mod.rs
```diff
@@ -269,12 +260,13 @@ impl JoinerSecret {
     /// `CommitSecret` as input. This should change with #224.
     pub(crate) fn new<'a>(
```
franziskuskiefer, line 261: Remove the lifetime if it's not needed anymore.

openmls/src/messages/mod.rs
```diff
@@ -312,6 +308,91 @@ impl From<Secret> for PathSecret {
...
+
+    /// Consume the `PathSecret`, returning the internal `Secret` value.
+    pub(crate) fn secret(self) -> Secret {
+        self.path_secret
+    }
+
+    /// Decrypt a given `HpkeCiphertext` using the `private_key` and `group_context`.
+    ///
+    /// Returns the decrypted `PathSecret`. Returns an error if the decryption
+    /// was unsuccessul.
```
franziskuskiefer, line 369: unsuccessful

openmls/src/messages/mod.rs
```diff
@@ -312,6 +308,91 @@ impl From<Secret> for PathSecret {
     }
 }
 
+impl PathSecret {
```
franziskuskiefer, line 311: Was this code moved? If so why? Was it changed?

openmls/tests/test_managed_group.rs
```diff
@@ -167,10 +167,21 @@ fn managed_group_operations(
         }
 
         // Check that the group now has two members
-        assert_eq!(alice_group.members().len(), 2);
+        assert_eq!(
+            alice_group
+                .members()
+                .expect(
+                    "error getting members from
```
franziskuskiefer, line 174: formatting hickup?

franziskuskiefer (changes requested):
Thanks that's a big one.
I left a couple coments. Nothing major I think.

openmls/src/group/mls_group/staged_commit.rs
```diff
@@ -243,52 +241,35 @@ impl MlsGroup {
...
         if &own_confirmation_tag != received_confirmation_tag {
-            // FIXME: reset nodes. This should get fixed with the tree rewrite.
-            provisional_tree.nodes = original_nodes;
             log::error!("Confirmation tag mismatch");
             log_crypto!(trace, "  Got:      {:x?}", received_confirmation_tag);
             log_crypto!(trace, "  Expected: {:x?}", own_confirmation_tag);
             return Err(StageCommitError::ConfirmationTagMismatch.into());
         }
 
-        // Verify KeyPackage extensions
```
raphaelrobert, line 254: Is this now done within TreeSync?

openmls/src/group/mls_group/staged_commit.rs
raphaelrobert: Let's tackle #424 in a follow-up PR

openmls/src/group/mls_group/staged_commit.rs
```diff
@@ -297,29 +278,21 @@ impl MlsGroup {
...
                 secret_tree: RefCell::new(secret_tree),
-                original_nodes,
+                staged_diff,
             }),
         })
     }
 
     /// Merges a [StagedCommit] into the group state.
-    pub fn merge_commit(&mut self, staged_commit: StagedCommit) {
+    pub fn merge_commit(&mut self, staged_commit: StagedCommit) -> Result<(), MlsGroupError> {
```
raphaelrobert, line 287: What kind of error do we expect here? Is it anything more meaningful than a library error?

openmls/src/group/mls_group/staged_commit.rs
```diff
@@ -297,29 +278,21 @@ impl MlsGroup {
                 epoch_secrets: provisional_epoch_secrets,
                 interim_transcript_hash,
                 secret_tree: RefCell::new(secret_tree),
-                original_nodes,
+                staged_diff,
```
raphaelrobert, line 281: For the line above: Does `secret_tree` still need to be a`RefCell`?

openmls/src/group/mls_group/staged_commit.rs
```diff
@@ -243,52 +241,35 @@ impl MlsGroup {
...
         }
 
         // Create a secret_tree, consuming the `encryption_secret` in the
         // process.
         let secret_tree = provisional_epoch_secrets
             .encryption_secret()
-            .create_secret_tree(provisional_tree.leaf_count());
+            .create_secret_tree(diff.leaf_count());
+
+        // Make the diff a staged diff.
```
raphaelrobert, line 271:
```suggestion
        // Make the diff a staged diff.
```
```suggestion
        // Make the diff a staged diff. This finalizes the diff and no more changes can be applied to it.
```

raphaelrobert (changes requested):
Did another pass with a few follow-up questions and small fixes.

openmls/src/treesync/treekem.rs
kkohbrok: Good catch! I checked and it turns out we can avoid cloning here and in a few other places. I refactored accordingly.

openmls/tests/test_managed_group.rs
kkohbrok: Yup!

kkohbrok (commented):
Thanks for all the great feedback! I made a bunch of changes, although there seems to be a bug to do with parent hash verification that only surfaces occasionally. I'll take a look, but it might delay merging for a bit.

openmls/src/treesync/node/parent_node.rs
```diff
@@ -0,0 +1,211 @@
...
+    /// intermediary `PathSecret`s and the [`CommitSecret`].
+    pub(crate) fn derive_path(
+        backend: &impl OpenMlsCryptoProvider,
+        ciphersuite: &Ciphersuite,
+        path_secret: PathSecret,
+        path_length: usize,
+    ) -> Result<PathDerivationResult, ParentNodeError> {
+        let mut path = Vec::new();
+        let mut update_path_nodes = Vec::new();
+        let mut path_secret_option = Some(path_secret);
```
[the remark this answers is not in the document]
franziskuskiefer: But why do you need an option for that? Just do a `let mut next_path_secret = path_secret;` here and get rid of all the `Option` stuff.

openmls/src/treesync/node/codec.rs
```diff
@@ -0,0 +1,138 @@
...
+        let unmerged_leaves = TlsSliceU32(self.unmerged_leaves());
+        self.public_key().tls_serialized_len()
+            + parent_hash.tls_serialized_len()
+            + unmerged_leaves.tls_serialized_len()
+    }
+}
+
+impl tls_codec::Size for &ParentNode {
+    fn tls_serialized_len(&self) -> usize {
+        let parent_hash = TlsByteSliceU8(self.parent_hash());
```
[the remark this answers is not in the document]
franziskuskiefer: Yeah, just make the fields accessible here.

openmls/src/treesync/diff.rs
```diff
@@ -0,0 +1,806 @@
...
+        for node_id in &full_path {
+            // If sibling is not a blank, return its HpkePublicKey.
+            let sibling_id = self.diff.sibling(*node_id)?;
+            let resolution = self.resolution(sibling_id, excluded_indices)?;
+            copath_resolutions.push(resolution);
+        }
+        Ok(copath_resolutions)
+    }
+
+    /// Verify the parent hashes of all nodes in the tree.
```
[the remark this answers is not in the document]
franziskuskiefer: I guess this is related to the other comment about the complex shenanigan you go through to call a simple tree function.

If I understand correctly the issue is that you have a bunch of functions that you'd like to call on the tree as well as the diff. And they are all on the diff.
I understand the issue but I don't think the solution right now is what we should be doing (way too complex).
Do you need a lot of member functions? At least this one here could work well standalone with the necessary inputs and an object that provides the required tree traversal functions.
Let me know if this makes sense or if you want to chat about it.

openmls/src/treesync/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+            ts_nodes.push(ts_node_option);
+        }
+        let tree = MlsBinaryTree::new(ts_nodes)?;
+        if let Some(leaf_index) = own_index_option {
+            let mut tree_sync = Self {
+                tree,
+                tree_hash: vec![],
+                own_leaf_index: leaf_index,
+            };
+            let diff = tree_sync.empty_diff()?;
```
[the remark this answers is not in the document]
franziskuskiefer: See the comment in diff.rs

openmls/src/treesync/node.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// Returns the public key of this node.
+    pub(crate) fn public_key(&self) -> &HpkePublicKey {
+        match self {
+            Node::LeafNode(ln) => ln.public_key(),
+            Node::ParentNode(pn) => pn.public_key(),
+        }
+    }
+
+    /// Returns the private key of this node.
+    pub(in crate::treesync) fn private_key(&self) -> &Option<HpkePrivateKey> {
```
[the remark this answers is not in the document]
franziskuskiefer: `&Option<T>` is most likely consuming the value of `T` (has to, depending on what it is doing). You don't want a reference to `Option` but a reference of `T` here. (And you can get from `Option<&T>` to `&Option<T>` but not the other way.)

franziskuskiefer (changes requested):
I get a bunch of warnings when building this. Can you have a look at them?

Thanks for addressing the comments. Only one issue left really

openmls/src/treesync/mod.rs
```diff
@@ -215,20 +215,55 @@ impl TreeSync {
...
     }
 
+    /// Populate the parent hash caches of all nodes in the tree.
+    fn populate_parent_hashes(
+        &mut self,
+        backend: &impl OpenMlsCryptoProvider,
+        ciphersuite: &Ciphersuite,
+    ) -> Result<(), TreeSyncError> {
+        let diff = self.empty_diff()?;
+        // Make the diff into a staged diff. This is implicitly computes the
```
franziskuskiefer, line 235:
```suggestion
        // Make the diff into a staged diff. This implicitly computes the
```

openmls/src/group/mls_group/staged_commit.rs
raphaelrobert: I filed #616 for that.

openmls/src/group/mls_group/create_commit.rs
raphaelrobert: I filed #617 for that.

openmls/src/group/mls_group/staged_commit.rs
raphaelrobert: I filed #618 for that.

raphaelrobert (approved):
I think this looks good enough for now and the PR is so big that it's become difficult to review further. Let's land this and deal with any potential fallout later.

openmls/src/treesync/diff.rs
```diff
@@ -83,13 +83,28 @@ impl<'a> TryFrom<&'a TreeSync> for TreeSyncDiff<'a> {
 }
 
 impl<'a> TreeSyncDiff<'a> {
-    /// Check if the right-most leaf is blank. If that is the case, remove the
-    /// right-most leaf until the right-most leaf is not blank anymore.
+    /// Check if the right-most leaf and its parent are blank. If that is the
```
franziskuskiefer, line 86: Please add something here that the code here differs from the spec and that this is an ongoing discussion.

franziskuskiefer (changes requested):
one nit, otherwise lets get this in.

franziskuskiefer (approved):

PR #615, opened 2021-12-10
Replace PskFetcher callback with key store

raphaelrobert:
This PR replaces the last remaining `PskFetcher` callback with the `KeyStore` and works towards #141.
It doesn't really introduce new functionality and the existing tests are just adapted accordingly.
This makes the library callback-free and once the spec is stable regarding resumption PSKs, it shouldn't be too difficult to finalise the PSK implementation.

How it works:
 - The KeyStore can now store and read `PskBundle`, referenced by `PreSharedKeyId`
 - The application just needs to make sure all relevant PSKs are stored in the key store
 - `new_from_welcome`, `create_commit` and `stage_commit` will fetch the PSKs directly from the key store


openmls/src/schedule/psk.rs
```diff
@@ -90,49 +109,33 @@ impl ExternalPsk {
     }
 }
 
-/// External PSK Bundle. This contains the secret part of the PSK as well as the
+/// Contains the secret part of the PSK as well as the
 /// public part that is used as a marker for injection into the key schedule.
-pub struct ExternalPskBundle {
+#[derive(Serialize, Deserialize)]
+pub struct PskBundle {
+    psk_id: PreSharedKeyId,
```
franziskuskiefer, line 116: Why do we need the psk id in here?
The only place I found where it is used it's not needed because we have the ids there.

openmls/src/schedule/psk.rs
```diff
@@ -207,12 +210,20 @@ pub struct PreSharedKeyId {
 
 impl PreSharedKeyId {
     /// Create a new `PreSharedKeyID`
-    pub fn new(psk_type: PskType, psk: Psk, psk_nonce: Vec<u8>) -> Self {
-        Self {
-            psk_type,
+    pub fn new(
+        ciphersuite: &Ciphersuite,
+        backend: &impl OpenMlsCryptoProvider,
```
franziskuskiefer, line 215: You could narrow this to the `rand` bit.

openmls/src/schedule/psk.rs
```diff
@@ -285,22 +296,29 @@ impl PskSecret {
...
-        let num_psks = psks.len();
+        // Check that we don't have too many PSKs
+        let num_psks = psk_ids.len();
         if num_psks > u16::MAX as usize {
             return Err(PskSecretError::TooManyKeys);
         }
-        let num_psks = num_psks as u16;
+
+        // Fetch the PskBundles from the key store and make sure we have all of them
+        let mut psk_bundles: Vec<PskBundle> = Vec::with_capacity(num_psks);
```
franziskuskiefer, line 307: It's generally a good idea to preallocate. But this could lead to DoS attacks depending on how this function is exposed and whether it can be called multiple times (throwing invalid psk ids on the function).
I'm not sure we have to do something about it here. But it should at least be mentioned in the rustdoc of the function.

openmls/src/schedule/psk.rs
```diff
@@ -285,22 +296,29 @@ impl PskSecret {
...
+
         let mls_version = ProtocolVersion::default();
         let mut psk_secret = Secret::zero(ciphersuite, mls_version);
-        for ((index, psk), psk_id) in psks.iter().enumerate().zip(psk_ids.iter()) {
+        for (index, psk_bundle) in psk_bundles.iter().enumerate() {
             let zero_secret = Secret::zero(ciphersuite, mls_version);
-            let psk_extracted = zero_secret.hkdf_extract(backend, psk)?;
-            let psk_label = PskLabel::new(psk_id, index as u16, num_psks)
+            let psk_extracted = zero_secret.hkdf_extract(backend, psk_bundle.secret())?;
+            let psk_label = PskLabel::new(psk_bundle.psk_id(), index as u16, num_psks as u16)
```
franziskuskiefer, line 321: You could convert `num_psks` once to `u16` before.

franziskuskiefer (changes requested):
Thanks, looks much better already.

openmls/src/schedule/psk.rs
raphaelrobert: How else would you identify what secret to use? Note that this is just a more generalised version of what we had before (now other types of PSKs are supported, not just `ExternalPsk`).

openmls/src/schedule/psk.rs
raphaelrobert: I see what you mean, but in order to exploit that an attacker would have to send a very large number of PSK proposals first (it's the only way to increase the number of PSKs we use for the allocation). I can change it though.

franziskuskiefer: Yeah, I think it's enough maybe say something in the doc for the function that callers are aware of the potential issue.

openmls/src/schedule/psk.rs
franziskuskiefer: The only function that I see using this gets a list of `psk_id`s as input. Do we use it anywhere else?

raphaelrobert: Well, the whole idea is to store this in the key store. Alternatively, we could only store the `Secret` in the key store and then the consumers have to bring their own out-of-band storage for `PreSharedKeyId`.

franziskuskiefer: Your description says "The KeyStore can now store and read PskBundle, referenced by PreSharedKeyId". So yes, you need to store the ID somewhere else. It's of no use having it in the stored value.

raphaelrobert: This is aligned with the logic we use for `KeyPackageBundle` and `CredentialBundle`: We not only store the private key material in the key store, but also the public keys and other non-secret data. I don't see how this is any different.
Is your proposal to only store the `Secret`? I don't understand what exactly you are proposing.

raphaelrobert:
I removed the `PreSharedKeyId` from `PskBundle` for now.

franziskuskiefer (approved):
Thanks!

franziskuskiefer (approved):

PR #620, opened 2021-12-10
Fix clippy warnings after tree rewrite

raphaelrobert:
This PR fixes the newly introduced clippy warnings.

### DIFF
--- a/openmls/src/ciphersuite/mod.rs
+++ b/openmls/src/ciphersuite/mod.rs
@@ -128,6 +128,11 @@ impl Ciphersuite {
         self.name
     }
 
+    /// Get the protocol version of the ciphersuite
+    pub fn protocol_version(&self) -> ProtocolVersion {
+        ProtocolVersion::Mls10
+    }
+
     /// Get the AEAD mode
     #[cfg(any(test, feature = "test-utils"))]
     pub fn aead(&self) -> AeadType {
--- a/openmls/src/ciphersuite/mod.rs
+++ b/openmls/src/ciphersuite/mod.rs
@@ -128,11 +128,6 @@ impl Ciphersuite {
         self.name
     }
 
-    /// Get the protocol version of the ciphersuite
-    pub fn protocol_version(&self) -> ProtocolVersion {
-        ProtocolVersion::Mls10
-    }
-
     /// Get the AEAD mode
     #[cfg(any(test, feature = "test-utils"))]
     pub fn aead(&self) -> AeadType {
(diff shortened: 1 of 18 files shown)


openmls/src/ciphersuite/mod.rs
```diff
@@ -128,6 +128,11 @@ impl Ciphersuite {
         self.name
     }
 
+    /// Get the protocol version of the ciphersuite
+    pub fn protocol_version(&self) -> ProtocolVersion {
```
franziskuskiefer, line 132: This will break any pre-release version

franziskuskiefer (changes requested):

openmls/src/ciphersuite/mod.rs
kkohbrok: While we should make this genuinely depend on the ciphersuite instead of returning a constant, I think I mentioned before, that we will need something like this for the external init anyway, since the `PublicGroupState` only contains the ciphersuite, not the MLS version.

raphaelrobert: The `PublicGroupState` does contain the protocol version. 
In general, the ciphersuites are specific to the protocol version (the first part of the ciphersuite name contains it in fact). Since we don't have support for migrating between pre-release versions anyway, what's your recommended strategy for now? I'm not sure what changes you're requesting.

franziskuskiefer: Well, first this is changing functionality and not fixing any clipping warning (directly), so I'd like to do this in a different PR.

Then the question is whether we want OpenMLS to be able to migrate between different non-RFC versions gracefully. If the answer is no, we should get rid of the explicit protocol version in general. If the answer is yes, we need to test it and make it usable.

kkohbrok: > The `PublicGroupState` does contain the protocol version. In general, the ciphersuites are specific to the protocol version (the first part of the ciphersuite name contains it in fact). Since we don't have support for migrating between pre-release versions anyway, what's your recommended strategy for now? I'm not sure what changes you're requesting.

My bad. I hadn't refreshed my tab with the spec in a while. I still think that it really doesn't have to in the same way as the Welcome message doesn't have to. 

Regarding pre-release versions: I'd say for now, we don't support migration between pre-release versions. I think an easy way of doing it in the future would be to simply add custom ciphersuites for pre-release versions. That would separate the versions nicely and we also have some testing in place.

raphaelrobert: I took it out so that we can make faster progress. Pre-release support is still broken though :)

franziskuskiefer: 👍🏻 can you open a follow up issue to discuss and fix/remove this?

franziskuskiefer (approved):
Thanks

PR #629, opened 2021-12-14
Extend crypto traits and implementations with setup_and_export one-shots

kkohbrok:
This PR introduces new two new functions to the `OpenMlsCrypto` trait that are required for the implementation of the external init feature of MLS. In particular, this PR introduces two HPKE one-shot functions that establish a context and export a secret from it. The function for use by the sender outputs the encapsulated context in addition to the exported secret. The corresponding receiver function takes the encapsulated context as additional input and outputs only the exported secret.


traits/src/crypto.rs
```diff
@@ -113,6 +113,50 @@ pub trait OpenMlsCrypto {
...
+    ///
+    /// The encapsulated secret is returned together with the exported secret.
+    /// If the secret key is missing in an authenticated mode, an error is
+    /// returned.
+    fn hpke_setup_sender_and_export(
+        &self,
+        config: HpkeConfig,
+        pk_r: &[u8],
+        info: &[u8],
+        psk: Option<&[u8]>,
```
franziskuskiefer, line 131: No need for `psk` and `psk_id`. They are never used by MLS.

traits/src/crypto.rs
```diff
@@ -113,6 +113,50 @@ pub trait OpenMlsCrypto {
...
+        psk_id: Option<&[u8]>,
+        sk_s: Option<&[u8]>,
```
franziskuskiefer, line 133: No need for the auth mode. It's not used by MLS.

traits/src/crypto.rs
```diff
@@ -113,6 +113,50 @@ pub trait OpenMlsCrypto {
...
+        exporter_context: &[u8],
+        exporter_length: usize,
+    ) -> Result<(Vec<u8>, Vec<u8>), CryptoError>;
```
franziskuskiefer, line 136: The return values could get an alias or something to distinguish which is which.

traits/src/crypto.rs
```diff
@@ -113,6 +113,50 @@ pub trait OpenMlsCrypto {
...
+    /// HPKE single-shot setup of a receiver and immediate export a secret.
+    ///
+    /// For the base and PSK modes this decapsulates `enc` with the secret key
+    /// `sk_r` of the receiver.
+    /// For the Auth and AuthPSK modes this decapsulates and authenticates `enc`
+    /// with the secret key `sk_r` of the receiver and the senders public key `pk_s`.
+    ///
+    /// Returns the exported secret. If the secret key is missing in an
+    /// authenticated mode, an error is returned.
+    fn hpke_setup_receiver_and_export(
```
franziskuskiefer, line 147: Same comments as above for the optional args.

evercrypt_backend/src/provider.rs
```diff
@@ -296,9 +296,56 @@ impl OpenMlsCrypto for EvercryptProvider {
...
+        exporter_length: usize,
+    ) -> Result<ExporterSecret, CryptoError> {
+        let hpke = Hpke::<HpkeEvercrypt>::new(
+            hpke::Mode::Base,
+            kem_mode(config.0),
+            kdf_mode(config.1),
+            aead_mode(config.2),
+        );
+        let context = hpke
+            .setup_receiver(enc, &sk_r.into(), info, None, None, None)
```
franziskuskiefer, line 338: Looking at the `&sk_r.into()` I feel like hpke-rs should have a byte slice API as well. 🤔

franziskuskiefer (changes requested):
One nit, otherwise this looks good to me.

evercrypt_backend/src/provider.rs
kkohbrok: It would be a bit more generic. But it wouldn't make that much of a difference, I guess?

franziskuskiefer: It avoids some allocations. But it's unrelated to this PR really 😄

franziskuskiefer:
I noticed just now that I did the single shot exporter functions in hpke a while back. We could use them as well
https://docs.rs/hpke-rs/0.0.12/hpke_rs/struct.Hpke.html#method.send_export

PR #631, opened 2021-12-14
Sender and Commit Types

kkohbrok:
This PR allows the creation of Commits, MlsPlaintexts and Proposals with a `SenderType` other than `Member`. This is in preparation of the external init feature.

### DIFF
--- a/openmls/src/framing/plaintext.rs
+++ b/openmls/src/framing/plaintext.rs
@@ -221,7 +221,7 @@ impl MlsPlaintext {
     }
 
     /// This constructor builds an `MlsPlaintext` containing a Commit. The given
-    /// `CommitType` determines the `SenderType`: If it's an `Internal` commit,
+    /// `CommitType` determines the `SenderType`: If it's an `Member` commit,
     /// it's `SenderType::Member` and `SenderType::NewMember` otherwise.
     pub fn commit(
         framing_parameters: FramingParameters,
--- a/openmls/src/framing/plaintext.rs
+++ b/openmls/src/framing/plaintext.rs
@@ -221,7 +221,7 @@ impl MlsPlaintext {
     }
 
     /// This constructor builds an `MlsPlaintext` containing a Commit. The given
-    /// `CommitType` determines the `SenderType`: If it's an `Member` commit,
+    /// `CommitType` determines the `SenderType`: If it's a `Member` commit,
     /// it's `SenderType::Member` and `SenderType::NewMember` otherwise.
     pub fn commit(
         framing_parameters: FramingParameters,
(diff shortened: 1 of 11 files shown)


openmls/src/group/mls_group/create_commit_params.rs
```diff
@@ -2,12 +2,20 @@
 
 use super::{proposals::ProposalStore, *};
 
+/// Can be used to denote the type of a commit.
+#[derive(Debug, Copy, Clone)]
+pub enum CommitType {
+    External,
+    Internal,
```
raphaelrobert, line 9: So far we don't use the name "internal" anywhere. I'm not super hung up on it, but in general the opposite of "external" has been "member", also in the spec. I'd vote for changing it to that.

raphaelrobert (commented):
Looks good, just one comment.

raphaelrobert (changes requested):
You did s/External/Member. What I suggested was s/Internal/Member 😀

franziskuskiefer (commented):
lgtm
But I'm going to change some of this again because the `Sender` has a `KeyPackageRef` nowadays (no `LeafIndex`). Let's get this in before though.

kkohbrok:
Sorry about that renaming mixup 😅. Should be correct now.

openmls/src/framing/plaintext.rs
```diff
@@ -188,19 +193,52 @@ impl MlsPlaintext {
...
+                content_type: ContentType::Proposal,
+            },
+            credential_bundle,
+            context,
+            backend,
+        )
+    }
+
+    /// This constructor builds an `MlsPlaintext` containing a Commit. The given
+    /// `CommitType` determines the `SenderType`: If it's an `Internal` commit,
```
raphaelrobert, line 224:
```suggestion
    /// `CommitType` determines the `SenderType`: If it's an `Member` commit,
```

openmls/src/group/mls_group/create_commit_params.rs
```diff
@@ -2,12 +2,20 @@
...
+    Member,
+}
+
 pub struct CreateCommitParams<'a> {
     framing_parameters: FramingParameters<'a>, // Mandatory
     credential_bundle: &'a CredentialBundle,   // Mandatory
     proposal_store: &'a ProposalStore,         // Mandatory
     inline_proposals: Vec<Proposal>,           // Optional
     force_self_update: bool,                   // Optional
+    commit_type: CommitType,                   // Optional (default is `Internal`)
```
raphaelrobert, line 18:
```suggestion
    commit_type: CommitType,                   // Optional (default is `Member`)
```

raphaelrobert (changes requested):
Almost there

raphaelrobert (approved):

raphaelrobert (approved):

Issue #563, opened 2021-11-18
[Feature request] Decrypt application messages from past epochs


raphaelrobert:
Currently, the key material is discarded during each epoch change. This means that application messages can no longer be decrypted if they belong to a past epoch. It would be good to keep older key material around (while being mindful of general FS implications) according to configurable parameters. Parameters could be a maximum number of epochs and a maximum time (whichever comes first).

adzialocha:
This is really interesting, we originally thought about a very similar logic like this on top of MLS :+1:

raphaelrobert:
Currently blocked by #423.

Issue #265, opened 2020-12-15
[Feature request] Use byte buffers for managed API


raphaelrobert:
**Is your feature request related to a problem? Please describe.**
The managed API should be able to interface with the DS directly. Right now `MLSMessage` still needs to be serialized/deserialized.

**Describe the solution you'd like**
The managed API should return octet strings instead of `Vec<MLSMessage>` and also accept the former as input.

**Describe alternatives you've considered**
We could have separate functions for the seralization/deserialization, so that `MLSMessage` remains easily inspectable, but I simply don't see the need for this.

**Additional context**
None

Issue #247, opened 2020-12-07
Don't derive Debug for secrets


franziskuskiefer:
Let's not derive `Debug` for keys and secrets but only when people really know what they are doing.
`Debug` formatting should only print `***` for secrets and keys (private parts) by default and only print the key if an environment variable like `OPENMLS_PRINT_KEYS` is set.

raphaelrobert:
This is implemented for `Secret` since #1277 and HPKE private keys use `SecretVLBytes`. There's also the `crypto-debug` feature that derives `Debug` on things like `AeadKey`.

PR #632, opened 2021-12-15
Add feature to decrypt application messages from past epochs

raphaelrobert:
This PR adds a new feature to store just enough key material to be able to decrypt and validate application messages from past epochs.
The feature is switched off by default and needs to be enabled by settings non-zero value in `ManagedGroupConfig.max_epochs`.
We discussed the potential impact of this feature with @kkohbrok and convinced ourselves that the impact should not be too bad in the usual, reasonable threat models. I'm happy to put this feature behind a hazmat feature flag though (in a follow-up PR).
Note that there is no time limit (as suggested in #563). I don't see an easy way to delete secrets after a certain time. This could only be done opportunistically when the group state is in memory anyway. If someone sees a need for this, we could always do a follow-up PR once we have discussed the strategy.

The PR got quite big, because I had to split the `EpochSecrets` into two parts: One part that is called `MessageSecrets` that gets stored for past epochs in order to be able to do decryption/validation and another part called `GroupEpochSecrets` that contains all remaining secrets.
The `SecretTree` in the `MlsGroup` is now also part of `MessageSecrets`, which has the side-effect that it no longer meets to be behind a `RefCell`.
This refactoring is mostly renaming and doesn't really change how things work, but it makes for quite a few code changes.

*Review guidance*

The new functionality is in the following files:
 - `openmls/src/group/mls_group/past_secrets.rs`
 - `openmls/src/group/mls_group/process.rs`
 - `openmls/src/group/mls_group/staged_commit.rs`
 - `openmls/src/group/managed_group/processing.rs`
 
The new functionality should be covered by the following tests:
 - `openmls/src/group/mls_group/test_past_secrets.rs`
 - `openmls/src/group/tests/test_past_secrets.rs`


openmls/src/group/tests/test_past_secrets.rs
```diff
@@ -0,0 +1,231 @@
...
+    framing::{MlsCiphertextError, ProcessedMessage},
+    group::{
+        GroupId, ManagedGroup, ManagedGroupConfig, ManagedGroupError, MlsGroupError, WireFormat,
+    },
+    key_packages::KeyPackageBundle,
+    tree::secret_tree::SecretTreeError,
+};
+
+#[apply(ciphersuites_and_backends)]
+fn test_past_secrets_in_group(
```
kkohbrok, line 22: Don't we have the test framework for these lengthy group setups? It should work pretty well here.

openmls/src/schedule/mod.rs
```diff
@@ -1077,9 +1097,89 @@ impl EpochSecrets {
         })
     }
 
-    #[cfg(any(feature = "test-utils", test))]
-    #[doc(hidden)]
-    pub(crate) fn sender_data_secret_mut(&mut self) -> &mut SenderDataSecret {
-        &mut self.sender_data_secret
+    pub(crate) fn split(
```
kkohbrok, line 1100: Maybe `split_secrets`? `split` seems a bit generic.

kkohbrok (changes requested):
Nice! Thanks for picking this up! Just a few comments.

Also, I'm getting quite a few unused code warnings, mostly for getters of `EpochSecrets`. Can you remove the corresponding functions?

raphaelrobert (commented):
Thanks for the review!

openmls/src/group/mls_group/past_secrets.rs
```diff
@@ -0,0 +1,60 @@
...
+            self.epoch_trees.pop_front();
+        }
+        self.epoch_trees.push_back(epoch_tree);
+    }
+
+    /// Get a mutable reference to a secret tree for a given epoch `group_epoch`.
+    /// If no tree is found for that epoch, `None` is returned.
+    pub fn get_epoch(&mut self, group_epoch: GroupEpoch) -> Option<&mut MessageSecrets> {
+        let GroupEpoch(epoch) = group_epoch;
+        for epoch_tree in self.epoch_trees.iter_mut() {
```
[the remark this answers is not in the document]
kkohbrok: Wait. Isn't that exactly what this functionality is? A mechanism that ensures we have secrets for the past X epochs? One would have to consider the case where the current epoch is smaller than X, but that should be doable. Or am I missing something here?

openmls/src/group/mls_group/past_secrets.rs
```diff
@@ -0,0 +1,60 @@
...
+        if self.max_epochs == 0 {
+            return;
+        }
+        let GroupEpoch(epoch) = group_epoch;
+        let epoch_tree = EpochTree {
+            epoch,
+            message_secrets,
+        };
+        while self.epoch_trees.len() >= self.max_epochs {
+            self.epoch_trees.pop_front();
```
franziskuskiefer, line 44: Should we be able to clear the store without adding? Not sure if it's needed. But I feel like there should be a way to delete secrets explicitely.

openmls/src/group/mls_group/process.rs
```diff
@@ -33,13 +33,37 @@ impl MlsGroup {
...
-                message,
-                self.ciphersuite(),
-                backend,
-                self.epoch_secrets(),
-                self.secret_tree_mut().deref_mut(),
-            )?,
+            WireFormat::MlsCiphertext => {
+                // If the message is older than the current epoch, we need to fetch the correct secret tree first
+                let ciphersuite = self.ciphersuite().clone();
+                let message_secrets = if message.epoch() < self.context().epoch() {
```
franziskuskiefer, line 39: So there are two different places with message secrets now. The one you get through `message_secrets_mut` and the store. This looks like a footgun.
How about we always use the `message_secrets_store` that has a minimum size 1 instead of 0?

openmls/src/group/mls_group/mod.rs
```diff
@@ -566,24 +564,23 @@ impl MlsGroup {
         self.tree.own_leaf_index()
     }
 
-    pub(crate) fn epoch_secrets(&self) -> &EpochSecrets {
-        &self.epoch_secrets
+    pub(crate) fn group_epoch_secrets(&self) -> &GroupEpochSecrets {
```
franziskuskiefer, line 567: While you're renaming these, can you add comments?

openmls/src/group/mls_group/staged_commit.rs
```diff
@@ -290,13 +289,39 @@ impl MlsGroup {
...
+
+    /// Merges a [StagedCommit] into the group state and optionally return a [`SecretTree`]
+    /// from the previous epoch. The secret tree is returned if the Commit does not contain a self removal.
+    ///
+    /// This function should not fail and only returns a [`Result`], because it
+    /// might throw a `LibraryError`.
+    pub fn merge_commit_take_message_secrets(
+        &mut self,
+        staged_commit: StagedCommit,
+    ) -> Result<Option<MessageSecrets>, MlsGroupError> {
```
franziskuskiefer, line 308: If there's no point in failing it can as well just return an `Option`.

openmls/src/schedule/message_secrets.rs
```diff
@@ -0,0 +1,89 @@
...
+    pub(crate) membership_key: MembershipKey,
+    pub(crate) confirmation_key: ConfirmationKey,
+    pub(crate) serialized_context: Vec<u8>,
+    pub(crate) secret_tree: SecretTree,
+}
+
+// Public functions
+impl MessageSecrets {
+    /// Get a reference to the message secrets's sender data secret.
+    pub fn sender_data_secret(&self) -> &SenderDataSecret {
```
franziskuskiefer, line 19: It's a little odd to have getter for all the fields but no way of setting them, i.e. for creating you rely on the `pub(crate)` visibility of the fields but not to get the values. Maybe make the fields private and add a constructor?

openmls/src/schedule/mod.rs
```diff
@@ -1077,9 +1104,89 @@ impl EpochSecrets {
         })
     }
 
-    #[cfg(any(feature = "test-utils", test))]
-    #[doc(hidden)]
-    pub(crate) fn sender_data_secret_mut(&mut self) -> &mut SenderDataSecret {
-        &mut self.sender_data_secret
+    pub(crate) fn split_secrets(
```
franziskuskiefer, line 1107: Can these new functions and structs get comments? 😊

franziskuskiefer (changes requested):
Thanks, generally looks good with a couple comments.

openmls/src/framing/ciphertext.rs
```diff
@@ -33,9 +34,8 @@ impl MlsCiphertext {
         mls_plaintext: &MlsPlaintext,
```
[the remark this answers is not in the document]
raphaelrobert: I had the same reflex, but I didn't do it because I thought it would blur the separation. I Don't have a very strong opinion, but changing it would probably mean it's not easy to test without having an actual group.

openmls/src/framing/message.rs
```diff
@@ -36,6 +36,14 @@ impl MlsMessageIn {
             MlsMessageIn::Plaintext(m) => m.epoch(),
         }
     }
+
+    /// Get the content type
+    pub fn content_type(&self) -> ContentType {
+        match self {
+            MlsMessageIn::Ciphertext(m) => m.content_type,
```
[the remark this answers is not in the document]
raphaelrobert: Fixed it right away.

openmls/src/framing/validation.rs
```diff
@@ -69,12 +70,10 @@ impl DecryptedMessage {
         inbound_message: MlsMessageIn,
         ciphersuite: &Ciphersuite,
         backend: &impl OpenMlsCryptoProvider,
-        epoch_secrets: &EpochSecrets,
-        secret_tree: &mut SecretTree,
+        message_secrets: &mut MessageSecrets,
     ) -> Result<Self, ValidationError> {
         if let MlsMessageIn::Ciphertext(ciphertext) = inbound_message {
```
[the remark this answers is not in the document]
raphaelrobert: It's not like we have to do all these checks, we just happen to do them. This code will change anyway with #265, which I want to tackle soon.

openmls/src/framing/validation.rs
```diff
@@ -188,22 +192,21 @@ impl UnverifiedContextMessage {
...
+        message_secrets: &MessageSecrets,
         backend: &impl OpenMlsCryptoProvider,
     ) -> Result<Self, ValidationError> {
         // Decompose UnverifiedMessage
         let (mut plaintext, credential_option) = unverified_message.into_parts();
 
         if plaintext.sender().is_member() {
             // Add serialized context to plaintext
-            plaintext.set_context(serialized_context);
+            plaintext.set_context(message_secrets.serialized_context().to_vec());
```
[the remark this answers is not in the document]
raphaelrobert: This is needed for the signature verification later on.

openmls/src/group/managed_group/config.rs
```diff
@@ -103,6 +111,17 @@ impl ManagedGroupConfigBuilder {
         self
     }
 
+    /// Sets the `max_past_epochs` property of the ManagedGroupConfig.
+    /// WARNING: This feature enables the storage of message secrets from past epochs.
+    /// This allows application messages from previous epochs to be decrypted.
+    /// It is a tradeoff between functionality and forward secrecy and should only be enabled
+    /// if the Delivery Service cannot guarantee that application messages will be sent in
+    /// the same epoch in which they were generated.
+    pub fn max_past_epochs(mut self, max_past_epochs: usize) -> Self {
```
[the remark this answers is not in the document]
raphaelrobert: I thought the same, but what is the right limit? I extended the comment a bit now.

openmls/src/group/mls_group/past_secrets.rs
```diff
@@ -0,0 +1,60 @@
+use std::collections::VecDeque;
+
+use crate::schedule::message_secrets::MessageSecrets;
+
+use super::*;
+
+// Internal helper struct
+#[derive(Debug, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
raphaelrobert: Agreed, but as you say: #247 will fix all that.

openmls/src/group/mls_group/past_secrets.rs
raphaelrobert: So far you can only drop the whole store if you want to delete secrets. I didn't add a more fine-grained method because I didn't want to write tests for something we don't need right now. Adding it with a test is easy though if we ever need it.

openmls/src/group/mls_group/past_secrets.rs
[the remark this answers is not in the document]
raphaelrobert: The mechanism you describe is what we have in `ManagedGroup`. But it is not enforced at this level. We could programmatically enforce it here, in the sense that you couldn't skip epochs etc. But I'm really not sure that's needed right now.

openmls/src/group/mls_group/process.rs
raphaelrobert: If we wanted to do that we would have to make `MessageSecretsStore` part of `MlsGroup` as well. So far it's only part of `ManagedGroup`. While not impossible, this would take away from the optional character it has now and it would make it harder to put it behind a feature flag. 
I'm also not convinced it would make things that much better, because the same branching logic would still have to exist – just elsewhere.
If we really wanted to do this, I'd prefer to do it in a follow-up PR because it would lead to a lot of code change.

openmls/src/group/mls_group/staged_commit.rs
raphaelrobert: No, because as the comment says, these are different cases. If we really throw a `LibraryError` but only return a `None`, applications could get confused if they expected a `Some(...)`. We'd make it their responsibility to detect it was a library error. I'd rather throw the error early on and give applications something they can easily react to.

openmls/src/schedule/message_secrets.rs
```diff
@@ -0,0 +1,89 @@
...
+use super::*;
+#[cfg(test)]
+use crate::tree::index::LeafIndex;
+/// Combined message secrets that need to be stored for later decryption/verification
+#[derive(Debug, Serialize, Deserialize)]
+pub struct MessageSecrets {
+    pub(crate) sender_data_secret: SenderDataSecret,
+    pub(crate) membership_key: MembershipKey,
+    pub(crate) confirmation_key: ConfirmationKey,
+    pub(crate) serialized_context: Vec<u8>,
```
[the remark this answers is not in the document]
raphaelrobert: Agreed, this is just a helper field and not a secret. But it's quite helpful to carry it along, because it often gets used with the other secrets.

raphaelrobert (commented):
Thanks for all the review comments!

openmls/src/framing/ciphertext.rs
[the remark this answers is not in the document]
kkohbrok: Maybe an intermediate struct that can readily be created from a reference to a group?

kkohbrok (dismissed):
Looks good to me now, thanks!

openmls/src/framing/ciphertext.rs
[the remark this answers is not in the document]
franziskuskiefer: Yes that's the downside. I'm fine with keeping it this way since you got rid of the `RefCell` anyway. Otherwise that would've been the solution.

openmls/src/framing/validation.rs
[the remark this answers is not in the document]
franziskuskiefer: Right we do them because the code isn't well structured 😉 Can you add a comment that this will be cleaned up in #265

openmls/src/framing/validation.rs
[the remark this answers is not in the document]
franziskuskiefer: Ok, I guess the match below needs to be cleaned up to make it clearer what's happening here.

openmls/src/group/mls_group/past_secrets.rs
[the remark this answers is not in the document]
franziskuskiefer: Right, but I was asking if we need to add more when we want to remove all of them. If you really need it, lets leave it to #247. If not, just drop it.

openmls/src/group/mls_group/process.rs
franziskuskiefer: Ok, then please file a follow up.

openmls/src/group/mls_group/staged_commit.rs
franziskuskiefer: Sure if you really want to. It's just not very idiomatic. Rust callers would expect an `Option` here as result.

openmls/src/schedule/message_secrets.rs
franziskuskiefer: The `secret_tree` can lose the `pub(crate)` as well.

franziskuskiefer (changes requested):
Thanks, a couple nits and follow ups to file. Then I think this is good to go

openmls/src/group/mls_group/past_secrets.rs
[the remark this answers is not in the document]
raphaelrobert: Yes, it's needed because `MlsGroup` derives `Debug`.

openmls/src/group/mls_group/process.rs
raphaelrobert: I filed #635.

openmls/src/group/mls_group/staged_commit.rs
raphaelrobert: I agree it doesn't look great. I'm happy to change it for something else, as long as it doesn't introduce dangerous behaviour. Maybe we can get rid of the library errors at some point.

openmls/src/schedule/message_secrets.rs
raphaelrobert: oops, fixed

franziskuskiefer (approved):
lgtm

PR #633, opened 2021-12-16
Assert that only members add a context to MlsPlaintextTBS

### LINKED ISSUE #477: [MLS Spec change] Explicit blank cases in MLSPlaintextTBS
Explicit blank cases in MLSPlaintextTBS

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/474

franziskuskiefer:
At this point the sender type is always member. This will change in future though.
But the context should only be present for members.
That's why we just assert this here.

Fixes #477

### DIFF
--- a/openmls/src/framing/codec.rs
+++ b/openmls/src/framing/codec.rs
@@ -148,6 +148,8 @@ pub(super) fn serialize_plaintext_tbs<'a, W: Write>(
     buffer: &mut W,
 ) -> Result<usize, tls_codec::Error> {
     let mut written = if let Some(serialized_context) = serialized_context.into() {
+        // Only a member should have a context.
+        debug_assert_eq!(sender.sender_type, SenderType::Member);
         buffer.write(serialized_context)?
     } else {
         0


raphaelrobert:
This is definitely good for outgoing messages, but I think that in order to fully fix #477 there should also be a validation step for incoming messages. Or do we consider incoming `MlsPlaintext` to be valid if their sender type is not member yet they still have a context field? The harm is probably low, but we should decide on that explicitly.

kkohbrok:
Two comments on this: I'm working on external inits, where the `SenderType` is `NewMember` and I don't quite see why a `NewMember` can't include a context here. The new member should have all the required information.
Secondly, instead of creating an assert, wouldn't it be more robust to enforce this structurally on the sender side as well? What I mean is that we could change the `MlsPlaintext` constructors such that a context can be passed as input only if the `SenderType` is `Member`. I realize, that I should have done that when I touched those constructors a few days ago, but I didn't know this was a thing until just now.

franziskuskiefer:
> Or do we consider incoming MlsPlaintext to be valid if their sender type is not member yet they still have a context field?

This is only for `MLSPlaintextTBS`.

> SenderType is NewMember and I don't quite see why a NewMember can't include a context here.

Well, that would be a spec issue. Do you want to open an issue if you think it should be included?

> Secondly, instead of creating an assert, wouldn't it be more robust to enforce this structurally on the sender side as well? 

Sure, if someone is rewriting how this works, that can be done. Otherwise there's no need to introduce code changes just to catch potential issues here.

kkohbrok (approved):
Then let's get this in for now and we'll deal with everything else either on the spec level or in follow-ups.

raphaelrobert (approved):
Can you file a follow-up issue for the validation? Let's get this in, that way the spec change is covered at least.

Issue #630, opened 2021-12-14
[Meta] External Init


kkohbrok:
This issue tracks the progress on the "External Init" functionality via external commits: https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#section-11.2.1.

- [x] Constructors for `NewMember` commits, proposals and MLSPlaintexts
- [x] one-shot HPKE functions for setup and key export by sender and receiver
- [x] creation of external init commits
- [x] application/processing of external init commits (`new_from_external_init`) by others
- [x] application/processing of external init commits (`new_from_external_init`) by the creator
- [x] #682
- [x] #680
- [x] #681

TODO: There are a number of spec-change issues that might be fixed by this. Before closing this issue, please check the following issues.
- [x] #556
- [x] #542
- [x] #536
- [x] #289
- [x] #286
- [x] #192

raphaelrobert:
All checks completed, closing this.

PR #634, opened 2021-12-16
Partial support for external init

kkohbrok:
This PR adds partial support for the external init operation, where a non-member can inject themselves into a group. In particular, this PR allows the creation of a group from a `PublicGroupState` object and the creation of a corresponding external commit. It also enables other group members to process this commit. However, it does not yet allow the externally committing group member to process their own commit. The implementation of this missing piece is non-trivial, but would be greatly simplified if we extended the way that staged commits are handled, such that a committing member does not need to process their own commit, but can instead just merge the diff that was created in the process of creating the commit. The overall effort of enabling external init in OpenMLS is tracked in #630.

TODO:
- [x] change the TreeSync constructor `from_nodes_without_leaf` such that it actually creates a `TreeSync` instance without an "owned leaf". This breaks the TreeSync invariant that there always has to be an "own index", but it allows us to verify the `tree_hash` given in the `PublicGroupState`. Maybe it would make sense to make the `own_index` optional such that we fail quickly if such a tree accidentally gets generated outside of the external init context.

Concretely, this PR introduces a new `new_from_external_init` function to the `MlsGroup` API, which results in a new group and the commit message. The other main changes are to `create_commit` and `stage_commit` to enable commits by non-members (i.e. `SenderType::NewMember`) and their processing by group members. Changes in other places should all be related to the changes in those functions. In addition, this PR also includes a test that mostly serves to show that the basic functionality is ensured. 

In the process of implementation, I noticed that one can't properly encrypt path secrets without a `confirmed_transcript_hash`, so I added it to the `PublicGroupState` struct for now. This will probably change in the near future after discussions in the WG, but since the `PublicGroupState` is only used in the course of an external init, there's no harm in changing it at the moment. Although it does mean that the message kat needs to be updated as part of this PR.


raphaelrobert (changes requested):
I did a first pass and flagged a few things. I will do another pass later.

openmls/src/ciphersuite/secret.rs
```diff
@@ -218,6 +218,11 @@ impl Secret {
     pub(crate) fn ciphersuite(&self) -> &'static Ciphersuite {
         self.ciphersuite
     }
+
+    /// Returns the version of the secret
+    pub(crate) fn version(&self) -> ProtocolVersion {
```
franziskuskiefer, line 223: I thought we were getting rid of this?

openmls/src/group/mls_group/new_from_external_init.rs
```diff
@@ -0,0 +1,171 @@
...
+                // We know we only have one element
+                .ok_or(ExternalInitError::LibraryError)?
+                .as_ratchet_tree_extension()?;
+            Some(extension)
+        } else {
+            // Throw an error if there is more than one ratchet tree extension.
+            // This shouldn't be the case anyway, because extensions are checked
+            // for uniqueness anyway when decoding them.
+            // We have to see if this makes problems later as it's not something
+            // required by the spec right now.
```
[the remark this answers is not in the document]
kkohbrok: https://github.com/mlswg/mls-protocol/issues/530

openmls/src/framing/plaintext.rs
```diff
@@ -222,7 +226,9 @@ impl MlsPlaintext {
 
     /// This constructor builds an `MlsPlaintext` containing a Commit. The given
     /// `CommitType` determines the `SenderType`: If it's a `Member` commit,
-    /// it's `SenderType::Member` and `SenderType::NewMember` otherwise.
+    /// it's `SenderType::Member` and `SenderType::NewMember` otherwise. If it
+    /// is an `External` commit, the context is not signed along with the rest
```
franziskuskiefer, line 230: So `SenderType` can not only be `Member` and `NewMember`? This comment is confusing.

openmls/src/group/mls_group/apply_proposals.rs
```diff
@@ -187,6 +222,36 @@ impl MlsGroup {
             diff.blank_leaf(remove_proposal.removed())?;
         }
 
+        // Process external init proposals
+        for queued_proposal in proposal_queue.filtered_by_type(ProposalType::ExternalInit) {
+            // If we are the originator of the external init, we don't need to
+            // get the init secret from the proposal. This branching will not be
+            // necessary after #617.
```
franziskuskiefer, line 229: This looks like a lot of duplicated code.

kkohbrok (commented):
Thanks for the review. I filed an issue with the spec regarding uniqueness of RatchetTreeExtensions.

openmls/src/group/mls_group/create_commit.rs
```diff
@@ -66,10 +72,41 @@ impl CoreGroup {
...
+            let key_package_bundle = KeyPackageBundle::new(
+                &[ciphersuite.name()],
+                params.credential_bundle(),
+                backend,
+                vec![],
+            )?;
+
+            let _leaf_index = diff.add_leaf(key_package_bundle.key_package().clone())?;
+            debug_assert_eq!(own_leaf_index, _leaf_index);
+            KeyPackageBundlePayload::from_rekeyed_key_package(
```
franziskuskiefer, line 97: You could do this outside of the conditional then you don't have to write it twice.

openmls/src/group/mls_group/create_commit.rs
```diff
@@ -66,10 +72,41 @@ impl CoreGroup {
 
         let proposal_reference_list = proposal_queue.commit_list();
 
-        let sender_index = self.sender_index();
         // Make a copy of the current tree to apply proposals safely
         let mut diff: TreeSyncDiff = self.treesync().empty_diff()?;
 
+        // If this is not an external commit we have to set our own leaf index
```
franziskuskiefer, line 78: So this function is already 200 lines too long. Can you factor out the new things you put in here?

openmls/src/group/mls_group/new_from_external_init.rs
```diff
@@ -0,0 +1,174 @@
...
+
+        let mut ratchet_tree_extensions = verifiable_public_group_state
+            .other_extensions()
+            .iter()
+            .filter(|e| e.extension_type() == ExtensionType::RatchetTree)
+            .collect::<Vec<&Extension>>();
+
+        let ratchet_tree_extension = if ratchet_tree_extensions.is_empty() {
+            None
+        } else if ratchet_tree_extensions.len() == 1 {
```
franziskuskiefer, line 62: Just pop out the first element then you don' need to check the length

openmls/src/group/mls_group/new_from_external_init.rs
```diff
@@ -0,0 +1,174 @@
...
+    ) -> ExternalInitResult {
+        let ciphersuite = Config::ciphersuite(verifiable_public_group_state.ciphersuite())?;
+        if !Config::supported_versions().contains(&verifiable_public_group_state.version()) {
+            return Err(ExternalInitError::UnsupportedMlsVersion);
+        }
+
+        let mut ratchet_tree_extensions = verifiable_public_group_state
+            .other_extensions()
+            .iter()
+            .filter(|e| e.extension_type() == ExtensionType::RatchetTree)
```
franziskuskiefer, line 57: You can just find the value and not get out all the matching values. (Then you don't need a heap allocation.)

franziskuskiefer (changes requested):
Some first comments. I'll have to look at this again with more time.

kkohbrok (commented):
Thanks again for the comments. Unfortunately, this is a bit of a moving targets due to the pending issues/PRs with the spec.

openmls/src/group/mls_group/new_from_external_init.rs
kkohbrok: I changed my mind and we are now detecting duplicate extensions again. However, that requires that I use `filter` instead of find here to make sure we're catching all of them (because you mentioned it in another comment..

openmls/src/ciphersuite/secret.rs
raphaelrobert: I think the way forward is to create an issue for pre-release compatibility and to flag it here.

openmls/src/group/core_group/apply_proposals.rs
```diff
@@ -187,6 +222,36 @@ impl CoreGroup {
             diff.blank_leaf(remove_proposal.removed())?;
         }
 
+        // Process external init proposals
+        for queued_proposal in proposal_queue.filtered_by_type(ProposalType::ExternalInit) {
+            // If we are the originator of the external init, we don't need to
+            // get the init secret from the proposal. This branching will not be
+            // necessary after #617.
+            if queued_proposal.sender().to_leaf_index() != self.treesync().own_leaf_index() {
+                // Unwrapping here is safe because we know the proposal type
```
raphaelrobert, line 231:
```suggestion
                // We know the proposal type, so this should not fail
```

openmls/src/group/core_group/new_from_external_init.rs
```diff
@@ -0,0 +1,148 @@
...
+        inline_proposals.push(external_init_proposal);
+
+        let params = CreateCommitParams::builder()
+            .framing_parameters(framing_parameters)
+            .credential_bundle(credential_bundle)
+            .proposal_store(&proposal_store)
+            .inline_proposals(inline_proposals)
+            .commit_type(CommitType::External)
+            // Populate the path
+            .force_self_update(true)
```
raphaelrobert, line 138: This should actually happen automatically. Can you adjust the `path_required` flag?

openmls/src/messages/public_group_state.rs
```diff
@@ -14,6 +14,7 @@ use crate::{
     },
     extensions::Extension,
     group::{CoreGroup, GroupEpoch, GroupId},
+    prelude::ProtocolVersion,
```
raphaelrobert, line 17: Don't use prelude for imports

openmls/src/schedule/mod.rs
```diff
@@ -208,6 +211,19 @@ impl From<Secret> for InitSecret {
...
 }
 
+/// Creates a string from the given MLS `ProtocolVersion` for the computation of
+/// the `init_secret` when creating or processing a commit with an external init
+/// proposal.
+fn hpke_info_from_version(version: ProtocolVersion) -> String {
+    match version {
+        ProtocolVersion::Reserved => "Reserved",
+        ProtocolVersion::Mls10 => "MLS 1.0",
+        ProtocolVersion::Mls10Draft11 => "MLS 1.0 Draft 11",
```
raphaelrobert, line 221: Please also flag the pre-release issue here, this conversion should happen elsewhere

openmls/src/group/core_group/new_from_external_init.rs
```diff
@@ -0,0 +1,148 @@
+use crate::{
+    ciphersuite::signable::Verifiable,
+    credentials::CredentialBundle,
+    group::errors::ExternalInitError,
+    messages::{
+        proposals::{ExternalInitProposal, Proposal},
+        public_group_state::{PublicGroupState, VerifiablePublicGroupState},
+    },
+    prelude::{plaintext::MlsPlaintext, KeyPackageBundle},
```
raphaelrobert, line 9: Don't use prelude for imports

raphaelrobert (changes requested):
One more round with things to change :)

openmls/src/schedule/mod.rs
kkohbrok: This conversion is specific to the derivation of the init secret. Why shouldn't it be here?

kkohbrok: Also, I'm not sure why the pre-release issue should be flagged here. This has to happen regardless of the pre-release issue.

raphaelrobert: True, but the string prefix "MLS 1.0" is also used for hash-based identifiers for example. I just thought it would be easier to have it somewhere closer to `ProtocolVersion`. I also realise now the spec is not consistent in that regard, sometimes "mls10" is also used. I'll follow up with the spec.

openmls/src/schedule/mod.rs
```diff
@@ -208,6 +211,19 @@ impl From<Secret> for InitSecret {
     }
 }
 
+/// Creates a string from the given MLS `ProtocolVersion` for the computation of
+/// the `init_secret` when creating or processing a commit with an external init
+/// proposal.
+fn hpke_info_from_version(version: ProtocolVersion) -> String {
+    match version {
```
franziskuskiefer, line 218: These are all static strings. No need to make this more expensive than necessary. You can return a `&'static str` here.

openmls/src/treesync/diff.rs
```diff
@@ -50,13 +50,14 @@ pub(crate) type UpdatePathResult = (KeyPackageBundle, Vec<PlainUpdatePathNode>,
...
 pub(crate) struct StagedTreeSyncDiff {
+    own_leaf_index: LeafIndex,
     diff: StagedMlsBinaryTreeDiff<TreeSyncNode>,
     new_tree_hash: Vec<u8>,
 }
 
 impl StagedTreeSyncDiff {
-    pub(super) fn into_parts(self) -> (StagedMlsBinaryTreeDiff<TreeSyncNode>, Vec<u8>) {
-        (self.diff, self.new_tree_hash)
+    pub(super) fn into_parts(self) -> (StagedMlsBinaryTreeDiff<TreeSyncNode>, Vec<u8>, LeafIndex) {
```
franziskuskiefer, line 59: Can we use the same order in struct and result to make this easier to use.

openmls/src/treesync/diff.rs
```diff
@@ -135,18 +136,9 @@ impl<'a> TreeSyncDiff<'a> {
...
-        &mut self,
-        leaf_node: KeyPackage,
-    ) -> Result<LeafIndex, TreeSyncDiffError> {
-        let node = Node::LeafNode(leaf_node.into());
+    /// Find and return the index of either the left-most blank leaf, or, if
+    /// there are no blank leaves, the leaf count.
+    pub(in crate::treesync) fn free_leaf_index(&self) -> Result<LeafIndex, TreeSyncDiffError> {
         // Find a free leaf and fill it with the new key package.
         let leaf_ids = self.diff.leaves()?;
         let mut leaf_index_option = None;
```
franziskuskiefer, line 144: I guess this is old code but why is there a `continue` in the loop?

openmls/src/treesync/mod.rs
```diff
@@ -225,6 +226,42 @@ impl TreeSync {
...
 
+    /// Create a [`TreeSync`] instance from a vector of nodes without expecting
+    /// there to be a [`KeyPackage`] that belongs to this particular MLS client.
+    /// WARNING: Some of the [`TreeSync`] invariants will not hold for this
+    /// tree, as the `own_leaf_index` does not point to a leaf with private key
+    /// material in it.
+    pub(crate) fn from_nodes_without_leaf(
+        backend: &impl OpenMlsCryptoProvider,
+        ciphersuite: &Ciphersuite,
+        node_options: &[Option<Node>],
```
franziskuskiefer, line 237: As long as the caller consumes this do we really need to use a slice here?
This function looks pretty expensive because you iterate over the entire tree (cloning and converting). We could keep it as is and file a follow up to figure out if it really is a performance issue if you prefer.

franziskuskiefer (changes requested):
Thanks for the changes. Only a couple nits left.

openmls/src/schedule/mod.rs
```diff
@@ -208,6 +211,19 @@ impl From<Secret> for InitSecret {
     }
 }
 
+/// Creates a string from the given MLS `ProtocolVersion` for the computation of
+/// the `init_secret` when creating or processing a commit with an external init
+/// proposal.
```
raphaelrobert, line 216:
```suggestion
/// proposal. TODO: #628.
```

openmls/src/group/core_group/new_from_external_init.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// `nodes`. After the group state is initialized, this function creates an
+    /// `ExternalInit` proposal and commits it along with the given proposals by
+    /// reference and by value.
+    ///
+    /// Returns the new `CoreGroup` object, as well as the `MlsPlaintext`
+    /// containing the commit.
+    pub fn new_from_external_init(
+        backend: &impl OpenMlsCryptoProvider,
+        framing_parameters: FramingParameters,
+        nodes_option: Option<Vec<Option<Node>>>,
```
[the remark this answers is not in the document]
kkohbrok: I think `Option` is not a bad choice here, since it make it explicit that providing a tree is optional, i.e. one can rely exclusively on the RatchetTreeExtension. You're right regarding the name, though.

raphaelrobert (changes requested):

openmls/src/group/core_group/new_from_external_init.rs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        // Build the ratchet tree
+
+        // Set nodes either from the extension or from the `nodes_option`.
+        // If we got a ratchet tree extension in the welcome, we enable it for
+        // this group. Note that this is not strictly necessary. But there's
+        // currently no other mechanism to enable the extension.
+        let (nodes, enable_ratchet_tree_extension) =
+            match try_nodes_from_extensions(&verifiable_public_group_state.other_extensions())? {
```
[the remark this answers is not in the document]
kkohbrok: Sorry, but I'm having a hard time wrapping my head around what this would look like with an `and_then`. Note, that the code had to change slightly since I change the input to be a slice of nodes rather than a vector.

openmls/src/schedule/mod.rs
kkohbrok: I did the concatenation manually for now. I think there is some macro magic that I could use to avoid appending the "external init" to each version, but I thought that was a bit overkill. Also, considering https://github.com/mlswg/mls-protocol/issues/531, we're probably going to touch this again soon anyway.

openmls/src/treesync/diff.rs
kkohbrok: I'm confused. What loop?

openmls/src/treesync/mod.rs
kkohbrok: Yes, this is very annoying. There are multiple functions handling these `Option<Node>` vectors or slices and we should be consistent. In the end, we invariably need to clone them to instantiate a tree, so we might as well use full vectors. I assume that in almost all cases, the caller doesn't need the nodes afterwards anyway. I chose the slice here, because we are using slices in pretty much all other places that we handle these tree copies. I'll file a follow-up to optimize it and if nothing else, make it consistent.

kkohbrok (commented):
Thanks for the feedback again!

openmls/src/treesync/diff.rs
franziskuskiefer: Line 150 (I can't comment there because it's old code). Not too important though

kkohbrok: That continue is there, because we only want the first free spot that we find and there might be more than one blank leaf.

franziskuskiefer: Ok, so you want a `break`? Because this isn't doing what you think it is.
But apperently there's no test actually excercising this.
Please add one and fix this.

openmls/src/treesync/mod.rs
franziskuskiefer: 👍🏻

openmls/src/group/core_group/new_from_external_init.rs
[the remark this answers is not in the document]
franziskuskiefer: No worries, just a though. No need to change this here.

franziskuskiefer (changes requested):
Looks like there's a bug, otherwise I'm fine with this.

raphaelrobert (dismissed):
Thanks for the fixes! Modulo the changes request by @franziskuskiefer I think this is good to go and we should get this in!

openmls/src/treesync/diff.rs
kkohbrok: Urgh. Good point. I think I'd rather fix it separately, though. I filed #653, which includes the fix and a test. Let's get that on main, then I'll merge it in here. It's delays this PR by a little, but it doesn't make the PR even bigger than it already is.

openmls/src/treesync/mod.rs
kkohbrok: Filed #652.

kkohbrok:
The bug in the free leaf computation was fixed on main and merged into this PR. It should work as intended now (and the tests should show that it does).

franziskuskiefer (approved):
Thanks :shipit:

Issue #645, opened 2021-12-19
Cleanup unused CoreGroup methods


raphaelrobert:
Quite a few methods are marked as test-only because they are no longer used and should be removed along with the tests.

Issue #644, opened 2021-12-19
Make CoreGroup integration tests unit tests


raphaelrobert:
Since the `CoreGroup` is not exposed publicly anymore, its integration tests should become unit tests.

Issue #643, opened 2021-12-19
Interop client should not use CoreGroup anymore


raphaelrobert:
It can probably use `MlsGroup` with some internals exposed.

raphaelrobert:
Closed vie #687.

Issue #642, opened 2021-12-19
DS & cli should not use CoreGroup anymore


raphaelrobert:
They should use `MlsGroup` or get dropped.

PR #646, opened 2021-12-19
Make CoreGroup private and shape API

raphaelrobert:
This PR makes the `CoreGroup` private and only exposes it through the `test-utils` feature flag.
It also cleans up the public API and the imports.

The public API is now exclusively exposed through `prelude.rs`.

Since the DS, cli, interop client & still need the `CoreGroup`, we need to expose it through `prelude_test.rs`. This is temporary and could get fixed like this:
 - The DS & cli should not use the `CoreGroup` anymore, or they should get dropped. See #642.
 - The interop client should also not use the `CoreGroup` but will most likely still need access to internals. See #643.
 - Integration tests for the `CoreGroup` should become unit tests. See #644.
 
Other remarks:
 - Unused code was deleted
 - CoreGroup methods are marked as test-only because they are now superseded by the validation logic. See #645.
 - Some methods appear to be unused because External Commits are not there yet, they have been marked as test-only for now
 - More things should become private once #265 is addressed.


openmls/src/group/core_group/mod.rs
```diff
@@ -117,11 +123,13 @@ impl CoreGroupBuilder {
         self
     }
     /// Set the [`Vec<PreSharedKeyId>`] of the [`CoreGroup`].
+    #[cfg(any(feature = "test-utils", test))]
     pub fn with_psk(mut self, psk_ids: Vec<PreSharedKeyId>) -> Self {
         self.psk_ids = psk_ids;
         self
     }
     /// Set the [`ProtocolVersion`] of the [`CoreGroup`].
+    #[cfg(any(feature = "test-utils", test))]
```
kkohbrok, line 132: Is this (or any of the functions now hidden behind config flags actually used? I check `with_version` and I couldn't find any place where we actually use it.

openmls/src/group/core_group/staged_commit.rs
```diff
@@ -282,6 +282,7 @@ impl CoreGroup {
     ///
     /// This function should not fail and only returns a [`Result`], because it
     /// might throw a `LibraryError`.
+    #[cfg(any(feature = "test-utils", test))]
     pub fn merge_commit(&mut self, staged_commit: StagedCommit) -> Result<(), CoreGroupError> {
```
kkohbrok, line 286: I'm a bit confused as to why we can hide this function behind a cfg. Shouldn't we be using this function in `MlsGroup` somewhere to merge staged commits?

openmls/src/lib.rs
```diff
@@ -21,39 +21,40 @@
...
+// === Modules ===
 
 #[macro_use]
-pub mod error;
+mod utils;
 
-#[cfg(any(feature = "test-utils", test))]
 #[macro_use]
-pub mod test_utils;
+pub mod error;
```
kkohbrok, line 43: Any reason we're not exposing this through `prelude.rs`?

openmls/src/utils.rs
```diff
@@ -242,8 +242,8 @@ fn level(index: u32) -> usize {
     k
 }
 
-fn root(size: u32) -> u32 {
-    (1 << log2(size)) - 1
+fn _root(size: u32) -> u32 {
+    (1 << _log2(size)) - 1
 }
 
 pub fn _print_tree(tree: &TreeSync, message: &str) {
```
kkohbrok, line 249: Can't we hide this behind the test/test-utils cfg as well?

kkohbrok (commented):
Thanks for taking care of this much-needed clean up and re-org! Just a few minor questions.

openmls/src/group/core_group/mod.rs
```diff
@@ -306,6 +314,7 @@ impl CoreGroup {
     // struct {
     //     PreSharedKeyID psk;
     // } PreSharedKey;
+    #[cfg(any(feature = "test-utils", test))]
     pub fn create_presharedkey_proposal(
```
franziskuskiefer, line 318: Some of these are exposed for tests, others are not. I wonder why and how we could oganise these better. Maybe moving the `pub` `test` functions to a separate `impl` block?

openmls/src/group/core_group/staged_commit.rs
franziskuskiefer: This is another one of these issues. (See my general comment.)

franziskuskiefer (changes requested):
Thanks for getting started on this.
But I'm afraid this won't work and needs some changes.

> The public API is now exclusively exposed through prelude.rs.

That's a bad idea. The prelude is only for convenience and not the only entry point. If you want to leverage it for organising the API, you should `pub use prelude` in `lib.rs`

> Since the DS, cli, interop client & still need the CoreGroup, we need to expose it through prelude_test.rs.

Similarly to the issue with the other prelude. But not such a big deal here as it's going away.

You guard a ton of functions with `#[cfg(any(feature = "test-utils", test))]` that will just dissapear in all other configurations. I don't think this is what you want.
A lot of these functions, e.g `create_presharedkey_proposal`, are only used in a (single) test. Either these functions are needed, then they should be used. Or they are not needed (and that's how it looks) then they should go away and the tests should use the functions that are actually supposed to be used.

Another class of functions such as `merge_commit` are needed and will need to be either guarded differently or get two implementations (one for the test feature and one for the regular build).

raphaelrobert:
Thanks for the reviews! I agree with all comments. The idea was to get this started, obviously this will need more work.
You are right that the public API should also be exposed in `lib.rs` again. I can do that as part of this PR.
Regarding the test-only flag for `CoreGroup` methods: That's why I flagged #645. Quite a few of them don't really have a purpose anymore, but I didn't want to make this PR bigger and harder to review.

openmls/src/group/core_group/mod.rs
raphaelrobert: This is what #645 is meant for.

openmls/src/group/core_group/staged_commit.rs
raphaelrobert: Again, #645.

openmls/src/lib.rs
raphaelrobert: We should, I wasn't sure about the macros. Usually they need to be exposed at the root of the crate.

openmls/src/group/core_group/mod.rs
raphaelrobert: Yes, I guess that could work. Though in this case it's marked as test-only because the feature is not yet fully implemented. (see #141).

openmls/src/lib.rs
franziskuskiefer: They don't need to be, they will be 😉 Macros will always get the high level scope, independent of the module where they are defined.

franziskuskiefer (approved):
This will need a number of follow ups but let's get started with this.

Issue #626, opened 2021-12-13
get rid of one of the apply_proposals functions


kkohbrok:

kkohbrok:
Duplicate of #565.

PR #648, opened 2021-12-20
Have `create_commit` return a `StagedCommit`

### LINKED ISSUE #617: Return an additional StagedCommit from create_commit
It would be handy to return a `StagedCommit` from `create_commit()` so that it can be easily merged once the DS has signalled there is no conflict.

kkohbrok:
This PR changes the return type of create_commit such that it returns a `staged_commit` in addition to the other results. This should make processing more efficient as we don't need to go through verification before merging our own commit.

Notes:
- This PR includes a conversion from `CreationProposalQueue` to `StagedProposalQueue`, which clones all proposals in the Queue. This is required for now, as the `StagedProposalQueue` includes the full proposal rather than a reference, which in turn leads to easier reviewability by a consumer if it's passed out for review in the course of commit processing. We might be able to improve the whole structure, but I leave that to #626.
- This PR is large mostly due to the change of signature of `create_commit`. Beside adding a simple test for processing the `StagedCommit` returned by `create_commit`, I have not changed any test files beyond fixing them to work with the new function signature.

Fixes #617.


raphaelrobert (changes requested):
Thanks for tackling this!
I think this only makes sense if we also forbid processing our own commits in `stage_commit()`. The tests should then be changed accordingly (meaning you merge your own commit but stage the commit for other members). That way we get much better test coverage for these changes and we end up having what we wanted anyway.

kkohbrok:
I agree. The idea was to split it up so that we don't have another huge PR. Having `create_commit` return a `StagedCommit` as a first step seemed like a decent first step to me.

openmls/src/group/core_group/staged_commit.rs
```diff
@@ -329,6 +329,46 @@ pub struct StagedCommit {
 }
 
 impl StagedCommit {
+    /// Create a new [`StagedCommit`] from the provisional group state created
+    /// during the commit process.
+    pub(crate) fn from_provisional_state(
```
raphaelrobert, line 334: I'm wondering if this split is really worth it at this point. The issues I see are:
 - The list of parameters is already above the clippy limit (generates a warning)
 - You re-calculate the serialized group context (you already did that in `create_commit()`)
 - The remaining code is just a few lines that you could also do in `create_commit()`

I think it would be better not to have this separate function for now and look at it again in the follow-up PR to see what the overlap with `stage_commit()` is.

openmls/src/group/core_group/staged_commit.rs
```diff
@@ -329,6 +329,46 @@ pub struct StagedCommit {
...
+            provisional_epoch_secrets.split_secrets(
+                provisional_group_context.tls_serialize_detached()?,
+                diff.leaf_count(),
+            );
+
+        let provisional_interim_transcript_hash = update_interim_transcript_hash(
+            ciphersuite,
+            backend,
+            &(&confirmation_tag).into(),
+            &confirmed_transcript_hash,
```
raphaelrobert, line 354: clippy warning

raphaelrobert (changes requested):
Generally looks good, just one change and then the rest can be done in the follow-up PR.

openmls/src/group/core_group/staged_commit.rs
kkohbrok: Good point about the clippy limit. I was doing it in `create_commit` beforehand, but @franziskuskiefer pointed out in PR #634 that this function is already too large, so I factored this out.

openmls/src/group/core_group/create_commit.rs
```diff
@@ -252,17 +253,36 @@ impl CoreGroup {
...
-                path_processing_result.key_package_bundle,
-            ))
-        }
+            None
+        };
+
+        let provisional_interim_transcript_hash = update_interim_transcript_hash(
+            ciphersuite,
+            backend,
+            &(&confirmation_tag).into(),
```
raphaelrobert, line 264:
```suggestion
            &MlsPlaintextCommitAuthData::from(&confirmation_tag),
```

raphaelrobert (dismissed):
Thanks for making the changes! Just a nit to make it look nicer.

raphaelrobert (approved):

PR #653, opened 2021-12-21
Fix bug in free leaf computation and add simple test

kkohbrok:
This fixes the bug found by @franziskuskiefer in the process of reviewing #634, where instead of adding a leaf in the first blank position, it instead adds it in the last. The test is very minimal and only tests the exact conditions under which the bug occurs. More expansive tests and proper test utils will be added in the course of #625 and #624.

### DIFF
--- a/openmls/src/treesync/tests_and_kats/tests/test_diff.rs
+++ b/openmls/src/treesync/tests_and_kats/tests/test_diff.rs
@@ -12,6 +12,7 @@ use crate::{
 
 use openmls_rust_crypto::OpenMlsRustCrypto;
 
+// Verifies that when we add a leaf to a tree with blank leaf nodes, the leaf will be added at the leftmost free leaf index
 #[apply(ciphersuites_and_backends)]
 fn test_free_leaf_computation(
     ciphersuite: &'static Ciphersuite,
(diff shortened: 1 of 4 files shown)


openmls/src/treesync/tests_and_kats/tests/test_diff.rs
```diff
@@ -0,0 +1,72 @@
...
+    ciphersuite::CiphersuiteName,
+    config::Config,
+    credentials::{CredentialBundle, CredentialType},
+    prelude::KeyPackageBundle,
+    prelude_test::{node::Node, Ciphersuite, TreeSync},
+};
+
+use openmls_rust_crypto::OpenMlsRustCrypto;
+
+#[apply(ciphersuites_and_backends)]
```
raphaelrobert, line 15:
```suggestion
// Verifies that when we add a leaf to a tree with blank leaf nodes, the leaf will be added at the leftmost free leaf index
#[apply(ciphersuites_and_backends)]
```

openmls/src/treesync/tests_and_kats/tests/test_diff.rs
```diff
@@ -0,0 +1,72 @@
...
+        "leaf3".as_bytes().to_vec(),
+        CredentialType::Basic,
+        ciphersuite.signature_scheme(),
+        backend,
+    )
+    .expect("error creating credential_bundle");
+    let kpb_3 = KeyPackageBundle::new(&[ciphersuite.name()], &cb_3, backend, vec![])
+        .expect("error creating kpb");
+
+    // Build a rudimentary tree with two empty leaf nodes.
```
raphaelrobert, line 41:
```suggestion
    // Build a rudimentary tree with two populated and two empty leaf nodes.
```

franziskuskiefer (approved):
Thanks

raphaelrobert (approved):

PR #659, opened 2021-12-23
Introduce `pending_commit`

kkohbrok:
This PR does a few things.

### CoreGroup
- It disallows the processing of own commit messages via `stage_commit`. Instead, clients now have to process the `StagedCommit` resulting from their own `create_commit` calls directly via `merge_staged_commit`.
- It gets rid of the logic in `stage_commit` that covers own commits (and some code that was unused as a consequence). There is probably more to be cleaned up, but I'll leave that as a follow-up (see below).
- It fixes all tests that use the `core_group` interface to implement the flow described above. In some cases, this means that a new party had to be added to the group, because alice can't process her own commits anymore to test things.

### MlsGroup
- It stores the `StagedCommit` resulting from the most recent call to `create_commit` in the `MlsGroup` as the `pending_commit` and implements functions for inspecting and clearing it. 
- It forces the consumer to clear the pending commit explicitly before creating another commit. If this is not done, commit-creating functions will return an error.
- If a commit is merged (be it the pending one, or one from another group member), the `pending_commit` is cleared.
- It introduces a `merge_pending_commit` function that merges the `pending_commit` and consumes it in the process.
- It fixes all tests using the `MlsGroup` interface. As with the `CoreGroup`, this means that in some cases, an extra group member had to be added.

### TODOs
- [x] explicit tests for the new API flows in `MlsGroup`

### Follow-ups
- [ ] clean up unneeded code/logic. For example, `apply_own_update_path` doesn't need to return a KPB anymore.


openmls/src/framing/test_framing.rs
```diff
@@ -411,12 +413,13 @@ fn unknown_sender(ciphersuite: &'static Ciphersuite, backend: &impl OpenMlsCrypt
         .create_commit(params, backend)
         .expect("Error creating Commit");
 
-    let staged_commit = group_alice
-        .stage_commit(&create_commit_result.commit, &proposal_store, &[], backend)
-        .expect("Could not stage Commit");
     group_alice
-        .merge_commit(staged_commit)
-        .expect("error merging commit");
+        .merge_staged_commit(
```
raphaelrobert, line 417: Why not use `.merge_commit()` here? Just flagging it here, but it happens in other places as well.

openmls/src/group/mls_group/mod.rs
```diff
@@ -160,6 +163,21 @@ impl MlsGroup {
...
+    /// commit. If there was no commit created in this epoch, either because
+    /// this commit or another commit was merged, it returns `None`.
+    pub fn pending_commit(&self) -> Option<&StagedCommit> {
+        self.pending_commit.as_ref()
+    }
+
+    /// Sets the `pending_commit` to `None`.
+    ///
+    /// This function should only be used if it is clear that the pending commit
+    /// will not be used in the group.
```
raphaelrobert, line 176: Maybe add a slightly stronger warning here, since this can break the group.

openmls/src/group/mls_group/test_mls_group.rs
```diff
@@ -443,3 +422,175 @@ fn test_invalid_plaintext(ciphersuite: &'static Ciphersuite) {
...
+        .expect("An unexpected error occurred.");
+    assert!(alice_group.pending_commit().is_none());
+
+    if let ProcessedMessage::ProposalMessage(staged_proposal) = alice_processed_message {
+        alice_group.store_pending_proposal(*staged_proposal);
+    } else {
+        unreachable!("Expected a StagedCommit.");
+    }
+
+    // There should be no pending commit after issueing and processing a proposal.
```
raphaelrobert, line 499:
```suggestion
    // There should be no pending commit after issuing and processing a proposal.
```

raphaelrobert (changes requested):
Thanks for tackling this, this will unblock a number of things!

I did a first pass and just flagged some nits.

I'm wondering about the following: Would it make sense to combine the use after eviction error with the pending commit error? We could have a state machine along the lines of:

```rust
pub enum GroupActivityState {
    // We are a member of the group and there is no pending commit
    Active,
    // We are a member of the group and there is a pending commit
    PendingCommit,
    // We are no longer a member of the group
    Inactive,
}
```

openmls/src/framing/test_framing.rs
kkohbrok: Good point. This indeed happens in pretty much all tests now. But now I'm confused why both calls exist, since `merge_commit` also takes a `StagedCommit` as input. Also, `merge_commit` seems to be `test`/`test-utils` only, so it's probably better to use `merge_staged_commit`, so we use code that's actually used by the `MlsGroup` API as opposed to test-only code.

kkohbrok:
Thanks for the review!

I'm not sure if mixing the two state variables (`is_active` and `pending_commit`) is a good idea. It helps us combine two checks in the beginning of the committing functions, but otherwise the states partially overlap. For example, a group is active with or without a pending commit. Also, it's conceivable that there's still a pending commit when the client is removed from the group (although we could probably clear it at that point). Not a super strong opinion here, but I'd prefer to keep the states separated at the cost of an extra explicit check at the start of the committing functions.

openmls/src/framing/test_framing.rs
raphaelrobert: I agree that this is in a messy state right now. The reason why `merge_staged_commit` existed was that `CoreGroup` was public. I'm thinking that this should get cleaned up as part of #645. We'll probably end up having just one function that replaces `merge_staged_commit`, `merge_commit` and `merge_commit_take_message_secrets` and that takes a `StagedCommit` as its only parameter. So in that sense, it doesn't really matter which one you pick, but you picked the one that introduces the biggest change in parameters.

kkohbrok: Fair enough. If we're going to "merge" the three functions and end up with only one that takes a `StagedCommit` as input, it probably makes sense that I change everything over to `merge_commit` right now, while I'm touching the code in all of these places anyway.

raphaelrobert (commented):
Thanks for the changes!

On the logic side: Shouldn't we also reject the creation of proposals when there is a pending commit? Generating application messages might still be ok I guess.

openmls/src/schedule/mod.rs
```diff
@@ -1190,7 +1184,7 @@ impl EpochSecrets {
         let secret_tree = self.encryption_secret.create_secret_tree(treesize);
         (
             GroupEpochSecrets {
-                init_secret: self.init_secret,
+                init_secret: Some(self.init_secret),
```
raphaelrobert, line 1187: I missed this before: There's probably also no reason to keep the `Option` in `GroupEpochSecrets`.

kkohbrok:
> On the logic side: Shouldn't we also reject the creation of proposals when there is a pending commit? Generating application messages might still be ok I guess.

Yes, that makes sense. I'll put it in.

raphaelrobert (dismissed):
Thanks for making the changes! Let's get this in!

raphaelrobert (approved):

PR #666, opened 2021-12-23
Re-introduce parallelization

### LINKED ISSUE #108: Add parallelisation
Some tree operations like `encrypt_to_copath` can be parallelised with `rayon`.

raphaelrobert:
This PR re-introduces parallelization for TreeKEM operations.

Fixes #108.


kkohbrok (commented):
Very nice! Although this gives us `itertools` as a non-optional non-dev dependency. I'm not sure how hard we should try to avoid it as a dependency, but I think the only place that you're using it in this PR could be replaced by a relatively simple loop. I haven't tried it, though. If it's really just that one loop, it might be worth it to avoid the dependency.

raphaelrobert:
> 

Good point, I can try to get rid of it.

kkohbrok (approved):

Issue #565, opened 2021-11-22
Duplicate code in `apply_proposals` and `apply_staged_proposals`


kkohbrok:
The plan is to merge the two functions as part of the tree rewrite.

Issue #347, opened 2021-02-09
[Feature request] Don't clone proposals of the proposal queue


raphaelrobert:
**Is your feature request related to a problem? Please describe.**
Proposals are cloned when they are converted into a specific proposal type.

**Describe the solution you'd like**
Not clone.

**Describe alternatives you've considered**
None

**Additional context**
None

franziskuskiefer:
> Having to collect something here sounds like the APIs are bad. But that's probably nothing to solve in this PR.

_Originally posted by @franziskuskiefer in https://github.com/openmls/openmls/pull/397#r620043832_

raphaelrobert:
This is somewhat outdated after #667. Cloning still happens, and it is also necessary because we keep proposals in both the `StagedCommit` and the original `ProposalQueue`. It is not obvious that we can improve performance here, there's also the risk that the serialization gets altered and we would have to introduce a migration mechanism. Beyond that, it is also unclear whether this constitutes a noteworthy performance bottleneck. However, there seems to be one `.clone()` that we could easily eliminate.

PR #667, opened 2021-12-27
Refactor proposal queues

kkohbrok:
This PR refactors proposal queues and in the process
 - gets rid of the `CreationProposalQueue` and the associated implementations and structs
 - gets rid of the `apply_proposal` function (thus fixing #565)
 - Removes the `Staged...` prefix whenever it's followed by `Proposal...`, as these structs are now used both for creation and staging

The advantage of the `CreationProposalQueue` was that it relied on references rather than full `Proposal` structs. However, in the course of #617 we started returning a `StagedProposalQueue` at the end of `create_commit`, so the cloning was taking place in any case.

This PR is mostly renaming an code removal with the exception of `filter_proposals`, where I had to clone a proposal where previously a reference was created.

### DIFF
--- a/cli/src/user.rs
+++ b/cli/src/user.rs
@@ -335,7 +335,7 @@ impl User {
             .borrow()
             .create_add_proposal(framing_parameters, credentials, key_package, &self.crypto)
             .expect("Could not create proposal.");
-        let proposal_store = ProposalStore::from_staged_proposal(
+        let proposal_store = ProposalStore::from_queued_proposal(
             QueuedProposal::from_mls_plaintext(
                 Config::ciphersuite(CIPHERSUITE).map_err(|e| format!("{}", e))?,
                 &self.crypto,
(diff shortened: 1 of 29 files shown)


raphaelrobert (changes requested):
Thanks for picking this up!
A few nits:
 - in comments, function names and parameters the word "staging" still appears, I think it would be good to be consistent with the naming
 - There a quite a few clippy warnings (clippy is still not fixed on the CI)
 - Can you check if this closes any of the tickets related to proposals/proposal queues?

kkohbrok:
Thanks for the review! I did some more renaming and documentation cleanup. I did fix a clippy complaint, but most of the clippy warnings are unrelated to this PR, so I'll leave them for now. I also checked the issues and this PR indeed fixes #565. Other than that there are no other issues I could find except maybe #347, which, if anything got worse 😅.

raphaelrobert (approved):
Thanks, this makes it much cleaner and shorter!

Issue #473, opened 2021-10-04
[Feature request] Provide API to inspect MlsPlaintext & MlsCiphertext message at the DS level


raphaelrobert:
The DS might have the need to inspect messages to process them (e.g. it should be able to see the group ID).

This is already partially possible with `VerifiableMlsPlaintext`, but it might be easier to expose more functions on `MlsMessageIn`.

raphaelrobert:
Now that the `PublicGroup` is there, we have everything we need.

PR #670, opened 2021-12-27
Include MlsMessageIn and MlsMessageOut in prelude

sgmenda:
Ref #473

### DIFF
--- a/openmls/src/framing/mod.rs
+++ b/openmls/src/framing/mod.rs
@@ -28,6 +28,7 @@ pub(crate) use sender::*;
 
 // Public
 pub use ciphertext::*;
+pub use message::*;
 pub use plaintext::*;
 pub use validation::*;
 
--- a/openmls/src/framing/mod.rs
+++ b/openmls/src/framing/mod.rs
@@ -23,7 +23,6 @@ pub(crate) mod validation;
 
 // Crate
 pub(crate) use errors::*;
-pub(crate) use message::*;
 pub(crate) use sender::*;
 
 // Public
--- a/openmls/src/prelude.rs
+++ b/openmls/src/prelude.rs
@@ -33,8 +33,7 @@ pub use crate::extensions::*;
 
 // Framing
 // TODO #265: This should mostly disappear
-pub use crate::framing::{errors::*, sender::Sender, *};
-pub use crate::framing::{MlsMessageIn, MlsMessageOut};
+pub use crate::framing::{errors::*, message::*, sender::Sender, *};
 
 // Key packages
 pub use crate::key_packages::*;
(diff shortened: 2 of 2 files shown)


openmls/src/prelude.rs
```diff
@@ -34,6 +34,7 @@ pub use crate::extensions::*;
 // Framing
 // TODO #265: This should mostly disappear
 pub use crate::framing::{errors::*, sender::Sender, *};
+pub use crate::framing::{MlsMessageIn, MlsMessageOut};
```
raphaelrobert, line 37: Is this really needed, given that we already expose `*` in the line above?

sgmenda: Apparently yes, if I remove that line it yells at me with error messages like the following.
```
error[E0433]: failed to resolve: use of undeclared type `MlsMessageOut`
   --> src/user.rs:120:13
    |
120 |             MlsMessageOut::Ciphertext(mls_ctxt_ptr) => *mls_ctxt_ptr,
    |             ^^^^^^^^^^^^^ use of undeclared type `MlsMessageOut`
```

Taking a step back, this might be because `MlsMessageIn` is defined in `crate::framing::message` and not `crate::framing`, so an alternative is to instead use
```
use crate::framing::message::*
```

raphaelrobert: Right, I guess it should be

```
pub use crate::framing::{errors::*, message::*, sender::Sender, *};
```

openmls/src/framing/mod.rs
```diff
@@ -28,6 +28,7 @@ pub(crate) use sender::*;
 
 // Public
 pub use ciphertext::*;
+pub use message::*;
```
raphaelrobert, line 31: You should remove the `pub(crate)` in line 26.

raphaelrobert (changes requested):

openmls/src/framing/mod.rs
sgmenda: I am not sure I follow, are you suggesting making it `pub` rather than `pub(crate)`?

raphaelrobert: You made it `pub` in line 31, which supersedes the `pub(crate)` in line 26. So you can delete line 26.

sgmenda: oops... good catch.

raphaelrobert (approved):
Thanks!

raphaelrobert:
If this is good to merge you can update the branch and then (squash) merge.

sgmenda:
> If this is good to merge you can update the branch and then (squash) merge.

I updated it, but looks like I cannot merge since I don't have write access, can you merge it?

sgmenda:
The coverage task seems broken?

PR #674, opened 2021-12-30
Remove unnecessary `KeyPackageBundle` return during commit processing

### LINKED ISSUE #673: clean up code that is unnecessary after the introduction of the `pending_commit` logic
For example, we no longer need to return key package bundles for storage until a commit is merged.

kkohbrok:
Since we're now relying on the stored `StagedCommit` to remember data from a commit, we don't need to return (and store in case of the `MlsGroup`) a `KeyPackageBundle` whenever we create a commit.

This PR removes the returned `KeyPackageBundle` throughout the `create_commit` chain of calls through the stack and thus fixes #673.

### DIFF
--- a/openmls/tests/test_group.rs
+++ b/openmls/tests/test_group.rs
@@ -89,14 +89,15 @@ fn create_commit_optional_path(
         .credential_bundle(&alice_credential_bundle)
         .proposal_store(&proposal_store)
         .build();
-    let create_commit_result = group_alice
-        .create_commit(params, backend)
-        .expect("error creating commit");
+    let create_commit_result =
+        match group_alice.create_commit(params /* No PSK fetcher */, backend) {
+            Ok(c) => c,
+            Err(e) => panic!("Error creating commit: {:?}", e),
+        };
     let commit = match create_commit_result.commit.content() {
-        MlsPlaintextContentType::Commit(commit) => Some(commit),
-        _ => None,
-    }
-    .expect("error getting commit content from create_commit_result");
+        MlsPlaintextContentType::Commit(commit) => commit,
+        _ => panic!(),
+    };
     assert!(commit.has_path());
 
     // Alice adds Bob without forced self-update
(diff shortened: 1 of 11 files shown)


openmls/tests/test_group.rs
```diff
@@ -454,8 +453,12 @@ fn group_operations(ciphersuite: &'static Ciphersuite, backend: &impl OpenMlsCry
         Err(e) => panic!("Error creating commit: {:?}", e),
     };
 
-    // Check that there is a new KeyPackageBundle
-    assert!(create_commit_result.key_package_bundle_option.is_some());
+    // Check that there is a path
+    let commit = match create_commit_result.commit.content() {
+        MlsPlaintextContentType::Commit(commit) => commit,
+        _ => panic!("Wrong content type"),
```
raphaelrobert, line 459: I'm never too sure if `panic!` in a branching statement isn't throwing codecov off course.
The alternative would be to return an `Option` that then gets unwrapped with an `expect()`. I don't have a strong opinion though.

raphaelrobert (approved):
Thanks, this looks neat! Just one question about testing.

openmls/tests/test_group.rs
kkohbrok: Only one way to find out. I removed a bunch of panics. Before, we had `Coverage   86.92%   87.08%   +0.15%` according to codecov..

kkohbrok: Looks like it actually got worse: to the point that codecov is not green anymore: `Coverage   86.92%   87.02%   +0.09%` 🤷🏻‍♂️

raphaelrobert: Thanks for trying! At least now we know. Do you want to roll back the changes?

kkohbrok: Why not. I think at this point it doesn't really matter.

raphaelrobert (approved):
Thanks!

PR #676, opened 2021-12-30
Enable processing of own external commits

kkohbrok:
This PR allows clients that want to join a group via an External Init to process their own staged commit. This was already possible after #659, but this PR introduces proper state transitions in the `MlsGroup`.

This PR is part of the larger effort to enable the external init process, which is tracked in #630.

### DIFF
--- a/openmls/src/group/mls_group/creation.rs
+++ b/openmls/src/group/mls_group/creation.rs
@@ -89,8 +89,8 @@ impl MlsGroup {
         Ok(mls_group)
     }
 
-    /// Creates a new group by joining an existing group through an External
-    /// Init. The resulting [`MlsGroup`] instance starts off with a pending
+    /// Join an existing group through an External Commit.
+    /// The resulting [`MlsGroup`] instance starts off with a pending
     /// commit (the external commit, which adds this client to the group).
     /// Merging this commit is necessary for this [`MlsGroup`] instance to
     /// function properly, as, for example, this client is not yet part of the
--- a/openmls/src/group/mls_group/creation.rs
+++ b/openmls/src/group/mls_group/creation.rs
@@ -100,7 +100,7 @@ impl MlsGroup {
     /// created using this function based on the latest `ratchet_tree` and
     /// public group state. For more information on the external init process,
     /// please see Section 11.2.1 in the MLS specification.
-    pub fn new_from_external_init(
+    pub fn join_by_external_commit(
         backend: &impl OpenMlsCryptoProvider,
         mls_group_config: &MlsGroupConfig,
         ratchet_tree: Option<Vec<Option<Node>>>,
@@ -114,7 +114,7 @@ impl MlsGroup {
             ResumptionSecretStore::new(mls_group_config.number_of_resumption_secrets);
         let framing_parameters = FramingParameters::new(aad, mls_group_config.wire_format());
         let ratchet_tree: Option<&[Option<Node>]> = ratchet_tree.as_deref();
-        let (group, create_commit_result) = CoreGroup::new_from_external_init(
+        let (group, create_commit_result) = CoreGroup::join_by_external_commit(
             backend,
             framing_parameters,
             ratchet_tree,
@@ -122,9 +122,7 @@ impl MlsGroup {
             proposals_by_reference,
             proposals_by_value,
             verifiable_public_group_state,
-        )
-        //FIXME: change new_from_external_init such that it returns a CoreGroupError
-        .map_err(|_| MlsGroupError::LibraryError("".into()))?;
+        )?;
 
         let mls_group = MlsGroup {
             mls_group_config: mls_group_config.clone(),
--- a/openmls/src/group/mls_group/creation.rs
+++ b/openmls/src/group/mls_group/creation.rs
@@ -1,3 +1,5 @@
+use crate::messages::VerifiablePublicGroupState;
+
 use super::*;
 
 impl MlsGroup {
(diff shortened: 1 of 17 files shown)


openmls/src/group/mls_group/creation.rs
```diff
@@ -81,11 +82,64 @@ impl MlsGroup {
...
-            active: true,
+            group_state: MlsGroupState::Operational,
             state_changed: InnerState::Changed,
-            pending_commit: None,
         };
 
         Ok(mls_group)
     }
+
+    /// Creates a new group by joining an existing group through an External
```
raphaelrobert, line 92:
```suggestion
    /// Join an existing group through an External Commit.
```

openmls/src/group/mls_group/creation.rs
```diff
@@ -81,11 +82,64 @@ impl MlsGroup {
...
+    /// Init. The resulting [`MlsGroup`] instance starts off with a pending
```
raphaelrobert, line 93:
```suggestion
    /// The resulting [`MlsGroup`] instance starts off with a pending
```

openmls/src/group/mls_group/creation.rs
```diff
@@ -81,11 +82,64 @@ impl MlsGroup {
...
+    /// commit (the external commit, which adds this client to the group).
+    /// Merging this commit is necessary for this [`MlsGroup`] instance to
+    /// function properly, as, for example, this client is not yet part of the
+    /// tree. As a result, it is not possible to clear the pending commit. If
+    /// the external commit was rejected due to an epoch change, the
+    /// [`MlsGroup`] instance has to be discarded and a new one has to be
+    /// created using this function based on the latest `ratchet_tree` and
+    /// public group state. For more information on the external init process,
+    /// please see Section 11.2.1 in the MLS specification.
+    pub fn new_from_external_init(
```
raphaelrobert, line 103: We should rethink the naming in general. This sounds very technical and something like `join_by_external_commit` might be more expressive. But we can do that in another PR.

openmls/src/group/mls_group/creation.rs
```diff
@@ -81,11 +82,64 @@ impl MlsGroup {
...
+            backend,
+            framing_parameters,
+            ratchet_tree,
+            credential_bundle,
+            proposals_by_reference,
+            proposals_by_value,
+            verifiable_public_group_state,
+        )
+        //FIXME: change new_from_external_init such that it returns a CoreGroupError
+        .map_err(|_| MlsGroupError::LibraryError("".into()))?;
```
raphaelrobert, line 127: We don't do any validation regarding the allowed proposals here or in the `CoreGroup`. We can do that in a follow-up PR, but it would be good to leave a comment here and make a note in the wiki. The spec mentions the following constraints:

The proposals included by value in an External Commit MUST meet the following conditions:
 - There MUST be a single ExternalInit proposal
 - There MUST NOT be any Add or Update proposals
 - If a Remove proposal is present, then the credential and endpoint_id of the removed leaf MUST be the same as the corresponding values in the Add KeyPackage.

The proposals included by reference in an External Commit MUST meet the following conditions:
 - There MUST NOT be any ExternalInit proposals

openmls/src/group/mls_group/processing.rs
```diff
@@ -141,18 +146,31 @@ impl MlsGroup {
...
+        match &self.group_state {
+            MlsGroupState::PendingCommit(_) => {
+                let old_state = mem::replace(&mut self.group_state, MlsGroupState::Operational);
+                if let MlsGroupState::PendingCommit(pending_commit_state) = old_state {
+                    self.merge_staged_commit(pending_commit_state.into())
+                } else {
+                    // We should never reach this state, as we previously
+                    // checked that there actually is a pending commit.
+                    Err(MlsGroupError::LibraryError(
+                        "Reached unreachable state during pending commit merge.".into(),
```
raphaelrobert, line 166: Why do we need the `else` clause at all?

raphaelrobert (changes requested):
Thanks for tackling this! I think it looks good in general, I just flagged some nits.

openmls/src/group/mls_group/creation.rs
kkohbrok: (Sorry about that FIXME. That was meant to go away.) Good point about the validation. There is a TODO/sub-issue in #630 for that. I'll make it a real issue and mention it here, though.

openmls/src/group/mls_group/processing.rs
kkohbrok: Because the `match` needs to return something if `old_state` is not a `PendingCommit` (which it can't be, hence the error). Is that what you meant?

openmls/src/group/mls_group/creation.rs
kkohbrok: You're right. I somehow thought that the canonical name for the process was "External Init", but the spec is not very clear on that. There is the notion of an "External Commit", which allows you to join (and which is why your suggested function name makes sense) and in the Key Schedule, there is "External Initialization". But you're right in that "join through an external commit" is probably the best way to go.

kkohbrok:
I did a bunch more refactoring an renaming. There are basic tests for the `core_group` version of `join_by_external_commit`, which I think are enough for now. More tests (also on the `mls_group` level will follow as soon as we have validation (#681).

openmls/src/group/mls_group/processing.rs
raphaelrobert: Since the function returns `()` you don't have to return something here. You can just return `Ok(())` in the end and abort in line 161 if there's an error and then delete the whole `else` clause since we know for sure what's happening.

kkohbrok: I see what you mean. That does indeed sound a bit cleaner.

raphaelrobert (approved):
Thanks!

PR #678, opened 2021-12-30
Improve FS when encrypting messages

### LINKED ISSUE #243: Improve Forward Secrecy for MLSCiphertext encryption

raphaelrobert:
This PR enforces immediate forward secrecy when encrypting messages, i.e. the encryption key gets deleted right away.
It introduces automatic storing of standalone proposals in the `MlsGroup`'s proposal store, since we can no longer decrypt our own proposals.
The alternative would be to hand out the proposals in the same way we hand out our own commits and expect the consumer to manually store them in the proposal store once the DS signaled there are no conflicts.
That would be more in line with how we treat other messages. I went for the simpler option for now, because the alternative would lead to quite a few code changes. If we have consensus on the alternative I can change it.

Fixes #243.

### DIFF
--- a/openmls/src/test_utils/test_framework/mod.rs
+++ b/openmls/src/test_utils/test_framework/mod.rs
@@ -240,11 +240,18 @@ impl MlsGroupTestSetup {
             CodecUse::StructMessages => MlsMessageIn::from(message.clone()),
         };
         let clients = self.clients.read().expect("An unexpected error occurred.");
-        // Distribute message to all members.
+        // Distribute message to all members, except to the sender in the case of application messages
         let results: Result<Vec<_>, _> = group
             .members
             .par_iter()
-            .map(|(_index, member_id)| {
+            .filter_map(|(_index, member_id)| {
+                if message.content_type() == ContentType::Application && member_id == sender_id {
+                    None
+                } else {
+                    Some(member_id)
+                }
+            })
+            .map(|member_id| {
                 let member = clients
                     .get(member_id)
                     .expect("An unexpected error occurred.")
--- a/openmls/src/tree/sender_ratchet.rs
+++ b/openmls/src/tree/sender_ratchet.rs
@@ -94,6 +94,7 @@ impl SenderRatchet {
         }
         // If generation is potentially within the window
         if generation <= self.generation {
+            // If the requested generation is within the window of past secrets, we should get a positive index
             let window_index =
                 self.past_secrets.len() as i32 - ((self.generation - generation) as i32) - 1;
             // We might not have the key material (e.g. we might have discarded it when generating an encryption secret)
--- a/openmls/src/tree/sender_ratchet.rs
+++ b/openmls/src/tree/sender_ratchet.rs
@@ -11,9 +11,6 @@ use crate::tree::{index::LeafIndex, secret_tree::*};
 use super::index::NodeIndex;
 use super::*;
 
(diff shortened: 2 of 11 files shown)


openmls/src/tree/sender_ratchet.rs
```diff
@@ -11,9 +11,6 @@ use crate::tree::{index::LeafIndex, secret_tree::*};
 use super::index::NodeIndex;
 use super::*;
 
-// TODO #265: This should disappear
```
kkohbrok, line 14: So this was meant to be #243 and not #265?

openmls/src/tree/sender_ratchet.rs
```diff
@@ -95,14 +92,20 @@ impl SenderRatchet {
         {
             return Err(SecretTreeError::TooDistantInThePast);
         }
-        // If generation is within the window
+        // If generation is potentially within the window
         if generation <= self.generation {
             let window_index =
-                (self.past_secrets.len() as u32 - (self.generation - generation) - 1) as usize;
+                self.past_secrets.len() as i32 - ((self.generation - generation) as i32) - 1;
```
kkohbrok, line 98: Can you explain the computation here in a code comment?

openmls/src/test_utils/test_framework/mod.rs
```diff
@@ -240,11 +240,18 @@ impl MlsGroupTestSetup {
             CodecUse::StructMessages => MlsMessageIn::from(message.clone()),
         };
         let clients = self.clients.read().expect("An unexpected error occurred.");
-        // Distribute message to all members.
+        // Distribute message to all members, except to the sender in the case of application messages
```
kkohbrok, line 243: Shouldn't this be a problem for handshake messages as well?

kkohbrok (changes requested):
Thanks for picking this up!

One comment regarding storing proposals ourselves vs. letting the consumer store them: The difference between proposals and commits here is that the DS doesn't really get to decide the validity of proposals in the same way as for commits, since they are not as epoch-sensitive. So I'm Ok with this approach, since in the end, the client has to inspect and validate all proposals anyway.

Otherwise there are just a few comments.

openmls/src/test_utils/test_framework/mod.rs
raphaelrobert: That's true, but the test framework is only used with plaintext messages currently. It was not adapted to the pending commit logic in #659, so it currently cannot do ciphertext messages I guess?

openmls/src/tree/sender_ratchet.rs
raphaelrobert: I guess I didn't predict the future correctly. It wasn't clear which issue would be implemented first (or whether they would both be implemented together).

openmls/src/test_utils/test_framework/mod.rs
kkohbrok: I'm pretty sure it was adapted. It could be that no test actually configures a group to encrypt handshake messages. I'll file a follow up to investigate.

kkohbrok (approved):
Thanks!

openmls/src/tree/sender_ratchet.rs
kkohbrok: Fair enough 😄

PR #683, opened 2022-01-02
Upgrade DS to accept MlsMessageIn

### LINKED ISSUE #642: DS & cli should not use CoreGroup anymore

sgmenda:
Close #642

* Expose converting `MlsMessageOut` to `MlsMessageIn`
* Use `MlsMessageIn` in the DS instead of `DsMlsMessage`
* Remove `DsMlsMessage`
* Update the DS tests to use `MlsGroup` instead of `CoreGroup`

### DIFF
--- a/cli/src/user.rs
+++ b/cli/src/user.rs
@@ -1,6 +1,6 @@
 use std::{cell::RefCell, collections::HashMap};
 
-use ds_lib::{ClientKeyPackages, DsMlsMessage, GroupMessage, Message};
+use ds_lib::{ClientKeyPackages, GroupMessage, Message};
 use openmls::prelude::*;
 use openmls_rust_crypto::OpenMlsRustCrypto;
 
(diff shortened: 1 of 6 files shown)


cli/src/user.rs
```diff
@@ -161,46 +150,33 @@ impl User {
...
-                                    );
-                                    continue;
-                                }
-                            };
-                            processed_message
+                    let unverified_message = match mls_group.parse_message(message, &self.crypto) {
+                        Ok(msg) => msg,
+                        Err(e) => {
+                            log::error!(
+                                "Error decrypting MlsCiphertext: {:?} -  Dropping message.",
```
raphaelrobert, line 157: The error message is not quite accurate, in theory it could also be an `MlsPlaintext`. If you want to enforce the wire format (like it was the case until now), you would have to look `message.wire_format()`.

cli/src/user.rs
```diff
@@ -161,46 +150,33 @@ impl User {
...
+                        unverified_message,
+                        None,
+                        &self.crypto,
+                    ) {
+                        Ok(msg) => msg,
+                        Err(e) => {
                             log::error!(
-                                "Received Plaintext message: {:?} -  Dropping message.",
-                                msg
+                                "Error verifying MlsPlaintext: {:?} -  Dropping message.",
```
raphaelrobert, line 172: Same here regarding the wire format.

delivery-service/ds/src/test.rs
```diff
@@ -208,38 +248,12 @@ async fn test_group() {
...
-        .proposal_store(&proposal_store)
-        .force_self_update(false)
-        .build();
-    let create_commit_results = group
-        .create_commit(params, crypto)
-        .expect("Error creating commit");
-    let welcome_msg = create_commit_results
-        .welcome_option
-        .expect("Welcome message wasn't created by create_commit.");
+    let (_out_messages, welcome_msg) = group
```
raphaelrobert, line 251: Re the comment above: this is now a Commit and no longer just a proposal

raphaelrobert (changes requested):
Nice, thanks for picking it up! Looks good in general, just a few nits.

I think this should also fix #642. If so, you can add it in the PR description and it'll close the ticket on merge.

cli/src/user.rs
sgmenda: Good catch! I'll update the error message.

raphaelrobert (approved):
Thanks for the changes, let's get this in!

sgmenda:
Thank you for the review! ✨

PR #686, opened 2022-01-03
Introduce byte buffers for messages

### LINKED ISSUE #265: [Feature request] Use byte buffers for managed API

raphaelrobert:
Fixes #265.

This PR introduces two opaque type for MLS messages: `MlsMessageIn` and `MlsMessageOut`. The types are identical, the distinction is for the added benefit of type safety.

While this doesn't change the current functionality, it is a step towards making `Mls(Plain|Cipher)text` messages crate-only in order to further restrict the public API.

### DIFF
--- a/openmls/src/framing/message.rs
+++ b/openmls/src/framing/message.rs
@@ -1,10 +1,12 @@
 use super::*;
 
-/// Unified message type for incoming MLS messages.
+/// Unified message type for MLS messages.
 /// Since the memory footprint can differ considerably between [`VerifiableMlsPlaintext`]
 /// and [`MlsCiphertext`], we use `Box<T>` for more efficient memory allocation.
-#[derive(Debug, Clone)]
-pub enum MlsMessageIn {
+/// This is only used internally, externally we use either [`MlsMessageIn`] or
+/// [`MlsMessageOut`], depending on the context.
+#[derive(PartialEq, Debug, Clone)]
+pub(crate) enum MlsMessage {
     /// An OpenMLS `VerifiableMlsPlaintext`.
     Plaintext(Box<VerifiableMlsPlaintext>),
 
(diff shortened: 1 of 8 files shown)


openmls/src/framing/message.rs
```diff
@@ -1,127 +1,212 @@
...
 
     /// Returns `true` if this is a handshake message and `false` otherwise.
-    pub fn is_handshake_message(&self) -> bool {
+    fn is_handshake_message(&self) -> bool {
         self.content_type().is_handshake_message()
     }
+
+    /// Tries to deserialize from a byte slice. Returns [`MlsMessageError::DecodingError`] on failure.
+    fn try_from_bytes(bytes: &[u8]) -> Result<Self, MlsMessageError> {
+        let mut reader = BufReader::new(bytes);
```
franziskuskiefer, line 61: Is using `BufReader` more efficient than just using the slice?
If not, why not just call `tls_deserialize` on the slice?

openmls/src/framing/message.rs
```diff
@@ -1,127 +1,212 @@
+use std::io::BufReader;
```
franziskuskiefer, line 1: Can you add here, or in the `framing` module, a description of the different types and the intended workflow? We need a) the one for the public API, and b) one for the internal types.

franziskuskiefer (changes requested):
Thanks, only two nits.

openmls/src/framing/message.rs
raphaelrobert: The problem with using it on the slice directly is that the slice needs to be mutable and I don't think that's a great requirement from a consumer perspective. I have replaced this with a `Cursor` now, because I realised that a `BufReader` would make a copy of the data.

franziskuskiefer: The slice doesn't need to be mutable. I also get confused by this sometimes. But the slice is only a pointer into a structure that's fine to modify. So something like this should work.
```rust
    fn try_from_bytes(mut bytes: &[u8]) -> Result<Self, MlsMessageError> {
        MlsMessage::tls_deserialize(&mut bytes).map_err(|_| MlsMessageError::DecodingError)
    }
```

raphaelrobert: Nice one!

franziskuskiefer (approved):
Thanks!

PR #687, opened 2022-01-05
Use MlsGroup in interop client

raphaelrobert:
This PR gets rid of `CoreGroup` in the interop client.

The functionality was not changed, however, I noticed by looking at the old code that the Commit checks (CommitRequest/CommitResponse) were probably not functional. This will have to be fixed when we work on the client again.

The client still uses the `test-utils` feature because it needs access to the KAT vectors. This can probably be changed at some point if we drop that feature and introduce a dedicated feature for the interop client. See  #691.

Fixes  #687.

### DIFF
--- a/interop_client/src/main.rs
+++ b/interop_client/src/main.rs
@@ -788,7 +788,7 @@ impl MlsClient for MlsClientImpl {
 
         // Proposals by value. These proposals are inline proposals. They should be converted into group operations.
 
-        // TODO: The interop client cannot process these proposals yet.
+        // TODO #692: The interop client cannot process these proposals yet.
 
         let (commit, welcome_option) = interop_group
             .group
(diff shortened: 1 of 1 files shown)


interop_client/src/main.rs
```diff
@@ -779,117 +759,45 @@ impl MlsClient for MlsClientImpl {
...
 
-        let params = CreateCommitParams::builder()
-            .framing_parameters(framing_parameters)
-            .credential_bundle(&interop_group.credential_bundle)
-            .proposal_store(&proposal_store)
-            .force_self_update(false)
-            .build();
+        // Proposals by value. These proposals are inline proposals. They should be converted into group operations.
+
+        // TODO: The interop client cannot process these proposals yet.
```
franziskuskiefer, line 791: An issue here with a number :)

franziskuskiefer (changes requested):
> This will have to be fixed when we work on the client again.

Can you file an issue for this so we don't forget about it?

franziskuskiefer (approved):

Issue #691, opened 2022-01-05
[Meta] Refactor CoreGroup/Internal structure


raphaelrobert:
- [ ] Identify tests for `CoreGroup` that are no longer needed and delete them
 - [ ] Remove the `test-utils` feature, introduce `interop-client` feature instead and only expose where needed
 - [ ] #645 
 - [ ] Remove the config of `CoreGroup`
 - [ ] Merge `CoreGroup` with `MlsGroup`, possibly by making it a submodule of the latter

kkohbrok:
I'm not sure it's a good idea to merge `CoreGroup` with `MlsGroup`. I think `CoreGroup` marks a good boundary of what's in the spec vs. what OpenMLS does on top of that.

raphaelrobert:
I think this is already covered enough by #645 and #739.

PR #693, opened 2022-01-05
Reduce visibility

raphaelrobert:
This PR reduces the visibility of `MlsPlaintext` and `MlsCiphertext` to crate-only. As a consequence, `CoreGroup` could also now become crate-only as intended.
Unused functions are dropped and minor changes are made to accommodate the change.
This will need more work as outlined in #691, but I want to keep it simple for now.

### DIFF
--- a/openmls/src/framing/validation.rs
+++ b/openmls/src/framing/validation.rs
@@ -286,7 +286,7 @@ impl UnverifiedExternalMessage {
     ) -> Result<VerifiedExternalMessage, ValidationError> {
         // ValSem10
         match self.plaintext.verify_with_key(backend, signature_key) {
-            Ok(plaintext) => Ok(VerifiedExternalMessage { plaintext }),
+            Ok(_plaintext) => Ok(VerifiedExternalMessage { _plaintext }),
             Err(e) => Err(e.into()),
         }
     }
@@ -299,30 +299,31 @@ pub struct VerifiedMemberMessage {
 
 impl VerifiedMemberMessage {
     /// Returns a reference to the inner [MlsPlaintext].
-    pub fn plaintext(&self) -> &MlsPlaintext {
+    pub(crate) fn plaintext(&self) -> &MlsPlaintext {
         &self.plaintext
     }
 
     /// Consumes the message and returns the inner [MlsPlaintext].
-    pub fn take_plaintext(self) -> MlsPlaintext {
+    pub(crate) fn take_plaintext(self) -> MlsPlaintext {
         self.plaintext
     }
 }
 
 /// External message, where all semantic checks on the framing have been successfully performed.
+/// Note: External messages are not fully supported yet #106
 pub struct VerifiedExternalMessage {
-    plaintext: MlsPlaintext,
+    _plaintext: MlsPlaintext,
 }
 
 impl VerifiedExternalMessage {
     /// Returns a reference to the inner [MlsPlaintext].
-    pub fn plaintext(&self) -> &MlsPlaintext {
-        &self.plaintext
+    pub(crate) fn _plaintext(&self) -> &MlsPlaintext {
+        &self._plaintext
     }
 
     /// Consumes the message and returns the inner [MlsPlaintext].
-    pub fn take_plaintext(self) -> MlsPlaintext {
-        self.plaintext
+    pub(crate) fn _take_plaintext(self) -> MlsPlaintext {
+        self._plaintext
     }
 }
 
--- a/openmls/src/group/core_group/mod.rs
+++ b/openmls/src/group/core_group/mod.rs
@@ -55,10 +55,9 @@ use serde::{
     ser::{SerializeStruct, Serializer},
     Deserialize, Deserializer, Serialize,
 };
-#[cfg(any(feature = "test-utils", test))]
-use std::collections::BTreeMap;
+#[cfg(test)]
 use std::convert::TryFrom;
-#[cfg(any(feature = "test-utils", test))]
+#[cfg(test)]
 use std::io::{Error, Read, Write};
 
 use tls_codec::Serialize as TlsSerializeTrait;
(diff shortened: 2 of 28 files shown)


openmls/src/framing/validation.rs
```diff
@@ -299,30 +299,31 @@ pub struct VerifiedMemberMessage {
...
+    pub(crate) fn take_plaintext(self) -> MlsPlaintext {
         self.plaintext
     }
 }
 
 /// External message, where all semantic checks on the framing have been successfully performed.
+/// Note: External messages are not fully supported yet #106
 pub struct VerifiedExternalMessage {
-    plaintext: MlsPlaintext,
+    _plaintext: MlsPlaintext,
```
franziskuskiefer, line 315: Is this to avoid unused code warning?
I'd prefer warnings rather than hiding them so we know there's something to fix.

openmls/src/group/core_group/mod.rs
```diff
@@ -55,10 +55,9 @@ use serde::{
     ser::{SerializeStruct, Serializer},
     Deserialize, Deserializer, Serialize,
 };
-#[cfg(any(feature = "test-utils", test))]
-use std::collections::BTreeMap;
+#[cfg(test)]
```
franziskuskiefer, line 58: How about we move all these things into a separate module that is guarded by `#[test]` and `impl` blocks guarded by `#[test]`? Then we can declutter all of this.

Same goes for all the other `test` changes in the other files.

openmls/src/group/core_group/mod.rs
```diff
@@ -127,24 +126,24 @@ impl CoreGroupBuilder {
...
+    #[cfg(test)]
+    pub(crate) fn with_psk(mut self, psk_ids: Vec<PreSharedKeyId>) -> Self {
         self.psk_ids = psk_ids;
         self
     }
     /// Set the [`ProtocolVersion`] of the [`CoreGroup`].
-    #[cfg(any(feature = "test-utils", test))]
-    pub fn with_version(mut self, version: ProtocolVersion) -> Self {
+    #[cfg(test)]
+    pub(crate) fn _with_version(mut self, version: ProtocolVersion) -> Self {
```
franziskuskiefer, line 141: `test` and it's still unused? Why do we need it then? 🤔

openmls/src/group/mod.rs
```diff
@@ -28,13 +27,13 @@ pub(crate) use group_context::*;
 
 // Public
 pub use mls_group::*;
+pub mod errors;
 
 // Tests
 #[cfg(any(feature = "test-utils", test))]
+pub(crate) use create_commit_params::*;
 pub(crate) mod tests;
```
franziskuskiefer, line 35: This module should still be guarded.

openmls/src/group/tests/mod.rs
```diff
@@ -1,5 +1,6 @@
 //! Unit tests for the core group
 
+#[cfg(any(feature = "test-utils", test))]
```
franziskuskiefer, line 3: These two `cfg` guards shouldn't be necessary.

franziskuskiefer (changes requested):
Thanks, we can probably make the visibility probably even stricter. But that's not for this PR.
Only a couple nits.
If you prefer, we can defer pulling out test-module to a follow-up PR (see inline comment).

openmls/src/framing/validation.rs
raphaelrobert: That's why I added the note above. I don't think it's a good idea to keep clippy warnings around for a long time (#106 is not part of milestone v0.4), it leads to warning fatigue and we might miss actually important warnings.

openmls/src/group/core_group/mod.rs
raphaelrobert: Fully agree, but I'd prefer to do it as part of a follow-up PR (in the context of #691).

openmls/src/group/tests/mod.rs
raphaelrobert: Now that the module is guarded, I could reduce the scope.

franziskuskiefer (approved):
Thanks

openmls/src/group/core_group/mod.rs
```diff
@@ -640,7 +640,7 @@ impl CoreGroup {
         if epoch < self.context().epoch() {
             self.message_secrets_store
                 .secrets_for_epoch_mut(epoch)
-                .ok_or_else(|| {
+                .ok_or({
```
franziskuskiefer, line 643: Why? `ok_or_else` is preferable for this use case.

openmls/src/group/core_group/test_past_secrets.rs
```diff
@@ -16,7 +16,7 @@ fn test_secret_tree_store(ciphersuite: &'static Ciphersuite, backend: &impl Open
 
     // Make sure we can access the message secrets we just stored
     assert!(message_secrets_store
-        .secrets_for_epoch(GroupEpoch(0))
+        .secrets_for_epoch_mut(GroupEpoch(0))
```
franziskuskiefer, line 19: Doesn't `secrets_for_epoch` work here? All these calls don't need mutability.

franziskuskiefer (changes requested):
Just two nits

openmls/src/group/core_group/mod.rs
raphaelrobert: That was clippy warning and I accepted the fix. I didn't check the previous PR for clippy warnings, but I guess it originated there.

openmls/src/group/core_group/test_past_secrets.rs
raphaelrobert: `secrets_for_epoch` was unused and I deleted it.

openmls/src/group/core_group/mod.rs
raphaelrobert: Here's the lint: https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_lazy_evaluations

franziskuskiefer (approved):

PR #695, opened 2022-01-05
Always have 3 digits in semantic validation steps for easy searching

sgmenda:
Also fix a typo where `ValSem108` was missing in docs.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -38,9 +38,7 @@ jobs:
           version: latest
       - name: Fuzz Welcome
         run: cargo fuzz run welcome_decode -- -runs=100000
-      - name: Fuzz MLSCiphertext
-        run: cargo fuzz run mls_ciphertext_decode -- -runs=100000
-      - name: Fuzz MLSPlaintext
-        run: cargo fuzz run mls_plaintext_decode -- -runs=100000
+      - name: Fuzz MLSMessageIn
+        run: cargo fuzz run mls_message_decode -- -runs=100000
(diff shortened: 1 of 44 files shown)


sgmenda:
I am not sure if this is the right approach, an alternative would be to always have 3 digits, like rename `ValSem1` to `ValSem001`.

raphaelrobert:
I'm not sure I understand how the backticks make search easier. Could you give a brief explanation? I know they change the formatting in the docs, which certainly makes it look nicer.

raphaelrobert:
> I am not sure if this is the right approach, an alternative would be to always have 3 digits, like rename `ValSem1` to `ValSem001`.

I went for the easiest possible scheme, but this can be changed. At some point @kkohbrok suggested to replace the labels with more expressive short strings. Either one is fine with me.

sgmenda:
Without the backticks, if you search for "ValSem1", you also get results for "ValSem101" for instance.

More expressive short strings also works. In the meantime perhaps it is easiest to just make everything 3 digits?

raphaelrobert:
> In the meantime perhaps it is easiest to just make everything 3 digits?

Agreed. Maybe even do 4 digits so we are sure we won't run out of number ranges. Backticks in documentation can stay of course.

sgmenda:
Since this is a temporary fix till we move over to human-readable short strings and we are currently only using a small portion of the 3-digit space, I think 3 digits might be enough for now.

raphaelrobert (approved):
Thanks, LGTM!

PR #697, opened 2022-01-11
Allow self-removal during external commit ("resync")

### LINKED ISSUE #682: add capability for removal of old self in group during external commit

kkohbrok:
This PR changes the external commit logic such that if a client does an external commit for a group and a client is present with the same `identity`, the client will automatically commit an inline remove proposal targeting its former self in the group.

Fixes #682.


openmls/src/group/core_group/apply_proposals.rs
```diff
@@ -57,6 +57,35 @@ impl CoreGroup {
...
+        // first proposal and ignore all others.
+        if let Some(queued_proposal) = proposal_queue
+            .filtered_by_type(ProposalType::ExternalInit)
+            .next()
+        {
+            let external_init_proposal = &queued_proposal
+                .proposal()
+                .as_external_init()
+                // We know the proposal type, so it's a library error if we're wrong.
+                .ok_or(CoreGroupError::LibraryError)?;
```
raphaelrobert, line 72: This could be replaced with an `if let` without an `else` clause to avoid the library error.

openmls/src/group/core_group/apply_proposals.rs
```diff
@@ -57,6 +57,35 @@ impl CoreGroup {
         let mut self_removed = false;
         let mut external_init_secret_option = None;
 
+        // Process external init proposals. We do this before the removes, so we
+        // know that removing "ourselves" (i.e. removing the group member in the
+        // same leaf as we are in) is valid in this case. We only care about the
+        // first proposal and ignore all others.
+        if let Some(queued_proposal) = proposal_queue
+            .filtered_by_type(ProposalType::ExternalInit)
```
raphaelrobert, line 65: Wouldn't this be a good place to plug in the validation check that there is at most one external init per commit?

openmls/src/group/core_group/create_commit.rs
```diff
@@ -38,9 +38,41 @@ impl CoreGroup {
...
+                    .find(|proposal| proposal.is_type(ProposalType::Remove));
+                // The external commit is processed like an add, so our new leaf
+                // index is the same as if we'd process an add after the remove
+                // proposal.
+                let leaf_index = if let Some(remove_proposal) = remove_proposal_option {
+                    let removed_index = remove_proposal
+                        .as_remove()
+                        .ok_or(CoreGroupError::LibraryError)?
+                        .removed();
+                    if removed_index < free_leaf_index {
```
raphaelrobert, line 66: I'm not sure I understand this logic... In the case of a "resync", are we moving ourselves in the tree to the left if possible?

openmls/src/group/core_group/create_commit.rs
```diff
@@ -60,20 +92,19 @@ impl CoreGroup {
...
         // Make a copy of the current tree to apply proposals safely
         let mut diff: TreeSyncDiff = self.treesync().empty_diff()?;
 
-        // If this is not an external commit we have to set our own leaf index
-        // and add our leaf. Also, we have to generate the
-        // [`KeyPackageBundlePayload`] slightly differently, because we can't
-        // just pull it from the tree if we're a `NewMember`.
-        let key_package_bundle_payload = self.prepare_kpb_payload(backend, &params, &mut diff)?;
+        // If this is not an external commit we have to set our own leaf index manually
+        if params.commit_type() == CommitType::External {
```
raphaelrobert, line 105: The comment contradicts the `if` statement here.

openmls/src/group/core_group/proposals.rs
```diff
@@ -150,7 +149,7 @@ impl ProposalQueue {
                 ProposalOrRef::Proposal(proposal) => {
                     // ValSem200
                     if let Proposal::Remove(ref remove_proposal) = proposal {
-                        if remove_proposal.removed() == sender.sender {
+                        if remove_proposal.removed() == sender.sender && sender.is_member() {
```
raphaelrobert, line 152: Could you add a comment here why the sender type has to be `member`?

openmls/src/group/core_group/test_external_init.rs
```diff
@@ -167,6 +169,107 @@ fn test_external_init(ciphersuite: &'static Ciphersuite, backend: &impl OpenMlsC
...
+            .expect("Error deserializing PGS");
+    let nodes_option = group_alice.treesync().export_nodes();
+
+    let proposal_store = ProposalStore::new();
+    let params = CreateCommitParams::builder()
+        .framing_parameters(framing_parameters)
+        .credential_bundle(&bob_credential_bundle)
+        .proposal_store(&proposal_store)
+        .build();
+    println!("\nBob joins externally.");
```
raphaelrobert, line 215: Do we need the print statements?

openmls/src/group/core_group/test_external_init.rs
```diff
@@ -167,6 +169,107 @@ fn test_external_init(ciphersuite: &'static Ciphersuite, backend: &impl OpenMlsC
...
+    let (mut new_group_bob, create_commit_result) = CoreGroup::join_by_external_commit(
+        backend,
+        params,
+        Some(&nodes_option),
```
raphaelrobert, line 219: Not part of this PR, but I'm wondering how you could join a group without providing any nodes.

openmls/src/group/core_group/test_external_init.rs
```diff
@@ -167,6 +169,107 @@ fn test_external_init(ciphersuite: &'static Ciphersuite, backend: &impl OpenMlsC
...
+                crate::messages::ProposalOrRef::Proposal(proposal) => {
+                    proposal.is_type(ProposalType::Remove)
+                }
+                _ => false,
+            }),
+        _ => panic!("Wrong content type."),
+    }
+    .is_some();
+    assert!(contains_remove);
+
```
raphaelrobert, line 240: Is there anything else that needs to checked here? E.g. Bob's index in the tree before and after.

raphaelrobert (changes requested):
Thanks for picking this up!

openmls/src/group/core_group/apply_proposals.rs
kkohbrok: Probably. I'm currently preparing a follow up that deals with validation (#680), but I haven't gone past signature verification yet. The validation flow of `CoreGroup` is not exactly straight-forward for external messages.

openmls/src/group/core_group/create_commit.rs
kkohbrok: We're not really moving ourselves. We're removing our old self and add ourselves into the group like a new member. They leaf index does not stay consistent.

openmls/src/group/core_group/proposals.rs
kkohbrok: Should that not go into the description of `ValSem200` instead?

openmls/src/group/core_group/test_external_init.rs
kkohbrok: You can't, same as with a `Welcome`, I think. This is only an `Option`, because the tree could be in a RatchetTreeExtension instead.

openmls/src/group/core_group/test_external_init.rs
kkohbrok: There is no real relationship between the indices before and after. The new index of Bob is the index a new joiner would get when they join the group after Bob was removed.

openmls/src/group/core_group/create_commit.rs
kkohbrok: Good point. The negation shouldn't be there.

kkohbrok (commented):
Thanks for the comments! The index pre-computation is the most annoying part of this PR, but I don't see a way around it without some major refactoring.

openmls/src/group/core_group/proposals.rs
raphaelrobert: My question was: why are you introducing this restriction?

raphaelrobert (commented):
Thanks for making the changes! Just one question left

openmls/src/group/core_group/proposals.rs
kkohbrok: Ah, the point is that as long as we're using leaf indices, the sender of the proposal might be the same as the committer. In case of the external commit, this is exactly what we want, so we have to allow it. This should go away with KeyPackage hashes as `sender` identifiers, though.

raphaelrobert (approved):
Thanks!

raphaelrobert (approved):

PR #698, opened 2022-01-11
Move book into repo & add user manual

raphaelrobert:
This PR moves the OpenMLS book into this repo so that it can use actual code samples from tests.
A first version of the user manual is also added and a few changes were made to the chapter about forward secrecy & validation.

Note that this PR does not yet deploy the book, this will require some cross-repo magic in a follow-up PR.


.github/workflows/gh-pages.yml
```diff
@@ -30,3 +30,16 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: openmls/target/doc
+  book:
```
franziskuskiefer, line 33: We could deploy it here as well (having a main version next to the main docs). Or do you want to do that in a follow up as well?

raphaelrobert: We could, but we'd lose the short URL from what I understand. My plan was to try building it here to make sure there are no errors and then trigger the deploy action on the `book` repo. That repo would then just contain a checkout/deploy action and no other content.
Btw. we could also do the same for the docs and have them deployed at openmls.tech/docs.

franziskuskiefer (approved):

PR #699, opened 2022-01-12
Remove all remaining unwraps

### LINKED ISSUE #428: [Bug] Potential panic when unwrapping in SignedStruct impl of KeyPackage

### LINKED ISSUE #438: Remove all unwraps

### LINKED ISSUE #591: Refactor ciphersuite to remove unwraps

raphaelrobert:
Fixes #428 
Fixes #438 
Fixes #591

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -6,10 +6,6 @@ on:
       - main
   workflow_dispatch:
 
-defaults:
-  run:
-    working-directory: openmls
-
 jobs:
   rustdoc:
     runs-on: ubuntu-latest
@@ -18,6 +14,7 @@ jobs:
       - uses: actions/checkout@master
       - name: Build docs
         run: |
+          cd openmls
           cargo doc --no-deps
           touch target/doc/.nojekyll
           cat > target/doc/index.html <<EOF
@@ -43,7 +40,7 @@ jobs:
           mdbook-version: 'latest'
       - name: Build mdBook
         run: |
-          cd openmls/book
+          cd book
           mdbook build
       - name: Deploy mdBook
         run: | 


openmls/src/ciphersuite/mod.rs
```diff
@@ -104,21 +104,26 @@ impl Ciphersuite {
             return Err(ConfigError::UnsupportedCiphersuite);
         }
 
-        Ok(Ciphersuite {
+        Ok(Self::new_from_supported(name))
+    }
+
+    /// This creates a Ciphersuite from one of the supported ciphersuite names.
+    /// This should only be used if it is clear the ciphersuite is used.
```
kkohbrok, line 111: I'm having trouble understanding the second line. Should it be "... the ciphersuite is supported."?

openmls/src/key_packages/mod.rs
```diff
@@ -108,31 +108,25 @@ impl KeyPackagePayload {
 pub struct KeyPackage {
     payload: KeyPackagePayload,
     signature: Signature,
-    encoded: Vec<u8>,
```
kkohbrok, line 111: @franziskuskiefer I think you introduced that at some point to reduce serialization operations. Can you remember how big the impact was?

Also: Should we maybe do this in a separate commit/PR?

openmls/src/tree/secret_tree.rs
```diff
@@ -168,21 +174,23 @@ impl SecretTree {
...
+            self.derive_down(ciphersuite, backend, n)?;
         }
         // Calculate node secret and initialize SenderRatchets
-        let node_secret = &self.nodes[index_in_tree.as_usize()]
-            .as_ref()
-            .unwrap()
-            .secret;
+        let node_secret = match &self.nodes[index_in_tree.as_usize()] {
+            Some(node) => &node.secret,
+            // We just derive all necessary nodes so this should not happen
```
kkohbrok, line 182:
```suggestion
            // We just derived all necessary nodes so this should not happen
```

kkohbrok (changes requested):
Thanks! Good to finally be unwrap-free 🙂. Not sure about removing the KeyPackage encoding optimization, though.

openmls/src/key_packages/mod.rs
raphaelrobert: I ran benchmarks on both versions and there is no measurable difference (i.e. the tests are actually a tad faster after this change).

openmls/src/ciphersuite/mod.rs
kkohbrok: I see. But shouldn't this throw an error then, if the ciphersuite is not supported? Which makes me wonder what currently happens if this is called with an unsupported ciphersuite.

Taking a closer look, the issue of ciphersuite support seems a bit inconsistent. There are two functions `supported_ciphersuites`, which accesses the `Config` and `supported_ciphersuite_names`, which accesses a constant that is independent of the config.

So either way, I think this function should at least get a comment saying what happens if the input ciphersuite is not supported. Regarding the general inconsistency of the config, I'll file a follow-up.

kkohbrok (changes requested):
Looks good to me with the exception of the `new_from_supported` issue.

raphaelrobert:
@kkohbrok Thanks for filing the issue! Any other changes you would like?

kkohbrok (commented):
I posted a comment under `new_from_supported`, but it probably got lost since the code is now outdated. I'll repost it here:

> I see. But shouldn't this throw an error then, if the ciphersuite is not supported? Which makes me wonder what currently happens if this is called with an unsupported ciphersuite.
> 
> Taking a closer look, the issue of ciphersuite support seems a bit inconsistent. There are two functions `supported_ciphersuites`, which accesses the `Config` and `supported_ciphersuite_names`, which accesses a constant that is independent of the config.
> 
> So either way, I think this function should at least get a comment saying what happens if the input ciphersuite is not supported. Regarding the general inconsistency of the config, I'll file a follow-up.

kkohbrok (approved):
Sounds good to me! Let's land it!

Issue #115, opened 2020-10-14
Sanitize input


raphaelrobert:
Right now the implementation doesn't have a lot of checks against malformed input values, especially when processing messages from other members.

franziskuskiefer:
Shall this become the issue for implementing everything in https://github.com/openmls/openmls/wiki/Message-validation ?

raphaelrobert:
Definitely part of it. I think the main issue is #133 though.

kkohbrok:
Syntactic validation is taken care of by the `tls_codec`.

Issue #78, opened 2020-10-08
Ensure upper bounds are checked


franziskuskiefer:
We currently don't check any upper bounds in struct vecs.

- [ ] write tests
- [ ] add checks in code

PR #703, opened 2022-01-13
Introduce thiserror

raphaelrobert:
This PR introduces the `thiserror` crate to replace our current error handling as discussed.

A few remarks:
 - I just picked one error to try it out
 - It looks like we can use like a drop-in replacement for the current errors, it plays nicely with what we have
 - Because of the above, we can incrementally change the existing errors while refactoring them, we don't have to do it all at once (there are 60+ errors to convert right now)
 - Backtraces are only supported with `nightly`, so that's not an option for now


kkohbrok (dismissed):
Seems nice. Looking forward to the debugging experience using this 😄

openmls/src/ciphersuite/signable.rs
```diff
@@ -54,12 +54,16 @@ pub trait Signable: Sized {
...
-    ) -> Result<Self::SignedOutput, CredentialError>
+    ) -> Result<Self::SignedOutput, CredentialBundleError>
     where
         Self::SignedOutput: SignedStruct<Self>,
     {
-        let payload = self.unsigned_payload()?;
-        let signature = credential_bundle.sign(backend, &payload)?;
+        let payload = self
+            .unsigned_payload()
+            .map_err(|_| CredentialBundleError::LibraryError)?;
```
kkohbrok, line 63: Do we want to keep the practice of annotating LibraryErrors with a reason on why they shouldn't occur in a particular situation? I think it would help code readability quite a bit.

openmls/src/ciphersuite/signable.rs
```diff
@@ -54,12 +54,16 @@ pub trait Signable: Sized {
...
+        let signature = credential_bundle
+            .sign(backend, &payload)
+            .map_err(|_| CredentialBundleError::SigningFailed)?;
```
kkohbrok, line 66: If I were a consumer getting this error, I think I'd be very confused. Any chance we can be more explicit here w.r.t. why signing failed? Or are there too many potential reasons to list?

kkohbrok (commented):
Thanks for putting this in motion! Just two questions.

openmls/src/ciphersuite/signable.rs
kkohbrok: I see. Then I would propose that until we have that error, we use code comments to annotate the error, so that we can easily populate the error string later.

In this particular case: What is hiding behind the "unlikely"? If there is a chance this errors out, it probably shouldn't be a library error.

raphaelrobert:
To make this more meaningful, I introduced the new `LibraryError` type that can be used anywhere in the code:

```rust
pub enum LibraryError {
    #[error(transparent)]
    TlsCodecError(#[from] TlsCodecError),
    #[error(transparent)]
    CryptoError(#[from] CryptoError),
    #[error("Custom library error: {0}")]
    Custom(&'static str),
}
```

That way we can simply propagate the typically meaningless codec/crypto errors, or add something custom. 
I had to make quite a few changes for this to work, but `CredentialError` should look clean now.

kkohbrok:
Just to make sure I understand what we're trying to do with `LibraryError` now. Before, `LibraryError` meant that there is a problem with our library, i.e. a bug that we need to fix. Now it's something like a global error that we can use in a `Result` wherever nominally the function should never return an error. Is that correct?

If so, I generally like that approach, because you can see from a function signature that the caller can expect that the function will not return an error (and thus either make the return type a `LibraryError` or use the `From<LibraryError>` of the relevant error enum. 

W.r.t. the codec error: Do I understand correctly, that the intent is to only use `LibraryError` for codec errors if we know a priori that it will not fail? Because in many cases it will depend on consumer input and we thus can't predict it.

raphaelrobert:
> Just to make sure I understand what we're trying to do with `LibraryError` now. Before, `LibraryError` meant that there is a problem with our library, i.e. a bug that we need to fix. Now it's something like a global error that we can use in a `Result` wherever nominally the function should never return an error. Is that correct?
> 
> If so, I generally like that approach, because you can see from a function signature that the caller can expect that the function will not return an error (and thus either make the return type a `LibraryError` or use the `From<LibraryError>` of the relevant error enum.
> 
> W.r.t. the codec error: Do I understand correctly, that the intent is to only use `LibraryError` for codec errors if we know a priori that it will not fail? Because in many cases it will depend on consumer input and we thus can't predict it.

Yes, your understanding is correct.
Regarding user-induced errors: This is a problem in general. So far we only returned a codec error (e.g. if a user uses more extensions than we can serialize). We should check for this separately, these changes don't address that problem.
That being said, not every codec error should become a `LibraryError`. In particular, deserialization should yield a different error still. Unfortunately, it's the same error type, otherwise it would be good enough if `LibraryError` onli covered serialization errors.

kkohbrok:
I see, so you propose that we add additional checks before the actual encoding is done to make sure that encoding can't fail. So I guess it's fine to have a codec error as a subtype of the LibraryError, as long as we use it only in places where we have a corresponding check. Regarding different types of Encoding error: One way to go about this would be to implement the `From` explicitly and return a different LibraryError depending on the the sub-type of codec error. If `thiserror` allows this, that is.

raphaelrobert:
> I see, so you propose that we add additional checks before the actual encoding is done to make sure that encoding can't fail. So I guess it's fine to have a codec error as a subtype of the LibraryError, as long as we use it only in places where we have a corresponding check. Regarding different types of Encoding error: One way to go about this would be to implement the `From` explicitly and return a different LibraryError depending on the the sub-type of codec error. If `thiserror` allows this, that is.

These additional checks would be part of #78 and #115. I wouldn't want to include this here to avoid scope creep of the PR.
Regarding your question: I don't think we can implement `From<>` for an `enum` variant. The TLS codec crate would have to be changed for that. However, I don't think this is strictly necessary right now (or ever).

openmls/src/error.rs
```diff
@@ -2,6 +2,82 @@
...
+
+impl Display for LibraryError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self.internal)
+    }
+}
+
+/// Internal enum to differentiate between the different types of library errors
+#[derive(Error, Debug, PartialEq, Clone)]
+enum InternalLibraryError {
```
kkohbrok, line 73: It's not obvious to me on a technical level, why we need to wrap this enum in a struct. I suspect that you explained it to me in the past, but in any case it might be worth adding a brief explanation here why this is necessary.

kkohbrok (approved):
Just one question/request, otherwise this looks good to me!

As a follow-up, it would be great if you could edit the Wiki with guidelines on how to apply this change to other places (i.e. the use of `thiserror` plus the use of `LibraryError` inasmuch as it is not explained through the docs of `LibraryError`).

openmls/src/error.rs
raphaelrobert: This was suggested by @franziskuskiefer because the consumers should not actually inspect it.

raphaelrobert:
> Just one question/request, otherwise this looks good to me!
> 
> As a follow-up, it would be great if you could edit the Wiki with guidelines on how to apply this change to other places (i.e. the use of `thiserror` plus the use of `LibraryError` inasmuch as it is not explained through the docs of `LibraryError`).

Yes, when I work on the public errors I'll also update the wiki accordingly.

franziskuskiefer:
After looking at this I still don't understand the purpose of this PR.
It introduces `thiserror` but doesn't really use it and, more importantly, doesn't tackle the issue of public facing errors. I have no objections to using `thiserror` but I don't see that this PR gets us any closer to defining errors for the public API.

Issue #536, opened 2021-11-11
[MLS Spec change] Constrain proposal in External Commit


raphaelrobert:
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/481

Issue #289, opened 2020-12-29
[MLS Spec change] External Commit 3


raphaelrobert:
**Description of the changes**
Sign the PublicGroupState

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/433

Issue #286, opened 2020-12-29
[MLS Spec change] External Commit 2


raphaelrobert:
**Description of the changes**
Specifies in more detail how external commits work.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/425

PR #706, opened 2022-01-14
Validate external commits

### LINKED ISSUE #556: Implement ExternalInit Proposal
https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#section-11.1.6

An ExternalInit proposal is used by new members that want to join a group by using an external commit. This proposal can only be used in that context.

```
struct {
  opaque kem_output<0..2^16-1>;
} ExternalInit;
```

A member of the group applies an ExternalInit message by initializing the next epoch using an init secret computed as described in Section 8.1. The `kem_output` field contains the required KEM output.

kkohbrok:
This PR introduces the following new Semantic Validation checks and corresponding unit tests on `MlsGroup` level:
* ValSem240: External Commit, inline Proposals: There MUST be at least one ExternalInit proposal.
* ValSem241: External Commit, inline Proposals: There MUST be at most one ExternalInit proposal.
* ValSem242: External Commit, inline Proposals: There MUST NOT be any Add proposals.
* ValSem243: External Commit, inline Proposals: There MUST NOT be any Update proposals.
* ValSem244: External Commit, inline Remove Proposal: The identity and the endpoint_id of the removed leaf are identical to the ones in the path KeyPackage.
* ValSem245: External Commit, referenced Proposals: There MUST NOT be any ExternalInit proposals.
* ValSem246: External Commit: MUST contain a path.
* ValSem247: External Commit: The signature of the MLSPlaintext MUST be verified with the credential of the KeyPackage in the included `path`.

The checks are due to the following requirements in the spec:
> External Commits work like regular Commits, with a few differences:
>
> The proposals included by value in an External Commit MUST meet the following conditions:
> * There MUST be a single ExternalInit proposal
> * There MUST NOT be any Add or Update proposals
> * If a Remove proposal is present, then the credential and endpoint_id of the removed leaf MUST be the same as the corresponding values in the Add KeyPackage.
>
> * The proposals included by reference in an External Commit MUST meet the following conditions:
>   * There MUST NOT be any ExternalInit proposals
> * External Commits MUST contain a path field (and is therefore a "full" Commit).
> * External Commits MUST be signed by the new member.
>
> * The signature on the enclosing MLSPlaintext MUST verify using the public key for the credential in the leaf_key_package of the path field.
> * When processing a Commit, both existing and new members MUST use the external init secret as described in Section 8.1.
> * The sender type for the MLSPlaintext encapsulating the External Commit MUST be new_member.

Fixes #556, #536, #289, #286, #192, #680 and #681.


openmls/src/framing/validation.rs
```diff
@@ -220,32 +256,25 @@ impl UnverifiedContextMessage {
...
-                    Err(ValidationError::LibraryError)
-                }
+            SenderType::Member | SenderType::NewMember => {
+                Ok(UnverifiedContextMessage::Group(UnverifiedGroupMessage {
+                    plaintext,
+                    credential,
+                }))
             }
-            // TODO #192: We don't support external senders yet
+            // TODO #151: We don't support preconfigured senders yet
```
raphaelrobert, line 265: I think the actual issue we want to tag here is #106. 151 is about ReInit specifically.

openmls/src/group/mls_group/mod.rs
```diff
@@ -239,6 +239,12 @@ impl MlsGroup {
         self.proposal_store.proposals()
     }
 
+    /// Inserts the given proposal into the proposal store.
+    #[cfg(test)]
+    pub fn add_pending_proposal(&mut self, queued_proposal: QueuedProposal) {
```
raphaelrobert, line 244: We already have the function `store_pending_proposal()` in `processing.rs` for that.

openmls/src/framing/validation.rs
```diff
@@ -145,15 +180,15 @@ impl DecryptedMessage {
...
+    credential: Credential,
     aad_option: Option<Vec<u8>>,
 }
 
 impl UnverifiedMessage {
     /// Construct an [UnverifiedMessage] from a [DecryptedMessage] and an optional [Credential].
     pub(crate) fn from_decrypted_message(
         decrypted_message: DecryptedMessage,
-        credential: Option<Credential>,
+        credential: Credential,
```
raphaelrobert, line 191: This was optional because external (preconfigured) messaged don't have a credential. This shouldn't change.

openmls/src/framing/validation.rs
```diff
@@ -178,23 +213,24 @@ impl UnverifiedMessage {
...
+/// Contains an [VerifiableMlsPlaintext] and a [Credential] if it is a message
+/// from a `Member` or a `NewMember`.  It sets the serialized group context and
+/// verifies the membership tag for member messages.  It can be converted to a
+/// verified message by verifying the signature, either with the credential or
+/// an external signature key.
 pub enum UnverifiedContextMessage {
-    Member(UnverifiedMemberMessage),
-    External(UnverifiedExternalMessage),
+    Group(UnverifiedGroupMessage),
+    Preconfigured(UnverifiedExternalMessage),
```
raphaelrobert, line 233: The renaming you propose is indeed more in line with the spec. But why not rename everything in that case? Now we have a weird mix.

openmls/src/group/core_group/validation.rs
```diff
@@ -264,4 +266,81 @@ impl CoreGroup {
...
+                return Err(ExternalCommitValidationError::ReferencedExternalInitProposal.into());
+            }
+        } else {
+            return Err(ExternalCommitValidationError::NoExternalInitProposals.into());
+        };
+
+        // ValSem241: External Commit, inline Proposals: There MUST be at most one ExternalInit proposal.
+        if let Some(additional_external_init_proposal) = external_init_proposals.next() {
+            // ValSem245: External Commit, referenced Proposals: There MUST NOT be any ExternalInit proposals.
+            if additional_external_init_proposal.proposal_or_ref_type()
```
raphaelrobert, line 298: Not sure we need this error, because we checked for that before. This check should just be about the fact that there is more than one ExternalInit.

openmls/src/group/core_group/validation.rs
```diff
@@ -264,4 +266,81 @@ impl CoreGroup {
...
+                return Err(ExternalCommitValidationError::ReferencedExternalInitProposal.into());
+            } else {
+                return Err(ExternalCommitValidationError::MultipleExternalInitProposals.into());
+            }
+        }
+
+        let add_proposals = proposal_queue.filtered_by_type(ProposalType::Add);
+        for proposal in add_proposals {
+            // ValSem242: External Commit, inline Proposals: There MUST NOT be any Add proposals.
+            if proposal.proposal_or_ref_type() == ProposalOrRefType::Proposal {
```
raphaelrobert, line 310: The spec isn't specific about inline proposals here. I guess it's more of a spec problem, but I'm not sure we should allow Add proposals by reference. The same is true for the Update and Remove proposals below.

raphaelrobert (changes requested):
Thanks for the PR! Nice to see the removal of the test-only flags in many places. It looks good in general, but a few things probably mandate some more discussion.

openmls/src/group/mls_group/mod.rs
kkohbrok: Good point, I totally missed that.

openmls/src/framing/validation.rs
kkohbrok: Good point. I'll reverse the change.

openmls/src/framing/validation.rs
kkohbrok: What do you mean by "everything"? The way I understand it is that there are places where we want to distinguish between `Member` and `NewMember`/`External` and other places (like this one) where we want to distinguish between `Group` and `Preconfigured`. Are there more occurrences of the latter that I should rename as well?

openmls/src/group/core_group/validation.rs
kkohbrok: Yeah, I thought it would be good to use the error to indicate what exactly went wrong. But I guess it doesn't really matter.

The check above only checks for the first ExternalInit proposal. This one checks for the second. But I agree, that we can simply return the second error here, i.e. `ExternalCommitValidationError::MultipleExternalInitProposals`.

openmls/src/group/core_group/validation.rs
kkohbrok: I'm not sure I understand. I think the spec is pretty specific. There is point point about proposals by value (i.e. inline), which is the following:
> * The proposals included by value in an External Commit MUST meet the following conditions:
>   * There MUST be a single ExternalInit proposal
>   * There MUST NOT be any Add or Update proposalsIf a Remove proposal is present, then the credential and endpoint_id of the removed leaf MUST be the same as the corresponding values in the Add KeyPackage.

Similarly, it's explicit about proposals by reference, where it states the following:
> * The proposals included by reference in an External Commit MUST meet the following conditions:
>   * There MUST NOT be any ExternalInit proposals

So I think the checks I'm proposing here line up exactly with what the spec says. Please correct me if I'm wrong. If we want to do more or less, I prefer we change the spec first and then act according to the spec changes.

kkohbrok (commented):
Thanks for the review! I think we should be able to get this in by the end of today.

openmls/src/framing/validation.rs
raphaelrobert: My point was that if you change the `enum` variant from `External` to `Preconfigured` you could consequently also rename `UnverifiedExternalMessage` to `UnverifiedPreconfiguredlMessage`.

openmls/src/group/core_group/validation.rs
raphaelrobert: Thanks for the explanation, I missed the last part. I guess it makes sense the way you implemented it.

openmls/src/group/tests/test_external_commit_validation.rs
```diff
@@ -0,0 +1,886 @@
...
+        mut alice_group,
+        bob_credential_bundle,
+        mut plaintext,
+        original_plaintext,
+    } = validation_test_setup(*PURE_PLAINTEXT_WIRE_FORMAT_POLICY, ciphersuite, backend);
+
+    let mut content = if let MlsPlaintextContentType::Commit(commit) = plaintext.content() {
+        commit.clone()
+    } else {
+        panic!("Unexpected content type.");
```
raphaelrobert, line 295: I'm still wondering how we can convince codecov that the `panic` is valid here and doesn't need additional testing. I don't have a suggestion though and this doesn't need to be changed.

raphaelrobert (commented):
Thanks for the changes! I added some clarifications to my comments.

I think this PR should also close the following issues: #556, #536, #289, #286, and #192.

openmls/src/framing/validation.rs
kkohbrok: Ah, sorry, I had totally missed that. Will change!

raphaelrobert (approved):
Thanks for the changes! I created #712 to finish the renaming you started so that we can get this in.

PR #709, opened 2022-01-14
Validate Local Commits and Add Basic Tests

sgmenda:
* Validate proposals within `CoreGroup::create_commit`
* Add basic tests for `ValSem100` and `ValSem103`

Ref #525


sgmenda:
I am not sure if we should be doing the checks at the `CoreGroup` level or the `MlsGroup` level. Currently, we do validation in `CoreGroup::stage_commit` so I continued with that tradition and put the new checks in `CoreGroup::create_commit`.

This PR breaks `group::tests::kat_messages::write_test_vectors_msg` and that is because the underlying `generate_test_vector` creates add proposals using [the same credential package](https://github.com/openmls/openmls/blob/35f74ac740ba53aa444e4982c33e7475ecab27c5/openmls/src/group/tests/kat_messages.rs#L106-L120). I wasn't sure if that was an indication that we should move these checks to `MlsGroup` or if I should just update the tests.

raphaelrobert:
Thanks for the PR! Implementing the checks at the `CoreGroup` level and testing them at the `MlsGroup` level is what we have been doing, so that's fine.

Re test vector: looks like the generation is wrong then and needs to be fixed.

sgmenda:
I updated the kat generation to work with the changes in this patch.

openmls/src/group/tests/kat_messages.rs
```diff
@@ -103,18 +103,36 @@ pub fn generate_test_vector(ciphersuite: &'static Ciphersuite) -> MessagesTestVe
...
-    let add_proposal = AddProposal {
-        key_package: key_package.clone(),
-    };
     let update_proposal = UpdateProposal {
         key_package: key_package.clone(),
     };
+
+    // Create proposal to add a user
+    let joiner_credential_bundle = CredentialBundle::new(
+        b"MLS rocks".to_vec(),
```
kkohbrok, line 117: 🤘🏻

openmls/src/group/tests/test_proposal_validation.rs
```diff
@@ -0,0 +1,190 @@
...
+    config::*,
+    credentials::*,
+    group::errors::*,
+    group::*,
+    key_packages::*,
+    prelude_test::{FramingParameters, MlsPlaintext, WireFormat},
+};
+
+/// Helper function to generate a CredentialBundle
+fn generate_credential_bundle(
```
kkohbrok, line 20: I used those functions as well for external commit validation and I think they're generally useful functions for testing. Not necessarily something we need to do in this PR, but maybe we should keep them in `utils.rs` or some other place, where they're more accessible?

openmls/src/group/tests/test_proposal_validation.rs
```diff
@@ -0,0 +1,190 @@
...
+        .create_commit(params, backend)
+        .expect_err("Created commit when the proposals have a duplicate identity!");
+    assert_eq!(
+        err,
+        CoreGroupError::ProposalValidationError(
+            ProposalValidationError::DuplicateIdentityAddProposal
+        )
+    );
+
+    // TODO #525: Also test that we are correctly validating incoming proposals.
```
kkohbrok, line 137: In the other validation tests, each test consists of two tests, a negative one (which is already here), where we check that the validation fails as specified and a positive one, where the validation goes through. It would be great, if you could add a simple positive case as well. Or is that why you flagged #525 here and mark the test as partly implemented? Same goes for the other test below.

openmls/src/group/core_group/create_commit.rs
```diff
@@ -107,6 +107,23 @@ impl CoreGroup {
...
+        // ValSem104
+        // ValSem105
+        // ValSem106
+        self.validate_add_proposals(&proposal_queue)?;
+        // ValSem107
+        // ValSem108
+        self.validate_remove_proposals(&proposal_queue)?;
+        // ValSem109
+        // ValSem110
+        self.validate_update_proposals(&proposal_queue, None)?;
```
kkohbrok, line 125: I'm wondering if we shouldn't do some of these validation checks before the client actually tries to commit them, e.g. when they are put into the proposal store rather than when the client tries to commit them. I think this makes it easier for a client to deal with invalid proposals by other members.

kkohbrok (changes requested):
Thanks for the PR! Looks great already. Just a few comments/questions.

openmls/src/group/tests/test_proposal_validation.rs
raphaelrobert: Agreed, we should do that in a follow-up. Do you want to file an issue?

openmls/src/group/core_group/create_commit.rs
raphaelrobert: That's a good question. I tend to agree with you that doing the checks earlier would make more sense.

openmls/src/group/tests/test_proposal_validation.rs
```diff
@@ -0,0 +1,190 @@
...
+use rstest::*;
+use rstest_reuse::{self, *};
+
+use crate::{
+    config::*,
+    credentials::*,
+    group::errors::*,
+    group::*,
+    key_packages::*,
+    prelude_test::{FramingParameters, MlsPlaintext, WireFormat},
```
raphaelrobert, line 16: You should use the `framing` module here. Prelude should never be used from the inside.

raphaelrobert (changes requested):
Thanks for the PR! Nice that you just picked a couple of tests for now, it makes the discussion easier, and adding the other tests in a follow-up PR should be fairly straightforward.
Just a few comments inline.

openmls/src/group/tests/test_proposal_validation.rs
kkohbrok: I don't think this needs an explicit issue. I'll do it in the external commit validation PR in a second, which I would expect will be merged before this one.

sgmenda: Sounds good, once that lands, I can merge and remove the helper functions here.

openmls/src/group/tests/test_proposal_validation.rs
sgmenda: Good catch, I forgot to add a positive test, I will add one.

openmls/src/group/core_group/create_commit.rs
sgmenda: @kkohbrok I agree, but just to make sure I understand your suggestion correctly:

1. we currently do checks in `CoreGroup::create_commit` (to prevent creating invalid proposals) and `CoreGroup::stage_commit` (to prevent accepting invalid proposals from other group members.)
2. you suggestion is to move the checks to `ProposalStore::add` so it works in both cases (we put our proposals and incoming proposals into the ProposalStore before committing them.)

openmls/src/group/tests/test_proposal_validation.rs
sgmenda: The TODO is for alice receiving a malformed proposal message from another group member bob. I left that as a TODO for now because to generate the malformed proposal, we need a test version of `create_commit` that does not run the validation checks. Alternatively, if we move the validation checks to `ProposalStore` as you suggested, we wouldn't have to do that.

openmls/src/group/core_group/create_commit.rs
kkohbrok: Sorry, I didn't make a clear suggestion of where exactly to put the check. Just to make sure we're on the same page, my understanding of our current way of doing validation is as follow:

1. It should be impossible for the consumer to not do the checks.
2. The checks should be at the level of the `CoreGroup` iff they are mandated by the spec and the `MlsGroup` otherwise.
3. `MlsGroup` should do as much of the work as possible for the user, where in the case of `CoreGroup`, we can expect the user to jump through a hoop or two.

We could do the check in `ProposalStore::add` as you suggest, but that would require passing all the information required for validation into the function. That would work, but it would bloat the otherwise relatively simple `ProposalStore` quite a bit. Another way would be to go the same route as for other `MlsMessages`, which would mean we create an intermediate type `UnverifiedProposal` (or `UnvalidatedProposal`), which allows a transformation to `QueuedProposal` that requires extra group-specific information and that does all the validation steps that we want. This would allow us to encapsulate the validation complexity somewhat and keep the `QueuedProposal`/`ProposalStore` types relatively simple. 

To force the user to perform this step, it would have to be impossible to circumvent the step from `MlsPlaintext` to `UnverifiedProposal`, which might pose problems in other places where we handle proposals.

sgmenda: > Sorry, I didn't make a clear suggestion of where exactly to put the check.

Oops, I am sorry, I misread your earlier message. 😬

Re: argument against putting it in `ProposalStore::add`: that is a good point. Then it might be better to put the checks in [`StagedCommit::new`](https://github.com/openmls/openmls/blob/bd0c52f064dc8edc5063cbeeca224a592ac1d4fa/openmls/src/group/core_group/staged_commit.rs#L342-L352) since it would not overcomplicate the `ProposalStore` nor mess with existing flows.

kkohbrok: I talked to Raphael about where to put these checks and it turns out what I propose doesn't actually line up with what the spec requires 😅

Long story short, the checks are exactly right in the place that they are right now. Checking earlier is possible only in  a few cases and we might do so on the `MlsGroup` level for consumer convenience, but that's nothing we have to worry about in this PR.

openmls/src/group/tests/test_proposal_validation.rs
kkohbrok: The PR has landed. Once you merge `main` into this branch, you should be able to find the `generate...` functions in the `utils.rs` submodule next to this one.

kkohbrok (changes requested):
This PR looks good to once you use the `generate...` functions mentioned in one of the comments. Please use GH's re-request review functionality to let us know if you want us to take a final look and r+ 🙂.

sgmenda:
I updated the validation checks to use the shared utils. (Sorry for the delay, I got caught up in flurry of stuff (start of term, etc.))

kkohbrok (approved):
Thanks for making the changes! Looks good to me now.

Don't worry about any delays. We're really just happy you're helping out!

raphaelrobert (approved):

Issue #701, opened 2022-01-13
"Supported ciphersuite" is unclear


kkohbrok:
This came up in the course of #699.

There are two functions `supported_ciphersuites`, which accesses the `Config` and `supported_ciphersuite_names`, which accesses a constant that is independent of the config. `supported_ciphersuite_names` should probably refer to `supported_ciphersuites` for its output.

franziskuskiefer:
The crypto provider trait should get functions for checking algorithm support. This should then be piped through unless the config explicitly disables cipher suites.

PR #711, opened 2022-01-17
Required capabilities validation

kkohbrok:


openmls/src/group/core_group/test_proposals.rs
```diff
@@ -447,7 +447,7 @@ fn test_group_context_extensions(
     .expect("Error joining group.");
 }
 
-#[apply(ciphersuites_and_backends)]
+//#[apply(ciphersuites_and_backends)]
```
sgmenda, line 450: This seems like a typo?

kkohbrok: Good catch! It kind of is a typo. I regularly comment out these macro invocations so `rust-analyzer` is a little more helpful when working on tests. I usually try not to commit these, but it seems like this one slipped through. Then again, this is not meant to be merged just yet.

sgmenda: Oops, I didn't notice the `draft` status. 😬 Yup, ++rust-analyzer.

kkohbrok:
I think it makes sense to wait with this until we deal with #701.

PR #714, opened 2022-01-18
Builds the documentation in one go

raphaelrobert:
We can do the CSS styling later


franziskuskiefer (approved):
We should probably put a script in there for the `run` and a nice html instead of the `cat` at some point.

Issue #705, opened 2022-01-13
Decrypting past epochs with key package references [#717]


franziskuskiefer:
The mechanism introduced in #563 to decrypt past epochs doesn't work with the changes from #541.
When #541 is fixed the sender of a message is identified by their key package reference. But this reference can change (e.g. through an update). Application messages from a sender that changed their key package in the meantime can not be decrypted because the leaf can not be identified.

In order to restore the functionality introduced in #563 not only message secrets from past epochs must be kept but the key package references from all members in order to identify the sender of application messages from past epochs.

Issue #652, opened 2021-12-21
Optimize and handling of `Node` vectors/slices and make it consistent


kkohbrok:
When joining a group, new members get the tree either via a RatchetTreeExtension of via an out-of-band channel as a vector of nodes. Functions currently take slices as input, but ultimately have to clone the nodes to instantiate the TreeSync instance. We might as well take the vector as input for the functions, at least in the cases, where we know that we end up cloning all the nodes anyway.

kkohbrok:
This was fixed by #1270.

PR #717, opened 2022-01-20
Use KeyPackageRef

### LINKED ISSUE #537: [MLS Spec change] Remove the notion of a 'leaf index'
https://github.com/mlswg/mls-protocol/pull/507 might re-introduce leaf indices.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/487

### LINKED ISSUE #541: [MLS Spec change] Use smaller, fixed-size hash-based identifiers

franziskuskiefer:
Apologies for this humongous PR but I'm afraid there's no other way.
This PR uses the `KeyPackageRef` and `ProposalRef` values instead of leaf indices
and the proposal hash.

Instead of leaf indices the `KeyPackageRef` value (16 bytes truncated hash of
the key package) is used to identify leaves now.
There should be no functional changes in this PR.

I filed a number of follow ups that are necessary (labeled with KeyPackageRef).
It's mostly tests that are either missing or don't work anymore the way they used
to.

Note that the secret tree still uses leaf indices.

Fixes #541
Fixes #537

follow ups
- #730 
- #727 
- #705
- #732 
- #731


openmls/src/treesync/mod.rs
```diff
@@ -87,7 +89,16 @@ impl TreeSync {
         let key_package = key_package_bundle.key_package();
         // We generate our own leaf without a private key for now. The private
         // key is set in the `from_nodes` constructor below.
-        let node: Node = Node::LeafNode(key_package.clone().into());
+        // let node: Node = Node::LeafNode(key_package.clone().into());
```
kkohbrok, line 92: Leftover code?

openmls/src/treesync/mod.rs
```diff
@@ -234,12 +249,9 @@ impl TreeSync {
     pub(crate) fn from_nodes_without_leaf(
         backend: &impl OpenMlsCryptoProvider,
         ciphersuite: &Ciphersuite,
-        node_options: &[Option<Node>],
+        mut nodes: Vec<Option<Node>>,
```
kkohbrok, line 252: This PR might fix [#652](https://github.com/openmls/openmls/issues/652) as well.

openmls/src/treesync/mod.rs
```diff
@@ -372,6 +402,51 @@ impl TreeSync {
...
+        let tsn = self.tree.leaf(self.leaf_index(key_package_ref)?)?;
+        Ok(match tsn.node() {
+            Some(node) => Some(node.as_leaf_node()?),
+            None => None,
+        })
+    }
+
+    /// Get the [`LeafIndex`] for a given [`KeyPackageRef`].
+    ///
+    /// This should go away and this tree should handle [`KeyPackageRef`] instead.
```
kkohbrok, line 423: If that is the case, can you file an issue, so we can keep track of it?

openmls/src/treesync/mod.rs
```diff
@@ -336,6 +348,24 @@ impl TreeSync {
         Ok(leaves)
     }
 
+    /// Returns a [`BTreeMap`] mapping [`KeyPackageRef`]s to the corresponding
+    /// [`KeyPackage`] instances in the leaves. The map only contains full
+    /// nodes.
+    pub(crate) fn leaves(
```
kkohbrok, line 354: There's now a `leaves` and a `full_leaves` function, where the difference is that `full_leaves` returns a map with `LeafIndex`. This can probably be cleaned up a bit.

openmls/src/treesync/treesync_node.rs
```diff
@@ -59,7 +62,10 @@ impl TreeSyncNode {
     pub(in crate::treesync) fn node_without_private_key(&self) -> Option<Node> {
         if let Some(node) = self.node() {
             match node {
-                Node::LeafNode(leaf_node) => Node::LeafNode(leaf_node.key_package().clone().into()),
+                Node::LeafNode(leaf_node) => Node::LeafNode(LeafNode::new_with_ref(
+                    leaf_node.key_package().clone(),
+                    leaf_node.key_package_ref().cloned(),
```
kkohbrok, line 67: Would it make sense to just implement `Copy` for `HashRef` and leave optimization to the compiler?

openmls/src/framing/sender.rs
```diff
@@ -61,20 +66,71 @@ impl TryFrom<u8> for SenderType {
...
 pub struct Sender {
     pub(crate) sender_type: SenderType,
-    // TODO: #541 replace sender with [`KeyPackageRef`] (and preconfigured/new)
-    pub(crate) sender: LeafIndex,
+    pub(crate) sender: SenderValue,
 }
 // Public functions
 impl Sender {
+    /// Build a [`Sender`] from [`MlsSenderData`].
+    pub(crate) fn from_sender_data(sender_data: MlsSenderData) -> Self {
```
kkohbrok, line 84: Would it make sense to implement `From` for `MlsSenderData` and `&KeyPackageRef` instead of these explicit functions?

kkohbrok (changes requested):
Thanks for fighting your way through the whole code base to push this through!

For now, I've only made an initial pass looking at everything except any tests, `core_group` and `mls_group`. I hope this helps already and I will continue later today with the `core_group`, taking it a bit at a time.

The way I understand it, the fact that this PR had to be this large is for a few reasons:
* Lots of bits of the code/protocol have to refer to leaves one way or another. I don't think this can be helped.
* `KeyPackageRef`s are functionally quite different from leaf indices, which leads to changes in logic in many places. I believe this also can't be helped.
* The way our code is modularized/structured is not great in many places. As discussed OOB, it would probably be a good idea to create a meta issue to track structural issues that were unearthed through this.

openmls/src/framing/sender.rs
franziskuskiefer: We could. I made them explicit because there has been confusion around conversation a lot recently. Having to type it makes it clearer what happens. I don't have strong feelins on this but wanted to be a little more explicit to see if it helps.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -1,20 +1,69 @@
...
+    ciphersuite::{hash_ref::KeyPackageRef, HpkePrivateKey, HpkePublicKey},
+    key_packages::{KeyPackage, KeyPackageBundle, KeyPackageError},
 };
 
 /// This struct implements the MLS leaf node and contains a [`KeyPackage`] and
 /// potentially a corresponding [`HpkePrivateKey`].
-#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct LeafNode {
+    key_package_ref: Option<KeyPackageRef>,
```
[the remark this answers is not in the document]
franziskuskiefer: We don't always need the key package reference and can easily compute it from the `KeyPackage`. So I don't think it makes sense to have it in there. Here it is an identifier for the leaf such that it would be annoying having to recompute it every time.
Unless you have strong feelings about it I'd keep it here.

openmls/src/treesync/mod.rs
franziskuskiefer: It's a step in towards fixing it. But `from_nodes` above is still using a slice.

openmls/src/treesync/mod.rs
franziskuskiefer: Yes, `full_leaves` should probably go away.
But I didn't want to change even more code. This PR really only touches what's necessary to make it work 😬. I filed #732 for this.

openmls/src/treesync/mod.rs
franziskuskiefer: I added a reference to #732

openmls/src/treesync/treesync_node.rs
franziskuskiefer: It implements `Copy`. But this is an `Option<&_>` that we have to convert, hence the `cloned`.

kkohbrok: Ah, I hadn't looked close enough. Fair point.

kkohbrok (commented):
`core_group` and `managed_group` look good! I did notice a few minor things, but I think those will be fixed in the process of making clippy a little happier.

I'll take a bit of a break and then I'll look at the tests next.

openmls/src/prelude_test.rs
```diff
@@ -14,4 +14,4 @@ pub use crate::group::tests::{
 };
 pub use crate::schedule::kat_key_schedule::{self, KeyScheduleTestVector};
 pub use crate::tree::tests_and_kats::kats::kat_encryption::{self, EncryptionTestVector};
-pub use crate::treesync::tests_and_kats::kats::kat_tree_kem::{self, TreeKemTestVector};
+// pub use crate::treesync::tests_and_kats::kats::kat_tree_kem::{self, TreeKemTestVector};
```
raphaelrobert, line 17: If we don't need it, we don't need it.

openmls/src/treesync/mod.rs
```diff
@@ -381,13 +456,16 @@ implement_error! {
             MissingKeyPackage = "Couldn't find our own key package in this tree.",
             DuplicateKeyPackage = "Found two KeyPackages with the same public key.",
             OutOfBounds = "The given `LeafIndex` is outside of the tree.",
+            LeafNotInTree = "The given `KeyPackageRef` is not in the tree.",
```
raphaelrobert, line 459: `LeafNotInTree` -> `KeyPackageRefNotInTree`

raphaelrobert (changes requested):
Last big milestone before v0.4, thanks for the PR!
I did a fist pass and left some comments.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -36,22 +89,36 @@ impl LeafNode {
     pub(crate) fn key_package(&self) -> &KeyPackage {
         &self.key_package
     }
+
+    /// Return a [`KeyPackageRef`] of this node('s key package).
+    /// Note that this returns `None` if the key package reference has not been
+    /// set explicitly with [`Self::set_key_package_ref()`].
+    pub(crate) fn key_package_ref(&self) -> Option<&KeyPackageRef> {
```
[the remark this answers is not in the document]
franziskuskiefer: This would require to always have a crypto backend available when calling this function, which is not the case.

openmls/src/group/core_group/mod.rs
```diff
@@ -608,8 +612,25 @@ impl CoreGroup {
...
+        Ok(self.treesync().leaf_index(key_package_ref)?)
+    }
+
+    /// Get the leaf index of this client.
+    pub(crate) fn own_leaf_index(&self) -> u32 {
+        self.treesync().own_leaf_index()
+    }
+
+    /// Get the [`KeyPackageRef`] of the client owning this group.
+    pub(crate) fn key_package_ref(&self) -> Option<&KeyPackageRef> {
```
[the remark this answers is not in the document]
franziskuskiefer: An error would indicate that something is wrong. But the group might not have computed the reference yet. So it's not really an error. It's a general issue with the way the code is structured. But I don't really mind. If you feel strongly about it, I can change it to an error.

openmls/src/group/mls_group/membership.rs
```diff
@@ -261,7 +265,17 @@ impl MlsGroup {
 
     /// Gets the current list of members
     #[cfg(any(feature = "test-utils", test))]
-    pub fn indexed_members(&self) -> Result<BTreeMap<LeafIndex, &KeyPackage>, MlsGroupError> {
+    pub fn leaves(
```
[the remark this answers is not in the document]
franziskuskiefer: It's a test function so I don't care.

openmls/src/group/mls_group/test_mls_group.rs
```diff
@@ -221,8 +226,10 @@ fn remover(ciphersuite: &'static Ciphersuite, backend: &impl OpenMlsCryptoProvid
 
     // === Alice removes Bob & Charlie commits ===
 
+    let bob_kpr = bob_group.key_package_ref().unwrap();
```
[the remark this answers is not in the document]
franziskuskiefer: Sure, makes no difference though

openmls/src/group/tests/test_external_commit_validation.rs
```diff
@@ -407,7 +408,7 @@ fn test_valsem243(ciphersuite: &'static Ciphersuite, backend: &impl OpenMlsCrypt
     let proposal_store = ProposalStore::new();
     let (_bob_group, message) = MlsGroup::join_by_external_commit(
         backend,
-        Some(&tree_option),
+        Some(&tree_option), // Note that this isn't actually used.
```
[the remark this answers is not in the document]
franziskuskiefer: Nope, it still uses a code path that ensures that the extension is used. It's not really testing it but since there's no test excercising this afaik I keep this.

openmls/src/group/tests/test_framing_validation.rs
```diff
@@ -325,22 +326,20 @@ fn test_valsem004(ciphersuite: &'static Ciphersuite, backend: &impl OpenMlsCrypt
...
         .parse_message(message_in, backend)
-        .expect_err("Could parse message despite wrong sender index.");
+        .expect_err("Could parse message despite wrong sender.");
 
     assert_eq!(
         err,
-        MlsGroupError::Group(CoreGroupError::FramingValidationError(
-            FramingValidationError::UnknownMember
-        ))
+        MlsGroupError::Group(CoreGroupError::TreeSyncError(TreeSyncError::LeafNotInTree))
```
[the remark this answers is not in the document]
franziskuskiefer: Yep, because we have to convert the key package reference into an index first we run into the issue that the leaf doesn't exist a lot before we get to some other check with a more useful message.

openmls/src/group/tests/test_past_secrets.rs
```diff
@@ -177,28 +179,30 @@ fn test_past_secrets_in_group(
...
                 err,
-                MlsGroupError::Group(CoreGroupError::MlsCiphertextError(
-                    MlsCiphertextError::SecretTreeError(SecretTreeError::TooDistantInThePast,),
+                MlsGroupError::Group(CoreGroupError::ValidationError(
+                    ValidationError::MlsCiphertextError(MlsCiphertextError::DecryptionError),
                 ))
             );
         }
 
+        // FIXME: It is impossible to decrypt this message now because the sender
```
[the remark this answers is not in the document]
franziskuskiefer: Yes. The issue didn't exist when I wrote this. I added the issue here.

openmls/src/test_utils/test_framework/mod.rs
```diff
@@ -21,6 +21,7 @@
 //! can be manipulated manually via the `Client` struct, which contains their
 //! group states.
 
+use crate::prelude_test::hash_ref::KeyPackageRef;
```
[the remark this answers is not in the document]
franziskuskiefer: yeah... rust analyzer....

openmls/src/prelude_test.rs
franziskuskiefer: We need it. But it doesn't work. I added the issue to make this work again.

openmls/src/test_utils/test_framework/mod.rs
```diff
@@ -504,33 +548,28 @@ impl MlsGroupTestSetup {
         Ok(())
     }
 
+    // /// Has the `re-
```
kkohbrok, line 551: This can probably go.

openmls/src/test_utils/test_framework/mod.rs
```diff
@@ -504,33 +548,28 @@ impl MlsGroupTestSetup {
...
+    /// error.
+    pub fn remove_clients(
         &self,
         action_type: ActionType,
         group: &mut Group,
         remover_id: &[u8],
-        target_indices: &[usize],
+        target_members: &[hash_ref::KeyPackageRef],
     ) -> Result<(), SetupError> {
+        // self.remove_clients_by_index(action_type, group, remover_id, &target_indices)?;
```
kkohbrok, line 564: ... and this as well.

kkohbrok (approved):
I think we can land this now. There's still a lot to clean up structurally, but I believe we have follow up issues to keep track of that. Now I'm really itching to clean up the TreeSync API so we don't have to juggle indices any more (or at least to a lesser extent).

kkohbrok (approved):
:shipit:

raphaelrobert (changes requested):
It looks like there is quite some refactoring left to do on treesync that leads to unnecessary code right now, but I guess we'll do that in #732. Let's try and get this in with the existing code structure, there's just some a bit of cleanup left to do:
 - We said that after #592 we don't want to use `unwrap()` anymore in the library and all tests should use `expect()` instead (seemed to lead to better code coverage and makes it easier to scan for undesired `unwrap()`s)
  - There are some clippy warnings left in the tests

franziskuskiefer:
> It looks like there is quite some refactoring left to do on treesync that leads to unnecessary code right now, but I guess we'll do that in #732. Let's try and get this in with the existing code structure, there's just some a bit of cleanup left to do:

Yes, not only treesync needs to be refactored in order to make this work nicely. But that's been known for a while.

> * We said that after #592 we don't want to use `unwrap()` anymore in the library and all tests should use `expect()` instead (seemed to lead to better code coverage and makes it easier to scan for undesired `unwrap()`s)

As you noted yourself, there's no difference in code coverage. But I made them `expects`, without a useful message though.

> * There are some clippy warnings left in the tests

Sure, I fixed them. I don't agree with them though.

raphaelrobert (approved):

raphaelrobert:
@franziskuskiefer you force-pushed, so you have to squash

Issue #723, opened 2022-01-21
[Bug] Unable to get Group epoch 


pulkitkothari:
**Describe the bug**
The group has group context which is not public, so application code is unable to get epoch from a MlsGroup.

**To Reproduce**
No method get epoch for a MlsGroup

**Expected behavior**
MlsGroup should have a method to get epoch

**Actual behavior**
No method

**Proposed fix:**
https://github.com/openmls/openmls/blob/main/openmls/src/group/mls_group/mod.rs#L180
```
MlsGroup {
...
    /// Return the epoch
    pub fn epoch(&self) -> GroupEpoch {
        return self.group.context().epoch();
    }

...
}
```

raphaelrobert:
This looks like a reasonable thing to expose indeed.

PR #724, opened 2022-01-22
#723: [Bug] Unable to get Group epoch

pulkitkothari:
#723: [Bug] Unable to get Group epoch

### DIFF
--- a/openmls/src/group/mls_group/mod.rs
+++ b/openmls/src/group/mls_group/mod.rs
@@ -234,6 +234,11 @@ impl MlsGroup {
         self.group.group_id()
     }
 
+    /// Return the epoch
+    pub fn epoch(&self) -> GroupEpoch {
+        self.group.context().epoch()
+    }
+
     /// Returns an `Iterator` over staged proposals.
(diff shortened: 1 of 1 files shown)


raphaelrobert (changes requested):
Thanks for the PR! Could make two separate PRs, one for each issue? That way we can discuss this separately.

pulkitkothari:
@raphaelrobert  Done this diff just have one bug fix.

openmls/src/group/mls_group/mod.rs
```diff
@@ -234,6 +234,11 @@ impl MlsGroup {
         self.group.group_id()
     }
 
+    /// Return the epoch
+    pub fn epoch(&self) -> GroupEpoch {
+        return self.group.context().epoch();
```
raphaelrobert, line 239:
```suggestion
        self.group.context().epoch()
```

raphaelrobert (changes requested):
Thanks, just one nit to make it more rusty.

pulkitkothari:
Still getting used to rust ;p thanks for the review and prompt replies.

raphaelrobert (approved):
Thanks!

Issue #722, opened 2022-01-21
[Bug] treesync::node::Node is private but public API needs it


pulkitkothari:
**Describe the bug**

OpenMls allows to export ratchet tree but the Node class is private

https://github.com/openmls/openmls/blob/main/openmls/src/group/mls_group/mod.rs#L305-L308
```
    /// Export the Ratchet Tree
    pub fn export_ratchet_tree(&self) -> Vec<Option<Node>> {
        self.group.treesync().export_nodes()
    }
```

**To Reproduce**
Call export_ratchet_tree on MLS Group and capture in variable
let ratchet_tree: Vec<Option<Node>> = group.export_ratchet_tree();

**Expected behavior**
Node class not found.

**Actual behavior**
Node class should be public and accessible outside crate

**Proposed fix:**
https://github.com/openmls/openmls/blob/main/openmls/src/lib.rs#L60
Update mod treesync; -> pub mod treesync;

https://github.com/openmls/openmls/blob/main/openmls/src/treesync/mod.rs#L49
Update pub(crate) mod node; -> pub mod node;

raphaelrobert:
That's correct and we should probably expose `Node`. Note however that `Node` is just an `enum` containing `LeafNode` and `ParentNode` which are also not exposed.
Can you tell a bit more about your use case, so that we can see what exactly needs to be exposed?

pulkitkothari:
I wanted to use 'export_ratchet_tree(&self)' and share the value with other user, so that other user can use the returned value  while joining the group from welcome message.

raphaelrobert:
As a workaround, you can probably just serialize the whole vector for now and pass the bytes to another user. I need to check a bit more about what exactly we want to expose here.

raphaelrobert:
Closed via #725.

pulkitkothari:
https://github.com/openmls/openmls/pull/783 PR has reopened this issue.

raphaelrobert:
Arf, sorry for that. I guess we should have an integration test for that. Would you like to provide a minimal example perhaps?

pulkitkothari:
Sure I can add one. The minimal example is using export_ratchet_tree method.

kkohbrok:
It looks to me like the `Node` struct is public and explicitly exported via `prelude.rs`. The `export_ratchet_tree()` function is used, for example, in the `mls_group_operations` test that lives in the `test/test_mls_group.rs` file as part of the integration tests.

[...]

PR #725, opened 2022-01-23
[Bug] treesync::node::Node is private but public API needs it

pulkitkothari:
#722 [Bug] treesync::node::Node is private but public API needs it
For now exposing treesync and node so that I can get unblocked.


raphaelrobert (approved):
Thanks, I guess we can merge it like this for now.

PR #726, opened 2022-01-23
Expose StagedCommit again

raphaelrobert:
This PR exposes `StagedCommit` and the necessary queued proposals. It also cleans up `prelude.rs` a bit.

### DIFF
--- a/openmls/src/prelude.rs
+++ b/openmls/src/prelude.rs
@@ -5,9 +5,7 @@
 pub use crate::group::*;
 
 // Group errors
-pub use crate::group::errors::{
-    CoreGroupError, CreateCommitError, FramingValidationError, StageCommitError, WelcomeError,
-};
+pub use crate::group::errors::*;
 
 // Indexes
 pub use crate::binary_tree::LeafIndex;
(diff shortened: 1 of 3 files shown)


openmls/src/prelude.rs
```diff
@@ -2,17 +2,10 @@
 //! Include this to get access to all the public functions of OpenMLS.
 
 // MlsGroup
-pub use crate::group::{
-    proposals::{ProposalStore, QueuedProposal},
-    EmptyInputError, GroupEpoch, GroupId, IncomingWireFormatPolicy, InnerState,
-    InvalidMessageError, MlsGroup, MlsGroupConfig, MlsGroupError, OutgoingWireFormatPolicy,
-    WireFormatPolicy, ALL_VALID_WIRE_FORMAT_POLICIES,
-};
+pub use crate::group::*;
```
kkohbrok, line 5: Shouldn't we export these things explicitly rather than via wildcard?

kkohbrok (commented):
Thanks! Looks good to me with one question.

openmls/src/prelude.rs
raphaelrobert: We export the right ones in the module now since this has been cleaned up. There is no need to be more fine-grained here anymore.

kkohbrok (approved):

PR #743, opened 2022-01-29
Add missing doc to public API

raphaelrobert:
This PR adds the missing documentation for the public API. It also does some minor renaming (`KeyID` has been renamed to `ExternalKeyID` in the spec at some point and we missed it) and restricts the visibility here and there. There are no other code changes.


openmls/src/key_packages/mod.rs
```diff
@@ -1,3 +1,24 @@
+//! Key Packages.
+//!
+//! In order to facilitate asynchronous addition of clients to a group,
+//! it is possible to pre-publish key packages that provide some public
```
kkohbrok, line 4:
```suggestion
//! it is necessary to pre-publish key packages that provide some public
```

openmls/src/key_packages/mod.rs
```diff
@@ -1,3 +1,24 @@
...
+//! information about a user. KeyPackage structures provide information
+//! about a client that any existing member can use to add this client
+//! to the group asynchronously.A KeyPackage object specifies a ciphersuite
```
kkohbrok, line 7:
```suggestion
//! to the group asynchronously. A KeyPackage object specifies a ciphersuite
```

openmls/src/key_packages/mod.rs
```diff
@@ -1,3 +1,24 @@
...
+//! that the client supports, as well as providing a public key that others
+//! can use for key agreement.The identity arising from the credential,
+//! together with the endpoint_id in the KeyPackage serve to uniquely identify
+//! a client in a group.When used as InitKeys, KeyPackages are intended to be
+//! used only once and SHOULD NOT be reused except in case of last
+//! resort. (See Section 15.4). Clients MAY generate and publish multiple InitKeys
+//!  to support multiple ciphersuites.KeyPackages contain a public key chosen by the
```
kkohbrok, line 14:
```suggestion
//!  to support multiple ciphersuites. KeyPackages contain a public key chosen by the
```

openmls/src/key_packages/mod.rs
```diff
@@ -1,3 +1,24 @@
...
+//! client, which the client MUST ensure uniquely identifies a given KeyPackage
+//! object among the set of KeyPackages created by this client.The value for
```
kkohbrok, line 16:
```suggestion
//! object among the set of KeyPackages created by this client. The value for
```

openmls/src/key_packages/mod.rs
```diff
@@ -1,3 +1,24 @@
...
+//! hpke_init_key MUST be a public key for the asymmetric encryption scheme
+//! defined by cipher_suite. The whole structure is signed using the client's
```
kkohbrok, line 18: I don't think we offer a possibility to insert self-generated KeyPackages, so OpenMLS will ensure that most of these requirements are met. Do we still need to list them here?

kkohbrok (changes requested):
Thanks for tackling this!

As noted in the comments, I'm not sure that the text from the spec makes good documentation for our code. It helps to provide a link, but a lot of the text is not really relevant for the consumer. For example in places where our code makes sure certain properties hold that we don't even expose to the consumer.

Also, while I understand that some of the enums are somewhat self-explanatory, I still think that adding a brief comment would be helpful.

openmls/src/credentials/mod.rs
```diff
@@ -1,3 +1,14 @@
+//! # Credentials
```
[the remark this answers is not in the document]
raphaelrobert: I changed the paragraph.

raphaelrobert:
> Thanks for tackling this!
> 
> As noted in the comments, I'm not sure that the text from the spec makes good documentation for our code. It helps to provide a link, but a lot of the text is not really relevant for the consumer. For example in places where our code makes sure certain properties hold that we don't even expose to the consumer.
> 
> Also, while I understand that some of the enums are somewhat self-explanatory, I still think that adding a brief comment would be helpful.

Regarding the enums: I skipped the documentation for those are come straight from the spec. All others are annotated.

openmls/src/credentials/mod.rs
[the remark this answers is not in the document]
kkohbrok: Thanks!

kkohbrok:
What is the rationale for skipping the ones that come from the spec? Do we assume that devs are familiar with the spec and thus the documentation is not necessary?

raphaelrobert:
> What is the rationale for skipping the ones that come from the spec? Do we assume that devs are familiar with the spec and thus the documentation is not necessary?

I only skipped it on variants of self-explanatory enums, not on the enums themselves. Things like ciphersuites, extensions, etc. I can copy over stuff form the spec, but that's also not what you wanted as per you initial comment.

kkohbrok (approved):
Thanks for making the changes! Let's get this in, then.

PR #745, opened 2022-01-31
use byte slice for key store key

### LINKED ISSUE #656: Key store - key bound
There have been concerns about the trait bound for the key values in the kay store.
Right now keys have to implement `Hash`.
https://github.com/openmls/openmls/blob/e9b65c651c56f101cff7d773717e261a3d74ef36/traits/src/key_store.rs#L25

One option would be to take a byte slice `k: &[u8]` instead. I think this would work. It might not improve performance though because the key might be a serialisation of a struct. But right now I think most keys can be cheaply converted to byte slices.

@OtaK thoughts or comments?

franziskuskiefer:
Note that the memory key store simply uses the provided bytes as key.
This is not recommended for production systems as this might get
large.

fixes #656

### DIFF
--- a/cli/src/identity.rs
+++ b/cli/src/identity.rs
@@ -18,10 +18,11 @@ fn store_key_package_bundle_in_keystore(
     crypto_backend
         .key_store()
         .store(
-            &key_package_bundle
+            key_package_bundle
                 .key_package()
                 .hash_ref(crypto_backend.crypto())
-                .expect("Failed to hash KeyPackage."),
+                .expect("Failed to hash KeyPackage.")
+                .as_slice(),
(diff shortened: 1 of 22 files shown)


raphaelrobert (commented):
Thanks, the changes look rather straightforward. Can we use this PR to also add some documentation in the book on how to use the key store?

franziskuskiefer:
> Can we use this PR to also add some documentation in the book on how to use the key store?

I can open another PR for that. I don't want to pollute this with unrelated content.

franziskuskiefer:
See #750

raphaelrobert (approved):
Thanks!

PR #747, opened 2022-01-31
Expose list of member KeyPackages in public API

### LINKED ISSUE #738: expose member list in public API

kkohbrok:
Fixes #738.

This PR changes the existing `members` function to return a vector of KeyPackages instead of credentials. This allows consumers to examine the corresponding credentials or compute KeyPackageRefs for use with other API endpoints.


openmls/src/group/mls_group/membership.rs
```diff
@@ -248,14 +248,14 @@ impl MlsGroup {
         self.plaintext_to_mls_message(remove_proposal, backend)
     }
 
-    /// Gets the current list of members
-    pub fn members(&self) -> Result<Vec<&Credential>, MlsGroupError> {
+    /// Gets a list of [`KeyPackage`]s of the current group members.
+    pub fn members(&self) -> Result<Vec<&KeyPackage>, MlsGroupError> {
```
franziskuskiefer, line 252: Can we get a test for this.
There are tests using this that could be extended.
Some scenarios how this could be used
- check that some member is some sender, i.e. check that the key package reference matches a certain value
- check a member's credential (this might be covered already)
- address members in the group, e.g. to remove

franziskuskiefer (changes requested):
Thanks.
Can you add some tests to make sure this works as expected?

openmls/src/group/mls_group/membership.rs
kkohbrok: Good point, I'll extend the tests as you describe.

raphaelrobert (changes requested):
Thanks for picking it up! The change looks pretty straightforward and I think it's the right minimal approach for now.

Could you also update/amend the book for this change?

book/src/user_manual/add_members.md
```diff
@@ -19,3 +19,11 @@ Members can also be added as a proposal (without the corresponding Commit messag
 ```
 
 In this case the the function returns an `MlsMessageOut` that needs to be fanned out to existing group members.
+
+# Retrieving a list of group members
```
raphaelrobert, line 23: I think this should get its own file because it has nothing to do with adding members. We will probably also expose more functions like looking up members by KPR or so.

raphaelrobert (changes requested):
Sorry, one more nit.

franziskuskiefer (dismissed):
Thanks

raphaelrobert (approved):
Thanks!

PR #748, opened 2022-01-31
Update secret tree derivations

### LINKED ISSUE #658: Generalize tree extend/truncate algorithms
Implement changes from spec PR [#522 ](https://github.com/mlswg/mls-protocol/pull/522).

- [ ] Update secret tree
- [ ] Update `LeafNodeHashInput`
- [ ] Update leaf/node indices

kkohbrok:
Fixes #658.

This PR changes the way that secrets are derived in the secret tree according to https://github.com/mlswg/mls-protocol/pull/522.

As far as I can tell, the rest of the changes are already implemented. In particular, the definition of `LeafNodeHashInput` already includes the `LeafIndex` rather than the `NodeIndex`.


kkohbrok:
@franziskuskiefer for some reason drone had trouble connecting to GH there for a second. Can you restart the `continuous-integration/drone/pr` check? I don't think I can.

openmls/src/tree/secret_tree.rs
```diff
@@ -51,23 +51,18 @@ impl From<&MlsPlaintext> for SecretType {
 pub(crate) fn derive_tree_secret(
```
franziskuskiefer, line 51: Looks like this is just an `kdf_expand_label` now and we can remove this.

franziskuskiefer (changes requested):
lgtm with one nit

openmls/src/tree/secret_tree.rs
raphaelrobert: Given that `DeriveTreeSecret` is explicitly defined in the spec before and after the spec change PR, I don't think it's a bad strategy to keep this.
Obviously it doesn't do anything functionally and should probably be `#[inline]` or so.

kkohbrok: I tend to agree with @raphaelrobert. I added an `#[inline]`, although I'm not sure it's going to do much, as well as a comment indicating that this function corresponds to the `DeriveTreeSecret` function. The latter might help later when correlating spec and implementation. @franziskuskiefer do you have a strong opinion on this?

franziskuskiefer: Not really. I'd rather remove it in the spec than keeping it here. But I wouldn't block on this.

kkohbrok: Then I'd say keep it in for now and file an issue with the spec. The secret tree needs a rewrite (or at least a major refactoring) anyway at some point and we can reconsider then if we feel like it's unnecessary.

franziskuskiefer (approved):

PR #755, opened 2022-02-02
pre 0.4 clean up and versions

franziskuskiefer:
In preparation for the 0.4 release I bumped some dependencies and set version numbers.

After this landed I'll go ahead and publish pre-release versions of all crates in this repo.


openmls_rust_crypto/src/provider.rs
```diff
@@ -409,4 +409,4 @@ impl OpenMlsRand for RustCrypto {
             .map_err(|_| "Unable to collect enough randomness.")?;
         Ok(out)
     }
-}
+}
```
kkohbrok, line 412: Shouldn't there be a newline here?

kkohbrok (commented):
Just one question, otherwise this looks good to me.

openmls_rust_crypto/src/provider.rs
franziskuskiefer: hm, I just moved it...  not sure what happened

kkohbrok (approved):

PR #757, opened 2022-02-02
Have removed members merge proposal changes to their TreeSync instance

### LINKED ISSUE #616: [Bug] Correct final state of the group membership upon removal
When a member is removed from a group, other proposals from that Commit are not applied to the tree/membership list which results in a wrong membership list.

kkohbrok:
This PR changes the bevahiour of `staged_commit` in `core_group` such that it does not immediately return after it notices that "self" was removed from the group. Instead, it includes a `StagedTreeSyncDiff` in the returned `StagedCommit`, which includes the results of all changes made in `apply_proposals`. The `StagedCommit` can then be merged after inspection, such that the membership list (but not the whole group state) of the removed client is the same as that of other clients after processing a given commit.

This PR also adds a test to test this new behaviour.

Reviewer notes:
- GH's diff is not very clear on this, but I didn't change any behaviour in the case that the group member was not removed.
- The enum/struct names I chose aren't great. Please feel free to suggest improved names!

Fixes #616.


openmls/src/group/core_group/test_core_group.rs
```diff
@@ -702,3 +702,199 @@ fn test_own_commit_processing(
...
+        backend,
+    )
+    .expect("Error joining group.");
+
+    // We can now check that Bob correctly processed his and applied the changes
+    // to his tree after he was removed by comparing membership lists. In
+    // particular, Bob's list should show that he was removed and Charlie was
+    // added.
+    let alice_members: Vec<&Credential> = alice_group
+        .treesync()
```
raphaelrobert, line 872: Why not use the `members()` function here?

raphaelrobert (changes requested):
Thanks for tackling this! Just some nits.

We need some documentation in the book for a) initiating a self-removal and b) processing a self-removal. It doesn't have to be in this PR, but since you are adding tests I was wondering if they could be used for the book right away.

openmls/src/group/core_group/test_core_group.rs
kkohbrok: Because this is on the `core_group` level. The `members()` function is on the `mls_group` level.

kkohbrok:
As mentioned in the comments, all of the changes here are on the `core_group` level, so I'd prefer not to put the `mls_group` documentation stuff into this PR. I filed #759 and will take care of it next.

PR #758, opened 2022-02-02
Clean up some more clippy warnings

raphaelrobert:
Turns out that one of the advantages of `thiserror` is that we can (and have to) document the errors.


openmls/src/credentials/errors.rs
```diff
@@ -2,12 +2,16 @@ use thiserror::Error;
 
 use crate::error::LibraryError;
 
+/// An error that occurs in methods of a [`Credential`].
 #[derive(Error, Debug, PartialEq, Clone)]
 pub enum CredentialError {
+    /// The type of credential is not supported.
     #[error("Unsupported credential type.")]
     UnsupportedCredentialType,
+    /// The signature that was verified with the credential was invalid.
```
kkohbrok, line 11: I think the `with` needs to be either `but` or `and`.

kkohbrok (changes requested):

PR #760, opened 2022-02-02
Expose update path key package separately in StagedCommit

### LINKED ISSUE #622: ManagedGroup API: Cannot distinguish between update and commit
In some cases it might be desirable to distinguish between an update proposal and a commit message

raphaelrobert:
This PR exposes the update path key package separately in a `StagedCommit`. Previously it was added as a fake Update. This way applications can actually distinguish between the two.

Fixes #622.


kkohbrok (commented):
Thanks for getting rid of this workaround. I stumbled over it a couple of times in the past, so it's good to finally see this go.

One improvement suggestion for the test to increase coverage a bit: In addition to checking against the sender's credential, we should check that the key package corresponds to the new leaf of the committer.

kkohbrok (approved):
Thanks!

PR #762, opened 2022-02-03
set metadata for all crate

franziskuskiefer:
Also, add `openmls_` prefix to memory key store crate name.


kkohbrok (approved):
Thanks for taking care of this!

PR #765, opened 2022-02-04
Add Proposal Validation Tests for ValSem101 and ValSem102

sgmenda:
Ref #525


openmls/src/group/tests/test_proposal_validation.rs
```diff
@@ -98,61 +152,191 @@ fn test_valsem100(ciphersuite: &'static Ciphersuite, backend: &impl OpenMlsCrypt
...
+
+/// ValSem102:
+/// Add Proposal:
+/// HPKE init key in proposals must be unique among proposals
+#[apply(ciphersuites_and_backends)]
+fn test_valsem102(ciphersuite: &'static Ciphersuite, backend: &impl OpenMlsCryptoProvider) {
+    for bob_and_charlie_share_hpke_init_keys in [
+        true,  // Negative Case: Bob and Charlie have same HPKE init key
+        false, // Positive Case: Bob and Charlie have different HPKE init keys
+    ] {
```
raphaelrobert, line 275: Not strictly necessary, but things become a little rustier by using enums because you can later `match` on them.

Example:

```rust
enum TestCase {
    PositiveDifferentKey,
    NegativeSameKey,
}

match test_case {
    TestCase::PositiveDifferentKey => {...},
    TestCase::NegativeSameKey => {...},
}
```

book/src/message_validation.md
```diff
@@ -29,8 +29,8 @@
 | ValidationStep | Description                                                                                 | Implemented    | Tested | Test File |
 | -------------- | ------------------------------------------------------------------------------------------- | -------------- | ------ | --------- |
 | `ValSem100`    | Add Proposal: Identity in proposals must be unique among proposals                          | ✅             | ❌<sup>2</sup> | `openmls/src/group/tests/test_proposal_validation.rs` |
-| `ValSem101`    | Add Proposal: Signature public key in proposals must be unique among proposals              | ✅             | ❌     | TBD       |
-| `ValSem102`    | Add Proposal: HPKE init key in proposals must be unique among proposals                     | ✅             | ❌     | TBD       |
+| `ValSem101`    | Add Proposal: Signature public key in proposals must be unique among proposals              | ✅             | ❌<sup>2</sup> | TBD       |
+| `ValSem102`    | Add Proposal: HPKE init key in proposals must be unique among proposals                     | ✅             | ❌<sup>2</sup> | TBD       |
```
raphaelrobert, line 33: I think the TODO for this is missing

raphaelrobert (changes requested):
Thanks for the PR! It looks good in general, but I noticed that the tests all happen at the level of `CoreGroup` (and that happened in a previous PR). The other validation checks all use the `MlsGroup` which also covers the `CoreGroup` internally. Since we want to make sure the validation works correctly for the public API (i.e. `MlsGroup`) we should also align it for the proposal validation. This could happen in a follow-up PR though, I'm fine merging it as-is.

sgmenda:
> Thanks for the PR! It looks good in general, but I noticed that the tests all happen at the level of `CoreGroup` (and that happened in a previous PR). The other validation checks all use the `MlsGroup` which also covers the `CoreGroup` internally. Since we want to make sure the validation works correctly for the public API (i.e. `MlsGroup`) we should also align it for the proposal validation. This could happen in a follow-up PR though, I'm fine merging it as-is.

Makes sense, I can do that next.

book/src/message_validation.md
sgmenda: It is the same TODO as `ValSem100`.

openmls/src/group/tests/test_proposal_validation.rs
sgmenda: Thank you for the suggestion. :)

raphaelrobert (approved):
Thanks!

PR #768, opened 2022-02-07
SenderRatchet rewrite for improved robustness and FS

### LINKED ISSUE #103: Improve FS in SecretTree
- [ ] Delete past path keys and only keep `RatchetSecrets`
 - [ ] Consume `RatchetSecrets`

kkohbrok:
Fixes #103.

This PR refactors SenderRatchet in several ways.
- there are now two types of sender ratchet: Encryption and Decryption
  - the `DecryptionRatchet` is essentially what we had before and it's used in all `SecretTree` leaves that are not the owner's. It can store secrets for out-of-order messages in case it needs to ratchet ahead to some epoch that's not the current one. Similarly, it has a limit on how far it can ratchet ahead in this way.
  - the `EncryptionRatchet` is in the leaf of the owner of the `SecretTree` and it can only ever ratchet one generation ahead. It will never store past secrets.
  - a result of this refactoring is that the `SecretTree` needs to know the owner's leaf index, so the constructor had to change
- the `DecryptionRatchet` now immediately derives key-nonce-pairs when ratcheting ahead. This allows us to achieve a more fine-grained FS, as we don't store past ratchet secrets that would allow an adversary to derive all keys from a given compromised secret to the current ratchet chain head.
- A lot of tests had to change, either because they relied on the missing FS (using a secret multiple times, once to encrypt and then again to decrypt), or they used the same tree to encrypt and decrypt messages from the same leaf.


raphaelrobert (changes requested):
Thanks for the PR, this definitely improves FS! Just a few nits.

openmls/src/tree/sender_ratchet.rs
```diff
@@ -53,152 +57,194 @@ impl Default for SenderRatchetConfiguration {
...
+            past_secrets: VecDeque::new(),
+            ratchet_head: RatchetSecret::initial_ratchet_secret(secret.clone()),
+        }
+    }
+
+    /// Remove elements from the `past_secrets` queue until it is within the
+    /// bounds determined by the [`SenderRatchetConfiguration`].
+    fn prune_past_secrets(&mut self, configuration: &SenderRatchetConfiguration) {
+        while self.past_secrets.len() > 0
+            && self.past_secrets.len() > configuration.out_of_order_tolerance() as usize
```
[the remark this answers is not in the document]
kkohbrok: 🤦🏻 👍🏻

openmls/src/tree/secret_tree.rs
```diff
@@ -102,6 +109,7 @@ impl SecretTree {
         });
 
         SecretTree {
+            own_index,
             nodes,
             handshake_sender_ratchets: vec![None; size.as_usize()],
```
franziskuskiefer, line 114: When we're rewriting this anyway we should try do something else here. Allocating two vectors of the size of the tree is pretty heavy and probably not really neded. Also, this requires you to allow cloning of keys, which is sort off the opposite of what you want to do with this PR.

openmls/src/tree/sender_ratchet.rs
```diff
@@ -53,152 +57,192 @@ impl Default for SenderRatchetConfiguration {
...
+        Self {
+            past_secrets: VecDeque::new(),
+            ratchet_head: RatchetSecret::initial_ratchet_secret(secret.clone()),
+        }
+    }
+
+    /// Remove elements from the `past_secrets` queue until it is within the
+    /// bounds determined by the [`SenderRatchetConfiguration`].
+    fn prune_past_secrets(&mut self, configuration: &SenderRatchetConfiguration) {
+        while self.past_secrets.len() > configuration.out_of_order_tolerance() as usize {
```
franziskuskiefer, line 173: We decided to use something like the `MessageSecretsStore::resize` function does to avoid `while` loops. I have no strong opinions on this. `while` is more concise, but also more error prone and harder to debug.

openmls/src/tree/sender_ratchet.rs
```diff
@@ -53,152 +57,192 @@ impl Default for SenderRatchetConfiguration {
...
         ciphersuite: &Ciphersuite,
         backend: &impl OpenMlsCryptoProvider,
-        generation: u32,
+        generation: Generation,
         configuration: &SenderRatchetConfiguration,
-    ) -> Result<RatchetSecrets, SecretTreeError> {
+    ) -> Result<RatchetKeyMaterial, SecretTreeError> {
         // If generation is too distant in the future
-        if generation > (self.generation + configuration.maximum_forward_distance()) {
+        if generation > (self.generation() + configuration.maximum_forward_distance()) {
```
franziskuskiefer, line 193: `elf.generation() + configuration.maximum_forward_distance()` can overflow

openmls/src/group/core_group/staged_commit.rs
```diff
@@ -278,8 +278,12 @@ impl CoreGroup {
         }
 
         let (provisional_group_epoch_secrets, provisional_message_secrets) =
-            provisional_epoch_secrets
-                .split_secrets(serialized_provisional_group_context, diff.leaf_count());
+            provisional_epoch_secrets.split_secrets(
+                serialized_provisional_group_context,
+                diff.leaf_count(),
+                // The index should be the same on TreeSync and Diff.
```
franziskuskiefer, line 284: Sounds like there should be a `debug_assert` making sure this is the case.

franziskuskiefer (changes requested):

openmls/src/tree/secret_tree.rs
kkohbrok: The rewrite was more focused on the `sender_ratchet` and the FS deficiency that results from keeping around past ratchet secrets rather than the aead keys and nonces. Here I'm just passing along an index. That being said, I enjoy this sort of work, so I'd be happy to rewrite the `secret_tree` as well, although it does distract somewhat from the 0.4 goal.

openmls/src/tree/sender_ratchet.rs
```diff
@@ -53,152 +57,192 @@ impl Default for SenderRatchetConfiguration {
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(any(feature = "test-utils", test), derive(PartialEq))]
+pub struct DecryptionRatchet {
+    past_secrets: VecDeque<Option<RatchetKeyMaterial>>,
+    ratchet_head: RatchetSecret,
+}
+
+impl DecryptionRatchet {
+    /// Creates e new SenderRatchet
+    pub fn new(secret: &Secret) -> Self {
```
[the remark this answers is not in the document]
kkohbrok: I thought about that, but I was afraid that I'd have to dig into `secret_tree` as well and not stop until I'm on the other end 😅. But as noted in another comment, I really enjoy refactoring, so I'd be happy to go for it.

openmls/src/tree/sender_ratchet.rs
kkohbrok: Ah, I was not aware that we had made a decision to avoid `while` functions generally. I'll take a look.

openmls/src/tree/sender_ratchet.rs
kkohbrok: Ugh. Good point. I keep missing these.

kkohbrok (commented):
Thanks for the reviews! I guess the question is if I should extend this PR to `secret_tree`, which would probably benefit from a rewrite.

openmls/src/tree/sender_ratchet.rs
raphaelrobert: Right, last time I was the one pushing for it. I still think not having `while` loops is generally preferable, because if they loop forever in production it can be nasty.

raphaelrobert (approved):
Thanks, I'll approve from my side and defer to @franziskuskiefer for the rest.

openmls/src/group/core_group/staged_commit.rs
kkohbrok: On second thought, tt's probably safer to just make `own_leaf_index` on `TreeSyncDiff` crate-public and use that one.

openmls/src/tree/secret_tree.rs
kkohbrok: Ok, I filed #789.

kkohbrok (commented):
Thanks again for the round of review. We should be getting there.

openmls/src/tree/secret_tree.rs
```diff
@@ -182,16 +191,32 @@ impl SecretTree {
...
-        self.handshake_sender_ratchets[index.as_usize()] = Some(handshake_sender_ratchet);
         let application_ratchet_secret = derive_tree_secret(
             node_secret,
             "application",
             0,
             ciphersuite.hash_length(),
             backend,
         )?;
-        let application_sender_ratchet = SenderRatchet::new(index, &application_ratchet_secret);
+        let (handshake_sender_ratchet, application_sender_ratchet) = if index == self.own_index {
```
franziskuskiefer, line 201: This is the only place where `own_index` is used. It feels like a lot of overhead carrying the value around for this one check.

openmls/src/ciphersuite/aead.rs
```diff
@@ -1,15 +1,15 @@
 use super::*;
 
 /// AEAD keys holding the plain key value and the AEAD algorithm type.
```
franziskuskiefer, line 3: You can drop the `Clone` here and for the Nonce and move it to test/test-utils only.

kkohbrok: It turns out that's not possible without some refactoring on the `SecretTree`. We need clone here, because we store these in the `DecryptionRatchet`, which in turn needs to implement clone, because the `SecretTree` uses `vec![None, size]` to initialize the vector of `SenderRatchet`s. I could replace that initialization, but since we're going to touch all of that anyway in #789, I'd prefer to leave a comment there indicating that the cascade of `Clone` derivations needs to be removed.

openmls/src/tree/sender_ratchet.rs
```diff
@@ -53,152 +57,207 @@ impl Default for SenderRatchetConfiguration {
...
+/// secrets around.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[cfg_attr(any(feature = "test-utils", test), derive(PartialEq))]
-pub struct SenderRatchet {
-    index: SecretTreeLeafIndex,
-    generation: u32,
-    past_secrets: Vec<Secret>,
+pub(crate) enum SenderRatchet {
+    EncryptionRatchet(RatchetSecret),
+    DecryptionRatchet(DecryptionRatchet),
```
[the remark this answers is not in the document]
franziskuskiefer: No worries... nothing you introduced. Let's keep it as is for now.

openmls/src/tree/secret_tree.rs
franziskuskiefer: 👍🏻  yeah that's nice. We could pass in something to distinguish the own index instead. But let's keep it as is.

openmls/src/ciphersuite/aead.rs
franziskuskiefer: Well, you can do something like  `(0..size.as_usize()).map(|_| None).collect()` instead of `vec!`. That's not super nice but would avoid requiring the clone until #789 gets fixed.

kkohbrok: Fair enough. I ended up choosing another method of initialization I found [here](https://stackoverflow.com/questions/65011997/why-cant-none-be-cloned-for-a-generic-optiont-when-t-doesnt-implement-clone), which I believe is equivalent in efficiency as `vec![...]`. Although it might all be equivalent in the end. In any case, it allowed us to finally get rid of the `Clone` derivations 🙂.

franziskuskiefer (approved):
Thanks. Let's get it in.

PR #777, opened 2022-02-07
Introduce RemoveOperation

### LINKED ISSUE #623: ManagedGroup API: Remove: cannot determine if someone left or was removed
We cannot determine if someone left or was removed (also: distinction between own member and other members)

raphaelrobert:
This PR introduces the `RemoveOperation` `enum` that helps interpret the semantic value of a remove proposal when inspecting proposals that are covered by a Commit.

Fixes #623.

### DIFF
--- a/openmls/src/group/tests/test_remove_operation.rs
+++ b/openmls/src/group/tests/test_remove_operation.rs
@@ -1,0 +1,316 @@
+//! This module tests the classification of remove operations with RemoveOperation
+
+use super::utils::{generate_credential_bundle, generate_key_package_bundle};
+use crate::{credentials::*, framing::*, group::*, test_utils::*, *};
+use openmls_rust_crypto::OpenMlsRustCrypto;
+
+// Tests the differen variants of the RemoveOperation enum.
+#[apply(ciphersuites_and_backends)]
+fn test_remove_operation_variants(
+    ciphersuite: &'static Ciphersuite,
+    backend: &impl OpenMlsCryptoProvider,
+) {
+    // We define two test cases, one where the member is removed by another member
+    // and one where the member leaves the group on its own
+    enum TestCase {
+        Remove,
+        Leave,
+    }
+
+    for test_case in [TestCase::Remove, TestCase::Leave] {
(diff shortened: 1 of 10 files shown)


openmls/src/group/tests/test_remove_operation.rs
```diff
@@ -0,0 +1,316 @@
...
+        // === Remove operation ===
+
+        let alice_kpr = *alice_group
+            .key_package_ref()
+            .expect("An unexpected error occurred.");
+        let bob_kpr = *bob_group
+            .key_package_ref()
+            .expect("An unexpected error occurred.");
+
+        // We differentiate between th two test cases here
```
kkohbrok, line 136:
```suggestion
        // We differentiate between the two test cases here
```

openmls/src/group/tests/test_remove_operation.rs
```diff
@@ -0,0 +1,316 @@
...
+        // === Remove operation from Alice's perspective ===
+
+        let alice_staged_commit = alice_group.pending_commit().expect("No pending commit.");
+
+        let remove_proposal = alice_staged_commit
+            .remove_proposals()
+            .next()
+            .expect("An unexpected error occurred.");
+
+        let remove_operation = RemoveOperation::new(remove_proposal, &alice_group)
```
kkohbrok, line 182: So the consumer inspecting the proposal has to create the `RemoveOperation` themselves? For inspection purposes, wouldn't it be more convenient for the consumer if they had all the information available immediately? Don't get me wrong, this is a great change already, I'm just wondering what we want the consumer API to look like eventually.

kkohbrok (commented):
Thanks for the PR! Looks good to me already with just one typo and one question.

openmls/src/group/tests/test_remove_operation.rs
raphaelrobert: As discussed, I think it's ok to just have it as an external helper for now. The current layout makes it easier to treat all remove proposals the same, which makes bulk parsing easier.
We could also add something similar for add and update proposals (although it's less needed there).

kkohbrok (approved):
Thanks!

PR #781, opened 2022-02-08
Delete keys from key store when joining a group

### LINKED ISSUE #778: Joining a group: Pick the right key
When joining a group in `new_from_welcome()`, the first available key is picked from the key store. There could be more than one valid key and there could be a discrepancy between the private key material of the client and the public group state.

In practice, this should not be a problem, but in tests where different clients use the same key store, this leads to the situation where clients join with the wrong identity.

raphaelrobert:
This PR addresses an issue discovered in #777. For better FS, `KeyPackageBundle`s should be deleted when joining a group from a `Welcome` message.

Fixes #778.


kkohbrok (approved):
This was indeed a good catch! We might want to revisit the key store design in the future, such that we have the option to store secrets in such a way that they can't be retrieved without removing them from the store entirely.

PR #782, opened 2022-02-08
Cleaned up rustdocs for extensions

### LINKED ISSUE #772: extensions rustdocs

franziskuskiefer:
Some cleanup on rustdocs for extensions.

fixes #772


kkohbrok (approved):
Looks good to me, thanks!

Since we're using text from the MLS spec in a lot of places, it might be worth thinking about how we maintain these docs while the spec is still changing quite a bit. While we could track each PR and update the docs in the same way as we do spec changes, it might be better to just do one big pass when the spec is released.

PR #783, opened 2022-02-08
Clean up visibility of treesync module

### LINKED ISSUE #770: treesync rustdocs

kkohbrok:
Fixes #770.

This PR does the following this:
- make `treesync` `pub(crate)` and thus gets rid of the majority of doc errors
- delete some unused errors (TreeError)
- change the error strings of `CoreGroupError` and `MlsGroupError` for errors out of the `treesync` module. This was necessary, because those errors are public, which I assume will change with #779 or in the general course of cleanup of the public API.
- move the `print_tree` function into the `group/tests` module and expose the function as part of the `CoreGroup` implementation, which I believe makes more sense anyway


openmls/src/framing/validation.rs
```diff
@@ -151,7 +151,7 @@ impl DecryptedMessage {
     ///  - ValSem246
     ///  - Prepares ValSem247 by setting the right credential. The remainder
     ///    of ValSem247 is validated as part of ValSem010.
-    pub fn credential(
+    pub(crate) fn credential(
```
raphaelrobert, line 154: I don't think we expose `DecryptedMessage` at all in public API, so all of it could probably `pub(crate)`.

raphaelrobert (changes requested):
Thanks for the PR, restricting is good in general. We can look into exposing the nodes again later. Just one nit regarding `DecryptedMessage`.

raphaelrobert (approved):

openmls/src/lib.rs
```diff
@@ -54,11 +54,11 @@ pub mod key_packages;
 pub mod key_store;
 pub mod messages;
 pub mod schedule;
-pub mod treesync;
 
 // Private
 mod binary_tree;
 mod tree;
+mod treesync;
```
pulkitkothari, line 61: This change again reopened issue: https://github.com/openmls/openmls/issues/722
@kkohbrok  @raphaelrobert

PR #786, opened 2022-02-08
Add documentation for leaving and getting removed

### LINKED ISSUE #759: Improve documentation for initiating and processing self-removal.
The corresponding documentation should go in the book along with a code example.

kkohbrok:
Fixes #759.

Adds some documentation to the book around leaving and getting removed from a group. Also, adds some new asserts to reflect what's written in the docs.


book/src/user_manual/remove_members.md
```diff
@@ -20,3 +20,18 @@ Members can also be removed as a proposal (without the corresponding Commit mess
 ```
 
 In this case the the function returns an `MlsMessageOut` that needs to be fanned out to existing group members.
+
+## Getting removed from a group
+A member is removed from a group if another member commits to a remove proposal targeting the member's leaf. Once that commit is merged via `merge_staged_commit()`, the group will apply all other proposals in that commit and mark the group as inactive. The member can still use the group, e.g. to examine the membership list after the final commit was processed, but it won't be able to create or process new messages.
```
raphaelrobert, line 25: Maybe clarify that this is from the perspective of the removed member.

book/src/user_manual/remove_members.md
```diff
@@ -20,3 +20,18 @@ Members can also be removed as a proposal (without the corresponding Commit mess
...
+
+```rust,no_run,noplayground
+{{#include ../../../openmls/tests/book_code.rs:getting_removed}}
+```
+
+### Remove self
```
raphaelrobert, line 31: We already have a category for leaving a group. This should be consolidated I think.

raphaelrobert (changes requested):
Thanks for the PR, it looks like there is some overlap with existing documentation.

book/src/user_manual/remove_members.md
kkohbrok: The headline should indicate this, but I rephrased the text to make this clearer as well.

openmls/tests/book_code.rs
```diff
@@ -693,6 +694,11 @@ fn book_operations(ciphersuite: &'static Ciphersuite, backend: &impl OpenMlsCryp
 
     // Check that Bob's group is no longer active
     assert!(!bob_group.is_active());
+    let bob_members = bob_group.members().expect("Error getting list of members.");
+    assert_eq!(bob_members.len(), 2);
+    assert_eq!(bob_members[0].credential().identity(), b"Alice");
+    assert_eq!(bob_members[1].credential().identity(), b"Charlie");
+    // ANCHOR_END: getting_removed
```
raphaelrobert, line 701:
```suggestion
    let members = bob_group.members().expect("Error getting list of members.");
    assert_eq!(members.len(), 2);
    assert_eq!(members[0].credential().identity(), b"Alice");
    assert_eq!(members[1].credential().identity(), b"Charlie");
    // ANCHOR_END: getting_removed
```

raphaelrobert (approved):
Thanks for the changes! Just two suggestions.

PR #787, opened 2022-02-08
Simplify Sender

### LINKED ISSUE #575: Find a consistent scheme to address to address members
Right now members are addressed by their `LeafIndex`, which is neither intuitive nor practical for applications.

Currently blocked by #541.

raphaelrobert:
This PR simplifies how `Sender` works and is exposed in the public API. The new `Sender` type replaces the old `Sender`/`SenderType`/`SenderValue` and uses the TLS codec discriminants instead of custom serialization/deserialization code.

The public API is now easier to use, since the new `Sender` is a public `enum` instead of an opaque `struct`. The improvement can be seen in the book's sample code.

Fixes #575.

### DIFF
--- a/openmls/src/group/core_group/apply_proposals.rs
+++ b/openmls/src/group/core_group/apply_proposals.rs
@@ -93,6 +93,7 @@ impl CoreGroup {
             if let Proposal::Update(update_proposal) = queued_proposal.proposal() {
                 // Check if this is our own update.
                 let sender = queued_proposal.sender();
+                // Only members can send update proposals
                 let hash_ref = match sender {
                     Sender::Member(hash_ref) => hash_ref,
                     _ => return Err(CoreGroupError::SenderError(SenderError::NotAMember)),
--- a/openmls/src/group/core_group/apply_proposals.rs
+++ b/openmls/src/group/core_group/apply_proposals.rs
@@ -4,7 +4,7 @@ use openmls_traits::OpenMlsCryptoProvider;
 
 use crate::{
     binary_tree::LeafIndex,
-    framing::{SenderError, SenderNew},
+    framing::{Sender, SenderError},
     group::CoreGroupError,
     key_packages::KeyPackageBundle,
     messages::{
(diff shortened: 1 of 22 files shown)


openmls/src/group/core_group/apply_proposals.rs
```diff
@@ -92,8 +93,13 @@ impl CoreGroup {
             if let Proposal::Update(update_proposal) = queued_proposal.proposal() {
                 // Check if this is our own update.
                 let sender = queued_proposal.sender();
+                // Only members can send update proposals
+                let hash_ref = match sender {
+                    Sender::Member(hash_ref) => hash_ref,
+                    _ => return Err(CoreGroupError::SenderError(SenderError::NotAMember)),
```
kkohbrok, line 99: Do we test for this somewhere already? Or was this error previously done in `as_key_package_ref()`?

openmls/src/group/mls_group/test_mls_group.rs
```diff
@@ -269,14 +268,10 @@ fn remover(ciphersuite: &'static Ciphersuite, backend: &impl OpenMlsCryptoProvid
...
-        assert_eq!(
-            staged_proposal
-                .sender()
-                .as_key_package_ref()
-                .expect("An unexpected error occurred."),
-            alice_kpr
-        );
+        assert!(matches!(
+            staged_proposal.sender(),
+            Sender::Member(member) if member == alice_kpr
```
kkohbrok, line 273: I think I've never seen this construction in rust before. What is going on here?

kkohbrok (commented):
That's some nice cleanup and refactoring, thanks!

Just two questions, otherwise this looks good to me.

openmls/src/group/core_group/apply_proposals.rs
raphaelrobert: The latter. I'm really just replacing `as_key_package_ref()` here. I was also wondering if this should become a validation criterion. I filed #788.

openmls/src/group/mls_group/test_mls_group.rs
raphaelrobert: This is specific to the `matches!` macro: https://doc.rust-lang.org/std/macro.matches.html

kkohbrok: Very cool! I had no idea this was a possibility. Seems useful, though.

kkohbrok (approved):
Thanks! No objections from my side :-)

PR #792, opened 2022-02-09
Add `member` endpoint

### LINKED ISSUE #784: Allow `mls_group` API users to fetch a KeyPackage using a `KeyPackageRef`

kkohbrok:
This PR adds a function to `mls_group` that allows the consumer to fetch a member's `KeyPackage` given the corresponding `KeyPackageRef`. It also adds some documentation to the book, as well as a small test to the `book_code`.

Fixes #784.


openmls/src/group/mls_group/membership.rs
```diff
@@ -259,6 +259,20 @@ impl MlsGroup {
             .collect())
     }
 
+    /// Get the [`KeyPackage`] of a member corresponding to the given
+    /// [`KeyPackageRef`]. Returns `None` if no matching [`KeyPackage`] can be
```
franziskuskiefer, line 263: But it's also returning an error. Either describe what that is, or preferably only return an `Option`.

openmls/tests/book_code.rs
```diff
@@ -478,6 +478,21 @@ fn book_operations(ciphersuite: &'static Ciphersuite, backend: &impl OpenMlsCryp
     assert_eq!(members[1].credential().identity(), b"Bob");
     assert_eq!(members[2].credential().identity(), b"Charlie");
 
+    // Check that the `member` and the `members` function are consistent
+    for member in members {
```
franziskuskiefer, line 482: Can you also check that the number of members is what we expect? Otherwise this could just skip members.

franziskuskiefer (changes requested):
Thanks, only a few nits.

raphaelrobert (changes requested):
Thanks for tackling this, just some nits.

openmls/src/group/mls_group/membership.rs
```diff
@@ -259,6 +259,20 @@ impl MlsGroup {
...
+    /// found in this group.
+    pub fn member(
+        &self,
+        key_package_ref: &KeyPackageRef,
+    ) -> Result<Option<&KeyPackage>, MlsGroupError> {
```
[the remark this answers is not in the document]
kkohbrok: It makes sense in some cases to return both. In my mind, it should give you a `None` if there is no such member and an error if something else went wrong. In this case, the only thing that can go wrong is an `ArchitectureError`, so maybe we can get away in this case to just return a `None` instead, but I disagree that this is always the case.

openmls/src/group/mls_group/membership.rs
kkohbrok: Good point, that should have been in the docs. I changed the function, though to return `None` when an error is bubbled up. (See my reply [here](https://github.com/openmls/openmls/pull/792#discussion_r804456019).)

franziskuskiefer (dismissed):
Thanks

raphaelrobert (dismissed):

franziskuskiefer (approved):

PR #797, opened 2022-02-11
First pass on cleaning up errors

raphaelrobert:
This PR does the following:
 - Convert errors into new syntax with `thiserror`
 - Gets rid of unused variants (after the conversion)
 - Refactors some functions to either not return Results or to return library errors only
 - Gets rid of some completely unused errors


openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -277,7 +281,10 @@ impl<'a, T: Clone + Debug> AbDiff<'a, T> {
             .iter()
             .position(|&direct_path_node_index| direct_path_node_index == subtree_root_node_index)
             // The shared subtree root has to be in the direct path of both nodes.
-            .ok_or(ABinaryTreeDiffError::LibraryError)
+            .ok_or_else(|| {
+                LibraryError::custom("ABDiff::node_by_index(): index should be in the direct path")
```
franziskuskiefer, line 285: Putting function names in comments is always dangerous as they might go out of date.

openmls/src/binary_tree/array_representation/tree.rs
```diff
@@ -169,25 +182,26 @@ impl<T: Clone + Debug> ABinaryTree<T> {
...
-        InvalidNumberOfNodes = "The given number of nodes does not allow the creation of a full, left-balanced binary tree.",
-        OutOfBounds = "The given index is outside of the tree.",
-        AddressCollision = "Found two nodes with the same address.",
-        InvalidNode = "Can't add the default node to the tree.",
-        NodeNotFound = "Can't find the node with the given address in the tree.",
-        LibraryError = "An unrecoverable error has occurred due to a bug in the implementation.",
-        ArchitectureError = "Architectures below 32bit are not suppored by OpenMLS.",
-        ABinaryTreeDiffError = "An error occurred while handling a diff.",
-    }
+/// Binary Tree error
```
franziskuskiefer, line 194: The comment is "Binary Tree error" but the enum is `ABinaryTreeError`. The comment doesn't help to understand what this is and only confuses.

openmls/src/binary_tree/array_representation/treemath.rs
```diff
@@ -1,16 +1,20 @@
...
-implement_error! {
-    pub enum TreeMathError {
-        LeafHasNoChildren = "Leaf nodes don't have children.",
-        RootHasNoParent = "Root nodes don't have parents.",
-        NotAParentNode = "Node index was not a parent node.",
-        LeafNotInTree = "The leaf index is larger than the tree size.",
-        NodeNotInTree = "The node index is larger than the tree size.",
-        InvalidInput = "The provided input is invalid for tree math.",
-    }
+/// Binary Tree error
```
franziskuskiefer, line 7: Copy paste? Public items should get useful comments now.

openmls/src/group/errors.rs
```diff
@@ -293,23 +273,39 @@ implement_error! {
...
-        ExistingPublicKeyAddProposal = "HPKE public key of the add proposal already existed in tree.",
-        UpdateProposalIdentityMismatch = "The identity of the update proposal did not match the existing identity.",
-        ExistingSignatureKeyUpdateProposal = "Signature key of the update proposal already existed in tree.",
-        ExistingPublicKeyUpdateProposal = "HPKE public key of the update proposal already existed in tree.",
-        DuplicateMemberRemoval = "Duplicate remove proposals for the same member.",
-        UnknownMemberRemoval = "The remove proposal referenced a non-existing member.",
-        UpdateFromNonMember = "Found an update from a non-member.",
-        CommitterIncludedOwnUpdate = "The Commit includes update proposals from the committer.",
-    }
+/// Proposal validation error
```
franziskuskiefer, line 276: When is this error thrown and what is the consumer supposed to do with it?

franziskuskiefer (changes requested):
Generally looks good. Two main things that should get addressed

- Don't use function/struct names etc. in error messages. They might change. Instead just describe what went wrong.
- To avoid having to touch all the documentation again, please add useful descriptions to all public errors.

raphaelrobert:
Thanks for the review! To your points:

 - I thought this PR is self-contained enough to get it in, so that we can review more incrementally and not have a big PR in the end. Obviously, everything is still WIP and this is not the end of it.
 - For now, I'm working from the bottom up. This PR is not useful for consumers at this point. My strategy is to simplify the intermediary errors, so that we don't have to surface all of them to the consumer.
 - This s also why I haven't added any docs to the errors yet. My hope is that I can turn many of them into `pub(crate)` instead of surfacing them. The final public errors should of course get proper documentation. I just wanted to avoid a lot of unnecessary c & p for now.
 - Re the function names: I'm not really happy about it either. I added them to get more context for potential debugging. Maybe this [can be done programmatically](https://rustc-dev-guide.rust-lang.org/backend/implicit-caller-location.html#caller-location-in-const), but I haven't investigated properly yet. I can take them out if you prefer, let me know.

franziskuskiefer:
> * Re the function names: I'm not really happy about it either. I added them to get more context for potential debugging. Maybe this [can be done programmatically](https://rustc-dev-guide.rust-lang.org/backend/implicit-caller-location.html#caller-location-in-const), but I haven't investigated properly yet. I can take them out if you prefer, let me know.

Putting a call stack in or something would be the best of course. More context for debugging is of course good but if something gets renamed might as well be super confusing for anyone not knowing the history. I'd prefer not adding names and maybe filing an issue to investigate adding call stacks.

Everything else we can then keep as is for now when it'll change in follow ups anyway.

raphaelrobert:
I dropped the function names and introduced backtraces instead that are appended to every custom error string.

franziskuskiefer (approved):
ok, let's get it in

PR #799, opened 2022-02-11
Add missing TreeSync diagram

bennycode:
I noticed that the [TreeSync Architecture](https://openmls.tech/book/tree_rewrite.html) image is missing in the [book](https://openmls.tech/book/) but available in the [wiki](https://github.com/openmls/openmls/wiki/Structure-and-APIs). 

My PR adds the SVG to the static folder, so that it can also be shown in the book.


raphaelrobert (approved):
Awesome, thanks for the PR Benny!

PR #801, opened 2022-02-13
Error cleanup 2

raphaelrobert:
This PR continues with the error cleanup:

 - I wrote a [wiki article](https://github.com/openmls/openmls/wiki/Error-management-in-OpenMLS) explaining the overall error management strategy (will be part of the book later)
 - `WelcomeError`/`ExternalCommitError` are public and in near-final state (except for documentation)
 - Refactoring in TreeSync/TreeSyncDiff/ABinaryTree API to return LibraryError more often
 - more than 50% of the errors now use `thiserror`

Sorry for the large PR again, most of it should be easy to review though.

### DIFF
--- a/openmls/src/group/core_group/new_from_welcome.rs
+++ b/openmls/src/group/core_group/new_from_welcome.rs
@@ -25,14 +25,15 @@ impl CoreGroup {
         }
 
         let ciphersuite_name = welcome.ciphersuite();
-        let ciphersuite = Config::ciphersuite(ciphersuite_name)?;
+        let ciphersuite = Config::ciphersuite(ciphersuite_name)
+            .map_err(|_| WelcomeError::UnsupportedCiphersuite)?;
 
         // Find key_package in welcome secrets
         let egs = if let Some(egs) = Self::find_key_package_from_welcome_secrets(
             key_package_bundle.key_package(),
             welcome.secrets(),
             backend,
-        )? {
+        ) {
             egs
         } else {
             return Err(WelcomeError::JoinerSecretNotFound);
@@ -43,32 +44,45 @@ impl CoreGroup {
             return Err(e);
         }
 
-        let group_secrets_bytes = backend.crypto().hpke_open(
-            ciphersuite.hpke_config(),
-            &egs.encrypted_group_secrets,
-            key_package_bundle.private_key().as_slice(),
-            &[],
-            &[],
-        )?;
-        let group_secrets = GroupSecrets::tls_deserialize(&mut group_secrets_bytes.as_slice())?
+        let group_secrets_bytes = backend
+            .crypto()
+            .hpke_open(
+                ciphersuite.hpke_config(),
+                &egs.encrypted_group_secrets,
+                key_package_bundle.private_key().as_slice(),
+                &[],
+                &[],
+            )
+            .map_err(|_| WelcomeError::CouldNotDecrypt)?;
+        let group_secrets = GroupSecrets::tls_deserialize(&mut group_secrets_bytes.as_slice())
+            .map_err(|_| WelcomeError::MalformedWelcomeMessage)?
             .config(ciphersuite, mls_version);
         let joiner_secret = group_secrets.joiner_secret;
 
         // Prepare the PskSecret
-        let psk_secret = PskSecret::new(ciphersuite, backend, group_secrets.psks.psks())?;
+        let psk_secret = PskSecret::new(ciphersuite, backend, group_secrets.psks.psks()).map_err(
+            |e| match e {
+                PskSecretError::LibraryError(e) => e.into(),
+                PskSecretError::TooManyKeys => WelcomeError::PskTooManyKeys,
+                PskSecretError::KeyNotFound => WelcomeError::PskNotFound,
+            },
+        )?;
 
         // Create key schedule
         let mut key_schedule = KeySchedule::init(ciphersuite, backend, joiner_secret, psk_secret)?;
 
         // Derive welcome key & nonce from the key schedule
         let (welcome_key, welcome_nonce) = key_schedule
-            .welcome(backend)?
-            .derive_welcome_key_nonce(backend)?;
+            .welcome(backend)
+            .map_err(|_| LibraryError::custom("Using the key schedule in the wrong state"))?
+            .derive_welcome_key_nonce(backend)
+            .map_err(LibraryError::unexpected_crypto_error)?;
 
         let group_info_bytes = welcome_key
             .aead_open(backend, welcome.encrypted_group_info(), &[], &welcome_nonce)
             .map_err(|_| WelcomeError::GroupInfoDecryptionFailure)?;
-        let group_info = GroupInfo::tls_deserialize(&mut group_info_bytes.as_slice())?;
+        let group_info = GroupInfo::tls_deserialize(&mut group_info_bytes.as_slice())
+            .map_err(|_| WelcomeError::MalformedWelcomeMessage)?;
 
         // Make sure that we can support the required capabilities in the group info.
         let group_context_extensions = group_info.group_context_extensions();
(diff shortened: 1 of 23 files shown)


openmls/src/group/errors.rs
```diff
@@ -8,269 +8,232 @@ use crate::{
...
+    #[error("We don't support the ciphersuite of the group we are trying to join.")]
+    UnsupportedCiphersuite,
+    #[error("We don't support all capabilities of the group.")]
+    UnsupportedCapability,
+    #[error("Sender not found in tree.")]
+    UnknownSender,
+    #[error("Malformed Welcome message.")]
+    MalformedWelcomeMessage,
+    #[error("Could not decrypt the Welcome message.")]
+    CouldNotDecrypt,
```
franziskuskiefer, line 111: Maybe `UnableToDecrypt`?

openmls/src/group/errors.rs
```diff
@@ -8,269 +8,232 @@ use crate::{
...
+    LibraryError(#[from] LibraryError),
+    #[error("The epoch of the group context and MlsPlaintext didn't match.")]
+    EpochMismatch,
+    #[error("stage_commit was called with an MlsPlaintext that is not a Commit.")]
+    WrongPlaintextContentType,
+    #[error("Unable to verify the key package signature.")]
+    PathKeyPackageVerificationFailure,
+    #[error("Unable to determine commit path.")]
+    RequiredPathNotFound,
+    #[error("Confirmation Tag is missing.")]
```
franziskuskiefer, line 164:
```suggestion
    #[error("The confirmation Tag is missing.")]
```

openmls/src/group/errors.rs
```diff
@@ -8,269 +8,232 @@ use crate::{
...
+    ConfirmationTagMissing,
+    #[error("Confirmation tag is invalid.")]
```
franziskuskiefer, line 166: Some messages are real sentences already. I think we should make all of them proper english.
```suggestion
    #[error("The confirmation tag is invalid.")]
```

openmls/src/group/errors.rs
```diff
@@ -308,28 +271,27 @@ pub enum ProposalValidationError {
...
+    #[error("Found inline Add or Update proposals.")]
+    InvalidInlineProposals,
+    // FIXME: this seems unused
+    #[error("Found multiple inline Remove proposals.")]
+    MultipleRemoveProposals,
+    #[error("Remove proposal targets the wrong group member.")]
+    InvalidRemoveProposal,
+    #[error("Found an ExternalInit proposal among the referenced proposals.")]
+    ReferencedExternalInitProposal,
+    // FIXME: this seems unused
```
franziskuskiefer, line 292: Remove then? Or file an issue and link here at least.

openmls/src/schedule/errors.rs
```diff
@@ -30,14 +30,12 @@ pub enum KeyScheduleError {
 /// PSK secret error
 #[derive(Error, Debug, PartialEq, Clone)]
 pub enum PskSecretError {
+    #[error(transparent)]
+    LibraryError(#[from] LibraryError),
```
franziskuskiefer, line 34: The errors in here appear to be in `WelcomeError` as well. Are they used somewhere else or could we re-use them?

openmls/src/group/core_group/new_from_welcome.rs
```diff
@@ -182,12 +217,14 @@ impl CoreGroup {
         key_package: &KeyPackage,
         welcome_secrets: &[EncryptedGroupSecrets],
         backend: &impl OpenMlsCryptoProvider,
-    ) -> Result<Option<EncryptedGroupSecrets>, KeyPackageError> {
+    ) -> Option<EncryptedGroupSecrets> {
```
kkohbrok, line 220: I know we make our lives easier by getting rid of a `Result`, but at the same time we're making debugging harder, because we "silently" return `None` instead of surfacing an error when something goes wrong. I did this as well in a recent PR, but I thought about it and I don't think this is a good idea anymore.

openmls/src/treesync/diff.rs
```diff
@@ -356,14 +356,24 @@ impl<'a> TreeSyncDiff<'a> {
         ciphersuite: &Ciphersuite,
         mut path_secret: PathSecret,
         sender_index: LeafIndex,
-    ) -> Result<CommitSecret, TreeSyncDiffError> {
-        let subtree_path = self.diff.subtree_path(self.own_leaf_index, sender_index)?;
+    ) -> Result<CommitSecret, TreeSyncSetPathError> {
+        // We assume both nodes are in the tree
```
franziskuskiefer, line 360: And what happens if not?

kkohbrok, line 360: Why would we do that? I'd prefer it if we were principled here and turn it into a library error on the caller side if we have checked there.

franziskuskiefer (changes requested):
Thanks. Looks good, just a few nits and questions.

openmls/src/treesync/diff.rs
```diff
@@ -389,20 +399,31 @@ impl<'a> TreeSyncDiff<'a> {
 
     /// A helper function that filters the unmerged leaves of the given node
     /// from the given resolution.
+    ///
+    /// Returns a LibraryError when the ParentNode is not in the tree or
```
kkohbrok, line 403: Same as above, I'd prefer to not make these assumptions.

openmls/src/treesync/diff.rs
```diff
@@ -466,13 +487,21 @@ impl<'a> TreeSyncDiff<'a> {
     /// Helper function computing the resolution of a node with the given index.
     /// If an exclusion list is given, do not add the public keys of the leaves
     /// given in the list.
+    ///
+    /// Returns The list of HPKE public keys.
+    /// In case node_id is not in the tree a LibraryError is returned.
```
kkohbrok, line 492: And again. At this point, why not have an error that is either a `NotInTree` error or a `LibraryError`?

openmls/src/treesync/mod.rs
```diff
@@ -161,19 +165,27 @@ impl TreeSync {
...
-    ) -> Result<(Self, Option<CommitSecret>), TreeSyncError> {
+    ) -> Result<(Self, Option<CommitSecret>), TreeSyncFromNodesError> {
         let mut tree_sync =
             Self::from_nodes(backend, ciphersuite, node_options, key_package_bundle)?;
 
         // Get the leaf index of the sender.
-        let sender_index = tree_sync.leaf_index(sender_kpr)?;
+        let sender_index = tree_sync
+            .leaf_index(sender_kpr)
+            .map_err(|_| LibraryError::custom("Sender not in tree"))?;
```
kkohbrok, line 175: Why is this a `LibraryError`? Similar to some of the other comments, I don't think we should start making assumptions about inputs by throwing `LibraryErrors` if they are malformed..

openmls/src/treesync/mod.rs
```diff
@@ -210,13 +223,11 @@ impl TreeSync {
                             if let Some(private_key) = private_key.take() {
                                 own_index_option =
                                     Some(u32::try_from(node_index / 2).map_err(|_| {
-                                        TreeSyncError::from(LibraryError::custom(
-                                            "Own leaf is outside of the tree",
-                                        ))
+                                        LibraryError::custom("Own leaf is outside of the tree")
```
kkohbrok, line 226: If this throws an error, the leaf is not outside of the tree. It just means that the `usize` couldn't be turned into a `u32`.

kkohbrok (changes requested):
Looks good, thanks for digging your way through all the things that can go wrong! I really like the new approach of treating the errors much more explicitly.

I think my main concern is that we should be principled about when we return which errors. I noted this in the comments, but to summarize:
- If there is an error, we should throw an error rather than "hiding" it under an option.
- We shouldn't make assumptions about caller input and throw a meaningful error rather than just a `LibraryError` when the inputs are not valid. 

The second point actually allows this new approach in more explicit error handling to shine: The caller can't just bubble the error up, because it's a LibraryError. Instead, the caller has to explicitly turn, for example, a `NotInTreeError` into a `LibraryError`.

openmls/src/group/core_group/new_from_welcome.rs
franziskuskiefer: I think we discussed this a couple times now. There are two reasons not to return errors in functions like this.
- There's no point in returning an error if it's nothing the caller can act upon.
- The idiomatic way is to return an `Option` for lookup functions. The main reason for this is again the first one. The caller wanted to get a value. The internals of the mechanisms are irrelevant.

kkohbrok: Then at least we should add some logging in case something went wrong during the lookup, or maybe a `debug_assert`. I understand that `Option` makes sense for lookup functions. But as a caller, I'd expect `None` if the value isn't there, not if there is something else that happened that prevented the lookup from happening in the first place. To me, it hints to the fact that this is more than a lookup function.

kkohbrok: Or looking at it in another way: If the caller gets a `LibraryError`, they at least know there's nothing they can do about that and they can decide to handle it in some way. If the caller gets a `None`, they can decide to do something different. However, if there is a `None` in both cases, the caller has no way of knowing if something unexpected went wrong or if the value is not there due to something else that they may have done.

openmls/src/schedule/errors.rs
raphaelrobert: Yes, and they are mapped to the ones in `WelcomeError`. I want to avoid having nested errors in the public errors as much as possible and therefore exposed the variants explicitly.

raphaelrobert:
Thanks for the quick review! I filed #804 for later after discussing with @kkohbrok.

franziskuskiefer (approved):
Thanks. Let's get it in and keep iterating.

kkohbrok (approved):
Thanks for the discussion and addressing my concerns! Let's get it in!

Issue #804, opened 2022-02-14
Improve the error return types in ABinaryTree/TreeSync


raphaelrobert:
Currently, several functions in those modules assume correct input and only return a `LibraryError` otherwise. Having more fine-grained error results would improve robustness.

PR #805, opened 2022-02-14
Error cleanup 3

raphaelrobert:
This PR 

 - refactors the framing errors (gets rid of MlsPlaintextError in the process) 
 - gets rid of the old verification function in `CoreGroup` that was only used in tests
 - refactors `MlsGroup`-level errors
 - makes `StageCommitError` & `ExternalCommitValidationError` public (refactoring in TreeSync/TreeKem needed for that)
 - makes `UnverifiedMessageError` public, as return value from `MlsGroup::process_unverified_message()`
 - Refactors a further 20% of the total errors to use `thiserror`

### DIFF
--- a/openmls/src/binary_tree/array_representation/diff.rs
+++ b/openmls/src/binary_tree/array_representation/diff.rs
@@ -182,12 +182,12 @@ impl<'a, T: Clone + Debug> AbDiff<'a, T> {
 
     /// Returns references to the leaves of the diff in order from left to
     /// right. This function should not throw an error. However, it might throw
-    /// an LibraryError error if there is a bug in the implementation.
+    /// a [`LibraryError`] error if there is a bug in the implementation.
     pub(crate) fn leaves(&self) -> Result<Vec<NodeId>, LibraryError> {
         let mut leaf_references = Vec::new();
         for leaf_index in 0..self.leaf_count() {
             let node_index = to_node_index(leaf_index);
-            // The node reference must be valid, since it is a vlaid leaf
+            // The node reference must be valid, since it is a valid leaf
             let node_ref = NodeId::try_from_node_index(self, node_index)
                 .map_err(|_| LibraryError::custom("Expected a valid node reference"))?;
             leaf_references.push(node_ref);
--- a/openmls/src/binary_tree/array_representation/diff.rs
+++ b/openmls/src/binary_tree/array_representation/diff.rs
@@ -182,12 +182,14 @@ impl<'a, T: Clone + Debug> AbDiff<'a, T> {
 
     /// Returns references to the leaves of the diff in order from left to
     /// right. This function should not throw an error. However, it might throw
-    /// an [`ABinaryTreeDiffError::OutOfBounds`] error if there is a bug in the implementation.
-    pub(crate) fn leaves(&self) -> Result<Vec<NodeId>, ABinaryTreeDiffError> {
+    /// an LibraryError error if there is a bug in the implementation.
+    pub(crate) fn leaves(&self) -> Result<Vec<NodeId>, LibraryError> {
         let mut leaf_references = Vec::new();
         for leaf_index in 0..self.leaf_count() {
             let node_index = to_node_index(leaf_index);
-            let node_ref = NodeId::try_from_node_index(self, node_index)?;
+            // The node reference must be valid, since it is a vlaid leaf
+            let node_ref = NodeId::try_from_node_index(self, node_index)
+                .map_err(|_| LibraryError::custom("Expected a valid node reference"))?;
             leaf_references.push(node_ref);
         }
         Ok(leaf_references)
--- a/openmls/src/framing/errors.rs
+++ b/openmls/src/framing/errors.rs
@@ -6,94 +6,84 @@
 use crate::{
     credentials::CredentialError, error::LibraryError, tree::secret_tree::SecretTreeError,
 };
-use openmls_traits::types::CryptoError;
-use tls_codec::Error as TlsCodecError;
+use thiserror::Error;
 
-implement_error! {
-    pub enum MlsPlaintextError {
-        Simple {
-            NotAnApplicationMessage = "The MlsPlaintext message is not an application message.",
-            UnknownSender = "Sender is not part of the group",
-            InvalidSignature = "The MlsPlaintext signature is invalid",
-            InvalidMembershipTag = "The MlsPlaintext membership tag is invalid",
-        }
-        Complex {
-            LibraryError(LibraryError) = "A LibraryError occurred",
-            CodecError(TlsCodecError) = "TLS Codec error",
-            CredentialError(CredentialError) = "See [`CredentialError`](`crate::credentials::CredentialError`) for details.",
-            VerificationError(VerificationError) = "See [`VerificationError`](`VerificationError`) for details.",
-            CryptoError(CryptoError) = "See [`CryptoError`](openmls_traits::types::CryptoError) for details.",
-        }
-    }
+/// MlsCiphertext error
+#[derive(Error, Debug, PartialEq, Clone)]
+pub enum MlsCiphertextError {
+    #[error(transparent)]
+    LibraryError(#[from] LibraryError),
+    #[error("Couldn't find a ratcheting secret for the given sender and generation.")]
+    GenerationOutOfBound,
+    #[error("An error occurred while decrypting.")]
+    DecryptionError,
+    #[error("The WireFormat was MLSPlaintext.")]
+    WrongWireFormat,
+    #[error("The content is malformed.")]
+    MalformedContent,
+    #[error(transparent)]
+    SecretTreeError(#[from] SecretTreeError),
+    #[error(transparent)]
+    SenderError(#[from] SenderError),
 }
 
-implement_error! {
-    pub enum MlsCiphertextError {
-        Simple {
-            InvalidContentType = "The MlsCiphertext has an invalid content type.",
-            GenerationOutOfBound = "Couldn't find a ratcheting secret for the given sender and generation.",
-            EncryptionError = "An error occurred while encrypting.",
-            DecryptionError = "An error occurred while decrypting.",
-            WrongWireFormat = "The WireFormat was MLSPlaintext.",
-        }
-        Complex {
-            PlaintextError(MlsPlaintextError) = "MlsPlaintext error",
-            SecretTreeError(SecretTreeError) = "SecretTree error",
-            CodecError(TlsCodecError) = "TLS codec error",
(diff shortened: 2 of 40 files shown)


openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -182,12 +182,14 @@ impl<'a, T: Clone + Debug> AbDiff<'a, T> {
 
     /// Returns references to the leaves of the diff in order from left to
     /// right. This function should not throw an error. However, it might throw
-    /// an [`ABinaryTreeDiffError::OutOfBounds`] error if there is a bug in the implementation.
-    pub(crate) fn leaves(&self) -> Result<Vec<NodeId>, ABinaryTreeDiffError> {
+    /// an LibraryError error if there is a bug in the implementation.
```
kkohbrok, line 185:
```suggestion
    /// a [`LibraryError`] error if there is a bug in the implementation.
```

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -182,12 +182,14 @@ impl<'a, T: Clone + Debug> AbDiff<'a, T> {
...
+    pub(crate) fn leaves(&self) -> Result<Vec<NodeId>, LibraryError> {
         let mut leaf_references = Vec::new();
         for leaf_index in 0..self.leaf_count() {
             let node_index = to_node_index(leaf_index);
-            let node_ref = NodeId::try_from_node_index(self, node_index)?;
+            // The node reference must be valid, since it is a vlaid leaf
```
kkohbrok, line 190:
```suggestion
            // The node reference must be valid, since it is a valid leaf
```

openmls/src/framing/errors.rs
```diff
@@ -6,94 +6,84 @@
...
-            VerificationError(VerificationError) = "See [`VerificationError`](`VerificationError`) for details.",
-            CryptoError(CryptoError) = "See [`CryptoError`](openmls_traits::types::CryptoError) for details.",
-        }
-    }
+/// MlsCiphertext error
+#[derive(Error, Debug, PartialEq, Clone)]
+pub enum MlsCiphertextError {
+    #[error(transparent)]
+    LibraryError(#[from] LibraryError),
+    #[error("Couldn't find a ratcheting secret for the given sender and generation.")]
```
kkohbrok, line 16: From the consumer's standpoint, wouldn't it be more helpful to get something like "The given message couldn't be decrypted as it was too far out-of-order relative to previously received messages."?

openmls/src/group/core_group/apply_proposals.rs
```diff
@@ -151,7 +152,9 @@ impl CoreGroup {
         // Extract KeyPackages from proposals
         let mut invitation_list = Vec::new();
         for add_proposal in add_proposals {
-            let leaf_index = diff.add_leaf(add_proposal.key_package().clone(), backend.crypto())?;
+            let leaf_index = diff
+                .add_leaf(add_proposal.key_package().clone(), backend.crypto())
+                .map_err(|_| LibraryError::custom("Tree full: cannot add more members"))?;
```
kkohbrok, line 157: This is something that can actually happen if the group gets too large. Is there already a check somewhere else that would make this a `LibraryError`? (Same for the other use of `add_leaf()` in `create_commit`.

openmls/src/group/core_group/process.rs
```diff
@@ -93,25 +95,29 @@ impl CoreGroup {
...
         own_kpbs: &[KeyPackageBundle],
         backend: &impl OpenMlsCryptoProvider,
-    ) -> Result<ProcessedMessage, CoreGroupError> {
+    ) -> Result<ProcessedMessage, UnverifiedMessageError> {
         // Add the context to the message and verify the membership tag if necessary.
         // If the message is older than the current epoch, we need to fetch the correct secret tree first.
-        let message_secrets = self.message_secrets_mut(unverified_message.epoch())?;
+        let message_secrets = self
+            .message_secrets_mut(unverified_message.epoch())
+            .map_err(|_| UnverifiedMessageError::NoPastEpochData)?;
```
kkohbrok, line 103: Don't we have these per-function error types such that we can actually map on the thrown error, rather than do "blind" `map_err` to a fixed error? The benefit of actually matching on the error would be that, should we refactor the function in the future, the type checker will make sure that callers properly handle any new errors. (Same for the other `map_err`s below.)

Or is this just a stop-gap, because the underlying function doesn't have a per-function error yet?

openmls/src/group/mls_group/errors.rs
```diff
@@ -6,102 +6,97 @@
...
-    pub enum EmptyInputError {
-        AddMembers = "An empty list of KeyPackages was provided.",
-        RemoveMembers = "An empty list of indexes was provided.",
-    }
+/// EmptyInput error
+#[derive(Error, Debug, PartialEq, Clone)]
+pub enum EmptyInputError {
+    #[error("An empty list of KeyPackages was provided.")]
+    AddMembers,
+    #[error("An empty list of indexes was provided.")]
```
kkohbrok, line 74: While you're at it ...
```suggestion
    #[error("An empty list of KeyPackage references was provided.")]
```

openmls/src/treesync/diff.rs
```diff
@@ -171,29 +178,39 @@ impl<'a> TreeSyncDiff<'a> {
...
+        let node = Node::LeafNode(LeafNode::new(leaf_node, backend)?);
         // Find a free leaf and fill it with the new key package.
         let leaf_index = self.free_leaf_index()?;
         // If the free leaf index is within the tree, put the new leaf there,
         // otherwise extend the tree.
         if leaf_index < self.leaf_count() {
-            self.diff.replace_leaf(leaf_index, node.into())?;
+            self.diff
+                .replace_leaf(leaf_index, node.into())
+                .map_err(|_| LibraryError::custom("no parent hash extension"))?;
```
kkohbrok, line 190: I'm confused. Why would an error thrown by `replace_leaf` indicate that a parent hash extension is missing?

Unrelared nit: In most other cases, you added a nice note on why we're mapping to a `LibraryError`. Would you mind adding that here as well?

openmls/src/treesync/diff.rs
```diff
@@ -290,31 +307,27 @@ impl<'a> TreeSyncDiff<'a> {
...
-                TreeSyncDiffError::CryptoError(e)
-            } else {
-                TreeSyncDiffError::LibraryError(LibraryError::custom("key package error"))
-            }
-        })?);
-        self.diff.replace_leaf(sender_leaf_index, node.into())?;
+        let node = Node::LeafNode(LeafNode::new(key_package, backend.crypto())?);
+        self.diff
+            .replace_leaf(sender_leaf_index, node.into())
+            .map_err(|_| LibraryError::custom("Expected sender leaf to be in the tree"))?;
```
kkohbrok, line 330: Could you add this as assumption on the input to the function docs? Just so the person working on #804 has it on their radar.

openmls/src/treesync/diff.rs
```diff
@@ -340,7 +353,9 @@ impl<'a> TreeSyncDiff<'a> {
 
         // Set the direct path. Note, that the nodes here don't have a tree hash
         // set.
-        self.diff.set_direct_path(leaf_index, direct_path)?;
+        self.diff
+            .set_direct_path(leaf_index, direct_path)
+            .map_err(|_| LibraryError::custom("Expected the leaf index to be in the tree"))?;
```
kkohbrok, line 358: Same here (function docs for #804).

openmls/src/treesync/treekem.rs
```diff
@@ -79,19 +78,24 @@ impl<'a> TreeSyncDiff<'a> {
...
             .get(path_position)
-            .ok_or(TreeKemError::UpdatePathNodeNotFound)?;
+            // We know the update path has the right length through validation, therefore there must be an element at this position
+            .ok_or_else(|| LibraryError::custom("Expected to find ciphertext in update path"))?;
 
-        let (decryption_key, resolution_position) =
-            self.decryption_key(params.sender_leaf_index, params.exclusion_list)?;
+        let (decryption_key, resolution_position) = self
+            .decryption_key(params.sender_leaf_index, params.exclusion_list)
+            .map_err(|_| LibraryError::custom("Expected sender to be in the tree"))?;
```
kkohbrok, line 94: Same here (function docs for #804).

openmls/src/treesync/treekem.rs
```diff
@@ -79,19 +78,24 @@ impl<'a> TreeSyncDiff<'a> {
...
+        let path_position = self
+            .subtree_root_position(params.sender_leaf_index, self.own_leaf_index())
+            // We know our own leaf is in the tree
+            .map_err(|_| LibraryError::custom("Expected own leaf to be in the tree"))?;
         let update_path_node = params
             .update_path
             .get(path_position)
-            .ok_or(TreeKemError::UpdatePathNodeNotFound)?;
+            // We know the update path has the right length through validation, therefore there must be an element at this position
+            .ok_or_else(|| LibraryError::custom("Expected to find ciphertext in update path"))?;
```
kkohbrok, line 90: Same here (function docs for #804).

openmls/src/treesync/treekem.rs
```diff
@@ -79,19 +78,24 @@ impl<'a> TreeSyncDiff<'a> {
...
-        let (decryption_key, resolution_position) =
-            self.decryption_key(params.sender_leaf_index, params.exclusion_list)?;
+        let (decryption_key, resolution_position) = self
+            .decryption_key(params.sender_leaf_index, params.exclusion_list)
+            .map_err(|_| LibraryError::custom("Expected sender to be in the tree"))?;
         let ciphertext = update_path_node
             .encrypted_path_secrets(resolution_position)
-            .ok_or(TreeKemError::EncryptedCiphertextNotFound)?;
+            // We know the update path has the right length through validation, therefore there must be a ciphertext at this position
+            .ok_or_else(|| LibraryError::custom("Expected to find ciphertext in update path"))?;
```
kkohbrok, line 98: Same here (function docs for #804).

kkohbrok (changes requested):
Thanks for another round of clean-up! Nice to see more functions with a `Result<T, LibraryError>`.

A few nits, mostly requests to add documentation on input assumptions for #804.

openmls/src/framing/errors.rs
raphaelrobert: This is not a public error (despite what it says). I'll look at that if and when I expose it.

openmls/src/treesync/diff.rs
raphaelrobert: I added TODO #804 in all these places

kkohbrok: 👍🏻 thanks!

PR #811, opened 2022-02-15
ciphersuite cleanup

### LINKED ISSUE #701: "Supported ciphersuite" is unclear

franziskuskiefer:
A ciphersuite cleanup removing a lot of unnecessary code.
The main change is that `Ciphersuite` is a simple enum now instead of a struct that needs to be passed around as static reference.

- remove config
- moving `Ciphersuite` to `traits::types`
- delete unnecessary code
- align ciphersuites with spec

Commits will be squashed before merging.

Fixes #701


openmls/src/group/core_group/new_from_external_init.rs
```diff
@@ -31,9 +31,8 @@ impl CoreGroup {
         tree_option: Option<&[Option<Node>]>,
         verifiable_public_group_state: VerifiablePublicGroupState,
     ) -> Result<ExternalCommitResult, ExternalCommitError> {
-        let ciphersuite = Config::ciphersuite(verifiable_public_group_state.ciphersuite())
-            .map_err(|_| ExternalCommitError::UnsupportedCiphersuite)?;
-        if !Config::supported_versions().contains(&verifiable_public_group_state.version()) {
+        let ciphersuite = verifiable_public_group_state.ciphersuite();
+        if verifiable_public_group_state.version() != ProtocolVersion::Mls10 {
```
kkohbrok, line 35: Does it make sense to even do the check here? If I understand correctly, the check is meant to determine if we support the version in the implementation in general rather than this being configurable in some way. But isn't the fact that we were able to deserialize the `verifiable_public_group_state` in the first place an indicator that we do support it? Probably not something we want to solve here, but if my understanding is correct, we might want to file this as a follow-up. 

(The same goes for other places where we check some variable against `ProtocolVersion::Mls10`.)

openmls/src/group/core_group/test_core_group.rs
```diff
@@ -70,122 +68,112 @@ pub fn flip_last_byte(ctxt: &mut HpkeCiphertext) {
 
 #[apply(ciphersuites_and_backends)]
 fn test_failed_groupinfo_decryption(
-    ciphersuite: &'static Ciphersuite,
+    ciphersuite: Ciphersuite,
     backend: &impl OpenMlsCryptoProvider,
 ) {
-    for version in Config::supported_versions() {
```
kkohbrok, line 76: This would then be something we'd do for all elements of the `ProtocolVersions` enum, right? Although we might want to do that with a macro in the future.

openmls/src/lib.rs
```diff
@@ -20,7 +20,7 @@
 //! [MLS RFC]: https://datatracker.ietf.org/doc/draft-ietf-mls-protocol/
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![cfg_attr(not(test), forbid(unsafe_code))]
-#![cfg_attr(not(feature = "test-utils"), warn(missing_docs))]
+// #![cfg_attr(not(feature = "test-utils"), warn(missing_docs))]
```
kkohbrok, line 23: This should probably be commented in again.

kkohbrok (changes requested):
Thanks for taking care of this! Nice to see the code get a little more trim 🙂. Some nits and questions, but otherwise this looks good.

interop_client/src/main.rs
```diff
@@ -92,10 +92,12 @@ fn to_ciphersuite(cs: u32) -> Result<&'static Ciphersuite, Status> {
...
     };
-    match Config::supported_ciphersuites()
-        .iter()
-        .find(|cs| cs.name() == cs_name)
-    {
+    let ciphersuites = &[
+        Ciphersuite::MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519,
+        Ciphersuite::MLS_128_DHKEMP256_AES128GCM_SHA256_P256,
+        Ciphersuite::MLS_128_DHKEMX25519_CHACHA20POLY1305_SHA256_Ed25519,
+    ];
```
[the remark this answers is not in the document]
franziskuskiefer: Because we don't have one here and I didn't want to change more than necessary. We should do it though when working on the interop client again.

openmls/src/ciphersuite/mod.rs
```diff
@@ -49,140 +43,9 @@ use self::signable::SignedStruct;
...
-    /// Build an [`HpkeConfi`] for this cipher suite.
-    pub(crate) fn hpke_config(&self) -> HpkeConfig {
-        HpkeConfig(self.hpke_kem, self.hpke_kdf, self.hpke_aead)
-    }
-}
-
 /// Compare two byte slices in a way that's hopefully not optimised out by the
 /// compiler.
-#[inline(always)]
+#[inline(never)]
```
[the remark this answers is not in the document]
franziskuskiefer: Yes. The function is best effort and there's a good chance it's being optimized away. This is less likely if it's not being inlined.

openmls/src/ciphersuite/secret.rs
```diff
@@ -4,19 +4,17 @@ use super::*;
...
     pub(in crate::ciphersuite) mls_version: ProtocolVersion,
 }
 
-implement_persistence!(Secret, value, mls_version);
-
 impl Default for Secret {
     fn default() -> Self {
         Self {
-            ciphersuite: Ciphersuite::default(),
+            ciphersuite: Ciphersuite::MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519,
```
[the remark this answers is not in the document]
franziskuskiefer: We can but I'm not sure that we want to. The places where we need one it's only a placeholder and any value should work. (And if not something is broken, so this is sort off a test :D)

openmls/src/extensions/life_time_extension.rs
```diff
@@ -1,11 +1,19 @@
...
 use std::time::{SystemTime, UNIX_EPOCH};
 
+/// This value is used as the default lifetime of `KeyPackage`s if no default
+/// lifetime is configured. The value is in seconds and amounts to 3 * 28 Days,
+/// i.e. about 3 months.
+const DEFAULT_KEY_PACKAGE_LIFETIME: u64 = 60 * 60 * 24 * 28 * 3; // in Seconds
+/// This value is used as the default amount of time (in seconds) the lifetime
+/// of a `KeyPackage` is extended into the past to allow for skewed clocks. The
+/// value is in seconds and amounts to 1h.
+const DEFAULT_KEY_PACKAGE_LIFETIME_MARGIN: u64 = 60 * 60; // in Seconds
```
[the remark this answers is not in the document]
franziskuskiefer: If this is something that's interesting for a consumer, it should be public and be part of the docs. And then it doesn't matter where it lives. I'd keep it here unless we'll have more things that we want to put in a file.

openmls/src/group/core_group/new_from_external_init.rs
franziskuskiefer: We have one other version we support in the enum but don't actually support in the code. So keeping the check makes sense I think. And it's a canary when starting to work on actually adding support for other versions.

openmls/src/group/core_group/test_core_group.rs
franziskuskiefer: Yes, I made it so that the code actually only supports MLS1.0 because this was the only place we actually support something else. When we start adding support for other versions we should indeed again test all of them, but with the macro and everywhere.

openmls/src/lib.rs
franziskuskiefer: Yes. I was interested in actual warnings ...

openmls/src/group/core_group/new_from_external_init.rs
kkohbrok: Then we should probably remove that version from the enum. But that's nothing we need to do in this PR.

kkohbrok (approved):
Thanks for addressing my concerns! Good to land from my side.

raphaelrobert (approved):
Thanks for the PR! I think this simplifies things going forward. It still doesn't allow for custom ciphersuites, but probably makes it easier to introduce them if we want to do that in the future.

kkohbrok (approved):

raphaelrobert (approved):

Issue #812, opened 2022-02-15
Better checks/error types when adding/removing members


raphaelrobert:
- [ ] `add_members()`: Early add proposal validation would be nice (instead of the validation that happens anyway as part of the commit creation
 - [ ] `remove_members()`: Check the members are actually part of the group, early remove proposal validation
 - [ ] `propose_add_member()`: Do full validation right away (right now only required extensions are checked)
 - [ ] `propose_remove_member()`: Do full validation right away

PR #813, opened 2022-02-15
Error cleanup 4

raphaelrobert:
This PR

 - Introduces `CreateCommitError`
 - Introduces individual error types for the membership functions in `MlsGroup`
 - Does some refactoring of the underlying layers
 - Discovered improvements: #812
 
Also re-exports `Node` in `MlsGroup` because of #722 (slightly out-of-scope, but has to do with the public API of `MlsGroup`).


openmls/src/framing/errors.rs
```diff
@@ -27,6 +29,19 @@ pub enum MlsCiphertextError {
     SenderError(#[from] SenderError),
 }
 
+/// Message encryption error
+#[derive(Error, Debug, PartialEq, Clone)]
+pub(crate) enum MessageEncryptionError {
+    #[error(transparent)]
+    LibraryError(#[from] LibraryError),
+    #[error("The WireFormat was MLSPlaintext.")]
```
franziskuskiefer, line 37: And what should it have been?

openmls/src/framing/validation.rs
```diff
@@ -81,7 +81,7 @@ impl DecryptedMessage {
             let ciphersuite = group.ciphersuite();
             let (message_secrets, old_leaves) = group
                 .message_secrets_and_leaves_mut(ciphertext.epoch())
-                .map_err(|_| MlsCiphertextError::DecryptionError)?;
+                .map_err(|_| MessageDecryptionError::DecryptionError)?;
```
franziskuskiefer, line 84: `MessageDecryptionError::DecryptionError` seems a little odd. I suppose all errors in that enum are "DecryptionErrors".

openmls/src/group/errors.rs
```diff
@@ -185,17 +187,35 @@ pub enum StageCommitError {
...
     #[error(transparent)]
     LibraryError(#[from] LibraryError),
+    #[error("Missing own key to apply proposal.")]
+    OwnKeyNotFound,
     #[error("The Commit tried to remove self from the group. This is not possible.")]
     CannotRemoveSelf,
+    #[error("The proposal queue is missing a proposal for the commit.")]
+    MissingProposal,
+    #[error("A proposal has the wrong sender type.")]
+    WrongProposalSender,
```
franziskuskiefer, line 202: The name indicates that something about the sender is wrong but the comment says sender type. That's confusing. The error could be read as the sender is not the expected sender or something else. Maybe `WrongProposalSenderType`?

openmls/src/group/core_group/create_commit.rs
```diff
@@ -121,18 +133,22 @@ impl CoreGroup {
         }
 
         // Apply proposals to tree
-        let apply_proposals_values =
-            self.apply_proposals(&mut diff, backend, &proposal_queue, &[])?;
+        let apply_proposals_values = self
+            .apply_proposals(&mut diff, backend, &proposal_queue, &[])
+            .map_err(|_| CreateCommitError::OwnKeyNotFound)?;
```
franziskuskiefer, line 138: Is this a legitimate case that can happen or is this a library error where there's a bug in OpenMLS? I think this should either be a library error or get more context in the error description what happened.

openmls/src/group/core_group/create_commit.rs
```diff
@@ -402,19 +448,23 @@ impl CoreGroup {
...
                 vec![],
-            )?;
+            )
+            .map_err(|_| LibraryError::custom("Unexpected KeyPackage error"))?;
 
             diff.add_leaf(key_package_bundle.key_package().clone(), backend.crypto())
                 .map_err(|_| LibraryError::custom("Tree full: cannot add more members"))?;
-            diff.own_leaf()?.key_package()
+            diff.own_leaf()
+                .map_err(|_| LibraryError::custom("Expecte own leaf"))?
```
franziskuskiefer, line 457:
```suggestion
                .map_err(|_| LibraryError::custom("Expected own leaf"))?
```

openmls/src/group/core_group/create_commit.rs
```diff
@@ -402,19 +448,23 @@ impl CoreGroup {
...
+                .key_package()
         } else {
-            self.treesync().own_leaf_node()?.key_package()
+            self.treesync()
+                .own_leaf_node()
+                .map_err(|_| LibraryError::custom("Expecte own leaf"))?
```
franziskuskiefer, line 462:
```suggestion
                .map_err(|_| LibraryError::custom("Expected own leaf"))?
```

openmls/src/group/mls_group/errors.rs
```diff
@@ -78,10 +66,14 @@ pub enum EmptyInputError {
...
 #[derive(Error, Debug, PartialEq, Clone)]
 pub enum MlsGroupStateError {
+    #[error(transparent)]
+    LibraryError(#[from] LibraryError),
     #[error("Tried to use a group after being evicted from it.")]
     UseAfterEviction,
     #[error("Can't create message because a pending proposal exists.")]
     PendingProposal,
+    #[error("Can't execute operation because a pending commit exists.")]
+    PendingCommit,
```
franziskuskiefer, line 76: Is it always clear which operation didn't succeed or do we need to put the operation in here?

openmls/src/group/mls_group/errors.rs
```diff
@@ -100,3 +92,79 @@ pub enum UnverifiedMessageError {
...
+}
+
+/// Propose add members error
+#[derive(Error, Debug, PartialEq, Clone)]
+pub enum ProposeAddMemberError {
+    #[error(transparent)]
+    LibraryError(#[from] LibraryError),
+    #[error("The own CredentialBundle could not be found in the key store.")]
+    NoMatchingCredentialBundle,
+    #[error("The new member does not support all required extensions.")]
```
franziskuskiefer, line 129: Here it could be interesting to know which extension type the offender is.

openmls/src/group/mls_group/errors.rs
```diff
@@ -100,3 +92,79 @@ pub enum UnverifiedMessageError {
...
+    GroupStateError(#[from] MlsGroupStateError),
+}
+
+/// Leave group error
+#[derive(Error, Debug, PartialEq, Clone)]
+pub enum LeaveGroupError {
+    #[error(transparent)]
+    LibraryError(#[from] LibraryError),
+    #[error("The own CredentialBundle could not be found in the key store.")]
+    NoMatchingCredentialBundle,
```
franziskuskiefer, line 167: Duplicating this error in all of these enums makes me think there should be a higher level error that can combine these.

openmls/src/group/mls_group/membership.rs
```diff
@@ -245,18 +286,16 @@ impl MlsGroup {
...
-            .group
-            .treesync()
-            .full_leaves()?
-            .iter()
-            .map(|(_, &kp)| kp)
-            .collect())
+    pub fn members(&self) -> Vec<&KeyPackage> {
+        match self.group.treesync().full_leaves() {
+            Ok(leaves) => leaves.iter().map(|(_, &kp)| kp).collect(),
+            // This should not happen, but this way we avoid returnung a library error
```
franziskuskiefer, line 296:
```suggestion
            // This should not happen, but this way we avoid returning a library error
```

openmls/src/treesync/treekem.rs
```diff
@@ -211,36 +209,46 @@ impl PlaintextSecret {
...
+    ) -> Result<Vec<Self>, LibraryError> {
         let mut plaintext_secrets = vec![];
         for (leaf_index, add_proposal) in invited_members {
             let key_package = add_proposal.key_package;
 
-            let direct_path_position =
-                diff.subtree_root_position(diff.own_leaf_index(), leaf_index)?;
+            let direct_path_position = diff
+                .subtree_root_position(diff.own_leaf_index(), leaf_index)
+                // This can only fail if the nodes are outside theree or identical
```
franziskuskiefer, line 231:
```suggestion
                // This can only fail if the nodes are outside the tree or identical
```

franziskuskiefer (changes requested):
Thanks. A couple nits and questions

openmls/src/group/mls_group/application.rs
```diff
@@ -15,31 +15,43 @@ impl MlsGroup {
...
+            return Err(CreateMessageError::GroupStateError(
                 MlsGroupStateError::PendingProposal,
             ));
         }
 
-        let credential = self.credential()?;
+        let credential = self
+            .credential()
+            // We checked we are in the right group state before
+            .map_err(|_| LibraryError::custom("Wrong group state"))?;
```
kkohbrok, line 33: How does failing to fetch a credential indicate a wrong group state? That's not quite clear to me.

openmls/src/group/mls_group/membership.rs
```diff
@@ -245,18 +286,16 @@ impl MlsGroup {
...
+            Err(_) => Vec::new(),
```
kkohbrok, line 297: If you really want to hide the library error here, can we add least trigger a debug log entry?

openmls/src/treesync/treekem.rs
```diff
@@ -44,13 +44,11 @@ impl<'a> TreeSyncDiff<'a> {
...
         key_package: KeyPackage,
-    ) -> Result<UpdatePath, TreeKemError> {
+    ) -> Result<UpdatePath, LibraryError> {
         let copath_resolutions = self.copath_resolutions(self.own_leaf_index(), exclusion_list)?;
 
         // There should be as many copath resolutions.
-        if copath_resolutions.len() != path.len() {
-            return Err(TreeKemError::PathLengthError);
-        }
+        debug_assert_eq!(copath_resolutions.len(), path.len());
```
kkohbrok, line 51: Sorry to be pedantic about this, but can you add a doc comment saying that this function assumes that the input `path` has the same length as the copath resolution of the own leaf?

openmls/src/treesync/treekem.rs
```diff
@@ -211,36 +209,46 @@ impl PlaintextSecret {
     /// Prepare the `GroupSecrets` for a number of `invited_members` based on a
     /// [`TreeSyncDiff`]. If a slice of [`PlainUpdatePathNode`] is given, they
     /// are included in the [`GroupSecrets`] of the path.
+    ///
+    /// Returns an error if
```
kkohbrok, line 213: Shouldn't the points on the list below be assumptions on the inputs rather than errors? In other cases, we don't document the cases where functions throw library errors.

openmls/src/treesync/treekem.rs
```diff
@@ -211,36 +209,46 @@ impl PlaintextSecret {
...
             // If a plain path was given, there have to be secrets for every new member.
             let path_secret_option = if let Some(plain_path) = plain_path_option {
                 Some(
                     plain_path
                         .get(direct_path_position)
                         .map(|pupn| pupn.path_secret())
-                        .ok_or(TreeKemError::PathSecretNotFound)?,
+                        .ok_or(TreeKemError::PathSecretNotFound)
+                        // This only fails if the supplied plain path is invalid
+                        .map_err(|_| LibraryError::custom("Invalid plain path"))?,
```
kkohbrok, line 242: I think that one's missing from the list in the doc string.

kkohbrok (changes requested):
Thanks for the PR! Just a few nits, mostly about documentation of assumptions in the function docs.

openmls/src/group/mls_group/errors.rs
raphaelrobert: A number of operations can fail when there is a pending commit, like adding & removing members. This is an immediate result of the call, so it should be clear from the context.

openmls/src/group/mls_group/errors.rs
raphaelrobert: Agreed, but there is much more missing and I filed #812 for that.

openmls/src/group/mls_group/errors.rs
raphaelrobert: This appears as a result of all operations that require signing. Since we try to flatten the errors, I think it makes sense to repeat it in every enum. If it was more than one variant, I agree that we should instead embed a generic error instead.

openmls/src/group/mls_group/application.rs
raphaelrobert: The own credential is removed after eviction, that's why you don't have access to it anymore:

```rust
    pub fn credential(&self) -> Result<&Credential, MlsGroupStateError> {
        if !self.is_active() {
            return Err(MlsGroupStateError::UseAfterEviction);
        }
        let tree = self.group.treesync();
        Ok(tree
            .own_leaf_node()
            .map_err(|_| LibraryError::custom("Own leaf node missing"))?
            .key_package()
            .credential())
    }
 ```

openmls/src/treesync/treekem.rs
raphaelrobert: I'm not sure I understand. This is the list of assumptions. We have this disclaimer on many (most?) functions now.

kkohbrok: Never mind. I thought it makes sense here between assumptions (breaking which results in a `LibraryError` and actual error cases, but since we don't really have those any more, it doesn't make much of a difference.

kkohbrok (approved):
Thanks! Let's get this in, then.

franziskuskiefer (approved):
Thanks

PR #814, opened 2022-02-16
 Add Proposal Validation Tests for ValSem104 and ValSem105

sgmenda:
Ref #525


openmls/src/group/tests/test_proposal_validation.rs
```diff
@@ -341,3 +341,187 @@ fn test_valsem103(ciphersuite: &'static Ciphersuite, backend: &impl OpenMlsCrypt
...
+fn test_valsem104(ciphersuite: &'static Ciphersuite, backend: &impl OpenMlsCryptoProvider) {
+    for alice_and_bob_share_keys in [
+        KeyUniqueness::NegativeSameKey,
+        KeyUniqueness::PositiveDifferentKey,
+    ] {
+        // 0. Initialize Alice and Bob
+        let alice_signature_keypair: SignatureKeypair;
+        let bob_signature_keypair: SignatureKeypair;
+
+        match alice_and_bob_share_keys {
```
kkohbrok, line 359: This is fine, but I think the slightly "rustier" way of doing it would be to have `match` block yield the tuple of keys. Something like:
```rust
let (alice_signature_keypair, bob_signature_keypair) = match alice_and_bob_share_keys { ... };
```

kkohbrok (dismissed):
Looks good, thanks for adding the tests! One proposal for style points, but otherwise this is good to go.

raphaelrobert:
> I did not dismiss @kkohbrok 's review, I am not sure why Github is suggesting that I did...

You pushed some changes, that's why the approval is no longer valid. Reviewers need to review your changes and approve them.

raphaelrobert (approved):
Thanks! Since @kkohbrok approved earlier and the diff looks good too, let's get this in!

PR #822, opened 2022-02-16
Proposal validation tests (receiver side) for ValSem100-103

kkohbrok:
As suggested by the title, this PR adds tests for ValSem100-ValSem103.

It also adds some test/test-utils only functions to facilitate the tests.

### DIFF
--- a/openmls/src/group/tests/test_proposal_validation.rs
+++ b/openmls/src/group/tests/test_proposal_validation.rs
@@ -275,8 +275,9 @@ fn test_valsem100(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider
         }
     }
 
-    // Before we can test reception of (invalid) proposals, we set up a new
-    // group with Alice and Bob.
+    // We now test if ValSem100 is also performed when a client receives a
+    // commit.  Before we can test reception of (invalid) proposals, we set up a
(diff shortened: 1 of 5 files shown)


openmls/src/group/tests/test_proposal_validation.rs
```diff
@@ -204,7 +423,84 @@ fn test_valsem101(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider
         }
     }
 
-    // TODO #525: Add test for incoming proposals.
+    // Before we can test reception of (invalid) proposals, we set up a new
```
franziskuskiefer, line 426: Can you describe first on a high level what you're testing here? Same for the other tests.

franziskuskiefer (changes requested):
Looks good, just a little more high level description for the tests would be nice.

openmls/src/group/tests/test_proposal_validation.rs
kkohbrok: I added a remark to each test, indicating that we're now testing the same validation logic on the receiver side. Is that what you meant?

franziskuskiefer (approved):
Thanks, yeah it tells the reader what's going on.

PR #826, opened 2022-02-17
Error cleanup 6

raphaelrobert:
This PR

 - Refactors `KeyPackage` related errors
 - Minor error refactoring
 - Adds more documentation
 - Eliminates all warnings from `cargo doc`

### DIFF
--- a/openmls/src/error.rs
+++ b/openmls/src/error.rs
@@ -74,8 +74,10 @@ impl Display for LibraryError {
 /// Internal enum to differentiate between the different types of library errors
 #[derive(Error, Debug, PartialEq, Clone)]
 enum InternalLibraryError {
+    /// See [`TlsCodecError`] for more details.
     #[error(transparent)]
     MissingBoundsCheck(#[from] TlsCodecError),
+    /// See [`CryptoError`] for more details.
     #[error(transparent)]
     CryptoError(#[from] CryptoError),
     #[error("Custom library error: {0}")]
@@ -85,6 +87,7 @@ enum InternalLibraryError {
 /// OpenMLS Error
 #[derive(Error, Debug, PartialEq, Clone)]
 pub enum OpenMlsError {
+    /// Unsupported MLS version.
     #[error("Unsupported MLS version.")]
     UnsupportedMlsVersion,
 }
--- a/openmls/src/error.rs
+++ b/openmls/src/error.rs
@@ -18,7 +18,6 @@ use tls_codec::Error as TlsCodecError;
 /// This error is returned when the library tries to serialize data that is too big for the
 /// MLS structs. In particular, when element lists contain more elements than the theoretical maximum
 /// defined in the spec, the serialization will fail. This should not happen when all input values are checked.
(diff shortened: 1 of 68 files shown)


openmls/src/error.rs
```diff
@@ -121,3 +112,29 @@ impl ErrorString {
...
+/*
+
+Note to maintainers
+
+The thiserror crate does not add any documentation to enum variants and this needs to be done manually.
+The best way to do this if you don't want to duplicate the string manually is to use the following regex:
+
+Add comments for naked variants:
+
+ ([,|{])\n(\s+)#\[error\("([a-z0-9 .,-_'^:]+)"\)\]
```
kkohbrok, line 125: This is certainly useful for the future. But can you add some indication how you applied this? Bash script? Python? The regex is certainly useful, but it would be more useful if I know how to interpret the $1..3, etc. below.

kkohbrok: Also, the wiki would probably be a better place for this.

kkohbrok (commented):
Looks good, thanks! Just one question regarding the removing of brackets in function docs.

openmls/src/error.rs
raphaelrobert: Re usage: I guess that depends on your favourite editor. In VS Code you can just copy & paste this.

Re wiki: Not sure what will happen to the wiki going forward. After we clean it up, not much will be left there. I thought it was safer here.

openmls/src/framing/validation.rs
```diff
@@ -214,7 +214,7 @@ impl DecryptedMessage {
 
 /// Partially checked and potentially decrypted message.
 /// Use this to inspect the [Credential] of the message sender
-/// and the optional `aad` if the original message was an [MlsCiphertext].
+/// and the optional `aad` if the original message was an MlsCiphertext.
```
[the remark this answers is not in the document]
raphaelrobert: A lot of things became `pub(crate)` recently and therefore it cannot be exposed in the public documentation anymore. That's also the case for `MlsCiphertext`.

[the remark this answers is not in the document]
kkohbrok: Right, then referring to MlsCiphertext makes sense in this particular case, because it's a struct from the MLS spec. I don't think that's the case in some of the other places. There, we're referring to structs that are not even visible to the reader. For example in line 264 in this file, we're referring to VerifiableMlsPlaintext. My proposal would be to remove those those references entirely rather than just remove the doc link.

openmls/src/error.rs
kkohbrok: Re usage: Then it would be useful to add the information that this is VS Code specific.

Re wiki: Fair point. Leave it here for now, then.

kkohbrok (changes requested):
Looks good to merge. Last request: Please add a note to the regex instructions indicating that they are VSCode specific. (And maybe how to use them with VSCode.)

kkohbrok (approved):
Thanks!

PR #829, opened 2022-02-18
clean up rustdoc front page

franziskuskiefer:
We could move this to the readme as well to have better integration.
But that should be done as part of #796


kkohbrok (approved):
Looks good, thanks!

raphaelrobert (changes requested):
This conflicts with #826, we should sync on that

PR #834, opened 2022-02-21
make key_store module private

franziskuskiefer:


raphaelrobert:
I'm wondering if this is a good idea since it precludes an application from managing items in the key store. For example, an application might want to delete expired `KeyPackageBundle`s from the store, and for that this module would be required.

kkohbrok:
Can't they do that via the backend? For example, if you instantiate the `RustCrypto` backend, it proves the full `OpenMlsCryptoProvider` using the `MemoryKeyStore` as key store backend. The application can then access the key store via `backend.key_store()`.

raphaelrobert (approved):

PR #838, opened 2022-02-22
clean up MlsGroup config and harder rustdoc requirements

franziskuskiefer:
This PR
- makes documentation of public items mandatory
- prohibits using private items in public documentation
- adds documentation for the MlsGroup config

### DIFF
--- a/openmls/src/lib.rs
+++ b/openmls/src/lib.rs
@@ -181,7 +181,9 @@
 //! [user Manual]: https://openmls.tech/book
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![cfg_attr(not(test), forbid(unsafe_code))]
-#![cfg_attr(not(feature = "test-utils"), warn(missing_docs))]
+#![cfg_attr(not(feature = "test-utils"), deny(missing_docs))]
+#![deny(rustdoc::broken_intra_doc_links)]
+#![deny(rustdoc::private_intra_doc_links)]
 
 // === Testing ===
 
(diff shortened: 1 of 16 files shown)


openmls/src/lib.rs
```diff
@@ -221,5 +223,8 @@ mod key_store;
 mod tree;
 mod treesync;
 
+// Re-export items from private modules.
+pub use treesync::errors::{ApplyUpdatePathError, PublicTreeError};
```
raphaelrobert, line 227: I think this should be re-exported in `group::mls_group::errors` instead. We already re-export other errors there, and I think it makes more sense to do it in the context in which they are needed.

raphaelrobert (changes requested):
Thanks for tackling this! Just one nit

kkohbrok (approved):
Looks good, thanks! Nice to get rid of that extra dependency 🙂.

openmls/src/lib.rs
raphaelrobert: You moved it to `error.rs` instead of `group::mls_group::errors.rs` where the other re-exports are. I think it would be good to have all of the sub-types there:

https://github.com/openmls/openmls/blob/main/openmls/src/group/mls_group/errors.rs#L9-L13

franziskuskiefer: The treesync module is not part of the mls_group module. So that doesn't really make sense.

raphaelrobert: Let's decide on a strategy later, since this is a bigger issue in other places as well

raphaelrobert (approved):

PR #839, opened 2022-02-22
Fix schedule rustdocs

### LINKED ISSUE #775: schedule rustdocs

raphaelrobert:
Fixes #775.

This PR also re-adjusts the exporting part of the `MlsGroup` API, so that `AuthenticationSecret` and `ResumptionSecret` are aligned.

### DIFF
--- a/openmls/src/schedule/mod.rs
+++ b/openmls/src/schedule/mod.rs
@@ -166,7 +166,7 @@ pub(crate) use psk::*;
 
 // Public types
 
-/// A group secret that can be used among members to prove that a member was part of a group.
+/// A group secret that can be used among members to prove that a member was part of a group in a given epoch.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[cfg_attr(test, derive(PartialEq))]
 pub struct ResumptionSecret {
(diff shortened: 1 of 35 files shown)


openmls/src/schedule/mod.rs
```diff
@@ -152,13 +157,66 @@ mod unit_tests;
...
+// Public
 pub use errors::*;
 
 // Crate
 pub(crate) use message_secrets::*;
 pub(crate) use psk::*;
 
+// Public types
+
+/// A group secret that can be used among members to prove that a member was part of a group.
```
kkohbrok, line 169:
```suggestion
/// A group secret that can be used among members to prove that a member was part of a group in a given epoch.
```

kkohbrok (changes requested):
Thanks for the PR! From a structural standpoint, wouldn't it make more sense to make the schedule (crate) private and instead re-export the Resumption and Authentication Secret explicitly as part of the `MlsGroup` API? In my mind, the `schedule`, much like `treesync` is more of an MLS-internal component that shouldn't really be visible to the consumer.

kkohbrok (approved):
Thanks!

PR #840, opened 2022-02-22
Remove tree rewrite chapter from book

kkohbrok:
I've also removed the corresponding pages from the wiki.


raphaelrobert (changes requested):
Thanks, you should also remove the entry from the summary

kkohbrok:
I was under the impression that that happens automatically with mdBook.

raphaelrobert:
> I was under the impression that that happens automatically with mdBook.

We have a `SUMMARY.md` file that still has that entry though.

raphaelrobert (approved):
Thanks!

PR #848, opened 2022-02-23
common readme

franziskuskiefer:
A common readme for GH, book, and docs.
I changed the readme to be a front page for consumers rather than developers and moved developer specific content to the Developer.md.


raphaelrobert (changes requested):
Thanks for picking it up, looks good in general! The only downside is that the documentation disappears from the GH landing page, because documentation is also relevant for folks who want to work *with* OpenMLS, not only *on* OpenMLS. Maybe you can find a way to bridge that gap?

book/README.md
```diff
@@ -0,0 +1 @@
+# Introduction
```
kkohbrok, line 1: Should there be more content here? I have trouble imagining what this will look like in the rendered book.

franziskuskiefer:
> documentation disappears from the GH landing page

yes, that's the idea. I don't think it should be there

kkohbrok (commented):
One question. Otherwise I was wondering the same thing as @raphaelrobert. I thought a direct link to the docs or the manual would be something that the average visitor of our repo is interested in.

book/README.md
franziskuskiefer: It's the readme.md

franziskuskiefer:
> I thought a direct link to the docs or the manual would be something that the average visitor of our repo is interested in.

If this is supposed to be a library that people use, the majority of people won't be interested in developer details.
One does not put information about how to work on a library on the front page.
There can be information on how to build and test. But since that's only `cargo build` and `cargo test` I removed it as it doesn't give any useful information.

kkohbrok:
Ok, I think I begin to understand. So we expect two types of visitors: Developers who want to use OpenMLS as MLS library in their code and other people who have a general interest, maybe because one of their applications is using OpenMLS as a dependency, or something similar. A third category is maybe people who want to work on OpenMLS itself. Does that roughly reflect the premise of the changes here?

raphaelrobert:
> > I thought a direct link to the docs or the manual would be something that the average visitor of our repo is interested in.
> 
> If this is supposed to be a library that people use, the majority of people won't be interested in developer details. One does not put information about how to work on a library on the front page. There can be information on how to build and test. But since that's only `cargo build` and `cargo test` I removed it as it doesn't give any useful information.

Like I said above, it's fine to hide the information that's relevant to folks who want to work *on* OpenMLS. But I still think that most visitors of the GH page are interested in how they can use OpenMLS in their projects. Keeping the Documentation section in the main README would solve that problem. The only downside is that it will also appear in the book and thus contain a self-reference (which is not pretty, but maybe acceptable)

book/README.md
kkohbrok: If I understand correctly, this file is included in the `SUMMARY.md`, where it appears as a new, empty chapter with the headline `Introduction`, but without content. Is that correct? Or is this meant to include the main README.md from the root of the repo?

franziskuskiefer:
> most visitors of the GH page are interested in how they can use OpenMLS in their projects

Release documentation is linked and what consumers are interested in.

raphaelrobert (approved):
Let's get this in then!

kkohbrok (approved):

PR #850, opened 2022-02-23
Move verification chapter to signable module docs

kkohbrok:
This PR doesn't literally move the content of the verification chapter, but instead deletes the page from the book and provides a similar overview over the technique (with a few more details) in the `signable` module docs.


raphaelrobert (approved):

kkohbrok:
Small clarification, as the title is a bit misleading: This PR doesn't literally move the content, but deletes the page from the book and instead provides a similar overview over the technique (with a few more details) in the `signable` module docs. I'll update the PR description accordingly.

PR #851, opened 2022-02-23
Ensure that only things that are actually needed are public

### LINKED ISSUE #809: ensure that only things that are actually needed are public

raphaelrobert:
Fixes #809.

The changes should be pretty straightforward.

As mentioned before, some refactoring still needs be done regarding unnecessary re-exports in these files:

```
schedule/mod.rs
messages/mod.rs
treesync/mod.rs
```


openmls/src/prelude.rs
```diff
@@ -42,9 +39,9 @@ pub use crate::treesync::{
 
 // PSKs
 // TODO #751
-pub use crate::schedule::psk::{
-    BranchPsk, ExternalPsk, PreSharedKeyId, PreSharedKeys, Psk, PskBundle, PskType, ReinitPsk,
-};
+// pub use crate::schedule::psk::{
```
kkohbrok, line 42: Those should probably go.

raphaelrobert:
> Looks good, with one bit of leftover commented-out code.
> 
> Regarding the renaming: I think we agreed at some point that that we want to avoid `mod.rs` files for newly added modules. We have the same naming scheme in a few other places where we have `kats_and_tests` submodules in case you want to make it uniform one way or the other.

I wasn't aware of that decision. I suggest we do this everywhere then if we really want it. Right now we have `mod.rs` everywhere, so in that sense it's consistent at least. As for the KATs, I think there's more work to be done there anyway, I didn't touch these parts at all in this PR.

kkohbrok (approved):
No worries. I don't mind that much either way.

Issue #852, opened 2022-02-23
Get rid of redundant re-exports


raphaelrobert:
Concerned files:

```
schedule/mod.rs
messages/mod.rs
treesync/mod.rs
```

PR #853, opened 2022-02-24
Remove re-exports from lib.rs

raphaelrobert:
This PR removes the re-exports from `lib.rs`, as discussed previously. It doesn't address #852 fully yet.


franziskuskiefer (approved):

kkohbrok (approved):
Thanks!

PR #857, opened 2022-02-24
Remove re-exports

### LINKED ISSUE #852: Get rid of redundant re-exports

raphaelrobert:
Fixes #852.

This PR removes re-exports (in particular on the crate-level).

### DIFF
--- a/openmls/src/extensions/mod.rs
+++ b/openmls/src/extensions/mod.rs
@@ -22,7 +22,7 @@
 //! - [`RequiredCapabilitiesExtension`] (GroupContext extension)
 
 use openmls_traits::crypto::OpenMlsCrypto;
-pub(crate) use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize};
 use std::{
     convert::TryFrom,
     fmt::Debug,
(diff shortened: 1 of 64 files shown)


openmls/src/extensions/mod.rs
```diff
@@ -22,25 +22,28 @@
...
 
+// Private
 mod capabilities_extension;
-mod errors;
 mod external_key_id_extension;
 mod life_time_extension;
 mod parent_hash_extension;
 mod ratchet_tree_extension;
 mod required_capabilities;
+use errors::*;
```
kkohbrok, line 40: Should this line not be above the `// Private` line?

openmls/src/key_packages/mod.rs
```diff
@@ -95,10 +95,10 @@ use tls_codec::{
 
 // Private
 mod codec;
-mod errors;
+use errors::*;
```
kkohbrok, line 98: Same here.

openmls/src/messages/mod.rs
```diff
@@ -15,45 +15,36 @@ use crate::{
...
-    types::{CryptoError, HpkeCiphertext},
+    types::{Ciphersuite, CryptoError, HpkeCiphertext},
     OpenMlsCryptoProvider,
 };
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
 // Private
-mod codec;
+use proposals::*;
```
kkohbrok, line 25: and here.

kkohbrok (changes requested):
Thanks for the going through all of that! Just some places where the `use` statements should be moved up a few lines, I think. I pointed out some of them, but I might have missed one or two.

openmls/src/extensions/mod.rs
raphaelrobert: I put the `use` of submodules in the private section. the other `use` statements are for things that are no submodules.

kkohbrok (approved):

PR #863, opened 2022-02-25
only expose hash_ref from ciphersuite

franziskuskiefer:


kkohbrok (approved):

raphaelrobert (approved):
Good call! There are probably some `use` statements in the tests that could now get cleaned up, but it's not mandatory.

raphaelrobert:
Actually, I think `prelude.rs` should also be adapted accordingly.

franziskuskiefer:
> Actually, I think `prelude.rs` should also be adapted accordingly.

Then I think I'd rather keep it as is. The entire module is useless. But this doesn't safe the day either.

PR #869, opened 2022-03-02
Enable android targets

pulkitkothari:
With the new added targets, openmls will always be ready for Android platform.

Source: https://crates.io/crates/cargo-ndk
https://github.com/bbqsrc/cargo-ndk/blob/main/README.md#L15-L19
```
rustup target add \
    aarch64-linux-android \
    armv7-linux-androideabi \
    x86_64-linux-android \
    i686-linux-android
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,6 +36,9 @@ jobs:
           - aarch64-unknown-linux-gnu
           # Android
           - aarch64-linux-android
+          - armv7-linux-androideabi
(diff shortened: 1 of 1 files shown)


franziskuskiefer (approved):
Thanks!
I re-triggered the CI get it unstuck, then this should be good to go.

franziskuskiefer:
Ah, I think it's not actually going to go through because it's from an external branch. I'll just merge as is.

PR #873, opened 2022-03-20
Create BasicCredentials from existing signature keypair

raphaelrobert:
This PR fixes a regression where `BasicCredential` did no longer expose a method to create it from an existing signature keypair.


openmls/src/ciphersuite/signature.rs
```diff
@@ -188,7 +197,8 @@ impl SignaturePrivateKey {
     }
 
     /// Returns the raw private key bytes as slice.
-    pub(crate) fn as_slice(&self) -> &[u8] {
+    #[cfg(feature = "crypto-subtle")]
+    pub fn as_slice(&self) -> &[u8] {
```
kkohbrok, line 201: Is it really necessary to expose this? Consumers who want to bring their own private keys can do so already. Maybe at this point we should add a chapter on the `crypto-subtle` API to the book. Just to keep an overview over what it enables.

kkohbrok (changes requested):
Thanks for the PR! If I read this correctly, it not only re-enables the creation of a BasicCredential from a signature key pair, but also the export of a signature private key (in the crypto-subtle API). If that is intentional and this is something we want to do, it should probably be documented somewhere.

book/src/user_manual/crypto-subtle.md
```diff
@@ -0,0 +1,41 @@
+# crypto-subtle feature
+
+This feature of the OpenMLS crate allows importing and exporting private signature keys that can be used with credentials.
+
+⚠️ Note that no checks are performed on the keys, use this feature at your own risk.
```
kkohbrok, line 5: Just out of an abundance of caution, can you add a note here that essentially says: If you create a credential from an existing key or if you export key material then you are responsible for the deletion of that key. If that key is kept around outside of OpenMLS, updating a leaf is not going to get you FS/PCS.

kkohbrok (changes requested):
Thanks for adding the documentation! Just one more nit.

kkohbrok (approved):
Thanks a lot!

PR #890, opened 2022-04-22
[FixUp] join_by_external_commit does not need to request proposal store

pulkitkothari:
ProposalStore::new() method visibility is for crate only, so no one is the application can call this method which is required by join_by_external_commit method.
Also there are no public methods which allow temporary proposal to be stored proposals that are received from the DS in between two commit messages by the application before creating group.

https://github.com/openmls/openmls/blob/main/openmls/src/group/core_group/proposals.rs#L27

### DIFF
--- a/openmls/src/group/mls_group/creation.rs
+++ b/openmls/src/group/mls_group/creation.rs
@@ -137,7 +137,6 @@ impl MlsGroup {
         mls_group_config: &MlsGroupConfig,
         aad: &[u8],
         credential_bundle: &CredentialBundle,
-        proposal_store: ProposalStore,
     ) -> Result<(Self, MlsMessageOut), ExternalCommitError> {
         let resumption_secret_store =
             ResumptionSecretStore::new(mls_group_config.number_of_resumption_secrets);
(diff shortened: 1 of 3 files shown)


openmls/src/group/mls_group/creation.rs
```diff
@@ -149,7 +148,7 @@ impl MlsGroup {
         let params = CreateCommitParams::builder()
             .framing_parameters(framing_parameters)
             .credential_bundle(credential_bundle)
-            .proposal_store(&proposal_store)
+            .proposal_store(&ProposalStore::new())
```
pulkitkothari, line 151: We may use this same proposal store instance in MlsGroup constructor as well.

openmls/src/group/mls_group/creation.rs
```diff
@@ -146,6 +145,7 @@ impl MlsGroup {
         let framing_parameters =
             FramingParameters::new(aad, mls_group_config.wire_format_policy().outgoing());
 
+        let proposal_store = ProposalStore::new();
```
pulkitkothari, line 148: We may use this same proposal store instance in MlsGroup constructor as well.

raphaelrobert (changes requested):
Thanks for the PR! This makes sense and it is also in line with the spec:

> The Commit MUST NOT include any proposals by reference, since an external joiner cannot determine the validity of proposals sent within the group

Technically this is a breaking API change, but since the API wasn't usable at all it is really just a fix.

Could you please make an entry in the `CHANGELOG.md` file?
Can you add a section called "Changed" under `[Unreleased]` and mention the fix in a bullet point?

raphaelrobert (approved):
Thanks!

Issue #245, opened 2020-12-06
[Meta] Improve serialization


raphaelrobert:
Areas of improvement:
 - [ ] Flexibility: support more formats than JSON
 - [ ] Efficiency: More compact serialization
 - [ ] #672
 - [ ] #1194

PR #900, opened 2022-06-15
[Fixup] Expose SerializedMlsGroup until issue #245 is done

pulkitkothari:
Expose SerializedMlsGroup because currently MlsGroup expose [load and save](https://github.com/openmls/openmls/blob/main/openmls/src/group/mls_group/mod.rs#L296-L308) methods but that uses serde_json which is not at all optimized. For 2 people group it spits around 60kb of bytes. 
For time being, until #245 is not done, SerializedMlsGroup can be used with [bincode](https://github.com/bincode-org/bincode) to serialize and it outputs ~4075 bytes for 2 people group.

### DIFF
--- a/openmls/src/group/mls_group/ser.rs
+++ b/openmls/src/group/mls_group/ser.rs
@@ -6,8 +6,12 @@ use serde::{
 };
 
 /// Helper struct that contains the serializable values of an `MlsGroup.
+#[deprecated(
+    since = "0.4.1",
+    note = "It is temporarily exposed, it will be private again after #245"
+)]
 #[derive(Serialize, Deserialize)]
-pub(crate) struct SerializedMlsGroup {
+pub struct SerializedMlsGroup {
     mls_group_config: MlsGroupConfig,
     group: CoreGroup,
     proposal_store: ProposalStore,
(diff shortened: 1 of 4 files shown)


franziskuskiefer (commented):
Generally fine with me.
I just want to note we have to be careful what to make public because it will be hard to make it private again. But being pre 1.0 I'm fine with this and having this public to allow for custom serialisations might be a reasonable API choice anyway.

kkohbrok (approved):
Sounds to me like a good temporary fix until we can tackle #245 properly.

raphaelrobert (changes requested):
I see how this solves the problem, but it also makes for a weird asymmetric API change.
In the spirit of making progress, could you please extend the documenting comment of `SerializedMlsGroup` to mention that this is strictly temporary? That way we could merge this for now while not documenting it any further.

openmls/src/group/mls_group/ser.rs
```diff
@@ -6,8 +6,9 @@ use serde::{
 };
 
 /// Helper struct that contains the serializable values of an `MlsGroup.
+#[deprecated(since="0.4.1", note="It is temporarily exposed, it will be private again after #245")]
```
pulkitkothari, line 9: This is my personal preference, please let me know if you want me to just add documentation instead of deprecated.

PR #902, opened 2022-06-23
Implement External Add proposal

beltram:
Part of #106 

This PR adds support for creating External Add Proposals. They allow non-group members to craft a proposal. Group members can later on commit that external Add proposal if they authorize the sender.

Only `Sender::NewMember` is supported where a client is creating a proposal for itself.

It can be used like this from a consumer POV:
```rust
let plaintext_message: MlsMessageOut = openmls::prelude::ExternalProposal::new_add(
            self_key_package,
            group_id,
            group_epoch,
            &credential_bundle,
            &backend
)?
```

I deliberately decided not to represent in the `ExternalProposal` enum other external proposal kinds to make illegal states not representable.

---
Also implements #949 

This replaces ~~`Sender::NewMember`~~ by `Sender::NewMemberProposal` and `Sender::NewMemberCommit` for external proposals and external commits repectively


franziskuskiefer:
Thanks for the PR @beltram.
Can you please have a look at the [contributing guidelines](https://github.com/openmls/openmls/blob/main/CONTRIBUTING.md) and update the PR accordingly?

beltram:
> Thanks for the PR @beltram. Can you please have a look at the [contributing guidelines](https://github.com/openmls/openmls/blob/main/CONTRIBUTING.md) and update the PR accordingly?

That should be in order now. If not please pinpoint the rules I missed

openmls/openmls.iml
```diff
@@ -0,0 +1,14 @@
...
+    <content url="file://$MODULE_DIR$">
+      <sourceFolder url="file://$MODULE_DIR$/benches" isTestSource="true" />
+      <sourceFolder url="file://$MODULE_DIR$/src" isTestSource="false" />
+      <sourceFolder url="file://$MODULE_DIR$/tests" isTestSource="true" />
+      <excludeFolder url="file://$MODULE_DIR$/target" />
+    </content>
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+</module>
```
raphaelrobert, line 14: What is this file?

raphaelrobert (changes requested):
Thanks for the PR! I think it's a good start, but on a high level it needs the following:

 - More testing than just ValSem113, we need at least a full integration test for a group with multiple members
 - A stronger distinction between Add proposal of type `NewMember` and `Preconfigured`. Right now it sounds like all external Add proposals are of type `NewMember`.
 - Documentation in the book, since this introduces a new feature and modifies the public API
 - An entry in `CHANGELOG.md` to document the API changes

Going forward please don't do force pushes because it makes reviewing changes impossible.

openmls/openmls.iml
beltram: woops, sorry, IntelliJ file

beltram:
> More testing than just ValSem113, we need at least a full integration test for a group with multiple members

Done [here](openmls/src/group/tests/external_proposal.rs)

>     * A stronger distinction between Add proposal of type `NewMember` and `Preconfigured`. Right now it sounds like all external Add proposals are of type `NewMember`.

I agree. I did not wanted to add it first because `ExternalSenders` are going to alter the API a lot. But I did this:
```rust
pub fn new_add(..., sender_id: Option<TlsByteVecU8>, ...)
```
And clearly documented that the API is subject to change (`TlsByteVecU8` will become a `u16` basically) and that for the moment the sender

> Documentation in the book, since this introduces a new feature and modifies the public API

Done

> An entry in `CHANGELOG.md` to document the API changes

Done

raphaelrobert:
Feel free to re-request a review whenever you are done with the commits.

> Going forward please don't do force pushes because it makes reviewing changes impossible.

I also saw that you force-pushed again. It potentially causes much more work for reviewers and we might just not review PRs with forced pushes going forward. Force pushing is only for emergencies, squashing should be used for all other cases.

duesee:
> I also saw that you force-pushed again. It potentially causes much more work for reviewers and we might just not review PRs with forced pushes going forward. Force pushing is only for emergencies, squashing should be used for all other cases.

I would phrase that a bit differently. Force-pushing can be benefitial in reviews. Maintainers of other open-source software projects may even ask for force-pushes. It kinda depends on the review style ...

Anyway, I do understand the shortcomings. Still, I think we should write that down somewhere and point people to that. I know from personal experience, sorry Franziskus, that it is kinda hard to break ones habits :-) Especially if it is not clear why this "rule" exists. Which might be the case when someone approached reviews differently.

beltram:
@raphaelrobert 
> Feel free to re-request a review whenever you are done with the commits

Sure I will do that in the future. I wasn't sure how I should reach out, either by requesting review or in Zulip. Now I know 😄 

> I also saw that you force-pushed again

Yes I did but just for amending my "PR review commit" and to leave you with the cleanest history possible. But I did not touch the commit you commented on.

> I think we should write that down somewhere

I think a PR template would be an excellent place for mentioning that 😉

raphaelrobert:
Fair points, I agree we should add that to our [guideline](https://github.com/openmls/openmls/blob/main/CONTRIBUTING.md).

The reason for not wanting force-pushes during review is that it invalidates notes (on GH or not) the reviewer made. It also makes it harder to assess if requested changes have been addressed and if so, how.
With force-pushes reviewers have to technically re-review all code, because anything (no matter how little) could have changed. In practice, reviewers will be inclined to just glance over the code parts that look familiar and approve them and that makes the whole process a lot less secure.

book/src/message_validation.md
```diff
@@ -53,21 +53,22 @@ The following is a list of the individual semantic validation steps performed by
...
+| `ValSem108`    | Remove Proposal: Removed member must be an existing group member                                  | ✅             | ✅ | `openmls/src/group/tests/test_proposal_validation.rs` |
+| `ValSem109`    | Update Proposal: Identity must be unchanged between existing member and new proposal              | ✅             | ✅ | `openmls/src/group/tests/test_proposal_validation.rs` |
+| `ValSem110`    | Update Proposal: HPKE init key must be unique among existing members                              | ✅             | ✅ | `openmls/src/group/tests/test_proposal_validation.rs` |
+| `ValSem111`    | Update Proposal: The sender of a full Commit must not include own update proposals                | ✅             | ✅ | `openmls/src/group/tests/test_proposal_validation.rs` |
+| `ValSem112`    | Update Proposal: The sender of a standalone update proposal must be of type member                | ✅             | ✅ | `openmls/src/group/tests/test_proposal_validation.rs` |
+| `ValSem113`    | External Add Proposal: The sender of a standalone external add proposal must be of type NewMember | ✅             | ✅ | `openmls/src/group/tests/test_proposal_validation.rs` |
```
raphaelrobert, line 71: I think it's more complicated than that. The sender of a standalone Add proposal can either be

 - An existing group member
 - A preconfigured sender (not implemented yet)
 - A sender of type `NewMember`

So I'm wondering where this validation rule comes from and what part of the spec it implements.

raphaelrobert (commented):
Not a full review yet, just a preliminary question.

book/src/message_validation.md
beltram: It is more a validation for **External** Add proposal only so I think in this case we can omit the `existing group member`.
Perhaps I should create a new ValSem "group" just for External proposals ?
Regarding `preconfigured` sender I preferred not to represent it yet since it's not implemented but I may be wrong

raphaelrobert: I agree it's good to leave out preconfigured senders for now and keep the PR lean. My point was that if every kind of sender is a valid sender then the whole validity check becomes meaningless. That's why I was wondering how it relates to the spec. So the question is: what exactly is being validated here?

openmls/src/framing/message.rs
```diff
@@ -156,6 +156,18 @@ impl MlsMessageIn {
...
+    pub fn is_external(&self) -> bool {
+        let is_external = match &self.mls_message.body {
+            MlsMessageBody::Plaintext(p) => {
+                matches!(p.sender(), Sender::NewMember | Sender::Preconfigured(_))
+            }
+            // external message cannot be encrypted
+            MlsMessageBody::Ciphertext(_) => false,
+        };
+        is_external && self.mls_message.is_handshake_message()
+    }
```
raphaelrobert, line 169: I think the check of whether a message is external only depends on its sender type. By design, external messages cannot be encrypted. The library cannot produce them, nor could it decrypt them. I'd simplify this only to check the sender type.

openmls/src/framing/plaintext.rs
```diff
@@ -228,6 +229,29 @@ impl MlsPlaintext {
         )
     }
 
+    /// This constructor builds an `MlsPlaintext` containing an External Proposal.
+    /// The sender type is either `Sender::Preconfigured` or `Sender::NewMember`.
+    pub(crate) fn member_external_proposal(
```
raphaelrobert, line 234: I'd make this exclusively about `NewMember` for now because this API doesn't force the caller to choose the right sender type and that's unsafe.

openmls/src/framing/validation.rs
```diff
@@ -186,16 +186,17 @@ impl DecryptedMessage {
...
+                    MlsContentBody::Commit(Commit { path, .. }) => path
+                        .as_ref()
+                        .map(|p| p.leaf_key_package().credential().clone())
+                        .ok_or(ValidationError::NoPath),
+                    MlsContentBody::Proposal(Proposal::Add(AddProposal { key_package })) => {
+                        Ok(key_package.credential().clone())
                     }
-                } else {
-                    Err(ValidationError::NotACommit)
+                    _ => Err(ValidationError::NotACommitOrExternalAddProposal),
```
raphaelrobert, line 199: This looks like things are being mixed now. In fact, the sender types are redefined in #949 and I'm starting to think that issue should also be covered by this PR. That way we'd be up to spec (modulo the preconfigured senders).

openmls/src/group/errors.rs
```diff
@@ -210,9 +210,9 @@ pub enum ValidationError {
     /// Message epoch differs from the group's epoch.
     #[error("Message epoch differs from the group's epoch.")]
     WrongEpoch,
-    /// The MlsPlaintext is not a Commit despite the sender begin of type NewMember.
-    #[error("The MlsPlaintext is not a Commit despite the sender begin of type NewMember.")]
-    NotACommit,
+    /// The MlsPlaintext is not a Commit or an external Add proposal despite the sender begin of type NewMember.
+    #[error("The MlsPlaintext is not a Commit or an external Add proposal despite the sender begin of type NewMember.")]
+    NotACommitOrExternalAddProposal,
```
raphaelrobert, line 215: This will also change with #949.

openmls/src/group/tests/test_proposal_validation.rs
```diff
@@ -21,11 +21,14 @@ use crate::{
         proposals::{AddProposal, Proposal, ProposalOrRef, RemoveProposal, UpdateProposal},
         Welcome,
     },
+    prelude::ExternalProposal,
```
raphaelrobert, line 24: We should only use `prelude` for integration tests, not in the library itself.

openmls/src/messages/external_proposals.rs
```diff
@@ -0,0 +1,84 @@
+//! External Proposal
+//!
+//! Contains the types and methods to build external proposal
+//! to add a client from a MLS group
+//! `Remove` & `ReInit are nto yet implemented`
```
raphaelrobert, line 5:
```suggestion
//! `Remove` & `ReInit are not yet implemented`
```

openmls/src/messages/external_proposals.rs
```diff
@@ -0,0 +1,84 @@
+//! External Proposal
```
raphaelrobert, line 1:
```suggestion
//! External Proposals
```

openmls/src/messages/external_proposals.rs
```diff
@@ -0,0 +1,84 @@
...
+/// External Proposal.
+/// External proposal allows parties outside a group to request changes to the latter.
+///
+/// This `enum` contains the different external proposals in its variants.
+/// Not yet implemented: `Remove` & `ReInit`
+#[non_exhaustive]
+pub enum ExternalProposal {
+    /// Proposes adding a client to a group.
+    /// Newly added client can be either the sender itself or not.
+    Add(AddProposal),
```
raphaelrobert, line 27: While we leave out preconfigured senders, we can still think about what the API should look like eventually. The spec is a bit annoying because there is some overlap between an Add proposal from a preconfigured sender (who will by design always remain an external sender) and an Add proposal from a sender of type NewMember. 

I'm wondering if we should really treat both as external proposals, risking confusing users of the library. I see NewMember Add proposals more of yet another way to join the group (we already have Welcome and ExternalCommit).

Maybe this should be renamed to `JoinProposal` to avoid all confusion. 

I'm also not sure if we should have this `enum` to cover all proposals, because neither NewMember Add proposals nor ReInit proposals are really external proposals. Did you see a commonality between them I missed?

raphaelrobert (changes requested):
Thanks for making the requested changes! I reviewed the code (not the tests yet because the code might change).

openmls/src/messages/external_proposals.rs
beltram: > Maybe this should be renamed to JoinProposal to avoid all confusion

I agree with you on this one, this will make the API clearer.

> neither NewMember Add proposals nor ReInit proposals are really external proposals

I don't follow you on this one. I understood external proposal as "sent by a non-member". IMHO, listing all the proposals under this enum makes the API straightforward for someone learning MLS from the draft. One can jump back-and-forth between the draft and the code ; it's simpler than if it were scattered across the codebase.

openmls/src/framing/validation.rs
beltram: Will include #949 in this PR

openmls/src/messages/external_proposals.rs
raphaelrobert: This part is complicated. 

These are the valid sender types for the various kinds of proposals where more than one sender type is allowed:

ReInit: `SenderType.member || SenderType.external`
Add: `SenderType.member || SenderType.external || SenderType.new_member_proposal`
Remove: `SenderType.member || SenderType.external`

Whenever the sender type is `external`, the group MUST use the `ExternalSender` extension. Note that this is not the case for `new_member_proposal`. This is where I see a big difference between "external" proposals and the JoinProposal.

We should use Rust's type safety to make it as easy and safe for consumers of the library, given how confusing this stuff is even for us.

I'd like to suggest that we limit this to an opinionated `JoinProposal` struct that has a `new(...)` method to safely create it. I don't see where the `ExternalProposal` enum is used anyway (unless I missed something). And we can revisit that when we introduce external/preconfigured senders.

On a similar note, we also need to think about what the inspection part of the API looks like. I think that part is missing so far. I'm thinking the easiest way would be to extend `StagedCommit` with a new category

book/src/message_validation.md
beltram: Ok I get it now. At the end of the day, from a receiver POV, an external proposal is just a proposal. So there's no need for this validation. I will completely remove it

openmls/src/messages/external_proposals.rs
beltram: > I'd like to suggest that we limit this to an opinionated JoinProposal struct that has a new(...) method

Done. However I really wonder how will the API look like for `Preconfigured` senders since anyway you can't tie them to a `MlsGroup` to verify the `RequiredCapabilities` & `ExternalSender` extension. But as you said we can revisit that later

> we also need to think about what the inspection part of the API looks like

I introduced `ProcessedMessage::ExternalJoinProposalMessage` for that. It will later have to also encompass external commits since applications might define a slightly different authorization policy for each (at least Wire does)

openmls/src/group/tests/external_proposal.rs
```diff
@@ -0,0 +1,323 @@
...
+                .parse_message(plaintext.into(), backend)
+                .unwrap_err(),
+            ParseMessageError::ValidationError(ValidationError::NotAnExternalAddProposal)
+        ));
+    } else {
+        panic!()
+    };
+    alice_group.clear_pending_proposals();
+
+    // Remove proposal cannot have a 'new_member_proposal' sender
```
raphaelrobert, line 309:
```suggestion
    // Update proposal cannot have a 'new_member_proposal' sender
```

openmls/src/group/tests/external_proposal.rs
```diff
@@ -0,0 +1,323 @@
...
+    assert!(matches!(
+        alice_group
+            .process_unverified_message(invalid_msg, None, backend)
+            .unwrap_err(),
+        UnverifiedMessageError::InvalidSignature
+    ));
+}
+
+#[apply(ciphersuites_and_backends)]
+fn new_member_proposal_sender_should_be_reserved_for_join_proposals(
```
raphaelrobert, line 257: I think we will need semantic validation for sender types once we have all external proposals, but it's good to have these tests for now.

openmls/tests/book_code.rs
```diff
@@ -1050,6 +1072,68 @@ fn book_operations(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvide
...
+    // Create a new KeyPackageBundle for Bob
+    let bob_credential_bundle = get_credential_bundle(
+        "Bob".into(),
+        CredentialType::Basic,
+        ciphersuite.signature_algorithm(),
+        backend,
+    )
+    .expect("Could not get CredentialBundle");
+    let bob_key_package =
+        generate_key_package_bundle(&[ciphersuite], &bob_credential_bundle.credential(), backend)
```
raphaelrobert, line 1086:
```suggestion
        generate_key_package_bundle(&[ciphersuite], bob_credential_bundle.credential(), backend)
```

raphaelrobert (changes requested):
Thanks for making the requested changes and sorry for the slow review (IETF115 is to blame).
I commented on a few nits inline and there's one test that I think is a good candidate for a ValSem check.

I suspect the API will change once we tackle external/preconfigured senders. This still feels a bit rough around the edges. Maybe I can think of a better way to express this partial overlap of proposals/commits/internal/external in the API. But for now, we should really get this PR merged.

openmls/src/group/tests/external_proposal.rs
```diff
@@ -0,0 +1,323 @@
...
+            .parse_message(proposal.clone().into(), backend)
+            .unwrap();
+
+        // message verification should fail with an invalid signature key i.e. != Charlie's one
+        let (_, attacker_key) = SignatureKeypair::new(ciphersuite.signature_algorithm(), backend)
+            .unwrap()
+            .into_tuple();
+        assert!(alice_group
+            .process_unverified_message(parsed.clone(), Some(&attacker_key), backend)
+            .is_err());
```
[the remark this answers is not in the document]
beltram: Good catch I forgot that one indeed

openmls/src/group/tests/external_proposal.rs
beltram: Agree. I added a TODO to remind us later

openmls/src/group/tests/external_proposal.rs
```diff
@@ -0,0 +1,342 @@
...
+        .unwrap();
+    assert!(matches!(
+        alice_group
+            .process_unverified_message(invalid_msg, None, backend)
+            .unwrap_err(),
+        UnverifiedMessageError::InvalidSignature
+    ));
+}
+
+// TODO: move this test to a dedicated external proposal ValSem test module once all external proposals implemented
```
raphaelrobert, line 274:
```suggestion
// TODO #1093: move this test to a dedicated external proposal ValSem test module once all external proposals implemented
```

raphaelrobert (commented):
Thanks! I think the CI didn't run because of conflicts, but otherwise we are good to go.

raphaelrobert:
Not sure why the CI doesn't indicate it, but `cargo doc` fails for me.

raphaelrobert:
It's also weird that GH displays 'This branch is out-of-date with the base branch` but doesn't offer a way to merge in the main branch. I wonder if the force pushes caused that.

raphaelrobert (approved):

PR #909, opened 2022-08-01
Use thiserror in support crates

zaynetro:
I've noticed that majority of errors have been migrated to thiserror. This PR transitions public errors from support crates to derive `std::error::Error` as well.

### DIFF
--- a/openmls/src/key_store.rs
+++ b/openmls/src/key_store.rs
@@ -1,7 +1,8 @@
 //! Serialization for key store objects.
 
 use crate::{
-    credentials::CredentialBundle, key_packages::KeyPackageBundle, schedule::psk::PskBundle,
+    credentials::CredentialBundle, key_packages::KeyPackageBundle, prelude::LibraryError,
+    schedule::psk::PskBundle,
 };
 
 use openmls_traits::key_store::{FromKeyStoreValue, ToKeyStoreValue};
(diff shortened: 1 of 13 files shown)


openmls/src/key_store.rs
```diff
@@ -1,53 +1,56 @@
...
+    schedule::psk::PskBundle,
 };
 
 use openmls_traits::key_store::{FromKeyStoreValue, ToKeyStoreValue};
 
 // === OpenMLS Key Store Types
 
 impl FromKeyStoreValue for KeyPackageBundle {
-    type Error = &'static str;
+    type Error = LibraryError;
```
zaynetro, line 13: I am not sure whether reusing `LibraryError` here is a good idea. Alternatively, I could define four structs to be more specific. It will be more verbose though.

franziskuskiefer: I think `LibraryError` is fine here for now. There's nothing to recover from the error information.

traits/src/key_store.rs
```diff
@@ -3,19 +3,19 @@
 use std::fmt::Debug;
 
 pub trait FromKeyStoreValue: Sized {
-    type Error: Debug + Clone + PartialEq + Into<String>;
```
franziskuskiefer, line 6: What about trait bounds that we lose here? `Into<String>` is not needed because we have `Display` now. But `Clone` and `PartialEq` might be nice to have, esp. the latter.

franziskuskiefer (changes requested):
Thanks for contributing!
The PR generally looks good. A few questions inline.

We haven't had changes on the supporting crates yet. So there's ne CHANGELOG.md in there yet (there's only one on the top level for the openmls library).
Can you add CHANGELOG.md files for these crates and add content according to https://keepachangelog.com/en/1.0.0/ (or just look at the one in this repo for how to fill it).

traits/src/key_store.rs
zaynetro: I actually I have a question about these extra traits. Why are they needed?

Let's say we have an `OpenMlsKeyStore` implementation and we want to store values in SQLite.

```rust
impl SqliteCryptoProvider {
    fn persist(&self, ...) -> Result<(), rusqlite::Error> {
        Ok(())
    }
}

impl OpenMlsKeyStore for SqliteCryptoProvider {
    type Error = KeyStoreError;

    fn store<V: ToKeyStoreValue>(&self, k: &[u8], v: &V) -> Result<(), Self::Error>
    where
        Self: Sized,
    {
        self.persist(k, v)?;
        Ok(())
    }
    
    // ...
}
```

We also need to define our error:

```rust

#[derive(thiserror::Error, Debug)]
pub(crate) enum KeyStoreError {
    #[error("{0}")]
    Database(#[from] rusqlite::Error),
    #[error("{0}")]
    Storage(#[from] anyhow::Error),
}
```

In this case I cannot derive `Clone` and `PartialEq` because neither `rusqlite::Error` nor `anyhow::Error` derive those traits! 

```
error[E0277]: can't compare `KeyStoreError` with `KeyStoreError`
  --> client/src/secrets.rs:71:18
   |
71 |     type Error = KeyStoreError;
   |                  ^^^^^^^^^^^^^ no implementation for `KeyStoreError == KeyStoreError`
   |
   = help: the trait `PartialEq` is not implemented for `KeyStoreError`
note: required by a bound in `openmls::prelude::OpenMlsKeyStore::Error`
  --> /../openmls/traits/src/key_store.rs:18:53
   |
18 |     type Error: std::error::Error + Debug + Clone + PartialEq;
   |                                                     ^^^^^^^^^ required by this bound in `openmls::prelude::OpenMlsKeyStore::Error`

error[E0277]: the trait bound `KeyStoreError: Clone` is not satisfied
  --> client/src/secrets.rs:71:18
   |
71 |     type Error = KeyStoreError;
   |                  ^^^^^^^^^^^^^ the trait `Clone` is not implemented for `KeyStoreError`
   |
note: required by a bound in `openmls::prelude::OpenMlsKeyStore::Error`
  --> /../openmls/traits/src/key_store.rs:18:45
   |
18 |     type Error: std::error::Error + Debug + Clone + PartialEq;
   |                                             ^^^^^ required by this bound in `openmls::prelude::OpenMlsKeyStore::Error`
```

I also can't easily derive those traits on my custom error type:

```
error[E0277]: the trait bound `rusqlite::Error: Clone` is not satisfied
   --> client/src/secrets.rs:143:22
    |
138 | #[derive(thiserror::Error, Debug, Clone, PartialEq)]
    |                                   ----- in this derive macro expansion
...
143 |     Database(#[from] rusqlite::Error),
    |                      ^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `rusqlite::Error`
    |
```

It seems like most errors from other libraries derive only `std::error::Error`.

Obviously, I can work around this by removing third-party errors from my custom enum.

Anyways, this is just something that I wanted to bring up and if you feel like this MR is not the right place to experiment I will happily add `Clone` and `PartialEq` trait requirements back.

zaynetro:
Changelog added.

traits/src/key_store.rs
franziskuskiefer: Sorry for the delay.

I'm fine with removing the trait bounds here. I can see that they make issues in the cases you describe.

franziskuskiefer (approved):
Thanks for adding the changelogs.

This looks good to me know.

Sorry again for the delay in reviewing.

PR #910, opened 2022-08-01
Remove Send + Sync requirement on OpenMlsKeyStore

zaynetro:
Send + Sync requirement is needed for rayon to work. It turns out that
OpenMlsKeyStore is never accessed in parallel so such bound is not
necessary.

Use case: rusqlite's Connection is not thread-safe so it is not possible
to save credential and key package bundles in SQLite out of the box.

Possible alternatives:

1. Unsafely implement Send + Sync on Connection.
    This is what Wire app uses ([Reference](https://github.com/wireapp/core-crypto/blob/develop/keystore/src/connection/platform/generic/mod.rs#L29))
2. Add a cargo feature to disable parallelization.

What do you think?

### DIFF
--- a/openmls/src/ciphersuite/secret.rs
+++ b/openmls/src/ciphersuite/secret.rs
@@ -138,12 +138,11 @@ impl Secret {
     /// HKDF expand where `self` is `prk`.
     pub(crate) fn hkdf_expand(
         &self,
-        backend: &impl OpenMlsCryptoProvider,
+        crypto: &impl OpenMlsCrypto,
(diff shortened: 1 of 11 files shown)


franziskuskiefer:
I think I'd rather prefer having this behind and not remove it completely. While I see that these trait bounds can make problems in some cases they are pretty nice in others.
Another option might be to have an AsyncKeyStore and a KeyStore to chose from.

zaynetro:
Alright! Thanks for checking this PR

PR #964, opened 2022-09-20
September cleanup

raphaelrobert:
Cleaning up warnings again after a while. It looks like clippy warnings are reported through `cargo check` now as well.

### DIFF
--- a/openmls_rust_crypto/src/provider.rs
+++ b/openmls_rust_crypto/src/provider.rs
@@ -269,12 +269,8 @@ impl OpenMlsCrypto for RustCrypto {
                 }
                 let mut sig = [0u8; ed25519_dalek::SIGNATURE_LENGTH];
                 sig.clone_from_slice(signature);
-                k.verify_strict(
-                    data,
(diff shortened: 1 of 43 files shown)


openmls_rust_crypto/src/provider.rs
```diff
@@ -269,12 +269,8 @@ impl OpenMlsCrypto for RustCrypto {
                 }
                 let mut sig = [0u8; ed25519_dalek::SIGNATURE_LENGTH];
                 sig.clone_from_slice(signature);
-                k.verify_strict(
-                    data,
-                    &ed25519_dalek::Signature::from_bytes(&sig)
-                        .map_err(|_| CryptoError::CryptoLibraryError)?,
-                )
-                .map_err(|_| CryptoError::InvalidSignature)
+                k.verify_strict(data, &ed25519_dalek::Signature::new(sig))
```
franziskuskiefer, line 272: We should move away from this crate in general.

franziskuskiefer (approved):
Thanks, lgtm

PR #1047, opened 2022-10-12
fix(docs): Fix description of validation.

duesee:
I think I noticed a bug in the documentation (`validation.rs`).

* `UnverifiedMemberMessage` -> `UnverifiedGroupMessage`
* `UnverifiedExternalMessage` -> `UnverifiedPreconfiguredMessage`


raphaelrobert (approved):
Good catch!

Issue #877, opened 2022-04-19
[MLS Spec change] TLS syntax updates (#594)


kkohbrok:
**Description of the changes**
This PR introduces variable-length vectors. This should probably be implemented in the TLS-codec crate, but in any case, OpenMLS needs to support it.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/594

franziskuskiefer:
Closing this for now. Everything is implemented in the TLS codec crate that we need. It should get a little more testing over there and needs a new release but the work is done.

raphaelrobert:
Not everything has been addressed yet, e.g. `HashReference` and parent hashes. I think we should keep this open for a bit longer until someone has done a thorough check.

PR #1048, opened 2022-10-13
Start transition to variable-length encodings

franziskuskiefer:
A first PR to update the TLS syntax to use variable-length encodings.
It's a small example to show the difference for byte vectors. It allows us to save remove and simplify some code 🎉 

I'll keep doing small PRs of this sort until #877 is done.


openmls/src/ciphersuite/mod.rs
```diff
@@ -42,6 +40,12 @@ pub(crate) use serde::{Deserialize, Serialize};
 #[cfg(test)]
 mod tests;
 
+/// A simple type for HPKE public keys using [`VLBytes`] for (de)serializing.
+pub type HpkePublicKey = VLBytes;
```
raphaelrobert, line 44: What is the reasoning behind dropping the wrapper and using `type` instead? I can think of a few pros and cons. If you want to drop the wrapper in order to reduce complexity, why do the type alias instead of renaming instances of `HpkePublicKey`?

One argument for keeping the wrapper: it provides strictly better type safety. In this particular case however the net benefit is probably not so big.

raphaelrobert (commented):
Thanks for getting this started! The CI seems unhappy for now, so just one design question for now.

openmls/src/ciphersuite/mod.rs
franziskuskiefer: Right, we could keep it for type safety. But I don't think the small benefits warrant the additional complexity of wrapping it.
I wouldn't use `VLBytes` directly because we then never know what it's supposed to be. The `type` is a hint for  how to use a variable.

franziskuskiefer (commented):
> The CI seems unhappy for now, so just one design question for now.

Yeah, I didn't path all the other crates ...

openmls/src/ciphersuite/mod.rs
duesee: Opinion: type aliases can be confusing and I would probably only use them to shorten complex generic types. In most cases you still want to know what was aliased and thus it introduces an extra mental indirection.

raphaelrobert: I tend to agree with @duesee on this one. Type aliases can be foot guns and don't really match the effort we try to make elsewhere in terms of type safety. They feel like something you would do in C but not in Rust. We've only used them rarely so far and never really had a good discussion about it. I'd like to propose that we don't use type aliases going forward because they can be dangerous.

franziskuskiefer: I strongly disagree. This is what a lot of our verification is based on and I'll make more use of types over time. It's rather bad that they aren't used more in the code yet.

raphaelrobert: Could explain a bit more? I'm not sure I follow. I count 14 instances of `type` (2 of which are public) in the code. If you disagree it would be good to have an explanation on why you think type aliases are not an issue.

franziskuskiefer: Sure. Types are hints for both the engineer as well as compilers to know what's going on. So by having a type there's strictly more information than without a type. Tools like F* use these types to perform their proofs because they can use the types to infer more information from it. Rust does not do anything like this.
Type aliases are only syntactic sugar that's not checked by the Rust compiler but might be by other tools. So they don't give you type safety from the Rust compiler. But the question is, as I said, whether you want or need that. For types like the ones have here there's no real benefit from wrapping this into a type to help the Rust compiler. But instead of using the plain type and not give any information about what this is one should use an alias to give hints (to the compiler, engineer, and other tools).

raphaelrobert (approved):
Discussed offline.

openmls/src/ciphersuite/mod.rs
duesee: I'd agree with @raphaelrobert.

For example,

```rust
struct StructB {
  pub_key: HpkePublicKey,
}
```

... may signal to a developer that someone had the intention to make `HpkePublicKey` to "more than bytes",  e.g., to enforce constraints on the allowed format, etc. 

Maybe you are thinking ahead of time and have plans to make good use of the aliases. However, please elaborate to make your thoughts/ideas understandable. For example, can you elaborate more on this:

> one should use an alias to give hints (to the compiler

An alias is, as you already said, syntactic sugar, so to ensure readability. As such, it gives hints to the developer, rather than to the compiler. But maybe you have here something in mind which may not be as obvious as it seems to you.

> engineer,

Where I would argue that this also has the potential to be misleading as above mentioned. Maybe it would be a good idea to follow here the advise of the Rust book to give aliases preferred to complex ('lengthy') types, as to avoid repetition?

>and other tools).

Which tools do you refer to here?  It would help to understand your point if you could be more specific.

duesee:
> Discussed offline.

Oh, sorry. I wrote something when I had to wait for an appointment and missed this comment. Would be happy to learn what was the conclusion 🐶 (Offline when you prefer.)

raphaelrobert:
> > Discussed offline.
> 
> Oh, sorry. I wrote something when I had to wait for an appointment and missed this comment. Would be happy to learn what was the conclusion 🐶 (Offline if you prefer.)

The gist of it was that type aliases are commonly used for formal verification (something that is in scope for OpenMLS going forward) and that therefore the benefits outweigh the downsides.

Issue #979, opened 2022-10-11
[MLS Spec change] Move wire_format to a separate tagged-union structure MLSMessage #523


franziskuskiefer:
**Description of the changes**
> The current use of wire_format is a bit weird to me: it is the first value of both MLSPlaintext and MLSCiphertext which allows the implementations to differentiate between them. Then it looks like it should be implemented as a regular tagged union, which is what this pull request does.
> 
> This does not imply any changes to existing implementations.
> 
> One thing that this pull-request changes is that MLSPlaintext is forced to have wire_format = mls_plaintext, but I don't see why it would have any other values in this case?

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/523

- [x] #958
- [x] #941
- [x] #881
- [x] #1134 
- [x] Fix the abstraction boundaries in the validation order
- [x] Update docs to reflect the new framing structure

Tracking required changes
- [x] Unambiguous signatures (#980)

duesee:
I'll try the following: In order to minimize the changes that have to be done manually and to avoid confusing types in certain places, I'll first only do renames. The hope is that the "places" where the renamed types are used, i.e., when receiving a message, etc. are roughly similar.

Then, in order to put a final target in place we want to arrie at, I will then drop the current definitions and replace them with the MLS draft ones. Obviously things will break here. Then, I'll try to go to the diffs again and see what steps are necessary to arrive at the final code (or cry for help).

### Changes

#### Structs and Enums

* rename `MLSPlaintext` to `MLSMessageContent`
  * after rename: remove `wire_format` field.
  * after rename: remove `signature`, ... (copy from draft)

* rename `MLSPlaintextTBS` to `MLSPlaintext`
  * adjust

* rename `MLSCiphertext` to `MLSMessage`
  * after: adjust struct (copy from RFC)

* rename `MLSPlaintextTBM` to `MLSMessageContentTBM`
  * adjust

* rename `MLSPlaintextCommitContent` to `MLSMessageCommitContent`

* rename `MLSPlaintextCommitAuthData` to `MLSMessageCommitAuthData`

* `WireFormat` (add fields)

* add: `MLSMessageContentAuth`
* add: `MLSMessageAuth`
* add: `MLSMessageContentTBS`
* add: `MLSCiphertext`

tmp

PR #1049, opened 2022-10-14
Work towards solving #979.

duesee:
Please see the commit messages for a description.

I am working towards solving #979 on branch `duesee/spec_issue_979` and figured that I can extract PR-able commits to slowly make some progress on `main`.

### DIFF
--- a/openmls/src/framing/mod.rs
+++ b/openmls/src/framing/mod.rs
@@ -65,6 +65,7 @@ impl<'a> FramingParameters<'a> {
             wire_format: wire_format.into(),
         }
     }
+
     pub(crate) fn aad(&self) -> &'a [u8] {
         self.aad
     }
--- a/openmls/src/framing/plaintext.rs
+++ b/openmls/src/framing/plaintext.rs
@@ -23,24 +23,10 @@ use tls_codec::{Serialize, TlsByteVecU32, TlsDeserialize, TlsSerialize, TlsSize}
 /// 9. Message framing
 ///
 /// ```c
+/// // draft-ietf-mls-protocol-16
+///
 /// struct {
-///     opaque group_id<0..255>;
-///     uint64 epoch;
-///     Sender sender;
-///     opaque authenticated_data<0..2^32-1>;
-///
-///     ContentType content_type;
-///     select (MLSPlaintext.content_type) {
-///         case application:
-///             opaque application_data<0..2^32-1>;
-///
-///         case proposal:
-///             Proposal proposal;
-///
-///         case commit:
-///             Commit commit;
-///     }
-///
+///     MLSContent content;
 ///     opaque signature<0..2^16-1>;
 ///     optional<MAC> confirmation_tag;
 ///     optional<MAC> membership_tag;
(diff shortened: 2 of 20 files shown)


openmls/src/framing/mod.rs
```diff
@@ -39,15 +39,68 @@ pub use validation::*;
 mod test_framing;
 
 /// Wire format of MLS messages.
+///
+/// # MLS Presentation Language
```
raphaelrobert, line 43: I think we still call it TLS presentation language in the spec because we never defined what the MLS presentation language is

openmls/src/framing/mod.rs
```diff
@@ -39,15 +39,68 @@ pub use validation::*;
...
+/// } WireFormat;
+/// ```
 #[derive(
     PartialEq, Eq, Clone, Copy, Debug, Serialize, Deserialize, TlsDeserialize, TlsSerialize, TlsSize,
 )]
 #[repr(u8)]
 pub enum WireFormat {
-    /// Plaintext message
+    /// Reserved
+    Reserved = 0,
```
raphaelrobert, line 64: We don't enumerate the reserved values, instead we rely on the decoder to fail if they are ever encountered

openmls/src/messages/proposals.rs
```diff
@@ -100,6 +117,7 @@ pub enum Proposal {
     PreSharedKey(PreSharedKeyProposal),
     ReInit(ReInitProposal),
     ExternalInit(ExternalInitProposal),
+    // TODO: `AppAck` is not in draft-ietf-mls-protocol-16.
```
raphaelrobert, line 120: When creating a TODO we adopted the rule that we reference a ticket. If no ticket exists for the TODO, we create a new one.

raphaelrobert (changes requested):
Thanks for the PR! Just a few mechanical things

duesee:
Nice, spot fuzzing is cool :-) I hoped a bit that this part of the code might not be reachable. Will convert this to a draft PR for now (with no review required).

openmls/src/framing/mod.rs
duesee: So maybe just "Presentation Language" as in "2.1. Presentation Language"?

openmls/src/framing/mod.rs
duesee: Ok 👍🏻 I'll put a note in the comment and remove the `Reserved` variant.

raphaelrobert:
> Nice, spot fuzzing is cool :-) I hoped a bit that the `unimplemented!()` part of the code will not be reachable and I can get away with it until I can remove it again. Will convert this to a draft PR for now and resolve later.

You are working on the most reachable part of the library (from a network standpoint), so this is sort of expected :)
I don't remember the fuzzing being triggered much in the past (if at all), but now it detected issues twice in one week.

duesee:
I'm working on another branch to fix #979. These are the changes so far that should be easy to review and low-risc to take. Making slow progress :-)

openmls/src/framing/mod.rs
duesee: I've removed the section everywhere.

franziskuskiefer (changes requested):
Thanks, generally looks good to me. Just a few questions and thoughts around `MlsContentBody`.

openmls/src/framing/plaintext.rs
```diff
@@ -54,14 +54,28 @@ pub(crate) struct MlsPlaintext {
...
+    content: MlsContentBody,
     signature: Signature,
     confirmation_tag: Option<ConfirmationTag>,
     membership_tag: Option<MembershipTag>,
 }
 
+#[allow(clippy::large_enum_variant)]
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+pub(crate) enum MlsContentBody {
+    Application(TlsByteVecU32),
```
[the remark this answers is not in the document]
duesee: Let's do it in a follow-up. Separating things may allow us to bisect later if needed.

openmls/src/framing/plaintext.rs
```diff
@@ -54,14 +54,28 @@ pub(crate) struct MlsPlaintext {
...
+    Proposal(Proposal),
```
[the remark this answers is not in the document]
duesee: This object is basically a "continuation" of `MlsContent` and I'll make that clear (also via the syntax.)

openmls/src/framing/plaintext.rs
```diff
@@ -54,14 +54,28 @@ pub(crate) struct MlsPlaintext {
...
     authenticated_data: TlsByteVecU32,
     content_type: ContentType,
-    content: MlsPlaintextContentType,
+    content: MlsContentBody,
     signature: Signature,
     confirmation_tag: Option<ConfirmationTag>,
     membership_tag: Option<MembershipTag>,
 }
 
+#[allow(clippy::large_enum_variant)]
```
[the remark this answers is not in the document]
duesee: It's not needed according to clippy. Removed.

openmls/src/framing/plaintext.rs
```diff
@@ -54,14 +54,28 @@ pub(crate) struct MlsPlaintext {
...
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+pub(crate) enum MlsContentBody {
```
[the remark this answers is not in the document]
duesee: As discussed privately, I've done that in this PR already.

openmls/src/framing/plaintext.rs
```diff
@@ -49,20 +35,82 @@ use tls_codec::{Serialize, TlsByteVecU32, TlsDeserialize, TlsSerialize, TlsSize}
...
+    #[tls_codec(discriminant = 1)]
+    Application(TlsByteVecU32),
+    #[tls_codec(discriminant = 2)]
+    Proposal(Proposal),
+    #[tls_codec(discriminant = 3)]
+    Commit(Commit),
+}
+
+impl MlsContentBody {
+    pub(crate) fn content_type(&self) -> ContentType {
```
franziskuskiefer, line 97: This function could be const.

openmls/src/framing/plaintext.rs
```diff
@@ -49,20 +35,82 @@ use tls_codec::{Serialize, TlsByteVecU32, TlsDeserialize, TlsSerialize, TlsSize}
...
+impl From<MlsPlaintext> for MlsContentBody {
+    fn from(plaintext: MlsPlaintext) -> Self {
+        plaintext.content.body
+    }
+}
+
 pub(crate) struct Payload {
-    pub(crate) payload: MlsPlaintextContentType,
-    pub(crate) content_type: ContentType,
+    pub(crate) payload: MlsContentBody,
```
franziskuskiefer, line 113: With the payload being only the `MlsContentBody` now we don't need it anymore.

openmls/src/framing/plaintext.rs
```diff
@@ -241,18 +286,22 @@ impl MlsPlaintext {
...
-    pub(crate) fn content(&self) -> &MlsPlaintextContentType {
-        &self.content
+    pub(crate) fn content(&self) -> &MlsContentBody {
+        &self.content.body
     }
 
     /// Get the content type of this message.
     pub(crate) fn content_type(&self) -> ContentType {
-        self.content_type
+        match self.content.body {
```
franziskuskiefer, line 295: Don't duplicate this code. This is `self.content.body.into()`.

openmls/src/framing/plaintext.rs
franziskuskiefer: You also implement this match at least 3 times. Pleas remove all but one.

franziskuskiefer (changes requested):
Almost there. A few more nits.

openmls/src/framing/plaintext.rs
duesee: Will do. I wanted to keep the changes small. But I guess this still fits in.

openmls/src/framing/plaintext.rs
```diff
@@ -342,12 +375,12 @@ impl TryFrom<u8> for ContentType {
     }
 }
 
-impl From<&MlsPlaintextContentType> for ContentType {
-    fn from(value: &MlsPlaintextContentType) -> Self {
+impl From<&MlsContentBody> for ContentType {
+    fn from(value: &MlsContentBody) -> Self {
         match value {
```
franziskuskiefer, line 380: Isn't this `value.content_type()`? Or the other way round, but it's the same match.

franziskuskiefer (changes requested):
I think you missed one redundant `match`.
Otherwise this looks good to me.

openmls/src/framing/plaintext.rs
duesee: Yeah. I vote to just delete this conversion. Generally, I'm a fan of using `{,Try}{From,Into}` but in this particular case it seems a bit off? What do you think? It is not used in the codebase and -- when I see correctly -- I havn't touched `MlsContentBody::into()` in this PR which means it wasn't used before.

duesee: Hm... can't tell for sure what feels so off about it really. So we can also keep it and I'll just simplify the match.

franziskuskiefer (approved):
Thanks.

This is good to go when @raphaelrobert added his review.

duesee:
@raphaelrobert I could rebase and fixup the commits before you do the review. Depends a bit on your review style (commit-by-commit) or "Files changed", I guess.

raphaelrobert (approved):
Thanks!

Issue #884, opened 2022-04-19
[MLS Spec change] Preconfigured senders extension and removal of signature key indirection (#617)


kkohbrok:
**Description of the changes**
Removes the indirection of signature keys for message verification and introduces a preconfigured senders extension.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/617

---

* [x] #1087
* [x] #1088
* [x] #1112

duesee:
* [ ] Preconfigured senders extension (#884_1)
* [ ] and removal of signature key indirection (#884_2)

PR #1050, opened 2022-10-17
Unambiguous signatures

### LINKED ISSUE #980: [MLS Spec change] Unambiguous signatures #526
> Signatures as specified in the MLS protocol document are ambiguous, in the sense that signatures produced for one purpose in the protocol can be (maliciously) used in a different place of the protocol.
Indeed, we found that for example, a KeyPackage can have the same serialization as MLSPlaintextTBS, which proves that one signature could be accepted for different structures.
> 
> The collisions we have found so far are quite artificial, so it is not super clear how to exploit this, but at least it makes it impossible to have provable security.
> 
> The solution is pretty simple, we simply prefix each signed value with a label, in a fashion similar to ExpandWithLabel.
That way, when calling SignWithLabel you can give the structure name in the Label argument.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/526

- [ ] #957

franziskuskiefer:
This PR fixes #980 

- add `SignContent`
- make `Signable` and `Verifiable` use labels

Some minor other changes working towards #877 
- rename `KeyPackagePaylod` to `KeyPackageTBS`
- fix miscellaneous warnings
- bump dependencies in the fuzzing crate


kkohbrok (changes requested):
Just two nits, otherwise looks good! 

FWIW I dug a bit on why we still need the `verify` (without label) and it turns out that will probably go away when fixing #884.

kkohbrok (approved):
🚢 it!

Issue #928, opened 2022-08-08
[MLS Spec change] Use a hash function for hash identifiers. (#688)


raphaelrobert:
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/688

**Description of the changes**
> Closes #640.

franziskuskiefer:
@nomaxg is taking a stab at this.

I believe you need to comment here for me to be able to assign it to you @nomaxg.

nomaxg:
Yes, work in progress [here](https://github.com/openmls/openmls/compare/main...nomaxg:openmls:hash-identifiers)

PR #1054, opened 2022-10-18
Switch to hash based identifiers (#928)

nomaxg:
https://github.com/openmls/openmls/issues/928

Updated HashReference to be computed using hash functions instead of a KDF. I put a few questions/notes in the PR.


openmls/src/ciphersuite/hash_ref.rs
```diff
@@ -58,45 +78,73 @@ pub type KeyPackageRef = HashReference;
...
 pub type ProposalRef = HashReference;
 
+#[derive(TlsSerialize, TlsSize)]
+struct HashReferenceInput<'a> {
+    label: TlsSliceU8<'a, u8>,
+    value: VLBytes,
+}
+
+/// Compute a new [`ProposalRef`] value for a `value`.
+pub fn make_proposal_ref(
```
nomaxg, line 88: To avoid some type-trickiness involved in passing different labels during KeyPackageRef/ProposalRef construction, I created two new functions `make_proposal_ref` and `make_key_package_ref` that construct references with the correct labels. I am open to other approaches here.

openmls/src/ciphersuite/hash_ref.rs
```diff
@@ -58,45 +78,73 @@ pub type KeyPackageRef = HashReference;
...
+        let value = backend.hash(ciphersuite.hash_algorithm(), &payload)?;
+        Ok(Self {
+            value: Value(VLBytes::new(value)),
+        })
     }
 
     /// Get a reference to the hash reference's value.
-    pub fn value(&self) -> &[u8; 16] {
-        &self.value
+    pub fn value(&self) -> &[u8] {
```
nomaxg, line 127: Now that values are variable length, should we remove `value` entirely?

openmls/src/ciphersuite/hash_ref.rs
```diff
@@ -58,45 +78,73 @@ pub type KeyPackageRef = HashReference;
...
-            LABEL,
-            VALUE_LEN,
-        )?;
-        let value: Value = okm.try_into().map_err(|_| CryptoError::InvalidLength)?;
-        Ok(Self { value })
+        let input = HashReferenceInput {
+            label: TlsSliceU8(label),
+            value: VLBytes::new(value.to_vec()),
+        };
+        let payload = input
```
nomaxg, line 117: Am I constructing the payload correctly? I noticed that this is how we construct `ParentNodeTree` hashes

openmls/src/ciphersuite/hash_ref.rs
```diff
@@ -43,8 +47,24 @@ type Value = [u8; VALUE_LEN];
     TlsDeserialize,
     TlsSerialize,
     TlsSize,
+    Ord,
     PartialOrd,
+)]
+struct Value(VLBytes);
```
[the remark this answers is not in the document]
nomaxg: No good reason! I'm happy to simplify

openmls/src/ciphersuite/hash_ref.rs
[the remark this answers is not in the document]
nomaxg: Yes I agree - I think structs would be ideal but it looked like a much larger refactor

openmls/src/ciphersuite/hash_ref.rs
nomaxg: I was just wondering whether it made sense to have both `fn value()` and `fn as_slice()` since they behave identically now (we can no longer return a fixed size array reference). We could switch to only using `as_slice`.

kkohbrok: Fair point. Let's get rid of `value` then.

kkohbrok (commented):
Sorry about the late reation.

I think once you remove `value` (in favor of `as_slice`) and get rid of the `Value` struct wrapper, this should be good to go.

nomaxg:
Got rid of `value` and the `Value` struct wrapper

kkohbrok (approved):
Thanks for making the changes! This looks good to merge as soon as the CI goes 🟢

EDIT: Looks like there's a call to `value()` in the DS that needs to be turned into `as_slice()`. These are always a bit annoying.

nomaxg:
> Thanks for making the changes! This looks good to merge as soon as the CI goes 🟢
> 
> EDIT: Looks like there's a call to `value()` in the DS that needs to be turned into `as_slice()`. These are always a bit annoying.

Thanks for the review @kkohbrok!

PR #1056, opened 2022-10-19
fix: group's creator could not join the group with an external commit

beltram:
Part of #767 

Any client within a group should be able to create an external commit to rejoin a group. This should create a commit with a `Remove` proposal to remove its previous self.

It was working for all clients except the group's creator (index 0). The problem happened [here](https://github.com/openmls/openmls/blob/main/openmls/src/group/core_group/apply_proposals.rs#L129-L134) because `self.key_package_ref()` will always return the creator's ref, causing `self_removed` to always be true when the external commit was created by the group's creator.

To fix this we simply adapt `CoreGroup#create_commit` not to fail when an external commit does `self_removed`

### DIFF
--- a/openmls/src/group/core_group/create_commit.rs
+++ b/openmls/src/group/core_group/create_commit.rs
@@ -136,7 +136,7 @@ impl CoreGroup {
                     CreateCommitError::OwnKeyNotFound
                 }
             })?;
-        if apply_proposals_values.self_removed {
+        if apply_proposals_values.self_removed && params.commit_type() != CommitType::External {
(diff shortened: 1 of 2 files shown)


raphaelrobert (approved):
Nice catch! External commits are probably still a bit brittle in general because they got in at the last minute before v0.4.

While you are here: could I ask you to join the [Zulip chat](https://openmls.zulipchat.com/) where we can better coordinate external PRs? We made an announcement there about what we will focus on in the next weeks and months.

PR #1060, opened 2022-10-21
docs: Describe creation of `backend` in the book.

duesee:


raphaelrobert:
This seems to break the tests.

raphaelrobert (approved):
Thanks, the user manual clearly lacks a good introduction and this is a good start.

Regarding forced pushes: Doing them after a review has been requested makes it kind of hard for the reviewers because it breaks their local checkout and can also invalidate some review comments (not in this PR though). They should be a last-resort mechanism, normally you can achieve the same effect by squashing when merging.

duesee:
> Regarding forced pushes: Doing them after a review has been requested makes it kind of hard for the reviewers because it breaks their local checkout and can also invalidate some review comments (not in this PR though). They should be a last-resort mechanism, normally you can achieve the same effect by squashing when merging.

Yeah, I had that discussion already with Franziskus :-) In this particular case I thought it makes the initial review more pleasent. But I will be careful with force-pushes and avoid them :+1:

Issue #1053, opened 2022-10-18
Eagerly `derive(TlsSerialize, TlsDeserialize, Size)` when possible.


duesee:
During refactoring it is currently not always straightforward to replace manual implementations. This is why I opened this issue for now.

duesee:
There is no `TODO(1053)` left anymore.

Issue #895, opened 2022-05-12
[MLS Spec change] Ensure that a signature public key is always intelligible #637


franziskuskiefer:
**Description of the changes**
This PR moves the signature key from Credential to LeafNode, to ensure that an MLS client always has a signature public key with which to verify messages, even if it does not support a member's credential type.

This is an alternative approach to https://github.com/mlswg/mls-protocol/pull/628, and IMO a bit cleaner. We no longer need the notion of a SignatureScheme, since BasicCredential doesn't stand alone. And since Credential only appears in LeafNode, common fields can be hoisted up into LeafNode. The Credential struct really only exists as a conceptual boundary, to make the exposition clearer; otherwise you could just shove the select into LeafNode.

Note that this does not obviate the need for https://github.com/mlswg/mls-protocol/pull/631. This PR only ensures that the signatures can be processed; that PR helps ensure that the signatures are meaningful.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/637

franziskuskiefer:
This has been addressed as part of #819 and #1222.
No changes are needed for this spec change but x509 will be implemented in #134 and address any changes introduced in this change.

PR #1061, opened 2022-10-21
refactor: Simplify and strengthen code around `Credential`.

duesee:
Specifically, the invariant that `credential_type` must match `credential` does not need to be maintained anymore.


openmls/src/credentials/codec.rs
```diff
@@ -48,6 +26,12 @@ impl tls_codec::Deserialize for Credential {
     }
 }
 
+// This is implemented manually because we call `SignaturePublicKey::new(...)`
+// and return a `DecodingError` in case when `SignaturePublicKey::new(...)` fails.
+
+// TODOs:
```
raphaelrobert, line 32: If there's a TODO, there should be an issue number.

raphaelrobert (changes requested):
Looks good in general. This might very well be temporary though, as discussed offline. Going forward we might want to be more flexible and support any kind of credential.

openmls/src/credentials/codec.rs
duesee: Hm... I guess I wasn't sure what would be the action this TODO would ask for... Currently, this implementation could be replaced by `derive(TlsDeserialize)`. Unless we want to check the public key at this step? I could open an issue for this check and replace this implementation for now.

duesee:
> Looks good in general. This might very well be temporary though, as discussed offline. Going forward we might want to be more flexible and support any kind of credential.

Hm, yeah. I think this PR still improves the current mechanism so I would be inclined to do the change now and revisit later. What do you think?

openmls/src/credentials/codec.rs
raphaelrobert: Well, I don't know what you had in mind exactly when you wrote the TODO. Whatever it was, that's what should go into the issue. Otherwise, drop the TODO.

The basic credential will sort of drop out anyway, so I wouldn't worry too much.

raphaelrobert (changes requested):
The PR does make sense in general, there was just an open change request.

franziskuskiefer (changes requested):
Please don't touch credential code. This will be completely rewritten and only makes that harder.

duesee:
> The PR does make sense in general, there was just an open change request.

I removed the dangling TODO and referenced #1053.

> Please don't touch credential code. This will be completely rewritten and only makes that harder.

Please say more. What do you propose to do?

franziskuskiefer:
> > Please don't touch credential code. This will be completely rewritten and only makes that harder.
> 
> Please say more. What do you propose to do?

Credentials will change as part of the spec changes. Please don't start doing changes without a spec change issue attached. It is unnecessary work and generates even more unnecessary workload for reviewers.

duesee:
This PR is basically finished and I don't see how it makes anything harder. If you don't want the changes, please close.

raphaelrobert (approved):

franziskuskiefer:
We discussed offline to close this to avoid conflicts with the ongoing work towards #895.

PR #1063, opened 2022-10-24
docs: Remove version from cipher suites in `README.md`.

### LINKED ISSUE #734: [MLS Spec change]  Remove version from ciphersuites (#544)
Remove the version prefix from MLS version identifiers.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/544

duesee:
This (hopefully) closes #734. From my understanding there should be nothing more to do at this point.


raphaelrobert (changes requested):
We have two READMEs that we need to keep in sync (we should really review this):

 - https://github.com/openmls/openmls/blob/main/README.md
 - https://github.com/openmls/openmls/blob/main/openmls/README.md

duesee:
Hm... I remember that there was a syntax that could be used to "redirect" one markdown document to another. But I suppose that this might break either `cargo doc` or the book. Thus, I've changed the other README, too, and synchronized them again.

raphaelrobert (approved):
Thanks for the changes!

PR #1066, opened 2022-10-25
ci: Eagerly cancel superseded workflows.

### LINKED ISSUE #1062: CI: Eagerly cancel unrelevant CI jobs.
We could use ...

```yaml
concurrency: 
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

... to eagerly cancel workflows that are superseded by another workflow. If I get the [documentation](https://docs.github.com/en/actions/using-jobs/using-concurrency) right we could use this to cancel, e.g., outdated workflows when a new commit is made in a PR. I would propose to get that in place to decrease CI work (and time).

I marked this as a question for now. Maybe there is a reason we want to finish these jobs.

duesee:
This closes #1062.

Explanation:

GitHub Actions allows to define a concurrency group -- as a string -- for a workflow. Two workflow runs in the same group will be sequenced by default. By using `cancel-in-progress: true` we cancel the superseded ones instead. This is useful because, often, we commit small changes such as reviewer suggesstions in the review process. These additional commits spawn new workflow runs that run *in addition* to the (now superseded) pending workflow runs.

I propose to cancel these as fast as possible to make the CI more responsive and to not waste computing power.

By putting the workflow into the concurrency group name ...

```yaml
  group: ${{ github.workflow }}-${{ github.ref }}
```

... we make sure that a workflow A does not cancel a workflow B. Thus, cancelations of a workflow should only be done by running that same workflow again.

Exclaimer:

I've done some sanity tests in my repos using this and it seems to work as expected. But I can imagine we will want to tweak this over time.

### DIFF
--- a/.github/workflows/benches.yml
+++ b/.github/workflows/benches.yml
@@ -1,5 +1,9 @@
 name: Benchmarks
 
+concurrency: 
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true 
+
 on:
   push:
     branches:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,9 @@
 name: Build
 
+concurrency: 
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 on:
   push:
     branches:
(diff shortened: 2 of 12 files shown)


kkohbrok (approved):
Sounds sensible. Let's try it out!

PR #1055, opened 2022-10-19
Work towards solving #979 (part 3)

duesee:
* docs: Add presentation language to `MlsCiphertext`.
* Refactor: Sync `MlsMessage` with draft-ietf-mls-protocol-16.
     * Rename `MlsMessage` to `MlsMessageBody`.
     * Introduce `MlsMessage` with (potential room) for `version` field (WIP).
* chore: Add `TODO(#1053)`.
* chore: Formatting.

PR #1069, opened 2022-10-25
Follow up on #902 & #1055

raphaelrobert:
Cleaning up nits that were missed.


kkohbrok (approved):
Thanks! I guess whoever does #979 will enevitably stumble over the `allow`, so this seems fine to me. No strong feelings about the typo, so this seems good to merge for me.

PR #1072, opened 2022-10-26
sync: Remove `ProtocolVersion::{Reserved,Mls10Draft11}`.

### LINKED ISSUE #1065: ProtocolVersion review
- [x] #1073
 - [ ] Remove draft 11

duesee:
This closes #1065.

### DIFF
--- a/openmls/src/ciphersuite/tests/test_secrets.rs
+++ b/openmls/src/ciphersuite/tests/test_secrets.rs
@@ -11,23 +11,10 @@ fn secret_init(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider) {
     // These two secrets must be incompatible
     let default_secret =
         Secret::random(ciphersuite, backend, None).expect("Not enough randomness.");
-    let draft_secret = Secret::random(ciphersuite, backend, ProtocolVersion::Mls10Draft11)
+    let draft_secret = Secret::random(ciphersuite, backend, ProtocolVersion::Mls10)
         .expect("Not enough randomness.");
 
     let derived_default_secret = default_secret.derive_secret(backend, "my_test_label");
     let derived_draft_secret = draft_secret.derive_secret(backend, "my_test_label");
     assert_ne!(derived_default_secret, derived_draft_secret);
-}
-
-#[should_panic]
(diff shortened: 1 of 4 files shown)


raphaelrobert (commented):
Getting rid of the `Reserved` variant would be consistent with other enums, but this PR also gets rid of a useful test. I think the idea was to have versioned secrets. If we shoot for spec compliance now right away, this mechanism might not be needed any more of course.

I would split this in two PRs:
a) Remove the `Reserved` variant
b) Remove the secret versioning altogether, pending further discussion

raphaelrobert (changes requested):
This still looks like you are getting rid of a useful test. You either have to explain why the test is not useful or change the code so that the test still works.

duesee:
Oh, I thought you meant this one https://github.com/openmls/openmls/pull/1072/commits/1bc1b6e34091c7a02e7467f5070ce1057b03a9b9#diff-0ea42bccd03b12e8c930af1f180aab3b8f45ce944af35f40ab16e0f58c906380L24.

Both, `KeyPackageTestVersion::WrongVersion` and `KeyPackageTestVersion::UnsupportedVersion` look to me like they are not useful anymore when the `Reserved` variant is gone. Will look into it.

franziskuskiefer:
> b) Remove the secret versioning altogether, pending further discussion

Drive-by: Right. We'll need to have support for different versions. We have to see if there's something in the spec by now to allow this, or if we do something custom, or suggest something for spec. But we should make sure not to remove  the infrastructure we have for this already.

duesee:
Maybe we can introduce a feature-gated pseudo version? For API consumers it should be when they don't have to deal with invalid versions.

franziskuskiefer:
> Maybe we can introduce a feature-gated (or test-gated) pseudo version? For API consumers it should be better when they don't have to deal with non-existent (currently invalid) versions.

MLS versions are necessarily exposed in APIs. Users should be able to pick the MLS version they want to use. If that's in an API or a configuration is a different question. But it will be exposed.

duesee:
I was just proposing to not expose a variant like `ProtocolVersion::OnlyForInternalUse` and try to hide that. Nothing more.

PR #1074, opened 2022-10-26
Sync: Align the padding mechanism with draft-ietf-mls-protocol-16.

### LINKED ISSUE #913: [MLS Spec change] Make padding arbitrary-size and all-zero (#650)
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/650

> Fixes #642 
 > 
 > Also requires that padding be all-zero.  Before, the sender was allowed to put arbitrary octets in there.

duesee:
This closes #913.

### DIFF
--- a/openmls/src/framing/codec.rs
+++ b/openmls/src/framing/codec.rs
@@ -210,3 +210,66 @@ impl tls_codec::Size for MlsMessage {
         }
     }
 }
+
+impl tls_codec::Size for MlsCiphertextContent {
+    fn tls_serialized_len(&self) -> usize {
+        self.content.tls_serialized_len() +
+           self.signature.tls_serialized_len() +
+            self.confirmation_tag.tls_serialized_len() +
+            // Note: The padding is appended as a "raw" all-zero byte slice
+            // with length `length_of_padding`. Thus, we only need to add
+            // this length here.
+            self.length_of_padding
+    }
+}
+
+impl tls_codec::Serialize for MlsCiphertextContent {
+    fn tls_serialize<W: Write>(&self, writer: &mut W) -> Result<usize, Error> {
+        let mut written = 0;
+
+        written += self.content.tls_serialize(writer)?;
+        written += self.signature.tls_serialize(writer)?;
+        written += self.confirmation_tag.tls_serialize(writer)?;
+        let padding = vec![0u8; self.length_of_padding];
+        writer.write_all(&padding)?;
+        written += self.length_of_padding;
+
+        Ok(written)
+    }
+}
+
+impl tls_codec::Deserialize for MlsCiphertextContent {
+    /// We first decode `content`, `signature`, and `confirmation_tag`, and then make sure
+    /// that the rest of the slice contains only zero bytes, i.e., is the padding.
+    /// Note: This always "terminates" the `Read` instance because we call `read_to_end`.
+    fn tls_deserialize<R: Read>(bytes: &mut R) -> Result<Self, Error>
+    where
+        Self: Sized,
+    {
+        let content = MlsContentBody::tls_deserialize(bytes)?;
+        let signature = Signature::tls_deserialize(bytes)?;
+        let confirmation_tag = Option::<ConfirmationTag>::tls_deserialize(bytes)?;
+
+        let padding = {
+            let mut buffer = Vec::new();
+            bytes
+                .read_to_end(&mut buffer)
+                .map_err(|_| Error::InvalidInput)?;
+            buffer
+        };
+
+        let length_of_padding = padding.len();
+
+        // ValSem011: MLSCiphertextContent padding must be all-zero.
+        if !padding.into_iter().all(|byte| byte == 0x00) {
+            return Err(Error::InvalidInput);
+        }
+
+        Ok(MlsCiphertextContent {
+            content,
+            signature,
+            confirmation_tag,
+            length_of_padding,
+        })
+    }
+}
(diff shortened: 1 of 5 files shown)


duesee:
I've wrote a bit down in the corresponding issue. This PR us marked as a draft for now and there are some TODOs.

openmls/src/framing/codec.rs
```diff
@@ -210,3 +210,66 @@ impl tls_codec::Size for MlsMessage {
...
+
+impl Size for MlsCiphertextContent {
+    fn tls_serialized_len(&self) -> usize {
+        self.content.tls_serialized_len() +
+           self.signature.tls_serialized_len() +
+            self.confirmation_tag.tls_serialized_len() +
+            // Note: The padding is appended as a "raw" all-zero byte slice
+            // with length `length_of_padding`. Thus, we only need to add
+            // this length here.
+            self.length_of_padding as usize
```
franziskuskiefer, line 222: Let's not use unsafe casts. You can do `usize::from` here.

openmls/src/framing/codec.rs
```diff
@@ -210,3 +210,66 @@ impl tls_codec::Size for MlsMessage {
...
+}
+
+impl Serialize for MlsCiphertextContent {
+    fn tls_serialize<W: Write>(&self, writer: &mut W) -> Result<usize, Error> {
+        let mut written = 0;
+
+        written += self.content.tls_serialize(writer)?;
+        written += self.signature.tls_serialize(writer)?;
+        written += self.confirmation_tag.tls_serialize(writer)?;
+        let padding = vec![0u8; self.length_of_padding as usize];
```
franziskuskiefer, line 233: With all the castion I wonder why you picked `u16` for `length_of_padding`? Why don't you just use `usize`?

openmls/src/framing/codec.rs
```diff
@@ -210,3 +210,66 @@ impl tls_codec::Size for MlsMessage {
...
+        writer.write_all(&padding)?;
+        written += self.length_of_padding as usize;
+
+        Ok(written)
+    }
+}
+
+impl MlsCiphertextContent {
+    /// Deserialize an `MlsCiphertextContent` from a slice (with known length).
+    /// Note: We use a slice here and not `impl Read` because we need to infer the padding.
```
franziskuskiefer, line 243: I'm not sure I understand why you can't use `Read`. It provides a `read_to_end` function as well.

openmls/src/framing/plaintext.rs
```diff
@@ -115,6 +115,11 @@ impl MlsPlaintext {
         &self.signature
     }
 
+    #[cfg(test)]
+    pub fn test_signature(&self) -> &Signature {
```
franziskuskiefer, line 119: Note that all this will go away again because the signature doesn't exist anymore.

franziskuskiefer (changes requested):
Thanks, generally looks good but I have a few general questions/remarks.

You can also revert the test_vector change (then you make labeller happy again :)

openmls/src/framing/codec.rs
duesee: Widening casts from `u16` to `usize` are safe. That's why I used `as` here and `try_from` in another place. But I like the idea to avoid `as` in general and can use `from`.

openmls/src/framing/codec.rs
duesee: I picked `u16` because it made the most sense to me for a padding. We do have `opaque ciphertext<V>;` and an implicit padding length now, though. Thus, in praxis, the maximum size of the padding is always 2^30 - the length of the other fields in `MLSCiphertextContent`. Using `u32` would be closer to the spec. `usize` is unnecessary large because it can not exceed 2^30 anyway. Thoughts?

openmls/src/framing/codec.rs
duesee: From my understanding all other deserialization instances are expected to read as many bytes as needed to construct a specific object and leave the `Read` instance operational.

This is different here as this function kinda always "terminates" the `Read` instance due to `read_to_end`. A slice thus felt more appropriate (even though I use a cursor for technical reasons inside the function).

Furthermore, when we change that to `Read` the caller has to wrap a slice in a cursor.

openmls/src/framing/plaintext.rs
duesee: Can you say more? I needed that for testing. What do you propose to do?

openmls/src/framing/codec.rs
franziskuskiefer: There's not really a reason not to use `usize`. It's not needed but you also don't safe anything by using something else.

openmls/src/framing/codec.rs
franziskuskiefer: From what I see you add complexity where it's not needed by using something else. Why do you think you need a cursor?

openmls/src/framing/plaintext.rs
franziskuskiefer: `MLSPlaintext` won't have a signature anymore. You can get it from the `MLSContentAuthData`. It's just a note here. Nothing to do.

openmls/src/framing/codec.rs
franziskuskiefer: With this reasoning we can put everything in an `unsafe` block because we know it's safe. `as` is generally a bad idea for casts that are safe as you have a `from` function in these cases.

openmls/src/framing/codec.rs
duesee: > Why do you think you need a cursor?

Nevermind. No need for a `Cursor` here.

duesee: I refactored to use `tls_codec::Deserialize` and added a note regarding termination of the `Read` instance.

franziskuskiefer (approved):
Thanks! 🚢

duesee:
Cleanup-up and force-pushed. Btw., the "force-pushed" text in ...

image

... is a link! When you click on it it shows what the force-push changed.

PR #1076, opened 2022-10-26
Include ciphersuite in group context

### LINKED ISSUE #952: [MLS Spec change] Include ciphersuite in group context (#720)
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/720

> This will guarantee that all group members agree on the ciphersuite used. It gives more meaning to statements like "Each MLS session uses a single ciphersuite".
 > 
 > Also, it prevents the following attack: the adversary (a malicious insider) invites someone and convinces them that the group uses a ciphersuite with say a signature scheme that is weaker but still compatible with the one actually used. Now the new member uses a weaker scheme and the adversary can impersonate them.

raphaelrobert:
Fixes #952.

The commits should reflect the two sub-tasks.


openmls/src/messages/tests/test_welcome.rs
```diff
@@ -15,10 +16,196 @@ use rstest_reuse::{self, *};
...
+    // Bob tries to join the group
+    let err = MlsGroup::new_from_welcome(
+        backend,
+        &mls_group_config,
+        welcome,
+        Some(alice_group.export_ratchet_tree()),
+    )
+    .expect_err("Error creating group from Welcome");
+
+    assert_eq!(err, WelcomeError::GroupInfoCiphersuiteMismatch);
```
kkohbrok, line 206: If it's not too much effort, it would be great if you could add a happy path in addition to the unhappy path.

raphaelrobert: Added below

kkohbrok (approved):
Thanks for addressing my comments. Let's get this in, then!

Issue #1051, opened 2022-10-18
Enable kat (interop) again


franziskuskiefer:
The interop kat tests are useless while reworking large parts of the code base.
This issue tracks re-enabling them again as soon as the code is more stable again.

- [x] kat_messages #1272 
- [x] kat_encryption
- [x] kat_transcript
- [x] kat_key_schedule

PR #1077, opened 2022-10-27
disable kats 2

franziskuskiefer:
Followup to #1052 to disable transcript and key schedule KATs to avoid churn as in #1074 and #1076.


kkohbrok (approved):
Fair enough. Can you add notes regarding transcript and key schedule kats to #1051?

PR #1082, opened 2022-10-28
Update most TLS syntax to variable length

franziskuskiefer:
Progress towards #877.

It updates all TLS syntax to use variable length encoding except for some places where code will change as part of larger spec changes. Hence this is not closing #877 yet but the remaining work will most likely happen in other issues. I'll keep #877 upon until that's happened.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -41,8 +41,8 @@ jobs:
           crate: cargo-fuzz
           version: latest
       - name: Fuzz Welcome
-        run: cargo fuzz run welcome_decode -- -runs=100000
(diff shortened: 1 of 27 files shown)


franziskuskiefer:
Exercising the variable length encoding more with this one we find some more issues there. I'll look into it (might be just an assert). It probably won't need changes to this PR.

franziskuskiefer:
I fixed the issue in `tls_codec` and reduced the number of fuzzing rounds in https://github.com/openmls/openmls/pull/1082/commits/a747b6bc6a5a191cf687166d42d817ea8b8d34ba (variable-length encoding can allocate a lot of memory and is therefore slow).

raphaelrobert (approved):
Thanks, LGTM

PR #1083, opened 2022-11-01
Add a marker byte to tree hash input structs

### LINKED ISSUE #927: [MLS Spec change] Add a marker byte to tree hash input structs. (#687)
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/687

> Closes #663. I'm not sure if this is actually necessary, but there usually seem to be issues in Merkle trees when the leaves are not explicitly distinct from the parents.

raphaelrobert:
Fixes #927.

Note that there are no tests because this PR only changes the input to a hash, this will be tested during interop.


franziskuskiefer (changes requested):
lgtm with one nit.

franziskuskiefer (approved):
Thanks!

PR #1086, opened 2022-11-02
docs: Add presentation language to `Extension{,Type}`.

### LINKED ISSUE #955: [MLS Spec change] Add `external_senders` to MLS Extension Types registry (#723)
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/723

> `external_senders` was missing from the Extension Types table so it didn't have an extension type value assigned

duesee:
This closes #955.

#955 is an editorial change as it only adds "external_senders" to the extension list.

However, given that we might want to update/improve the presentation language due to an editorial change, I think that we can use this issue as a reason to introduce the presentation language. But we can also just close #955 without linking it to this PR.

### DIFF
--- a/openmls/src/extensions/mod.rs
+++ b/openmls/src/extensions/mod.rs
@@ -55,9 +55,19 @@ use crate::treesync::node::Node;
 #[cfg(test)]
 mod test_extensions;
 
-/// # Extension types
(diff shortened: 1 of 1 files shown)


franziskuskiefer (approved):
Thanks

PR #1090, opened 2022-11-04
use cargo workspaces

beltram:
This enables: 
* running cargo commands on root project. It makes things easier because a change to `openmls` can lead to compilation errors in other crates which you can only spot by either running the CI or `cd` in every folder & `cargo check`
* facilitates using the project on some IDEs (IntelliJ)
* is more idiomatic and could lead to some simplifications in CI scripts
* simpler `patch` declaration just on the root project


Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+    "memory_keystore",
+    "evercrypt_backend",
+    "delivery-service/ds",
+    "delivery-service/ds-lib"
+]
+resolver = "2"
+
+# Patching unreleased crates
+[patch.crates-io.tls_codec]
+git = "https://github.com/RustCrypto/formats.git"
```
franziskuskiefer, line 18: Missing new line.

openmls_rust_crypto/Cargo.toml
```diff
@@ -26,8 +26,4 @@ rand_chacha = { version = "0.3" }
 hpke = { version = "0.1.0", package = "hpke-rs", default-features = false, features = ["hazmat", "serialization"] }
 hpke-rs-crypto = { version = "0.1.1" }
 hpke-rs-rust-crypto = { version = "0.1.1" }
-thiserror = "1.0"
-
-[patch.crates-io]
-openmls_traits = { path = "../traits" }
-openmls_memory_keystore = { path = "../memory_keystore" }
+thiserror = "1.0"
```
franziskuskiefer, line 29: Missing new line.

.github/workflows/fuzz.yml
```diff
@@ -16,10 +16,6 @@ concurrency:
 env:
   CARGO_TERM_COLOR: always
 
-defaults:
-  run:
-    working-directory: openmls
```
franziskuskiefer, line 21: What about the `working-directory`s in the other workflows? We could update them as well and operate on the workspace instead.

franziskuskiefer (changes requested):
Thanks.
I think we can safely move to a workspace here again with the gitignore change.
We'll have to discuss within the core team how to move forward with the demo and helper crates. But for now this is fine.

Please add documentation to contributing.md how to use `.cargo/config.toml` patches for locally overriding dependencies for development.

.gitignore
```diff
@@ -6,6 +6,8 @@ delivery-service/ds/target
 **/Cargo.lock
 *.code-workspace
 /test_vectors/
+fuzz/corpus
+fuzz/artifacts
```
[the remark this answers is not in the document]
beltram: I disagree on this one. `.cargo/config.toml` should be sourced since it can contain command aliases as well as compiler configuration.
I think local patches should go in root `Cargo.toml` and changed to a remote branch when pushed in order for the CI to work

openmls_rust_crypto/Cargo.toml
```diff
@@ -26,8 +26,4 @@ rand_chacha = { version = "0.3" }
 hpke = { version = "0.1.0", package = "hpke-rs", default-features = false, features = ["hazmat", "serialization"] }
 hpke-rs-crypto = { version = "0.1.1" }
 hpke-rs-rust-crypto = { version = "0.1.1" }
-thiserror = "1.0"
-
-[patch.crates-io]
```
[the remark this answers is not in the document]
beltram: They have been moved to the root `Cargo.toml` since anyway [they wouldn't be recognized in a workspace environment](https://doc.rust-lang.org/cargo/reference/workspaces.html#key-features)

.github/workflows/fuzz.yml
beltram: Done. I left that for another PR since it's not easy for me to trigger CI jobs. I kept feature parity because I would like to improve the CI in another PR. For example currently, evercrypt tests are not executed in the CI and I think they should

.gitignore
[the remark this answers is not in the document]
franziskuskiefer: That's not how things work.
`.cargo` is not supposed on version control. If you don't want to this, we won't have a workspace in this repository.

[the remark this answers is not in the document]
beltram: I added .cargo to gitignore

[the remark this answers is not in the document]
OtaK: > .cargo is not supposed on version control. If you don't want to this, we won't have a workspace in this repository.

Some things to keep in mind: 
* The [docs](https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure) mention that .cargo *can* be checked in source control
* From my experience, there are many targets that simply do not compile and/or link without the proper `.cargo/config.toml` configuration. For example, Android targets typically silently fail compiling (i.e. rustc returns 0 but no library is produced) because they require a valid Android NDK to be set as the linker. If you want reproducible builds while targeting "exotic" platforms, you will need to check those files into source control.
* The `[patch]` section is a part of the [Cargo manifest format](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section), not the [configuration format](https://doc.rust-lang.org/cargo/reference/config.html#configuration-format). You **cannot** patch dependencies using `.cargo/config.toml`. 

Otherwise I agree with you that it's *usually* not needed and creates more places to look at configuration, making it unnecessarily complicated :(

franziskuskiefer (changes requested):
Also, looks like CI isn't entirely happy.

.gitignore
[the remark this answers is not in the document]
beltram: Fair point. I did not know you can patch this. It is limited though compared to `[patch]` or as the [documentation](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#paths-overrides) says `As a result usage of a path override is typically isolated to quick bug fixes rather than larger changes`.  

Anyway I'm not opinionated on this, it's also fine not to source `.cargo`. We can move on.

duesee:
What's the state of this? I would like to have that as fast as possible 😺

.github/workflows/ds.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os:
+        arch:
```
duesee, line 24: Huch?

beltram: fixed

cli/Cargo.toml
```diff
@@ -10,15 +10,15 @@ reqwest = { version = "0.11", features = ["blocking", "json"] }
...
 openmls_rust_crypto = { path = "../openmls_rust_crypto" }
 openmls_memory_keystore = { path = "../memory_keystore" }
 
-# Patching unreleased crates
-[patch.crates-io]
-tls_codec = { git = "https://github.com/RustCrypto/formats.git", features = ["derive", "serde_serialize"] }
+[dependencies.termion]
+version = "1.5"
+git = "https://gitlab.redox-os.org/Jezza/termion.git"
+branch = "windows-support"
```
duesee, line 24: Please add a newline.

cli/src/main.rs
```diff
@@ -189,6 +189,7 @@ fn main() {
     }
 }
 
+#[ignore]
```
duesee, line 192: Should this be ignored now?

.github/workflows/ds.yml
duesee: Sorry, could you please change both to be "os"? Not mission-critical but more consistent with other workflows and GitHub actions documentation.

cli/src/main.rs
beltram: Yes it always fails

duesee: When I read the workflow files correctly, it was tested before and hasn't failed? I am wondernig what in this PR makes it fail.

duesee: Ah, there is another #[ignore] here. I'm confused :-)

beltram: Yes another `#[ignore]` was merged from another branch. I've removed one and kept one ;)

.github/workflows/ds.yml
```diff
@@ -30,11 +30,13 @@ jobs:
       - uses: actions/checkout@v2
         with:
           ref: ${{ github.event.pull_request.head.sha }}
+      - uses: actions-rs/toolchain@v1
```
duesee, line 33: Can't we just remove (not add) this whole `uses` statement?

.github/workflows/ds.yml
duesee: Discussed in https://github.com/openmls/openmls/pull/1090#discussion_r1027878324.

.github/workflows/ds.yml
beltram: done

duesee:
Not sure how much I have to say here (yet) :-) But for me this all looks very good :+1: @franziskuskiefer, can you have a quick look when you have time?

franziskuskiefer:
> What's the state of this? I would like to have that as fast as possible 😺

It's pretty low on the priority list. So I'll get around to it but probably not this week.

duesee:
> It's pretty low on the priority list. So I'll get around to it but probably not this week.

It would help going forward with other PRs because it makes refactorings work in certain IDEs across projects. That's one reason I want to have it (and wanted to make a preliminary review). Happy to speed this up when I can.

duesee:
From my understanding, this is ready-to-be-reviewed (if not already ready-to-be-merged). Could we move on here, @openmls/core?

fuzz/Cargo.toml
```diff
@@ -12,11 +12,7 @@ cargo-fuzz = true
 libfuzzer-sys = "0.4"
 
 [dependencies.openmls]
-path = ".."
-
-# Prevent this from interfering with workspaces
-[workspace]
-members = ["."]
+path = "../openmls"
```
franziskuskiefer, line 15: Pulling the fuzzing crate into the workspace isn't great.
With this we are benchmarking the fuzzers when running `cargo bench` in the root directory.  How do you think this should be used?

.github/workflows/docs.yml
```diff
@@ -21,4 +21,4 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: doc
-          args: --manifest-path openmls/Cargo.toml --message-format json
+          args: --message-format json
```
franziskuskiefer, line 24: This looks like we're building all docs now? We don't want to have docs published for the PoC crates.

franziskuskiefer (changes requested):
There are a few more items. Especially pulling out the fuzzing crate is problematic.

Generally, workspaces are not well suited for a project like this because they are pretty limited. I'm fine with getting something like this in for now. But it might change again in future when we notice that it doesn't work well.

The other simple option would be to put the `Cargo.toml` in the root into `.gitignore`. Then everyone can design the workspace to their liking.

cli/Cargo.toml
```diff
@@ -10,15 +10,15 @@ reqwest = { version = "0.11", features = ["blocking", "json"] }
...
 openmls_traits = { path = "../traits" }
 openmls_rust_crypto = { path = "../openmls_rust_crypto" }
 openmls_memory_keystore = { path = "../memory_keystore" }
 
-# Patching unreleased crates
-[patch.crates-io]
-tls_codec = { git = "https://github.com/RustCrypto/formats.git", features = ["derive", "serde_serialize"] }
+[dependencies.termion]
+version = "1.5"
+git = "https://gitlab.redox-os.org/Jezza/termion.git"
```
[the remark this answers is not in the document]
beltram: The cli wasn't compiling anymore on windows because of termion. This requires an unreleased branch of termion. That's what this change is about. Without this, the project wouldn't simply compile on windows

.github/workflows/clippy.yml
```diff
@@ -17,4 +13,4 @@ jobs:
       - uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --manifest-path openmls/Cargo.toml --tests
+          args: --tests
```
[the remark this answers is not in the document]
beltram: Sure. I've restrained it to `openmls` crate. Btw I figured out it was just not contraining (the job never failed even when there was some clippy warnings) so I fixed that if that's ok

fuzz/Cargo.toml
beltram: Nice catch, I've disable `cargo bench` lookup of the fuzzer binaries manually with the appropriate flags (`bench = false` & `harness = false`)

franziskuskiefer (approved):
Thanks for doing the changes.

Let's get this in and see how it works for everyone.

franziskuskiefer (approved):

PR #1091, opened 2022-11-04
Access AAD from UnverifiedMessage

zaynetro:
Implemented as discussed on Zulip: https://openmls.zulipchat.com/#narrow/stream/262730-general/topic/AAD


zaynetro:
Based on the latest discussion we might not want to access AAD from UnverifiedMessage.

PR #1099, opened 2022-11-15
fix: external commit with pure_ciphertext

beltram:
external commit were using group's wire_format which could fail when it was set to pure ciphertext.

This was transitively fixed by #902 with [this line](https://github.com/openmls/openmls/blob/main/openmls/src/group/mls_group/processing.rs#L32) but it could have failed with another MLS implementation.  

Here it is more about adding tests to guard against accidentally changing the wire_format of an external commit.


kkohbrok (approved):
Seems generally pretty straight-forward and looks good to me. Thanks for another follow up to #902!

Issue #916, opened 2022-08-08
[MLS Spec change] Remove AppAck proposal. (#654)


raphaelrobert:
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/654

**Description of the changes**
>

raphaelrobert:
AppAck is only partially implemented so far. Since the AppAck proposal spec moves to `draft-mls-extensions` we might still implement it and therefore we should not remove it from the code just yet.

franziskuskiefer:
I moved this out of spec changes to extension work.

kkohbrok:
When fully implementing AppAck, it should be added as a test case to `test_valsem242`.

franziskuskiefer:
@wysiwys is there anything else that needs to get done or should we close this?

Issue #817, opened 2022-02-16
[MLS Spec change] Forbid proposals by reference in external commits (#562)


kkohbrok:
**Description of the changes**
As the title says.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/562

beltram:
I'd like to be assigned this issue

duesee:
Hm... I can't assign @augustocdias :-) I guess need to comment here first?

augustocdias:
Please assign me this issue.

franziskuskiefer:
To ensure this on the radar for this issue.

So, basically, what I think we should do is to remove the "inline" from all ValSems covering external commits and remark that "inline" is implied by "external commit". 

_Originally posted by @duesee in https://github.com/openmls/openmls/issues/986#issuecomment-1330517758_

There is a working test for ValSem242 that covers inline Add proposals (but not referenced Add proposals.) Thus, I think we can close this issue or link to https://github.com/openmls/openmls/issues/817.

_Originally posted by @duesee in https://github.com/openmls/openmls/issues/986#issuecomment-1330532664_

PR #1104, opened 2022-11-16
feat: use an allowlist for validating external commit proposals

### LINKED ISSUE #878: [MLS Spec change] Allowlist approach for proposals in external commit (#595)
This PR changes the logic of what is allowed in an external commit by explicitly allowing proposals, rather than disallowing all others individually.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/595

beltram:
This PR validates proposals included in an external commit by using an allowlist rather than a denylist as before. Proposals allowed in an external commit are : `ExternalInit`, `Remove` & `PreSharedKey`.
Note that I merged (and rephrased) ValSem242 & ValSem243 and also shifted the others ValSem[244..247]

Note: **There are no tests to verify PSK proposals are supported (and their cardinality) since it is not supported by external commits.** And I couldn't find any issue tracking this. Is this in draft-16 scope ? If not we should probably create an issue to track that down.

Fixes #878

### DIFF
--- a/book/src/message_validation.md
+++ b/book/src/message_validation.md
@@ -83,13 +83,12 @@ The following is a list of the individual semantic validation steps performed by
 
 ### External Commit message validation
 
-| ValidationStep | Description                                                                            | Implemented | Tested | Test File                                                    |
-| -------------- | -------------------------------------------------------------------------------------- | ----------- | ------ | ------------------------------------------------------------ |
-| `ValSem240`    | External Commit must cover at least one inline ExternalInit proposal                   | ✅          | ✅     | `openmls/src/group/tests/test_external_commit_validation.rs` |
-| `ValSem241`    | External Commit must cover at most one inline ExternalInit proposal                    | ✅          | ✅     | `openmls/src/group/tests/test_external_commit_validation.rs` |
-| `ValSem242`    | External Commit must not cover any inline Add proposals                                | ✅          | ✅     | `openmls/src/group/tests/test_external_commit_validation.rs` |
-| `ValSem243`    | External Commit must not cover any inline Update proposals                             | ✅          | ✅     | `openmls/src/group/tests/test_external_commit_validation.rs` |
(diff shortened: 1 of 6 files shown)


openmls/src/messages/proposals.rs
```diff
@@ -265,7 +265,7 @@ impl From<Vec<u8>> for ExternalInitProposal {
     Debug, PartialEq, Clone, Serialize, Deserialize, TlsDeserialize, TlsSerialize, TlsSize,
 )]
 pub struct AppAckProposal {
-    received_ranges: TlsVecU32<MessageRange>,
+    pub(crate) received_ranges: TlsVecU32<MessageRange>,
```
kkohbrok, line 268: Not sure this change belongs in this PR. Or is there something I'm missing here?

kkohbrok (changes requested):
Thanks for picking this up! Looks good to me with just a small nit.

If you're looking for a follow-up issue to tackle, #817 is very close to this one.

openmls/src/messages/proposals.rs
beltram: It's used for building this proposals in tests. I hesitated between covering it or not since anyway AppAck is going to be removed. 
I'll remove AppAck coverage, simpler

kkohbrok (approved):

openmls/src/messages/proposals.rs
kkohbrok: Ah, so I did indeed miss something. I commented [here](https://github.com/openmls/openmls/issues/916#issuecomment-1318570071) to make sure that AppAck is added when/if it is implemented.

franziskuskiefer (changes requested):
Looks like the force push added some files again.

beltram:
> Looks like the force push added some files again

Yes, sorry. Should stop with #1090

kkohbrok (dismissed):
Let's get this in, then. @franziskuskiefer you'll have to r+ as well.

franziskuskiefer (dismissed):

franziskuskiefer:
Looks like there have been changes again in the latest force push. I'll leave it to @kkohbrok to review again.

kkohbrok (approved):

PR #1107, opened 2022-11-17
spec: Use a random group ID by default when creating groups.

duesee:
This PR makes it so that the public-facing API will always create a group with a random group ID. This is in alignment with MLS (draft 16) and makes the API simpler to use.

This PR also updates the book and documentation accordingly. It does not change test code, though. If you feel that we could simplify the tests, too, I can make this change. However, as things might change, I figured it is not the right time to touch the tests.


kkohbrok (commented):
Thanks for taking a stab at this. I have to say that forcing a random group id feels a bit restrictive. Is there a reason why we can't just take an optional GroupId as input upon group creation and generate a random one if none is given? At least that's the first thing that came to my mind when I first looked at the issue.

franziskuskiefer:
> In my mind "by default" implies that there are other options.

Yes, it must still be possible to use other group IDs. Only the "default"/most obvious way of creating one should generate a random one.

duesee:
> I have to say that forcing a random group id feels a bit restrictive. Is there a reason why we can't just take an optional GroupId as input upon group creation and generate a random one if none is given? At least that's the first thing that came to my mind when I first looked at the issue.

Are there good reasons to use custom (non-random) group IDs? (I tend to think of them as client/server_random.) I can imagine that someone wants to plug their own random but this should rather happen through the backend? I can also imagine that someone would like to use a counter. But from my understanding, synchronizing a counter seems difficult and might open up other problems.

If we agree that it makes sense to make this configurable:

We could pass `Option<GroupId>` but I am not sure if this would make things better. It's always a bit weird to expect `None` as default. Also, when we do so, it could still happen that it is confused with a title and we make this parameter more complicated than before without the effect of nudging people into random group IDs.

We could make it configurable by introducing a builder, but, again, this makes things more complicated.

Thus, the easiest way would be to "unguard" `Group::new_with_group_id` and `GroupId::from_slice`?

beltram:
> Are there good reasons to use custom (non-random) group IDs

For my use case this allows us to use a backend generated id which is guaranteed to be unique. Otherwise, it is still doable but it would involve doing retries on the client side when the id conflicts (probably will never happen but more code to write)

> the easiest way would be to "unguard" Group::new_with_group_id and GroupId::from_slice

Yes I personally prefer that over an optional parameter

duesee:
Sounds like a good reason to keep it customizable. I am not sure I understand the implications of colliding group IDs and the operational model of MLS. But I guess the backend has to discard group IDs it doesn't know then? If the group ID is (in practice) managed by the service, I wonder if this PR should have been a no-op?

kkohbrok:
You can always change the group id by doing a re-init, although this can be quite expensive.

There are a variety of reasons why you might want a unique-yet-structured group id. The example @beltram gives for example might you to determine which domain/server "owns" the group and thus gives you information allowing you to route messages and authenticate the responsible server.

duesee:
Okay, thanks for the explanation :-)

duesee:
I wanted to proceed here. @franziskuskiefer, you assigned @kkohbrok. Should we coordinate something before I continue?

franziskuskiefer:
> @franziskuskiefer, you assigned @kkohbrok. Should we coordinate something before I continue here?

Nope, I just assigned him for review.

book/src/user_manual/create_group.md
```diff
@@ -10,14 +10,10 @@ In practice, the supported and required extensions are set by adding them to the
 {{#include ../../../openmls/tests/book_code.rs:create_key_package_bundle}}
 ```
 
-Every group has a unique group ID that needs to be specified at the time of the group creation. The group ID cannot be changed after the group creation and therefore remains immutable throughout the group's lifetime. It should be chosen so that it doesn't collide with any other group IDs in the same system:
```
kkohbrok, line 13: Can you update this text (and/or the note below) to reflect that there are now two constructors and what they do? Ideally, you can also provide examples by using the constructors in the book code.

openmls/src/group/mls_group/creation.rs
```diff
@@ -11,12 +11,32 @@ use super::*;
 impl MlsGroup {
     // === Group creation ===
 
-    /// Creates a new group from scratch with only the creator as a member. This
-    /// function removes the `KeyPackageBundle` corresponding to the
+    /// Creates a new group with the creator as the only member.
```
kkohbrok, line 14: Can you note here that the group id is randomly generated?

kkohbrok (changes requested):
Thanks for implementing the changes! Just two nits regarding the docs, otherwise this looks good to go!

book/src/user_manual/create_group.md
duesee: Ah, yeah. That slipped through :+1:

kkohbrok:
Thanks for implementing the changes! Generally looks good now, but the CI isn't happy.

kkohbrok (approved):

Issue #751, opened 2022-02-01
[Feature request] Expose API for PSKs


raphaelrobert:
Currently PSKs are implemented, but we might need a better API to use them.

Issue #566, opened 2021-11-22
Validate GroupContextExtensions Proposal


franziskuskiefer:

kkohbrok:
Re-opening this at least as long as we don't add GCE proposals into the `valid_proposals` pool in `filter_proposals`. As long as they're not in the pool, they are not included in commits and can't properly be validated. Once that happens, we can include GCE proposals into validation tests, for example when testing for path presence [here](https://github.com/openmls/openmls/pull/1108#discussion_r1031569123).

franziskuskiefer:
This should've been closed by #1475

PR #1108, opened 2022-11-17
feat: path_required proposals

### LINKED ISSUE #879: [MLS Spec change] Require Commit.path by default (#597)
This PR changes the way that the spec indicates if a path is required or not. There has been no change in terms of semantics, I think, but we might want to make sure that we're implementing it the way it's described in the spec.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/597

beltram:
Adds a methods on proposals to determine if the commit they are included into MUST have a path or not.

It can only be used internally using `CoreGroup` since by default partial commits are not supported.

It has to be completed in:
* #916 when/if AppAck proposal are implemented (with their respective [Proposal Validation List](https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#name-proposal-list-validation))
* #751 when ReInit proposal validation are implemented (with their respective [Proposal Validation List](https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#name-proposal-list-validation))
* when `GroupContextExtensions` are implemented (could not find an issue tracking this)

Fixes #879


openmls/src/group/core_group/apply_proposals.rs
```diff
@@ -168,15 +164,16 @@ impl CoreGroup {
...
-        // proposals or if it is a "full" commit. A commit is full if it refers
-        // to proposal types other than Add, PreSharedKey and/or ReInit
-        // proposals.
-        let path_required = has_updates
-            || has_removes
-            // The fact that this is some implies that there's an external init
-            // proposal.
+        let proposals_require_path = proposal_queue
+            .queued_proposals()
+            .any(|p| p.proposal().is_path_required());
```
kkohbrok, line 169: The drawback with this approach is that we have to iterate over all proposals again. As opposed to just setting a flag when we're iterating over them anyway. I don't think it'll have that much of an impact, though, so it's probably fine.

openmls/src/group/tests/test_commit_validation.rs
```diff
@@ -226,217 +251,172 @@ fn test_valsem200(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider
...
-        .process_unverified_message(unverified_message, None, backend)
-        .expect("Unexpected error.")
-    {
-        charlie_group
-            .merge_staged_commit(*staged_commit)
-            .expect("error merging staged commit");
-    } else {
-        panic!("Expected a staged commit.");
+    let add_proposal = || {
+        let debbie_credential = generate_credential_bundle(
```
kkohbrok, line 276: The traditional fourth participant is called Dave, but Debbie works, too 😄.

openmls/src/group/tests/test_commit_validation.rs
```diff
@@ -226,217 +251,172 @@ fn test_valsem200(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider
...
+        (vec![remove_proposal()], true),
+        (vec![gce_proposal()], true),
+        // !path_required + !path_required = !path_required
+        (vec![add_proposal(), psk_proposal()], false),
+        // path_required + !path_required = path_required
+        (vec![remove_proposal(), add_proposal()], true),
+        // path_required + path_required = path_required
+        (vec![update_proposal(), remove_proposal()], true),
+        // TODO: this should work if GCE proposals were implemented
+        // (vec![add_proposal(), gce_proposal()], true),
```
kkohbrok, line 349: Should this be commented in? If not, the TODO should probably link to an issue.

kkohbrok (approved):
Looks very nice! I especially like the improved test setup. Much more methodical than before. Let's get this in.

openmls/src/messages/proposals.rs
```diff
@@ -125,19 +125,31 @@ pub enum Proposal {
...
+            Self::AppAck(ref _r) => ProposalType::AppAck,
+            Self::GroupContextExtensions(ref _r) => ProposalType::GroupContextExtensions,
         }
     }
+
     pub(crate) fn is_type(&self, proposal_type: ProposalType) -> bool {
         self.proposal_type() == proposal_type
     }
+
+    /// Indicates whether a [Commit](crate::prelude::Commit) containing this [Proposal] requires a path
```
kkohbrok, line 143:
```suggestion
    /// Indicates whether a [Commit](crate::prelude::Commit) containing this [Proposal] requires a path.
```

kkohbrok (changes requested):
Sorry, the approve was a little premature. The comment in my previous review about the commented-out code and TODO should probably get a fix.

openmls/src/group/tests/test_commit_validation.rs
beltram: Let's go for Dave 😆

openmls/src/group/core_group/apply_proposals.rs
beltram: Yes I think it's fine, it does not add that much overhead. The biggest problem here would be to miss some proposal type (for example if it's not implemented yet like GCE proposals).

Anyway in this method `proposal_queue.filtered_by_type` already iterates over the proposal queue. I think the `CoreGroup::apply_proposals` method should be rewritten like this to do what you suggest.

```rust
let mut is_path_required = false;
proposal_queue
    .queued_proposals()
    .into_iter()
    .for_each(|p| {
        is_path_required |= p.proposal().is_path_required();
        match p.proposal() {
            Proposal::Add(add_proposal) => {
                todo!()
            }
            _ => todo!(),
        }
    });
```

openmls/src/group/tests/test_commit_validation.rs
beltram: I agree, it should link to issue #566. If you agree and dare reopening it, I'll link it in the TODO

kkohbrok (approved):

PR #1110, opened 2022-11-21
Disable concurrency groups for clippy/coverage

franziskuskiefer:
I suspect that github.ref is from main for pull_request_target jobs such that these jobs get cancelled all the time.

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -4,10 +4,6 @@ on:
   - push
   - pull_request_target
 
-concurrency: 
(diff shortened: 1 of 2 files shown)


kkohbrok (approved):
Sure, let's see if this works.

duesee (approved):
I wonder if we can not use a more specific ref instead of removing the cancelation completely.

franziskuskiefer:
pushed this straight to main without waiting for (broken) CI to unblock the other PRs.

PR #1111, opened 2022-11-21
Struct renaming and small framing bug fix

kkohbrok:
This PR is another PR on the way to #979.

* rename MlsPlaintextTbm -> MlsContentTbm
* rename MlsPlaintextCommitContent -> ConfirmedTranscriptHashInput
* rename MlsPlaintextCommitAuthData -> InterimTranscriptHashInput
* fix small encoding bug, where the content type is encoded as part of MlsCiphertextContent. This is solved by encoding/decoding the enum manually where necessary.

My apologies for the messy commit history, but the changes should be easy enough to review as a whole.


openmls/src/framing/ciphertext.rs
```diff
@@ -445,6 +452,10 @@ impl MlsSenderDataAad {
 /// ```
 #[derive(Debug, Clone)]
 pub(crate) struct MlsCiphertextContent {
+    // The `content` field is serialized and deserialized manually without the
+    // `content_type`, which is not part of the struct as per MLS spec. See the
+    // implementation of `TlsSerialize` for `MlsCiphertextContent`, as well as
+    // `deserialize_ciphertext_content`.
     pub(crate) content: MlsContentBody,
```
franziskuskiefer, line 459: The other option would be to have a different type here without discriminant and this way avoid manual (de)serialisation. But both ways aren't great. So this is fine.

franziskuskiefer (approved):
Thanks.

It's not great that we have to do all this manual (de)serialisation. It might be worth putting this into the tls_codec when we see it more often. But not for this one case.

openmls/src/framing/ciphertext.rs
kkohbrok: Yes, I think both cases lead to a bit of code duplication. Let's wait and see how often we run into these special cases for serialization/deserialization and then try to come up with a somewhat unified solution.

franziskuskiefer (approved):

PR #1114, opened 2022-11-22
feat: Validate that external commits don't contain reference proposals

### LINKED ISSUE #817: [MLS Spec change] Forbid proposals by reference in external commits (#562)

augustocdias:
Closes #817 

I'm taking over the issue from @beltram (we agreed on it).

I've added the validation described in the spec change by checking if any of the proposals in the queue are of type `Reference` and returning an error.
I've also added a test to validate that it is working in the `openmls/src/group/tests/test_external_commit_validation.rs` following the example of the other tests in there.


openmls/src/group/errors.rs
```diff
@@ -339,6 +339,9 @@ pub enum ExternalCommitValidationError {
     /// The remove proposal referenced a non-existing member.
     #[error("The remove proposal referenced a non-existing member.")]
     UnknownMemberRemoval,
+    /// External commit contains referenced proposal
+    #[error("Found a referenced proposal.")]
```
raphaelrobert, line 343:
```suggestion
    #[error("Found a referenced proposal in an External Commit.")]
```

openmls/src/group/tests/test_external_commit_validation.rs
```diff
@@ -6,16 +6,16 @@ use openmls_rust_crypto::OpenMlsRustCrypto;
...
-use rstest::*;
-use rstest_reuse::{self, *};
-
 use crate::{
     ciphersuite::signable::{Signable, Verifiable},
     credentials::{errors::*, *},
     framing::*,
     group::{errors::*, *},
     messages::{proposals::*, public_group_state::VerifiablePublicGroupState},
+    prelude::hash_ref::ProposalRef,
```
raphaelrobert, line 15: `prelude` should not be used in unit tests

openmls/src/group/tests/test_external_commit_validation.rs
```diff
@@ -869,3 +869,84 @@ fn test_pure_ciphertest(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoPr
...
+        &[ciphersuite],
+        bob_credential_bundle.credential(),
+        vec![],
+        backend,
+    )
+    .unwrap();
+
+    alice_group
+        .propose_add_member(backend, &bob_key_package)
+        .unwrap();
```
raphaelrobert, line 901: What does this do?

openmls/src/group/tests/test_external_commit_validation.rs
```diff
@@ -869,3 +869,84 @@ fn test_pure_ciphertest(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoPr
...
+        .sign(backend, &bob_credential_bundle)
+        .unwrap();
+
+    // Set old confirmation tag
+    signed_plaintext.set_confirmation_tag(original_plaintext.confirmation_tag().unwrap().clone());
+
+    let verifiable_plaintext: VerifiableMlsAuthContent =
+        VerifiableMlsAuthContent::from_plaintext(signed_plaintext, None);
+
+    // Have alice process the commit resulting from external init.
```
raphaelrobert, line 929:
```suggestion
    // Have Alice process the commit resulting from external init.
```

raphaelrobert (changes requested):
Thanks for the PR! It looks good in general and should be pretty straightforward since it is similar to other rules. There are a few nits and one question.

openmls/src/group/tests/test_external_commit_validation.rs
augustocdias: This creates a proposal, effectively adding it to the internal ProposalQueue. That way I can reference it when I make a commit with a proposal reference.

raphaelrobert: Right, but that's not great. The goal is to have a mechanism that reliably prevents External Commits from being processed if they reference proposals. In the test you first create such an invalid Commit (which is fine) but then you also manipulate your target so that it gives back the error you want. 
I think the detection can and should be done much earlier for this particular validation step.

augustocdias: But when should it be if not when processing a commit containing such invalid reference? I also didn't understand why the target is being manipulated... The only way to create a reference is when the referenced proposal is already in the queue, otherwise adding this validation is pointless because the reference itself is not valid.

raphaelrobert: Well, the stricter option would be not to allow External Commits that cover proposals by reference, regardless of whether the proposal is valid or not. You could check that on a purely syntactic level, once you have established what kind of a Commit it is (I think the sender type is what determines that).
Your check is implemented at a much later point in time, after the proposals have been processed into a queue already, and references have been correlated with standalone proposals.
I'm not saying that your check is completely wrong, but out of the two options, it is the less strict one.

augustocdias: I see. I'll try to implement according to your suggestion

augustocdias: I moved it to right before the queue is created. Because of this `valsem244` doesn't make sense anymore, so I removed it and renamed this new validation with the same name, so there's no skip between the ids. 

I see that the `valsem246` was removed from the MD file doc but it is still being validated in the code... In some of the conflicts resolution I made it wrong and it was back there... Should I remove or leave it? Isn't the idea to have all validations documented there?

openmls/src/group/errors.rs
```diff
@@ -342,6 +342,9 @@ pub enum ExternalCommitValidationError {
     /// The remove proposal referenced a non-existing member.
     #[error("The remove proposal referenced a non-existing member.")]
     UnknownMemberRemoval,
+    /// External commit contains referenced proposal
+    #[error("Found a referenced proposal in an External Commit.")]
+    ReferencedProposal,
```
raphaelrobert, line 347: You should remove `ReferencedExternalInitProposal` since it's no longer needed.

openmls/src/group/core_group/validation.rs
```diff
@@ -393,29 +393,22 @@ impl CoreGroup {
...
+        if count_external_init_proposals == 0 {
+            // ValSem240: External Commit, inline Proposals: There MUST be at least one ExternalInit proposal.
             return Err(ExternalCommitValidationError::NoExternalInitProposals);
-        };
-
-        // ValSem241: External Commit, inline Proposals: There MUST be at most one ExternalInit proposal.
-        if external_init_proposals.next().is_some() {
-            // ValSem244: External Commit, referenced Proposals: There MUST NOT be any ExternalInit proposals.
+        } else if count_external_init_proposals > 1 {
+            // ValSem241: External Commit, inline Proposals: There MUST be at most one ExternalInit proposal.
```
raphaelrobert, line 408: Was there anything wrong with the original code (except for `ExternalCommitValidationError::ReferencedExternalInitProposal` that's no longer needed)?

openmls/src/group/tests/test_external_commit_validation.rs
```diff
@@ -586,37 +590,32 @@ fn test_valsem244(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider
...
         VerifiableMlsAuthContent::from_plaintext(signed_plaintext, None);
 
-    // Have alice process the commit resulting from external init.
+    // Have Alice process the commit resulting from external init.
     let message_in = MlsMessageIn::from(verifiable_plaintext);
 
     let err = alice_group
         .process_message(backend, message_in)
-        .expect_err("Could process message despite the external commit including an external init proposal by reference.");
+        .unwrap_err();
```
raphaelrobert, line 606: Since you removed the string, can you add some comments instead that explain what we test and what we expect?

raphaelrobert (changes requested):
Thanks for making the changes, I think this is much better.

Just some nits and we should be good to go.

openmls/src/group/tests/test_external_commit_validation.rs
raphaelrobert: > I moved it to right before the queue is created. Because of this `valsem244` doesn't make sense anymore, so I removed it and renamed this new validation with the same name, so there's no skip between the ids.

Good thinking with repurposing valsem244, obviously it's outdated now. 

> I see that the `valsem246` was removed from the MD file doc but it is still being validated in the code... In some of the conflicts resolution I made it wrong and it was back there... Should I remove or leave it? Isn't the idea to have all validations documented there?

Where is valsem246 removed? I still see it in the the [message_validation.md](https://github.com/openmls/openmls/blob/main/book/src/message_validation.md?plain=1#L94)

augustocdias: > Where is valsem246 removed? I still see it in the the [message_validation.md](https://github.com/openmls/openmls/blob/main/book/src/message_validation.md?plain=1#L94)

My mistake. The diff is showing as "readded" that's why I was confused.
image

openmls/src/group/core_group/validation.rs
augustocdias: Nothing wrong, just the logic would become more complicated because of removing that one validation.

raphaelrobert (approved):

PR #1115, opened 2022-11-23
handle conflicting commits

### LINKED ISSUE #939: [MLS Spec change] Resolve conflicting commit messages (#704)
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/704

>

beltram:
No sources modification have been required for this one as 2 commits with the same epoch were already failing when the message was parsed. Anyway, added a test to make sure it happened.  

Then, documented in the book that both commits and welcome messages should be dealt with caution and nothing should be merged locally before the Delivery Service approves the message.

Fixes #939

### DIFF
--- a/book/src/forward_secrecy.md
+++ b/book/src/forward_secrecy.md
@@ -24,6 +24,10 @@ committer is stored in the diff. It exists alongside the key material of the
 ratchet tree before the commit until the client merges the diff, upon which the
 key material in the original ratchet tree is dropped.
 
+Because the client cannot know if the commit it creates will conflict with another commit created by another client
+for the same epoch, it MUST wait for the acknowledgement from the Delivery Service before merging the diff and dropping
+the previous ratchet tree.
+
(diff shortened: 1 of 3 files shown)


raphaelrobert (commented):
Thanks for the PR! Good thinking regarding the test.

In the negative case, only `parse_message()` is used and the subsequent `process_unverified_message()` is omitted. I know it's not really part of this PR, the problem existed before. In #1117 I'm rectifying this. 

If you agree, I think it would be good to rebase this on #1117 (which should land shortly). Alternatively, you should add the two missing `process_unverified_message()` to make sure the negative test case works as intented.

beltram:
+1 for merged parse & process

raphaelrobert (approved):
Thanks for rebasing. Going forward please don't use force pushing, as we already discussed. It forces me to re-review the whole PR, which is not an option for larger PRs.

PR #1116, opened 2022-11-23
Enforce that all extension types in extension list are unique.

duesee:
This PR introduces an `Extensions` type that enforces uniqueness of extension types in the list. Throughout the codebase there is no `Vec<Extension>` anymore (despite in the Wrapper). This PR also already rectifies some cases where we borrow extensions, i.e., `&[Extension]` and directly clone them. This was changed by requiring `Extensions` that must be moved.


openmls/src/extensions/mod.rs
```diff
@@ -195,6 +195,183 @@ pub enum Extension {
...
+                return false;
+            }
+        }
+
+        true
+    }
+}
+
+impl TryFrom<Vec<Extension>> for Extensions {
+    type Error = &'static str;
```
franziskuskiefer, line 303: Don't use string errors. There's an error for this already you should re-use (`InvalidExtensionError`).

openmls/src/extensions/mod.rs
```diff
@@ -195,6 +195,183 @@ pub enum Extension {
...
+    pub fn multi(extensions: Vec<Extension>) -> Result<Self, &'static str> {
+        extensions.try_into()
+    }
+
+    // ---------------------------------------------------------------------------------------------
+
+    /// Add an extension to the extension list.
+    ///
+    /// Returns an error when there already is an extension with the same extension type.
+    pub fn add(&mut self, extension: Extension) -> Result<(), &'static str> {
```
franziskuskiefer, line 231: Use the `InvalidExtensionError` error type everywhere.

openmls/src/extensions/mod.rs
```diff
@@ -195,6 +195,183 @@ pub enum Extension {
...
+            Ok(())
+        } else {
+            Err("Cannot add duplicate extension.")
+        }
+    }
+
+    /// Add an extension to the extension list (or silently replace an existing one.)
+    pub fn add_or_replace(&mut self, extension: Extension) {
+        let _ = self.remove(extension.extension_type());
+        self.add(extension).unwrap();
```
franziskuskiefer, line 243: Don't unwrap, throw an error.

openmls/src/extensions/mod.rs
```diff
@@ -195,6 +195,183 @@ pub enum Extension {
...
+    }
+
+    /// Remove an extension from the extension list.
+    ///
+    /// Returns an error when there is no extension with the given extension type.
+    pub fn remove(&mut self, extension_type: ExtensionType) -> Result<(), &'static str> {
```
franziskuskiefer, line 249: Remove functions are generally expected to return the value they removed. I don't think we need it anywhere right now. But we should probably keep the convention.

franziskuskiefer (changes requested):
Thanks for the PR.
I left some first comments on the `Extension` itself. I didn't look at the rest of the code yet.

openmls/src/extensions/mod.rs
```diff
@@ -195,6 +195,183 @@ pub enum Extension {
...
+            .any(|e| e.extension_type() == extension_type)
+    }
+
+    /// Check that the candidate extension list is valid.
+    ///
+    /// Valid means:
+    ///
+    /// * Does not contain duplicate extension types (ValSem012)
+    pub fn validate(candidate: &[Extension]) -> bool {
+        // We use a [`HashSet`] to identify duplicate extension types by ...
```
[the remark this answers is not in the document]
duesee: Yeah. I used sort + dedup before but it seemed to me that this one is easier to understand and reason about.

openmls/src/extensions/mod.rs
```diff
@@ -195,6 +195,183 @@ pub enum Extension {
...
+#[derive(
+    Debug, Clone, PartialEq, Eq, Serialize, Deserialize, TlsSize, TlsSerialize, TlsDeserialize,
+)]
+pub struct Extensions {
+    inner: Vec<Extension>,
+}
+
+impl Extensions {
+    /// Create an extension list that is empty.
+    pub fn empty() -> Self {
```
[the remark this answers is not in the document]
duesee: As discussed offline, I'll see if making extensions optional is a better was to go about this.

openmls/src/extensions/mod.rs
duesee: This can't fail. I propose to add a `# Safety` note here instead that explains the reason.

openmls/src/extensions/mod.rs
```diff
@@ -195,6 +195,183 @@ pub enum Extension {
...
+        } else {
+            Err("List of extensions must not contain duplicate extension types.")
+        }
+    }
+}
+
+impl Extensions {
+    /// Get a reference to the inner value.
+    pub fn inner(&self) -> &Vec<Extension> {
+        &self.inner
```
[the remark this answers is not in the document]
duesee: This one, I don't understand :-)

openmls/src/extensions/mod.rs
raphaelrobert: Drive-by: We have a strict "no panic" policy that precludes us from using unwraps anywhere.

openmls/src/extensions/mod.rs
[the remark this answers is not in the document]
franziskuskiefer: Ugh, this isn't in the right line. It should be in `application_id`.
There's no safe unwrap. So no unwraps in OpenMLS.
But as I say, you don't need it because you can use `find_map` instead of `find` and `map` then there's no need to catch the error.

openmls/src/extensions/mod.rs
```diff
@@ -195,6 +195,183 @@ pub enum Extension {
...
+    pub fn capabilities(&self) -> Option<&CapabilitiesExtension> {
+        // Safety: `.unwrap()` is safe here.
+        self.inner
+            .iter()
+            .find(|e| e.extension_type() == ExtensionType::Capabilities)
+            .map(|e| e.as_capabilities_extension().unwrap())
+    }
+
+    /// Get a reference to the [`LifetimeExtension`] if there is any.
+    pub fn lifetime(&self) -> Option<&LifetimeExtension> {
```
raphaelrobert, line 366: Why not have a generic lookup function where you have a parameter of type `ExtensionType` to look up the type you want? This is a rather long hard-coded list.

Having a generic lookup would also be a first step in the direction of being able to use proprietary extensions later on.

duesee: Do you mean the generic lookup function as a replacement for the provided methods or as a way to refactor this a bit?

The former, i.e., only having an `ExtensionType -> Extension` function, would require the caller to further call `as_<extension_type>()` and take care of (unnecessary) error handling.

openmls/src/extensions/mod.rs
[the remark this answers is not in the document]
duesee: @franziskuskiefer, what do you think of 13c8be3, and 82234e2?

openmls/src/extensions/mod.rs
raphaelrobert: Discussed offline: The best course of action is to refactor here so that there is no result whatsoever that needs to be handle.

duesee:
I'm closing this, because it will be largely outdated after the key package cleanup. Sorry for the wasted time. On a pro side, there are certainly some things I might be able to cherry-pick later.

Issue #946, opened 2022-08-08
[MLS Spec change] Improve parent hash guarantees (#713)


raphaelrobert:
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/713

**Description of the changes**
> # A nice security property for parent-hash
 > 
 > The goal of #527 was to have a property similar to:
 > 
 > > If there is a parent-hash chain from the leaf L to a node N, then the signature of L binds the tree Canonicalize(N)
 > 
 > (where Canonicalize(N) is N with its unmerged leaves removed, i.e. the subtree of N the last time a Commit went through it)
 > 
 > However, this is currently not true, there are a few places where we can do (hopefully benign) modifications on N.
 > 
 > Example 1:
 > ```
 >               W
 >         ______|_____
 >        /             \
 >       _               Y
 >     __|__           __|__
 >    /     \         /     \
 >   T       _       X       Z
 >  / \     / \     / \     / \
 > A   B   _   _   E   F   G   H
 > ```
 > 
 > with the parent-hash chain A-T-W (happening with filtered update path).
 > 
 > We can modify the blank subtree:
 > 
 > ```
 >               W
 >         ______|_____
 >        /             \
 >       _               Y
 >     __|__           __|__
 >    /     \         /     \
 >   T       V       X       Z
 >  / \     / \     / \     / \
 > A   B   C   D   E   F   G   H
 > ```
 > 
 > and the parent-hash chain A-T-W is still valid
 > 
 > Example 2:
 > 
 > ```
 >               W
 >         ______|_____
 >        /             \
 >       U               Y
 >     __|__           __|__
 >    /     \         /     \
 >   T       V       X       Z
 >  / \     / \     / \     / \
 > A   B   _   D   E   F   G   H
 > ```
 > 
 > with the parent-hash chain A-T-U-W.
 > 
 > We can add unmerged leaves:
 > 
 > ```
 >               W
 >         ______|_____
 >        /             \
 >       U[C]            Y
 >     __|__           __|__
 >    /     \         /     \
 >   T       V[C]    X       Z
 >  / \     / \     / \     / \
 > A   B   C   D   E   F   G   H
 > ```
 > 
 > and the parent-hash chain A-T-U-W is still valid.
 > 
 > # How to get this property
 > 
 > These examples are not possible if we require the parent-hash check to verify other invariants that exists in MLS:
[...]

PR #1120, opened 2022-11-25
Require that unmerged_leaves be ordered

### LINKED ISSUE #947: [MLS Spec change] Require that unmerged_leaves be ordered (#714)
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/714

> While reviewing #713, I realized that we have a potential interop problem around `ParentNode.unmerged_leaves`.  This struct feeds into the tree hash, so everyone in the group needs to have the entries in the same order or else they'll disagree on the tree hash.  But the order isn't specified anywhere: All the document says right now is "add L's leaf index to the unmerged_leaves list", which could be implemented as an insert anywhere in the list.  This PR resolves the ambiguity by requiring the list be sorted.

raphaelrobert:
Fixes #947.

These changes ensure that the list of unmerged leaves of a `ParentNode` is always sorted by acting on the two places where the list is changed:

 - when deserializing a `ParentNode`
 - when adding a new entry to the unmerged leaves list of a `ParentNode`


duesee:
Have you thought about introducing an `UnmergedLeafs` type that guarantees that the list is sorted? Not sure how long unmerged leafs will stay in the standard but it seems easier to get right so we don't forget to do this check?

raphaelrobert:
> Have you thought about introducing an `UnmergedLeafs` type that guarantees that the list is sorted? That would align with the "make good use of the type system" principle and seems easier to get right so we don't forget to do this check?

That was in fact my first thought, to leverage the type safety to get better guarantees. I suspected it would be much more involved than what I ended up doing. I think however that it could be a good improvement going forward, once we are done with the spec compliance. Do you want to create an issue with what you have in mind, and we take it from there?

duesee:
> I think however that it could be a good improvement going forward, once we are done with the spec compliance. Do you want to create an issue with what you have in mind, and we take it from there?

Yeah. I think this is good enough given that there are roughly only these places where we need to be a bit careful. I can take care to create an issue once this is merged :+1:

raphaelrobert:
> > I think however that it could be a good improvement going forward, once we are done with the spec compliance. Do you want to create an issue with what you have in mind, and we take it from there?
> 
> Yeah. I think this is good enough given that there are roughly only these places where we need to be a bit careful. I can take care to create an issue once this is merged 👍

On second thoughts and after digging a bit, there's a chance that #946 and others might mandate changes to unmerged leaves, so I went ahead and extracted a separate type for unmerged leaves as discussed.

duesee:
Awesome 🤩

openmls/src/treesync/node/codec.rs
```diff
@@ -112,7 +116,34 @@ impl tls_codec::Deserialize for ParentNode {
...
         let public_key = HpkePublicKey::tls_deserialize(bytes)?;
         let parent_hash = VLBytes::tls_deserialize(bytes)?;
-        let unmerged_leaves = Vec::tls_deserialize(bytes)?;
+        let unmerged_leaves = UnmergedLeaves::tls_deserialize(bytes)?;
+
         Ok(Self::new(public_key, parent_hash, unmerged_leaves))
     }
 }
+
+impl tls_codec::Size for UnmergedLeaves {
```
duesee, line 125: This could be derived?

openmls/src/treesync/node/codec.rs
```diff
@@ -112,7 +116,34 @@ impl tls_codec::Deserialize for ParentNode {
...
+    fn tls_serialized_len(&self) -> usize {
+        self.list().tls_serialized_len()
+    }
+}
+
+impl tls_codec::Serialize for UnmergedLeaves {
```
duesee, line 131: This could be derived, too?

openmls/src/treesync/node/codec.rs
```diff
@@ -112,7 +116,22 @@ impl tls_codec::Deserialize for ParentNode {
...
     }
 }
+
+impl tls_codec::Deserialize for UnmergedLeaves {
+    fn tls_deserialize<R: std::io::Read>(bytes: &mut R) -> Result<Self, tls_codec::Error>
+    where
+        Self: Sized,
+    {
+        let list = Vec::tls_deserialize(bytes)?;
+        Self::try_from(list).map_err(|e| match e {
```
franziskuskiefer, line 131: I'd generally like to get rid and void any additional deserialization failures that are not due to the serialization.
It's of course some tradeoff because it requires an additional call on the outside and some wrapper struct(s).
I'd be fine with this if you don't want to change it. But we should probably do some nice helpers for something like this (e.g. verify functions that can be called during deserialization) and make use of them here.

franziskuskiefer (approved):
Thanks, lgtm.
I have one comment. But I think it's something we should solve in general and not in this PR.

openmls/src/treesync/node/codec.rs
raphaelrobert: Right, we have this in a few places now. Happy to discuss this if you think combining this is problematic. Personally, I'm not too worried about this, as long as the extra validation is happening separately and not in the codec part of the code. 
I created #1132 to that effect.

franziskuskiefer (approved):

PR #1106, opened 2022-11-17
Use `GroupInfo` instead of `PublicGroupState`.

duesee:
This PR does the following:

* Replace the usage of `PublicGroupState` with `GroupInfo`.
* Reorder extensions to align them with draft-16
  * Note: changed wire representation
* Rename `Extension::LifeTime` to `Extension::Lifetime` (to make it consistent with the `ExtensionType`)
* Introduce the `ExternalPub` extension that holds an `external_pub` HPKE public key that will be used in group_info
* Make the tests compile again/run successfully

Issue #1027, opened 2022-10-11
[MLS Spec change] Remove dangling references to PublicGroupState (#601)


raphaelrobert:
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/601

**Description of the changes**
> PublicGroupState was removed as an independent struct in #529.  This PR fixes some dangling references to it.

---

Part of #719

Issue #719, opened 2022-01-21
[MLS Spec change]  Refactor Group Joining and fix External Inits (#529)


kkohbrok:
**Description of the changes**
PGS and GroupInfo are joined into a single struct. Also fixes a big in External Commit that we had fixed locally before.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/529

---

- [x] #1027

PR #1121, opened 2022-11-28
Use GroupInfo instead of PublicGroupState. (1/2)

duesee:
This PR introduces the `external_pub` extension needed to make progress on #719 and #1027. This is basically the (unchanged) start of closed #1106.

### DIFF
--- a/openmls/src/extensions/external_pub_extension.rs
+++ b/openmls/src/extensions/external_pub_extension.rs
@@ -1,0 +1,29 @@
+use tls_codec::{TlsDeserialize, TlsSerialize, TlsSize};
+
+use super::{Deserialize, Serialize};
+use crate::prelude::HpkePublicKey;
+
+/// ```c
+/// // draft-ietf-mls-protocol-16
+/// struct {
+///     HPKEPublicKey external_pub;
+/// } ExternalPub;
+/// ```
+#[derive(
+    PartialEq, Eq, Clone, Debug, Serialize, Deserialize, TlsSerialize, TlsDeserialize, TlsSize,
+)]
+pub struct ExternalPubExtension {
+    external_pub: HpkePublicKey,
+}
+
+impl ExternalPubExtension {
+    /// Create a new `external_pub` extension.
+    pub fn new(external_pub: HpkePublicKey) -> Self {
+        Self { external_pub }
+    }
+
+    /// Get a reference to the HPKE public key.
+    pub fn external_pub(&self) -> &HpkePublicKey {
+        &self.external_pub
+    }
+}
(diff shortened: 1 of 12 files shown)


openmls/src/extensions/external_pub_extension.rs
```diff
@@ -0,0 +1,29 @@
...
+/// ```c
+/// // draft-ietf-mls-protocol-16
+/// struct {
+///     HPKEPublicKey external_pub;
+/// } ExternalPub;
+/// ```
+#[derive(
+    PartialEq, Eq, Clone, Debug, Serialize, Deserialize, TlsSerialize, TlsDeserialize, TlsSize,
+)]
+pub struct ExternalPubExtension {
```
franziskuskiefer, line 15: Wasn't there a test for this extension? Would be nice to have at least something along the lines of the others in `test_extensions`.

openmls/src/extensions/mod.rs
```diff
@@ -85,30 +88,37 @@ pub enum ExtensionType {
...
+    /// tree.
+    RatchetTree = 2,
+
+    /// The required capabilities extension defines the configuration of a group
+    /// that imposes certain requirements on clients in the group.
+    RequiredCapabilities = 3,
+
+    /// To join a group via an External Commit, a new member needs a GroupInfo
+    /// with an ExternalPub extension present in its extensions field.
+    ExternalPub = 4,
```
franziskuskiefer, line 105: Can you add external sender while you're at it? Then we have all values from the RFC.

openmls/src/extensions/mod.rs
```diff
@@ -85,30 +88,37 @@ pub enum ExtensionType {
     /// Reserved. This must not be used.
     Reserved = 0,
```
franziskuskiefer, line 89: Can you drop this one? It should not be able to instantiate and OpenMLS moved towards not defining the reserved value.

franziskuskiefer (changes requested):
Thanks for splitting this out. Just a few nits.

openmls/src/extensions/external_pub_extension.rs
duesee: No there wasn't. But I will add something :+1:

duesee: Ah, I think you thought about this one https://github.com/openmls/openmls/pull/1085/files#diff-7f947c709686d3816215e94503d38aafeb8990e9a8b60230fedbbc1aed50bfd8R31 I was also confused for a moment :-)

franziskuskiefer (approved):
Thanks!

franziskuskiefer (approved):

Issue #819, opened 2022-02-16
[MLS Spec change] Split LeafNode from KeyPackage (#575)


kkohbrok:
**Description of the changes**
As the title says. Result are two distinct structs for pre-publishing key material and populating leaf nodes in trees.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/575

- [x] #894
- [x] #818 
- [ ] #893
- [x] #968
- [x] #1088

**Tracking required changes**
- [x] Unambiguous signatures (#980)

**Open items**
- [ ] High level API design
- [x] ~~Re-introduce caching of tree hashes~~
- [ ] Avoiding excessive cloning of leaf nodes
- [x] #1140
- [x] Ensure sign/verify flows for leaf nodes is clean and safe
- [x] #727
- [x] #885
- [x] Cleanup extensions
- [x] Cleanup setting of lifetimes for leaf nodes.
- [x] #1226

franziskuskiefer:
I'm closing this now that most of the work is done.
#893 is one open follow up spec change and there are a few things like API design we want to do, but that's either tracked separately or not a spec change.

PR #1125, opened 2022-11-28
Introduce new leaf nodes

### LINKED ISSUE #894: [MLS Spec change] fix the switch condition for LeafNodeSource #632
Fix `LeafNode` struct.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/632

This should be done as part of #819.

### LINKED ISSUE #968: [MLS Spec change] Include leaf index in LeafNodeTBS for better parent-hash guarantees (#731)
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/731

> # A problem in parent-hash proofs
 > 
 > The goal of #527 and #713 was to have a property similar to:
 > 
 > > If there is a parent-hash chain from the leaf L to a node N, then the signature of L binds the tree Canonicalize(N)
 > 
 > In #713, I proved that the pull-request allowed to prove the following property:
 > 
 > > If Canonicalize(U1) = Canonicalize(U2),
 > > and there is a parent-hash link from U1 to P1 and from U2 to P2,
 > > then Canonicalize(P1) = Canonicalize(P2)
 > 
 > With this property, the hope is to inductively show that the leaf signature binds the canonicalization of subtrees it is parent-hash linked from.
 > The induction step works fine thanks to the previous theorem, however the base case is wrong.
 > 
 > The hypothesis "Canonicalize(U1) = Canonicalize(U2)" says that the content of the trees are equal, and also says they have the same location (i.e. they have the same node index). This is important for the induction step.
 > However, since the signature of a leaf doesn't bind its position in the tree, we can't initialize the induction.
 > 
 > This PR resolves this problem.
 > 
 > # A concrete counter-example
 > 
 > Assume we have the following tree:
 > 
 >           Y
 >         __|__
 >        /     \
 >       _       Z
 >      / \     / \
 >     A   _   C   D
 > 
 > If there is a parent-hash link from A to Y, then the following tree is also valid:
 > 
 >           Y
 >         __|__
 >        /     \
 >       _       Z
 >      / \     / \
 >     _   A   C   D
 >     
 > # The solution
 > 
 > We only need to add the leaf index inside the signature when the source is a commit.
 > I also added it in the update case because it doesn't hurt, but that's not necessary.

franziskuskiefer:
This PR introduces the new `LeafNode`s.
Sorry, it's a bit large and touches a lot of code. It also needs a number of follow ups. Here some thoughts to guide you through the review.

* Creating new group API requires credential bundle and lifetime now. I think this should be made nicer as an API. But it should be done in a follow up.
* This is because the node is signed now and the lifetime is in the node.
* The OpenMlsLeafNode is the leaf node the code works on, which includes optional secret values and the leaf index. The LeafNode is the one from the RFC.
* The leaf node structs are already a mess and I think they could get cleaned up a little. But the way they are defined in the RFC is a little awkward because the Tbs gets information the leaf doesn't. This should probably be improved in a follow up.
* When changing the path, the node is updated in the diff directly instead of added later. This requires some chances to everything that's computed on the diff. But I think this is cleaner.
* I removed caching of the tree hash because the way the diff is modified now breaks the caching. We should look at adding it again but make sure that it actually provides a performance benefit.
* The key package wasn't cleaned up yet. A lot of the code will be deleted.
* I left some XXX in as potential follow ups. But nothing that needs addressing now.
* I didn't solve the question who's setting the capabilities. This can be passed in every time. But it adds yet another argument. We can have a builder for leaves and use a default or put it into a config.
* I had to touch a lot of tests. I hope I didn't break too many ...

Fixes #894
Fixes #968

Follow-up items are collected in #819.


openmls/src/group/core_group/mod.rs
```diff
@@ -186,17 +194,36 @@ impl CoreGroupBuilder {
     /// [`OpenMlsCryptoProvider`].
     pub(crate) fn build(
         self,
+        credential_bundle: &CredentialBundle,
+        life_time: LifetimeExtension,
```
kkohbrok, line 198: Passing in a LifetimeExtension could probably be an optional build step `with_lifetime_extension`, where by default we just use the default one.

openmls/src/group/core_group/test_proposals.rs
```diff
@@ -469,28 +480,31 @@ fn test_group_context_extension_proposal_fails(
...
         e,
-        CreateGroupContextExtProposalError::KeyPackageExtensionSupport(
-            KeyPackageExtensionSupportError::UnsupportedExtension
+        CreateGroupContextExtProposalError::TreeSyncError(
+            crate::treesync::errors::TreeSyncError::UnsupportedExtension
         )
     );
 
-    // Well, this failed luckily.
+    // // Well, this failed luckily.
```
kkohbrok, line 507: This change should probably be reversed.

openmls/src/group/mls_group/creation.rs
```diff
@@ -20,13 +20,17 @@ impl MlsGroup {
     pub fn new(
         backend: &impl OpenMlsCryptoProvider,
         mls_group_config: &MlsGroupConfig,
+        life_time: LifetimeExtension,
         key_package_hash: &[u8],
+        credential_bundle: &CredentialBundle,
```
kkohbrok, line 25: As discussed OOB, this could probably be pulled from the key store.

openmls/src/group/mls_group/creation.rs
```diff
@@ -20,13 +20,17 @@ impl MlsGroup {
     pub fn new(
         backend: &impl OpenMlsCryptoProvider,
         mls_group_config: &MlsGroupConfig,
+        life_time: LifetimeExtension,
```
kkohbrok, line 23: This is an API choice in the end, so @raphaelrobert probably has an opinion, but since we're creating a new group, we can just go with the default. If I understand correctly, the extension is just used in the founder's initial leaf KeyPackage, which is not going to expire anyway.

openmls/src/treesync/mod.rs
```diff
@@ -441,24 +463,52 @@ impl TreeSync {
...
         // Our own leaf should be inside of the tree and never blank.
         self.leaf(self.own_leaf_index)?
             .ok_or_else(|| LibraryError::custom("Own leaf is outside of the tree").into())
     }
 
+    /// Returns the [`LeafNode`] of this client.
+    ///
+    /// This function should not fail and only returns a [`Result`], because it
+    /// might throw a [LibraryError](TreeSyncError::LibraryError).
+    pub(crate) fn own_leaf_node_mut(&mut self) -> Result<&mut OpenMlsLeafNode, TreeSyncError> {
```
kkohbrok, line 476: This kind of breaks the pattern where all changes to the TreeSync instance have to happen by merging a diff.

openmls/src/group/mls_group/updates.rs
```diff
@@ -33,10 +36,20 @@ impl MlsGroup {
 
         // Create Commit over all proposals. If a `KeyPackageBundle` was passed
         // in, use it to create an update proposal by value. TODO #751
-        let create_commit_result = match key_package_bundle_option {
-            Some(kpb) => {
+        let create_commit_result = match key_pair {
+            Some(key_pair) => {
+                let group_id = self.group_id().clone();
+                let own_leaf = self.group.treesync_mut().own_leaf_node_mut().map_err(|_| {
```
kkohbrok, line 42: As noted at the `own_leaf_node_mut` definition, this might be problematic. In particular, here we change the tree before the commit has been merged. I'm actually curious why this doesn't lead to problems when a commit is created but not merged. But maybe I'm overlooking something.

openmls/src/group/tests/test_proposal_validation.rs
```diff
@@ -1551,20 +1590,38 @@ fn test_valsem109(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider
...
+        .tls_serialize_detached()
+        .expect("Error serializing signature key.");
+    // Store the credential bundle into the key store so OpenMLS has access
+    // to it.
+    backend
+        .key_store()
+        .store(&credential_id, &new_cb)
+        .expect("An unexpected error occurred.");
+
+    // let new_kp = backend.crypto().derive_hpke_keypair(
```
kkohbrok, line 1605: This and the commented-out code below can probably be deleted.

openmls/src/group/tests/test_proposal_validation.rs
```diff
@@ -1706,11 +1752,36 @@ fn test_valsem110(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider
...
+        .rekey(
+            bob_group.group_id(),
+            ciphersuite,
+            ProtocolVersion::Mls10,
+            &bob_credential_bundle,
+            backend,
+        )
+        .unwrap();
+
+    // update_leaf_node.set_public_key(
```
kkohbrok, line 1766: Again, probably remove here and a few lines below.

kkohbrok (changes requested):
Thanks for tackling this big one! Looks generally very good to me, although I'm a bit dubious about the direct TreeSync modification as per comments. It's probably faster to discuss some of the issues OOB, so feel free to ping me.

openmls/src/group/mls_group/creation.rs
raphaelrobert: I think the lifetime should be part of the `MlsGroupConfig`.

openmls/src/group/core_group/test_proposals.rs
kkohbrok: Again? Just kidding. No need to fix this one.

kkohbrok (approved):
Thanks for implementing the changes and, again, for dealing with this massive change! Let's leave the rest for follow-ups and get this in.

kkohbrok:
Not sure why this turned in to a double r+, but there you go.

PR #1127, opened 2022-11-28
Use GroupInfo instead of PublicGroupState. (2/2) 

### LINKED ISSUE #719: [MLS Spec change]  Refactor Group Joining and fix External Inits (#529)

### LINKED ISSUE #1027: [MLS Spec change] Remove dangling references to PublicGroupState (#601)

duesee:
* Closes #1027
* Closes #719

This PR is very similar to the second half of #1106 but adds an additional `with_ratchet_tree` flag to `export_group_info`.


openmls/src/messages/mod.rs
```diff
@@ -305,6 +317,71 @@ impl SignedStruct<GroupInfoTBS> for GroupInfo {
...
+    fn unsigned_payload(&self) -> Result<Vec<u8>, tls_codec::Error> {
+        self.payload.tls_serialize_detached()
+    }
+
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn label(&self) -> &str {
+        "GroupInfoTBS"
```
franziskuskiefer, line 365: Use `SIGNATURE_GROUP_INFO_LABEL` here as well.

openmls/src/group/core_group/new_from_external_init.rs
```diff
@@ -117,14 +125,30 @@ impl CoreGroup {
         );
         let message_secrets_store = MessageSecretsStore::new_with_secret(0, message_secrets);
 
+        let interim_transcript_hash = {
+            if group_info.group_context().epoch() == GroupEpoch::from(0) {
```
franziskuskiefer, line 129: I'm not sure this can ever happen.
Joining a group by external commit implies that there is a group already. The commit that is generated for this join moves the epoch forward. I therefore don't see how the epoch can be 0 here.
Is this actually used? Shouldn't we throw an error instead (if the above is correct)?

openmls/src/group/core_group/new_from_external_init.rs
```diff
@@ -117,14 +125,30 @@ impl CoreGroup {
...
+            } else {
+                // New members compute the interim transcript hash using
+                // the confirmation_tag field of the GroupInfo struct.
+                update_interim_transcript_hash(
+                    ciphersuite,
+                    backend,
+                    &InterimTranscriptHashInput::from(group_info.confirmation_tag()),
+                    group_info.group_context().confirmed_transcript_hash(),
+                )
+                .unwrap()
```
franziskuskiefer, line 140: Throw a library error instead of unwrapping.

openmls/src/messages/tests/test_pgs.rs
```diff
@@ -1,98 +1,42 @@
...
 
-use crate::{
-    credentials::*,
-    framing::*,
-    group::*,
-    key_packages::*,
-    messages::{public_group_state::*, *},
-    test_utils::*,
-};
+use crate::{credentials::*, key_packages::*, messages::*, test_utils::*};
```
franziskuskiefer, line 3: This file should probably get a new name.

openmls/src/messages/tests/test_pgs.rs
```diff
@@ -1,98 +1,42 @@
...
-    let verifiable_pgs = VerifiablePublicGroupState::tls_deserialize(&mut encoded.as_slice())
-        .expect("Could not decode");
-
-    let pgs_decoded: PublicGroupState = verifiable_pgs
+    let _: GroupInfo = verifiable_group_info
         .verify(backend, alice_credential_bundle.credential())
-        .expect("error verifiying public group state");
-
-    assert_eq!(pgs, pgs_decoded)
+        .unwrap();
```
franziskuskiefer, line 41: We have the convention to use `expect` in tests for cases, at least where the checks happen.
This also serves as documentation for what the test is checking.

franziskuskiefer (changes requested):
Thanks. Generally looks good.
A few items and questions inlien.

openmls/src/messages/mod.rs
```diff
@@ -283,6 +282,19 @@ impl GroupInfo {
     ) -> Result<Self, LibraryError> {
         self.payload.sign(backend, credential_bundle)
     }
+
+    #[cfg(test)]
+    pub(crate) fn into_verifiable_group_info(self) -> VerifiableGroupInfo {
```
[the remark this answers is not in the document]
duesee: `into_verifiable_group_info` is used in tests, see [https://github.com/openmls/openmls/pull/1127/files#diff-d1d10ba424715e9e5529dc7069797c74a8d946448d844a9ab0bd994ab2e88352R113] or [https://github.com/openmls/openmls/pull/1127/files#diff-d1d10ba424715e9e5529dc7069797c74a8d946448d844a9ab0bd994ab2e88352R185].

What was done before in the `PublicGroupState` tests was to serialize and deserialize it. But I found this code to be distracting and (if it was for this reason)  (de)serialization should be tested more systematically elsewhere.

openmls/src/group/core_group/mod.rs
```diff
@@ -526,6 +527,48 @@ impl CoreGroup {
...
+
+            if with_ratchet_tree {
+                vec![ratchet_tree_extension(), external_pub_extension()]
+            } else {
+                vec![external_pub_extension()]
+            }
+        };
+
+        // Create to-be-signed group info.
+        let group_info_tbs = GroupInfoTBS::new(
```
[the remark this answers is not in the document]
duesee: Yeah. That's generally something I would like to change over the time (but havn't done in this PR for now.) Whenever we need ownership of an object, we should let the caller decide how to obtain ownership and not take references and clone internally. https://rust-lang.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control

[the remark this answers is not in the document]
duesee: The next PR in the pipeline will change that. https://github.com/openmls/openmls/pull/1116/files#diff-e7948aa030a226922756322ba4c0e6748caf5fcd80cf8dcd4d11cf92f3d73cb1R210 Do you still want to make the change now or can we wait for #1116 to land?

openmls/src/messages/mod.rs
```diff
@@ -305,6 +317,71 @@ impl SignedStruct<GroupInfoTBS> for GroupInfo {
     }
 }
 
+/// A type that represents a group info of which the signature has not been verified.
+/// It implements the [Verifiable] trait and can be turned into a group info by calling
```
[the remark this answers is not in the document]
duesee: I'm thinking if it would not be easier to open an issue for things like that and do a single PR that rectifies all of this close to some release. I will rectify it here, though. Just a thought.

openmls/src/group/core_group/new_from_external_init.rs
duesee: Without this condition, tests fail with a `InvalidCommit(ConfirmationTagMismatch)`. But, to be honest, I am not entirely sure the draft tells us what to do in this case.

In the failing tests, alice creates a group first (with no other members).

According to draft-16, the group should have ...

> Initialize a one-member group with the following initial values:
> [...]
> epoch: 0
> [...]
> Confirmed transcript hash: The zero-length octet string
> Interim transcript hash: The zero-length octet string

(In OpenMLS, the `interim_transcript_hash` is set to `b"\x00" * 32` and `confirmed_transcript_hash` is non-zero. I assume the `confirmed_transcript_hash` is correct.)

Then, alice exports a group state and bob joins by an external commit. Here, the epoch is 0 (and triggers the above condition).

Now, I am not sure, how the `interim_transcript_hash` should be calculated really. From my understanding of draft-16, I would expect the hash of a zero-length string (empty interim || empty confirmation). However, debugging the "old code" (with `PublicGroupState`) it seemed to me that `b""` was used instead. Given that all tests pass having this condition, I assumed it to be fine.

Could you shed some light on this`

openmls/src/messages/mod.rs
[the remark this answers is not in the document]
duesee: Sorry, I still don't understand. From my understanding this is a method available throughout the crate during testing only. And it should be only available during tests, because, in real life, clients will deserialize a `VerifiableGroupInfo` and not create it from a `GroupInfo`.

> Is there an integration test covering this? Not sure if I've seen one.

Sorry, again. I don't understand what "this" refers to.

openmls/src/group/core_group/new_from_external_init.rs
duesee: I've created https://github.com/openmls/openmls/issues/1129.

openmls/src/messages/mod.rs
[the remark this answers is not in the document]
duesee: As discussed OOB, I've added an integration test to see how a member can join by using `export_group_info`, and `join_by_external_commit`.

openmls/src/group/core_group/new_from_external_init.rs
franziskuskiefer: Can you link the issue here in the code as well?

openmls/src/messages/mod.rs
```diff
@@ -305,6 +317,71 @@ impl SignedStruct<GroupInfoTBS> for GroupInfo {
...
+        self.payload.group_context.ciphersuite()
+    }
+
+    /// Get signer of the verifiable group info.
+    pub(crate) fn signer(&self) -> u32 {
+        self.payload.signer
+    }
+
+    /// Get extensions of the verifiable group info.
+    pub(crate) fn extensions(&self) -> &[Extension] {
```
franziskuskiefer, line 344: Accessing values in the verifable object is always a little tricky because the idea is of course that you shouldn't get access until it's verified.
But in many cases, like this one, you need to get to some values in order to actually verify.
Can you add the reason for each of these getters why they are needed so no one has to guess in the future?

franziskuskiefer (changes requested):
Thanks for the test!
Generally looks good to me. Just a few nits, then this is good to go.

openmls/src/messages/mod.rs
duesee: Very good point. Maybe we can also go a step further and make it even more explicit by using a `_unverified` suffix for these?

openmls/src/group/core_group/new_from_external_init.rs
```diff
@@ -71,40 +63,56 @@ impl CoreGroup {
...
-        if treesync.tree_hash() != verifiable_public_group_state.tree_hash() {
+        let group_info: GroupInfo = {
+            let group_info_signer_leaf = treesync
+                .leaf(verifiable_group_info.signer())
+                .map_err(|_| ExternalCommitError::UnknownSender)?
+                .ok_or(ExternalCommitError::UnknownSender)?
+                .key_package()
+                .credential();
+
+            Verifiable::verify(verifiable_group_info, backend, group_info_signer_leaf)
```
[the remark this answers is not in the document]
duesee: Type inference :P

openmls/src/messages/mod.rs
duesee: Ping @franziskuskiefer. (Will re-request a review later.)

openmls/src/group/core_group/new_from_external_init.rs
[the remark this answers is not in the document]
franziskuskiefer: `verifiable_group_info.verify` works just fine 🙂

openmls/src/messages/mod.rs
duesee: Alright :+1:

franziskuskiefer (approved):
Thanks, lgtm.

> Rebase?

I'd just squash and merge.

duesee:
Sorry, had to reformat the code which dismissed the review.

franziskuskiefer (approved):

franziskuskiefer (approved):

Issue #1126, opened 2022-11-28
Design: Organize Validation.


duesee:
An MLS implementation needs to perform a variety of syntactical, semantical, and cryptographic validations. Currently, these tasks are tracked as [ValSemXXX values](https://openmls.tech/book/message_validation.html). However, it is not obvious what checks could be missing (or are too tight). Ideally, we want to have tooling in place to extract, collect, cluster, and reference validation tasks from MLS in OpenMLS. This could be done, e.g., by annotating the MLS draft and cross-referencing to the OpenMLS implementation (or tests that ensure compliance).

This issue is a reminder to come up with a proposal how to do that.

Notes:

* [x] ~HackMD~
* [x] ~https://github.com/doccano/doccano~
* [x] ~Taguette~
* [x] ~CATMA~
* [x] ~Atlas.ti~
* [x] ~Dovetail~
* [ ] https://www.notion.so/desktop
* [ ] https://github.com/CriticMarkup/CriticMarkup-toolkit
* [ ] https://github.com/icann/rfc-annotations
* [ ] Google Docs

duesee:
We've created an experimental tool here: https://github.com/openmls/annotations

Let's see how this works and close this issue for now.

PR #1133, opened 2022-12-02
test refactoring proposal

beltram:
Here's my view and proposition on how to refactor openmls tests.

This is not aimed to be merged ! It's just a nicer way to comment and debate.

Now let's DEBATE this ! 😆


openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,172 @@
...
+
+To achieve these goals, here are the "rules"/recommendations to set up:
+
+1. Any test expected to fail by altering a valid input should also keep this valid input and verify that, contrary to
+   the
+   tampered one, it succeeds.
+2. All errors should be namely asserted i.e. `assert!(result.is_err())` is prohibited. When matching an error, one
+   should `assert!(matches!(result.unwrap_err(), MlsError::MyError))` because errors are not supposed to impl Eq (it can
+   impose hard to fulfill constraints on the data they wrap)
+3. Do not use `.expect("...")` but only use `.unwrap()`. The former adds too much overhead for too little value. Another
```
duesee, line 42: In some cases, you can really feel how motivated the creator of a test was about using `expect` instead of `unwrap`: `.expect("An unexpected error occurred.")` 🙈 

So, I generally agree. But let's not phrase it this strict: Does the extra information in `expect` help to debug the problem? If yes, go for it. If no, maybe use `unwrap`. Let's just be sensible about what we use when.

duesee: There are also different [message styles](https://doc.rust-lang.org/std/error/index.html#common-message-styles) for `expect`.

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,172 @@
...
+## Bonus
+
+* We could use mutation testing to "test out tests". A good crate available for doing that
+  is [mutagen](https://github.com/llogiq/mutagen). After annotating all the methods in the [`tree` mod](src/tree) it
+  gives the [following result](https://github.com/beltram/openmls/blob/mutation/openmls/MUTATION.txt) with ~= 1/3 of
+  mutants surviving i.e. not rightfully tested
+* If WASM is a target we could already use [wasm-bindgen-test](https://crates.io/crates/wasm-bindgen-test) macros on all
+  tests methods (no async migration required at that point, only a few things to fix: rayon, time, rng)
+* Use [nextest](https://crates.io/crates/cargo-nextest) to execute test in the CI because it is faster (19s vs 30s for
+  cargo test)
```
duesee, line 172: Add to bonus: Implement `Arbitrary` for every type and use structured fuzzing.

openmls/TEST-DESIGN.md
beltram: Okay to cool down the rule :)

duesee: Maybe we should first discuss this before you invest the effort to make changes. The above is only how I think about this and maybe others view it differently :-)

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,175 @@
...
+       ```rust
+       let (mut alice_group, ..) = MlsGroup::try_init(&case, "Alice").unwrap();
+       let [(mut bob_group, ..), (mut charly_group, ..)] = alice_group.try_invite(&case, ["Bob", "Charly"]).unwrap();
+       ```
+    3. `MlsGroup::try_talk_to<const N: usize>(&mut self, case: &TestCase, others: &mut [MlsGroup ; N])` e.g.
+       ```rust
+       let (mut alice_group, ..) = MlsGroup::try_init(&case, "Alice").unwrap();
+       let [(mut bob_group, ..), (mut charly_group, ..)] = alice_group.try_invite(&case, ["Bob", "Charly"]).unwrap();
+       assert!(alice_group.try_talk_to(&case, &mut [bob_group, charly_group]).is_ok());
+       ```
```
kkohbrok, line 74: I tried to write a test framework at some point, although the result is arguably not great and can/should be revisited. In any case, I agree that we need something along those lines.

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,175 @@
...
+11. Fixtures: in order for being more flexible, they could just require a single `case` variable. It would look like
+    this:
```
kkohbrok, line 76: This relates to point number 6. As long as this allows us to cover a range of cases where necessary, I have no objections.

kkohbrok (commented):
Thanks for writing this up! Looks good so far. Just a few comments.

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,175 @@
...
+   should `assert!(matches!(result.unwrap_err(), MlsError::MyError))` because errors are not supposed to impl Eq (it can
+   impose hard to fulfill constraints on the data they wrap)
+3. Prefer `.unwrap()` over `.expect("...")`. The latter adds overhead and most of the time does not add enough debugging
+   information. If, on the opposite, an `.expect(...)` can help a developer troubleshooting a failing test with a very
+   complex setup which this `expect` message could help understanding then go for it ! Another
+   argument [here](https://twitter.com/timClicks/status/1584676737572487169).
+4. It is okay (recommended) to `use crate::prelude::*;` in all test mod (to debate ?)
+5. Have a public test framework in each top level mod inside `src/{module}/test_utils(.rs)`. Avoid scattered utils,
+   factorize them.
+6. Each test should test one and only one thing. It should not iterate over multiple cases.
```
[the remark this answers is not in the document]
beltram: No I have nothing against test fixtures, quite the opposite actually. It is more about [this kind](https://github.com/openmls/openmls/blob/9d101db6be14fbe5b28d78d40f4ec17d20f2af53/openmls/src/group/tests/test_remove_operation.rs#L12-L17) of pattern

openmls/TEST-DESIGN.md
beltram: Most importantly, it's extensible. For example, once we introduce x509 Credentials, it's likely to become another fixture. By doing this, we don't have to change all the test methods signature

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,175 @@
...
+7. test method names should
+    1. follow the template `should_{expected_behaviour}_when_{state_tested}`
+       e.g. `should_succeed_when_...` & `should_fail_when_...` and use natural language
+    2. not be prefixed by `test_`
+    3. `ValSem` should not be in the test method name but kept in comment
+8. There should not be a `test-utils` feature (best effort). All those helpers should be `#[cfg(test)]` instead and used
+   only in unit tests
```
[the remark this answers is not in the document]
beltram: Okay then what about a clear segregation between tests and interop requirements ? Anyway, most of the time it ends up in `#[cfg(any(feature = "test-utils", test))]`
* rename ~~`test-utils`~~ -> `interop`
* Do not use `interop` feature in unit/integration tests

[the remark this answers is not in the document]
kkohbrok: Sure, there might be something we can do, but I don't think we can separate it entirely. More concretely, we will have to generate and run test vectors both in tests and in the interop client. Also, some of the functions we use to scaffold the groups in integration tests will be useful in the interop client.

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Rules
+
+To achieve these goals, here are the "rules"/recommendations to set up:
+
+1. Any test expected to fail by altering a valid input should also keep this valid input and verify that, contrary to
+   the
+   tampered one, it succeeds.
+2. All errors should be namely asserted i.e. `assert!(result.is_err())` is prohibited. When matching an error, one
+   should `assert!(matches!(result.unwrap_err(), MlsError::MyError))` because errors are not supposed to impl Eq (it can
```
franziskuskiefer, line 40: This is a requirement we currently don't have and that should not be made in this context. It's a different discussion.

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,175 @@
...
+   impose hard to fulfill constraints on the data they wrap)
+3. Prefer `.unwrap()` over `.expect("...")`. The latter adds overhead and most of the time does not add enough debugging
```
franziskuskiefer, line 42: OpenMLS currently has the opposite. What's the rationale behind this?
Generally, when there's information to be given, it should be given in an `expect`. So I don't agree with the general `prefer` statement.

openmls/TEST-DESIGN.md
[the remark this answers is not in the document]
franziskuskiefer: This only makes sense if there's a way to get a common setup. If writing multiple tests duplicates a lot of code, testing only one thing at a time is not a good approach.

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,175 @@
...
+   argument [here](https://twitter.com/timClicks/status/1584676737572487169).
+4. It is okay (recommended) to `use crate::prelude::*;` in all test mod (to debate ?)
+5. Have a public test framework in each top level mod inside `src/{module}/test_utils(.rs)`. Avoid scattered utils,
+   factorize them.
+6. Each test should test one and only one thing. It should not iterate over multiple cases.
+7. test method names should
+    1. follow the template `should_{expected_behaviour}_when_{state_tested}`
+       e.g. `should_succeed_when_...` & `should_fail_when_...` and use natural language
+    2. not be prefixed by `test_`
+    3. `ValSem` should not be in the test method name but kept in comment
```
franziskuskiefer, line 54: This will depend on #1126.

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,175 @@
...
+        let [(mut bob_group, ..), (mut charly_group, ..)] = alice_group.try_invite(&case, ["Bob", "Charly"]).unwrap();
+        assert!(alice_group.try_talk_to(&case, &mut [bob_group, charly_group]).is_ok());
+    }
+    ```
+    1. Does it make sense to keep the backend in fixtures ? I don't see openmls supporting anything else other
+       than `MemoryKeyStore`. Also should we keep evercrypt as an option or discontinue it ?
+    2. `MlsGroupConfig` is introduced to tune the `WireFormatPolicy` to make sure everything also works
+       in `PURE_CIPHERTEXT` mode.
+12. Test file layout
+    * Do not hesitate nesting your tests within mod. They provide better readability. They can be collapsed/expanded.
```
franziskuskiefer, line 96: Generally there's no need for this. Putting the tests in a module under the file can cover almost all tests for a module and doesn't require collapsing anything.
But tests can of course be added to the file itself if it makes things easier.

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,175 @@
...
+
+        // snip
+        #[test]
+        fn should_bla_when_blabla_n() {}
+    }
+}
+```
+
+* Crate layout
+    * unit tests in `src` in the file they relate to
```
franziskuskiefer, line 145: As I said above, module should be preferred but both is fine to do.

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,175 @@
+# OpenMLS testing proposition
+
+The purpose of this document is to improve tests in openmls and try to agree on a common standard for writing them.
+
+## Current status
+
+Here are some factual remarks about the current state of tests.
+
+1. erratic: no clear pattern emerges. It looks like tests have been piled up at different times by different persons.
```
franziskuskiefer, line 9: That's how testing works, everyone adds tests for their code.

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,175 @@
...
+   There are very few unit tests, some integration tests in `src/` and some other integration tests in `tests/`
```
franziskuskiefer, line 10: That's simply wrong. Most code is tested.

franziskuskiefer (commented):
Thanks for getting started on this.
For the final document almost everything should be recommendations (SHOULD) not requirements (MUST).

The document focuses mostly on unit tests. Can you add considerations for integration tests? That's the more important one really.

openmls/TEST-DESIGN.md
beltram: I've precised that one

openmls/TEST-DESIGN.md
beltram: The idea is that most of the time, the reason why those Result panic is straightforward, even without looking at the backtrace. Having `expect` bloats the method a bit and most of the time there's no information to be given.  

There seemed to be a rule of thumb in openmls to always use `expect` in tests. It is also a common practice to prefer `unwrap` so it can only make the code more idiomatic

openmls/TEST-DESIGN.md
beltram: Okay removed

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,175 @@
...
+   are duplicated all across the codebase
+5. some tests actually test more than one thing. It makes them harder to write/understand. Also harder to debug when
+   they
+   fail (in which iteration am I ?). This results mostly from the test setup being verbose which incentivizes reuse
+
+## Goals
+
+This proposition aims at:
+
+1. Write less verbose tests and quicker
```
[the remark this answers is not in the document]
beltram: Okay rephrased into "Write less test code" which we can all agree on

openmls/TEST-DESIGN.md
beltram: You mean unit tests in separate files ? Like currently ?

openmls/TEST-DESIGN.md
beltram: I am not sure I understand what you mean by that.  
To be more specific, I was thinking about those files with a lot of methods hence a lot of tests e.g. [this one](https://github.com/openmls/openmls/blob/9b1747179a71feaae6cc8ea90be04258cf2aceee/openmls/src/group/mls_group/membership.rs#L15). When writing unit tests for it (so in the same file), I propose the file to look like this:

```rust
impl MlsGroup {
    pub fn add_members();
    pub fn own_leaf();
    pub fn remove_members();
    pub fn propose_add_member();
    pub fn propose_remove_member();
    pub fn leave_group();
    pub fn members();
    pub fn member();
}

#[cfg(test)]
mod tests {
    use super::*;

    mod add_members {
        use super::*;
    }
    mod own_leaf {
        use super::*;
    }
    mod remove_members {
        use super::*;
    }
    mod propose_add_member {
        use super::*;
    }
    mod propose_remove_member {
        use super::*;
    }
    mod leave_group {
        use super::*;
    }
    mod members {
        use super::*;
    }
}
```

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,177 @@
+# OpenMLS testing proposition
+
+The purpose of this document is to define a testing infrastructure and guidelines how to write test.
```
dconnolly, line 3:
```suggestion
The purpose of this document is to define a testing infrastructure and guidelines how to write tests.
```

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,176 @@
...
+## Rules
+
+To achieve these goals, here are the "rules"/recommendations to set up:
+
+1. Any test expected to fail by altering a valid input SHOULD also keep this valid input and verify that, contrary to
+   the tampered one, it succeeds.
+2. Prefer `.unwrap()` over `.expect("...")`. The latter adds overhead and most of the time does not add enough debugging
+   information. If, on the opposite, an `.expect(...)` can help a developer troubleshooting a failing test with a very
+   complex setup which this `expect` message could help understanding then go for it ! Another
+   argument [here](https://twitter.com/timClicks/status/1584676737572487169).
```
raphaelrobert, line 42: As discussed offline: `.unwrap()` is fine in tests, but should be commented on when used for non-trivial checks.

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,176 @@
...
+3. It is okay (recommended) to `use crate::prelude::*;` in all test mod (to debate ?)
```
raphaelrobert, line 43: Unit tests must never use `prelude`. Integration tests can use `prelude`, ideally some of them also use the public API directly.

openmls/TEST-DESIGN.md
raphaelrobert: We should introduce this as soon as we have more than just the backend and the ciphersuite.

openmls/TEST-DESIGN.md
```diff
@@ -0,0 +1,176 @@
...
+## Bonus
+
+* We could use mutation testing to "test out tests". A good crate available for doing that
+  is [mutagen](https://github.com/llogiq/mutagen). After annotating all the methods in the [`tree` mod](src/tree) it
+  gives the [following result](https://github.com/beltram/openmls/blob/mutation/openmls/MUTATION.txt) with ~= 1/3 of
+  mutants surviving i.e. not rightfully tested
+* If WASM is a target we could already use [wasm-bindgen-test](https://crates.io/crates/wasm-bindgen-test) macros on all
+  tests methods (no async migration required at that point, only a few things to fix: rayon, time, rng)
+* Use [nextest](https://crates.io/crates/cargo-nextest) to execute test in the CI because it is faster (19s vs 30s for
+  cargo test)
```
raphaelrobert, line 175: I always use that locally, we could consider it for the CI as well if there are no downsides.

That being said, the overall speedup on the CI should only be marginal, since building is the slowest part.

raphaelrobert (commented):
Adding some comments from the meeting yesterday.

beltram:
Probably already implemented

kkohbrok:
This has (sadly) not yet been implemented and it's still on our radar. Increasing test quantity and quality is clearly something we want to improve in the future. We will hopefully get back to it once we have finished interop testing with the other MLS implementations, which is what has consumed pretty much all of our time in the recent weeks.

PR #1137, opened 2022-12-06
remove unnecessary lifetimes

franziskuskiefer:
The lifetimes can be infered by now and the typetag crate isn't used anymore.


duesee (approved):
I think, I'm qualified to do this review :-) :P LGTM!

Issue #1131, opened 2022-11-30
Improve Forward Secrecy


franziskuskiefer:
Ensure that all secrets are dropped at the earliest possible.

- [ ] `leaf_secret` in the `OpenMlsLeafNode`

raphaelrobert:
Closing since we don't use `OpenMlsLeafNode` any longer.

Issue #876, opened 2022-04-19
[MLS Spec change] Clean up the description of tree evolution. (#593)


kkohbrok:
**Description of the changes**
This PR notes that it's possible to generate the leaf keypairs independently from the rest of the path. We could implement this to save a few derivation operations, but the current solution of deriving the whole path at once is probably easier.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/593

franziskuskiefer:
Thanks for tackling this @nomaxg. I assigned you the issue.

franziskuskiefer:
Has this been fully addressed with #1139, i.e. can we close this, or is there another PR coming up? @kkohbrok can you check as well please as you reviewed #1139.

kkohbrok:
This has indeed been addressed with #1139. Closing.

PR #1139, opened 2022-12-06
Sample leaf node path secret randomly

nomaxg:
Addresses #876. The spec changes [here](https://github.com/mlswg/mls-protocol/pull/593) mention that the leaf node path secret (which is used to generate new HPKE keys) can be sampled randomly, instead of from the leaf secret.


openmls/src/treesync/diff.rs
```diff
@@ -252,17 +252,18 @@ impl<'a> TreeSyncDiff<'a> {
...
+    fn derive_path(
         &self,
         backend: &impl OpenMlsCryptoProvider,
         ciphersuite: Ciphersuite,
-        leaf_secret: Secret,
     ) -> Result<PathDerivationResult, LibraryError> {
-        let leaf_path_secret = PathSecret::from(leaf_secret);
-        let path_secret = leaf_path_secret.derive_path_secret(backend, ciphersuite)?;
+        let path_secret = PathSecret::from(
+            Secret::random(ciphersuite, backend, None)
```
nomaxg, line 264: Should this be versioned?

openmls/src/treesync/diff.rs
```diff
@@ -252,17 +252,18 @@ impl<'a> TreeSyncDiff<'a> {
         Ok(())
     }
 
-    /// Derive a new direct path for our own leaf from the given `leaf_secret`.
+    /// Derive a new direct path for our own leaf
```
kkohbrok, line 255:
```suggestion
    /// Derive a new direct path for our own leaf.
```

openmls/src/treesync/diff.rs
kkohbrok: I think at this point, there is no need. There is currently no clear story for how we handle version agility. That should be tackled with #1065.

openmls/src/treesync/diff.rs
```diff
@@ -290,19 +291,8 @@ impl<'a> TreeSyncDiff<'a> {
         credential_bundle: &CredentialBundle,
     ) -> Result<UpdatePathResult, LibraryError> {
         debug_assert!(self.own_leaf().is_ok(), "Tree diff is missing own leaf");
-        debug_assert!(
```
kkohbrok, line 293: Until we properly fix #1131, the assert should probably stay here. Or am I missing a reason why it has to go now?

kkohbrok (changes requested):
Thanks for taking a stab at this!

openmls/src/treesync/diff.rs
nomaxg: I removed it because it doesn't look like any of the code that branches from here depends on the existence of `leaf_secret` anymore, but I can definitely throw it back in if it's still a valuable check

franziskuskiefer:
FYI: #1142 is a follow up to this and the leaf node changes from #1125 that removes the leaf secrets.

kkohbrok:
Sounds good. @nomaxg do you want to update the branch and put the period at the end of that sentence? Then we can get this in.

nomaxg:
> Sounds good. @nomaxg do you want to update the branch and put the period at the end of that sentence? Then we can get this in.

done!

kkohbrok (approved):

PR #1141, opened 2022-12-07
Cleanup `KeyPackage`.

duesee:
This PR cleans up `KeyPackage` extensions, removes `Lifetime` and `Capabilities` extensions, and introduces `KeyPackageBundleBuilder`. It also switches to a single ciphersuite in `KeyPackage` for which I used a cleanup branch started by @franziskuskiefer to avoid introducing a version of `KeyPackageBundleBuilder` that would need immediate changes.


openmls/src/group/core_group/validation.rs
```diff
@@ -172,50 +171,6 @@ impl CoreGroup {
 
                 return Err(ProposalValidationError::InsufficientCapabilities);
             }
-
```
duesee, line 175: There is neither a `capabilities` (nor `required_capabilities`) extension for key packages. Thus, I think this can go.

openmls/src/group/tests/test_proposal_validation.rs
```diff
@@ -1155,25 +1153,10 @@ fn test_valsem106(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider
         };
         match key_package_version {
             KeyPackageTestVersion::WrongCiphersuite => {
-                test_kpb_payload.set_ciphersuite(wrong_ciphersuite)
+                test_kpb_payload.set_ciphersuite(wrong_ciphersuite);
             }
-            KeyPackageTestVersion::UnsupportedVersion => {
```
duesee, line 1160: There are no extensions for key packages. Thus, I think, this can go, too.

openmls/src/key_packages/codec.rs
```diff
@@ -41,7 +41,7 @@ impl tls_codec::Deserialize for KeyPackage {
         let hpke_init_key = HpkePublicKey::tls_deserialize(bytes)?;
         let leaf_node = LeafNode::tls_deserialize(bytes)?;
         let credential = Credential::tls_deserialize(bytes)?;
-        let extensions = Vec::<Extension>::tls_deserialize(bytes)?;
+        let extensions = Vec::<KeyPackageExtension>::tls_deserialize(bytes)?;
```
duesee, line 44: I could revert this to use `Extension`.

openmls/src/key_packages/mod.rs
```diff
@@ -130,7 +130,40 @@ struct KeyPackageTBS {
     init_key: HpkePublicKey,
     leaf_node: LeafNode,
     credential: Credential, // TODO[FK]: remove
-    extensions: Vec<Extension>,
+    extensions: Vec<KeyPackageExtension>,
```
duesee, line 133: I could revert this to use Extension. The dummy tls_codec impls would go then. If there will be an extension in the future, we could use `#[derive(Tls...)]`

openmls/src/key_packages/mod.rs
```diff
@@ -286,42 +307,6 @@ impl KeyPackage {
         })
     }
 
-    /// Get the application ID of this key package as byte slice.
-    /// See [`ApplicationIdExtension`](`crate::extensions::ApplicationIdExtension`)
-    /// for more details on the application ID extension.
-    ///
-    ///
-    /// Returns a [`ExtensionError`] if no application ID extension is present.
-    pub fn application_id(&self) -> Result<&[u8], ExtensionError> {
```
duesee, line 295: Not a key package extension.

openmls/src/key_packages/mod.rs
```diff
@@ -389,25 +374,14 @@ impl KeyPackage {
             init_key: hpke_init_key,
             leaf_node,
             credential: credential_bundle.credential().clone(),
-            extensions,
+            extensions: vec![],
```
duesee, line 377: Here, I think, we don't want to include the extensions because there are no key package extensions defined.

openmls/src/key_packages/mod.rs
```diff
@@ -445,26 +419,24 @@ pub(crate) struct KeyPackageBundlePayload {
 }
 
 impl KeyPackageBundlePayload {
-    /// Add (or replace) an extension to the KeyPackage.
-    #[cfg(any(feature = "test-utils", test))]
-    pub fn add_extension(&mut self, extension: Extension) {
```
duesee, line 450: This should be handled by an Extensions type, IMO.

openmls/src/key_packages/mod.rs
```diff
@@ -633,46 +605,6 @@ impl KeyPackageBundle {
             return Err(error);
         }
 
-        // First, check if one of the input extensions is a capabilities
```
duesee, line 636: As key packages don't have extensions, this is not needed.

openmls/src/extensions/capabilities_extension.rs
```diff
@@ -94,45 +94,50 @@ impl CapabilitiesExtension {
...
-            .proposals()
-            .iter()
-            .any(|p| !self.proposals().contains(p))
-        {
-            return false;
-        }
-        true
-    }
+
+    // /// Check if this [`CapabilitiesExtension`] supports all the capabilities
```
duesee, line 118: Unused now.

openmls/src/group/core_group/new_from_welcome.rs
```diff
@@ -100,10 +100,12 @@ impl CoreGroup {
                 .map_err(|_| WelcomeError::UnsupportedCapability)?;
             // Also check that our key package actually supports the extensions.
             // Per spec the sender must have checked this. But you never know.
-            key_package_bundle
-                .key_package()
-                .check_extension_support(required_capabilities.extensions())
-                .map_err(|_| WelcomeError::UnsupportedExtensions)?
+
+            // XXX: There are currently no key package extensions in draft-ietf-mls-protocol-16.
```
franziskuskiefer, line 104: This should probably be a check on the leaf as well.

openmls/src/extensions/capabilities_extension.rs
franziskuskiefer: The capabilities extension can be completely removed. These checks are on the `Capabiblities` struct in the `LeafNode` now.

openmls/src/key_packages/mod.rs
```diff
@@ -286,42 +272,6 @@ impl KeyPackage {
...
-        for required in required_extensions.iter() {
-            if !my_extension_types.clone().any(|e| &e == required) {
-                return Err(KeyPackageExtensionSupportError::UnsupportedExtension);
-            }
-        }
-        Ok(())
-    }
-
     /// Get a reference to the [`Credential`].
     pub fn credential(&self) -> &Credential {
```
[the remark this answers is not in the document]
duesee: I'll do one bit at a time.

franziskuskiefer:
I see much more added here now. Let's not do that and only clean up here first. This is already way too big (as the labeller tells you).

duesee:
Apologies for the force-push. It's best to start clean since I isolated all extension-related changes and removed anything that can be handled separately. The PR is still large, but primarily due to required test changes. Not sure what the best review strategy is, but I recommend spending less time on repeated `KeyPackageBundle::new` changes as it's just a refactoring. Reviewing each commit individually may help. Note, however, that in this case `Lifetime` changes will need to be inspected twice due to consecutive `KeyPackageBundle::new` changes.

duesee:
We discussed this offline and I'll close this PR as it is too difficult to review and will try to make smaller PRs.

PR #1142, opened 2022-12-09
leaf secret cleanup

franziskuskiefer:
Cleanup of leaf secrets after #1139 and #1125 removing the leaf secret in the leaf node as well as the key package.

This is part of #876 and #819.


openmls/src/key_packages/mod.rs
```diff
@@ -749,18 +736,17 @@ impl KeyPackageBundle {
         }
 
         let ciphersuite = ciphersuites[0];
-        let leaf_node_secret = derive_leaf_node_secret(&leaf_secret, backend)
-            .map_err(LibraryError::unexpected_crypto_error)?;
+        // let leaf_node_secret = derive_leaf_node_secret(&leaf_secret, backend)
+        //     .map_err(LibraryError::unexpected_crypto_error)?;
```
kkohbrok, line 740: More code to remove :-)

kkohbrok (changes requested):
Hooray for lots of removed code! Thanks for cleaning this up. One nit and also clippy seems to be a bit confused.

franziskuskiefer:
> Hooray for lots of removed code! Thanks for cleaning this up. One nit and also clippy seems to be a bit confused.

Deleting code 🎉 

The `pull_request_target` appears to take code from main (not only the workflow description). The error you see on the PR run is from main and the push job goes through (also see message on Zulip).

kkohbrok (approved):

PR #1151, opened 2022-12-15
chore: #[allow(clippy::large_enum_variant)] on `Node` and `ProposalOrRef`.

duesee:
This PR silences two (remaining) Clippy warnings about the large size difference in `Node`s and `ProposalOrRef`s enum variants.

Should we `Box` the large variant? Or would it be sufficient for now to file an issue that reminds us to re-evaluate later?

### DIFF
--- a/openmls/src/messages/proposals.rs
+++ b/openmls/src/messages/proposals.rs
@@ -354,6 +354,7 @@ pub(crate) enum ProposalOrRefType {
 )]
 #[repr(u8)]
 #[allow(missing_docs)]
+#[allow(clippy::large_enum_variant)]
 pub(crate) enum ProposalOrRef {
     #[tls_codec(discriminant = 1)]
     Proposal(Proposal),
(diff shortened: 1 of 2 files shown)


duesee:
It seems that Clippy would also be happy when https://github.com/openmls/openmls/pull/1136 is merged. So maybe the best thing to do is to ignore Clippy for a bit :-)

kkohbrok:
Not sure what's going on with clippy right now. It's a bit strange since #1136 doesn't even change the enums you touch here.

duesee:
> Not sure what's going on with clippy right now. It's a bit strange since https://github.com/openmls/openmls/pull/1136 doesn't even change the enums you touch here.

I assumed that your PR must change some "children" structs of `Node` or `ProposalOrRef`? If so, it could be that reducing the the size of some of the children was already to silence the lint?

duesee:
This can be closed because Clippy already seems to be happy on main.

PR #1152, opened 2022-12-16
spec: Remove key package extensions.

duesee:
This PR removes the key package extensions and makes it so that all required checks (`ExtensionType::Capabilities`) are made against the leaf node capabilities.

This is the first part of a series of PRs to cleanup the key package. Franziskus and I pair programmed this one.


franziskuskiefer (approved):
:shipit:

PR #1155, opened 2022-12-18
Cleanup around tree

raphaelrobert:
This PR cleans up code around the ratcheting tree to correct previous design choices and make the transition to a full tree easier:

 - Limits valid platforms to 32, 64, and 128 bits of pointer width
 - Avoids unnecessary LibraryErrors in the binary tree
 - Improves how leaves are returned, making the public `MlsGroup.members()` a nice iterator that's no longer a `Result`
 - Improves merges, making `MlsGroup.merge_staged_commit()` `LibraryError`-free
 - Deletes now unnecessary code
 
The PR is split into commits that can be reviewed separately. Each commit contains a small set of code changes and the corresponding churn.
 
Note that this is just a partial cleanup and that more needs to be done.


openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -171,19 +169,66 @@ impl<'a, T: Clone + Debug> AbDiff<'a, T> {
...
+                    Some((*index / 2, leaf))
+                } else {
+                    None
+                }
+            })
+            .peekable();
+
+        let mut combined = Vec::new();
+
+        loop {
```
kkohbrok, line 190: I think getting rid of Itertools is nice, but `loop` is a bit dangerous. I seem to remember that we decided against using it in another instance due to the potential for an infinite loop, but I can't remember the specifics. This one seems pretty straight-forward due to how `next` is used in each iteration indicating that the loop must terminate at some point, so I don't have a strong opinion here.

kkohbrok (commented):
I can't remember why we didn't just use config flags to restrict the architecture to begin with. I only remember that we discussed it at length and that I spent quite a bit of time to implement the current solution. 

That being said, I think the solution in this PR is quite a bit more elegant than using library errors.

openmls/src/treesync/mod.rs
```diff
@@ -354,65 +344,46 @@ impl TreeSync {
...
-                encryption_key.as_slice().to_vec(),
-                signature_key.as_slice().to_vec(),
-                identity.to_vec(),
-            ))
-        }
-        Ok(out)
+    pub(crate) fn full_leave_members(&self) -> impl Iterator<Item = Member> + '_ {
+        self.tree
+            .leaves()
+            // Filter out blank nodes (should not be necessary in a valid tree)
```
raphaelrobert, line 366:
```suggestion
            // Filter out blank nodes
```

kkohbrok (approved):

PR #1154, opened 2022-12-16
spec: Remove `LifetimeExtension`.

duesee:

PR #1156, opened 2022-12-19
Remove LifetimeExtension

franziskuskiefer:
This PR replaces #1154 to only remove the lifetime extension.
The lifetime is part of the leaf node now instead.

Note that there are different mechanisms to set the lifetime that are not aligned well. This will be cleaned up in follow ups.

### DIFF
--- a/openmls/src/key_packages/test_key_packages.rs
+++ b/openmls/src/key_packages/test_key_packages.rs
@@ -15,38 +15,10 @@ fn generate_key_package(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoPr
     .expect("An unexpected error occurred.");
 
     // Generate a valid KeyPackage.
-    let lifetime_extension = Extension::Lifetime(LifetimeExtension::new(60));
-    let kpb = KeyPackageBundle::new(
-        &[ciphersuite],
-        &credential_bundle,
-        backend,
-        vec![lifetime_extension],
-    )
-    .expect("An unexpected error occurred.");
+    let kpb = KeyPackageBundle::new(&[ciphersuite], &credential_bundle, backend, vec![])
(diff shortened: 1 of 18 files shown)


openmls/src/key_packages/test_key_packages.rs
```diff
@@ -15,38 +15,10 @@ fn generate_key_package(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoPr
...
-
-    // Now with two lifetime extensions, the key package should be invalid.
-    let lifetime_extension = Extension::Lifetime(LifetimeExtension::new(60));
-    let kpb = KeyPackageBundle::new(
-        &[ciphersuite],
-        &credential_bundle,
-        backend,
-        vec![lifetime_extension.clone(), lifetime_extension],
-    );
-    assert!(kpb.is_err());
```
kkohbrok, line 49: As I understand it, there is no way to set a non-default lifetime in a KeyPackage for the time being, so we can't test this at the moment, although we definitely should. Do you plan to deal with this in a follow-up?

kkohbrok (changes requested):
Thanks for following up on the previous leaf node work! Looks good with one nit/question.

openmls/src/key_packages/test_key_packages.rs
franziskuskiefer: Yes, setting lifetimes is all over the place right now. This will be cleaned up in one of the next PRs. I added it as one of the items to #819.

franziskuskiefer: Also note that key packages don't have lifetimes anymore but the leaf node has the lifetime now.

kkohbrok (approved):

kkohbrok (approved):

PR #1160, opened 2022-12-21
Drop key package bundle

franziskuskiefer:
This PR exercises the approach of using the MLS specs rather than "enriched" OpenMls structs.

`KeyPackageBundle` is mostly gone and only used internally as a shallow wrapper to pass around `KeyPackage`s together with their private keys. It is also used in tests for simplicity but could be removed there when rewriting tests.

Instead, the `KeyPackage` is used everywhere now and the corresponding private key is stored in the key store.

I introduce a `CryptoConfig` to pass around the `Ciphersuite` and `ProtocolVersion` together. Not sure if this is the best name and where it should live (I'm open for suggestions). It's just a simple struct for simplifying arguments.

Sorry for the large PR but it's touching a lot of code. It should be relatively easy to review though.


openmls/src/group/mls_group/creation.rs
```diff
@@ -149,6 +143,14 @@ impl MlsGroup {
             key_package,
             private_key: private_key.into(),
         };
+
+        // Delete the [`KeyPackage`] and the corresponding private key from the
+        // key store
+        key_package_bundle
+            .key_package
+            .delete(backend)
+            .map_err(|_| WelcomeError::KeyStoreDeletionError)?;
```
kkohbrok, line 152: Can you maybe add a brief test for this? We probably have to tackle deletion more thoroughly at some point, but it would be good to be sure this works.

openmls/src/group/tests/test_proposal_validation.rs
```diff
@@ -431,13 +431,6 @@ fn test_valsem101(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider
     let dave_credential_bundle =
         CredentialBundle::from_parts("Dave".into(), charlie_credential_bundle.key_pair());
 
-    // TODO[FK]: What's really being tested here?
-    // let mut kpb_payload: KeyPackageTBS = charlie_key_package.into();
-    // kpb_payload.set_credential(dave_credential_bundle.credential().clone());
-    // let dave_key_package_bundle = kpb_payload
-    //     .sign(backend, &dave_credential_bundle)
-    //     .expect("error signing credential bundle");
```
kkohbrok: Can this go entirely, then?

kkohbrok (changes requested):
Thanks for the PR! Just two nits and the rest can be done in follow-ups.

kkohbrok (approved):
Thanks for adding the test! 🚢 it!

franziskuskiefer:
I had to fix some merge conflicts.

kkohbrok (approved):

PR #1163, opened 2022-12-22
Leaf node index refactoring

raphaelrobert:
This PR refactors how indexing nodes in the tree works:

 - `LeafNode` is replaced by `LeafNodeIndex`
 - `ParentNodeIndex` is introduced to reference parent nodes exclusively
 - `NodeTreeIndex` replaces `NodeIndex` temporarily
 - Tree math now uses these new types for better type safety
 - More code is deleted

The main benefit is that operations around the tree should now be easier and error-free (a few `LibraryError`s are still left). Introducing the full tree should also be easier now.

Potential future work: If we split the nodes vector into two vectors: one for leaf nodes and one for parent nodes, we can eliminate the remaining `LibraryError`s and get better type safety/correctness.

For the review: Most changes are mere text replacements. The interesting changes are in `binary_tree` and `treesync`.


openmls/src/binary_tree/array_representation/treemath.rs
```diff
@@ -43,106 +151,113 @@ fn level(index: NodeIndex) -> usize {
...
+    debug_assert!(k > 0);
+    x ^ (0x01 << (k - 1))
+}
+
+pub(super) fn right(index: ParentNodeIndex, size: u32) -> TreeNodeIndex {
+    debug_assert!(size > 0);
+    debug_assert!(index.to_tree_index() < size);
+    let x = index.to_tree_index();
+    let k = level(x);
+    debug_assert!(k > 0);
```
kkohbrok, line 178: I understand how the level of k has to be > 0 due to the implementation of `ParentNodeIndex`, but `size` has no protection at all.

openmls/src/binary_tree/array_representation/treemath.rs
```diff
@@ -43,106 +151,113 @@ fn level(index: NodeIndex) -> usize {
...
+    let x = x.u32();
+    let k = level(x);
+    let b = (x >> (k + 1)) & 0x01;
+    let index = (x | (1 << k)) ^ (b << (k + 1));
+    ParentNodeIndex::from_tree_index(index)
+} */
+
 // The parent here might be beyond the right edge of the tree.
-pub(super) fn parent_step(x: NodeIndex) -> NodeIndex {
+pub fn parent_step(x: u32) -> u32 {
```
kkohbrok, line 208:
```suggestion
fn parent_step(x: u32) -> u32 {
```

openmls/src/binary_tree/array_representation/tree.rs
```diff
@@ -127,27 +121,32 @@ impl<T: Clone + Debug> ABinaryTree<T> {
...
             }
         }
     }
 
-    /// Return a reference to the leaf at the given `LeafIndex`.
-    ///
-    /// Returns an error if the leaf is outside of the tree.
-    pub(crate) fn leaf(&self, leaf_index: LeafIndex) -> Result<&T, ABinaryTreeError> {
+    /// Return a reference to the leaf at the given `LeafNodeIndex`, or the default
+    /// value if the leaf is not found..
```
kkohbrok, line 145:
```suggestion
    /// value if the leaf is not found.
```

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -171,66 +161,43 @@ impl<'a, T: Clone + Debug> AbDiff<'a, T> {
...
-    pub(crate) fn set_direct_path(
-        &mut self,
-        leaf_index: LeafIndex,
-        path: Vec<T>,
-    ) -> Result<(), ABinaryTreeDiffError> {
-        let node_index = to_node_index(leaf_index);
-        let direct_path = direct_path(node_index, self.tree_size())?;
-        if path.len() != direct_path.len() {
-            return Err(ABinaryTreeDiffError::PathLengthMismatch);
-        }
```
kkohbrok, line 229: This check goes away, but this function is crate-public and it's not clear where this check is made instead. If we want to remove this Result, we could have a `Path` type that gives us guarantees w.r.t. its length.

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -253,11 +220,12 @@ impl<'a, T: Clone + Debug> AbDiff<'a, T> {
         // is never empty, since if there is a second leaf index (that is within
         // the bound of the tree), there is a non-leaf root node that is in the
         // direct path of all leaves.
-        self.leaf_pair_check(leaf_index_1, leaf_index_2)?;
+        debug_assert!(leaf_index_1 != leaf_index_2);
+        debug_assert!(leaf_index_1.to_tree_index() < self.size);
+        debug_assert!(leaf_index_2.to_tree_index() < self.size);
```
kkohbrok, line 225: This is another place where we have a crate-public function where this PR replaces a runtime check with a debug-time one without getting any guarantee from the input type. I'd much prefer keeping the check here, potentially return an error and then hide the error at the same place where we do the actual check that these two indices don't collide.

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -348,150 +289,109 @@ impl<'a, T: Clone + Debug> AbDiff<'a, T> {
...
-    /// an error when the given [`NodeIndex`] points to the root node or to a
-    /// node not in the tree.
-    pub(crate) fn sibling(&self, node_index: NodeIndex) -> Result<NodeIndex, ABinaryTreeDiffError> {
-        Ok(sibling(node_index, self.tree_size())?)
-    }
-
-    /// Returns a [`NodeIndex`] to the left child of the referenced node.
-    /// Returns an error when the given [`NodeIndex`] points to a leaf node or
+    /// Returns a [`TreeNodeIndex`] to the left child of the referenced node.
+    /// Returns an error when the given [`TreeNodeIndex`] points to a leaf node or
```
kkohbrok, line 346: This line can be removed, I think.

kkohbrok (changes requested):
So far I have only looked at the changes in `array_representation`. It's great that we're getting better type guarantees from the new index types and in many places it's certainly better than it was before. However, there are a couple of places where I feel the changes favor API niceness (no Result types) over correctness/robustness. I agree that it's important to have a clean, top-level API and that bubbling up LibraryErrors everywhere is not great. But there are a few places where actual errors can occur due to invalid inputs to crate-public functions. For those functions I would propose we keep the error types and hide the errors in the same places where we can actually provide the guarantee that the error can't occur.

openmls/src/binary_tree/array_representation/diff.rs
raphaelrobert: That should be covered by `ValSem202`.

openmls/src/binary_tree/array_representation/treemath.rs
```diff
@@ -1,26 +1,134 @@
...
+    }
+
+    /// Return the inner value as `usize`.
+    pub fn usize(&self) -> usize {
+        self.u32() as usize
+    }
+
+    /// Return the index as a TreeNodeIndex value.
+    pub(crate) fn to_tree_index(self) -> u32 {
+        self.0 * 2
```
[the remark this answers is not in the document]
raphaelrobert: True, but as discussed offline this is not really new: previously we had `LeafIndex` and also converted it to `NodeIndex`.

openmls/src/binary_tree/array_representation/treemath.rs
```diff
@@ -43,106 +151,113 @@ fn level(index: NodeIndex) -> usize {
...
+pub(super) fn left(index: ParentNodeIndex) -> TreeNodeIndex {
+    let x = index.to_tree_index();
     let k = level(x);
-    if k == 0 {
-        return Err(TreeMathError::LeafHasNoChildren);
-    }
-    Ok(x ^ (0x01 << (k - 1)))
+    debug_assert!(k > 0);
+    let index = x ^ (0x01 << (k - 1));
+    TreeNodeIndex::new(index)
```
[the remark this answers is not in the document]
raphaelrobert: This will go away with the full tree rewrite, so I'm not too bothered.

openmls/src/binary_tree/array_representation/treemath.rs
raphaelrobert: I introduced `TreeSize` now.

openmls/src/binary_tree/array_representation/treemath.rs
```diff
@@ -1,26 +1,168 @@
...
+    TlsSerialize,
+    TlsSize,
+)]
+pub struct LeafNodeIndex(u32);
+
+impl LeafNodeIndex {
+    /// Create a new `LeafNodeIndex` from a `u32`.
+    pub fn new(index: u32) -> Self {
+        LeafNodeIndex(index)
+    }
```
kkohbrok, line 27: A `From` implementation would probably be a bit more ergonomic.

openmls/src/treesync/diff.rs
```diff
@@ -83,59 +92,61 @@ impl<'a> From<&'a TreeSync> for TreeSyncDiff<'a> {
...
-            // have to stop here.
-            if self.leaf_count() == 1 {
-                return Ok(());
+
+        // We reverse the nodes here to make sure we start with the right-most
+        let mut nodes = self
+            .diff
+            .nodes()
+            .collect::<Vec<(TreeNodeIndex, &TreeSyncNode)>>();
+        nodes.reverse();
```
kkohbrok, line 107: Since everything is done via an iterator now, have you considered implementing `DoubleEndedIterator` for `SortedIter` and just using `rev()` here? Not sure if it's worth the effort, but it would probably fit the theme better than collecting and looping over `chunks`.

kkohbrok (approved):
I still don't feel great about the removal of some of the checks, but as discussed OOB, it's a trade off I'll have to live with. Let's see what this looks like when we have the full tree implementation. Otherwise just two comments.

PR #1164, opened 2022-12-22
Remove stale tests

raphaelrobert:
Removes stale tests decommissioned during the tree rewrite a year ago but never deleted.


franziskuskiefer (commented):
Thanks, just two questions
- The treemath kat we might want to have for the treesync module?
- Do we have enough tests on the treesync module or would it make sense to move some of the unit tests there?

raphaelrobert:
> Thanks, just two questions
> 
>     * The treemath kat we might want to have for the treesync module?

The treemath KAT was there twice. I deleted the stale and unused one.

>     * Do we have enough tests on the treesync module or would it make sense to move some of the unit tests there?

These tests seem to be very specific to the old code structure. Since treesync has a completely different API, they are pretty worthless.

franziskuskiefer (dismissed):
👍🏻 :shipit:

raphaelrobert:
I had to address merge conflicts.

franziskuskiefer (approved):

PR #1165, opened 2022-12-23
Remove credential from KeyPackage

franziskuskiefer:
A small PR towards closing #819. This PR only removes the `credential` field from the `KeyPackage` (actually `KeyPackageTBS`).

Later, the `signature_key()` function will be removed and the key has to be retrieved from the `LeafNode`. This is blocked on #1160 and equivalent changes to the `Credential`.


openmls/src/key_packages/mod.rs
```diff
@@ -320,11 +305,6 @@ impl KeyPackage {
         Ok(())
     }
 
-    /// Get a reference to the [`Credential`].
-    pub fn credential(&self) -> &Credential {
```
kkohbrok, line 324: Wouldn't it be easier to just redirect this function to point to  `&self.payload.leaf_node.credential`?

kkohbrok (changes requested):
Looks simple enough. Just one question.

openmls/src/key_packages/mod.rs
franziskuskiefer: There are two different philosophies for this. Either make it explicit where it's coming from and avoid confusion (with the cost of calling the additional getter), or making it easier to get and have a "semantic" getter (hiding where the value is coming from).

I made it explicit here with the addition `leaf_node()` to highlight where we do this (and it's not in too many places). I think the places where this is needed should maybe be refactored because `credential` is not a property of `KeyPackage` anymore (the semantics changed). Keeping this function here feels like trying to keep the old semantic alive that doesn't really make sense anymore.

kkohbrok: Thanks for the explanation. That makes sense to me.

kkohbrok (approved):

Issue #995, opened 2022-10-11
[MLS Spec change] Stronger parent hashes for authenticated identities (#527)


raphaelrobert:
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/527

**Description of the changes**
> Our proposal
 > ============
 > 
 > The parent hash construction currently authenticates only the public keys in the ratchet tree and does not protect other tree contents, notably the credentials at the leaves.
 > Consequently, a malicious member can tamper with the credentials before sending a Welcome package to a new member, even if it cannot tamper with the public keys.
 > 
 > Instead of using the tree resolution in the parent hash computation, we propose to use the tree hash.
 > 
 > The last time this was discussed on the list, it was agreed that it would be better to use tree hashes from a security point of view, but it was felt that complexity-wise that the recipient of a Welcome message would have to do O(n^2) work.
 > A more thorough analysis shows that processing a Welcome message can be done in O(n*log(log(n))).
 > The result is that each signature captures the full tree contents including leaf credentials, which provides better authentication guarantees for member identities.
 > 
 > Efficient algorithm to process a Commit
 > =======================================
 > 
 > When processing Commits, the implementation can store the tree hashes of every subtree.
 > That way, with n being the number of leaves, you can process each proposal in O(log(n)) and each UpdatePath in O(log(n)).
 > Then you can check the parent hash when processing a Commit in O(p*log(n)) where p is the number of proposals.
 > 
 > Efficient (enough) algorithm to process a Welcome
 > =================================================
 > 
 > When processing a Welcome message, we need to do the parent hash verification procedure.
 > To do it with the new parent hash, we need to compute tree hashes of subtrees at the last time they were updated.
 > This can be done using the following modified tree hash:
 > 
 >     OldTreeHash(leaf, excluded_leaves) = Hash(leaf) if leaf not in excluded_leaves
 >     OldTreeHash(leaf, excluded_leaves) = Hash(blank leaf) if leaf in excluded_leaves
 >     OldTreeHash(node, excluded_leaves) = Hash(node.content | OldTreeHash(node.left, left_excluded_leaves) | OldTreeHash(node.right, right_excluded_leaves))
 >     where left_excluded_leaves = leaves that are in excluded_leaves and in node.left (same for right_excluded_leaves).
 >     and `excluded_leaves` are removed from `node.content.unmerged_leaves`.
 > 
[...]

PR #1166, opened 2022-12-28
More treesync refactoring: split nodes

raphaelrobert:
This PR does the following:
  - This is the final PR that concludes the separation between leaf nodes & parent nodes (including the indexing)
  - Splits the internal nodes vector into two distinct ones: leaf nodes and parent nodes
  - The former `NodeIndex` (now known as `TreeNodeIndex`) is only used within `treemath`, all hazmat functions are private
  - Re-enables tree hash caching
  - Deactivates parent hash verification for now. This will be enabled again with #995.


openmls/src/binary_tree/test_binary_tree.rs
```diff
@@ -147,63 +156,54 @@ fn test_diff_merging() {
...
-        let mut diff = tree.empty_diff();
-
-        assert_eq!(
-            diff.add_leaf(666, 667)
-                .expect_err("no error adding beyond u32 max"),
-            MlsBinaryTreeDiffError::TreeTooLarge
-        )
-    }
-}
-
```
kkohbrok, line 182: Is there a particular reason why this test is removed?

kkohbrok (changes requested):
Looks good in general with lots of nice improvements due to the new split between parents and leaves. A few somewhat minor comments, then we can get this in.

openmls/src/treesync/diff.rs
```diff
@@ -479,57 +464,50 @@ impl<'a> TreeSyncDiff<'a> {
         &self,
         node_index: TreeNodeIndex,
```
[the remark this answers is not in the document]
raphaelrobert: It's still there, but purely as an `enum`, which makes it safe to use.

openmls/src/binary_tree/test_binary_tree.rs
raphaelrobert: The unsafe code no longer works, because we now do something with the nodes. But I think I can re-add the removal check.

kkohbrok (approved):
Thanks for making those changes. Let's get it in!

PR #1167, opened 2022-12-29
Add ProtocolVersion to MlsMessage and improve context handling

kkohbrok:


openmls/src/framing/message.rs
```diff
@@ -214,7 +214,10 @@ impl From<PublicMessage> for MlsMessageOut {
         let body = MlsMessageBody::PublicMessage(plaintext);
 
         Self {
-            mls_message: MlsMessage { body },
+            mls_message: MlsMessage {
+                body,
+                protocol_version: ProtocolVersion::Mls10,
```
franziskuskiefer, line 219: Why don't you use the version from the `FramedContentTbs`?

openmls/src/framing/message.rs
```diff
@@ -224,7 +227,10 @@ impl From<PrivateMessage> for MlsMessageOut {
         let body = MlsMessageBody::PrivateMessage(ciphertext);
 
         Self {
-            mls_message: MlsMessage { body },
+            mls_message: MlsMessage {
+                body,
+                protocol_version: ProtocolVersion::Mls10,
```
franziskuskiefer, line 232: Here we don't have the version in the message. But we should be able to pass it in without too much churn. I just want to avoid hard coding the version in too many places when we can easily get to it.

openmls/src/framing/message.rs
```diff
@@ -30,6 +30,7 @@ use crate::error::LibraryError;
 /// ```
 #[derive(PartialEq, Debug, Clone, TlsSerialize, TlsSize, TlsDeserialize)]
 pub(crate) struct MlsMessage {
+    pub(crate) protocol_version: ProtocolVersion,
```
franziskuskiefer, line 33: You call it `version` in `FramedContentTbs`, and that's what it is in the spec as well. Maybe let's use only version?

openmls/src/framing/mls_content.rs
```diff
@@ -234,6 +236,7 @@ impl FramedContentTbs {
             body,
         };
         FramedContentTbs {
+            version: ProtocolVersion::Mls10,
```
franziskuskiefer, line 239: This should probably get a builder to set all the things. But not in this PR.

openmls/src/framing/public_message.rs
```diff
@@ -45,7 +45,7 @@ pub(crate) struct MembershipTag(pub(crate) Mac);
 /// ```
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
 pub(crate) struct PublicMessage {
-    content: FramedContent,
+    content: FramedContentTbs,
```
franziskuskiefer, line 48: In the spec this is `FramedContent`. Why do we have a Tbs here?

openmls/src/framing/public_message.rs
```diff
@@ -112,17 +117,23 @@ impl PublicMessage {
...
         Self {
-            content: v.tbs.content,
+            content: v.tbs,
             auth: v.auth,
             membership_tag: None,
         }
     }
 }
 
+impl From<PublicMessage> for VerifiableAuthenticatedContent {
```
franziskuskiefer, line 127: Should this be possible? Isn't the `VerifiableAuthenticatedContent` there to enforce verification?

openmls/src/framing/public_message.rs
```diff
@@ -196,47 +218,24 @@ impl PublicMessage {
...
-    /// Turns this [`PublicMessage`] into a [`VerifiableAuthenticatedContent`]. The
-    /// `serialized_context` is added only if the [`Sender`] of the plaintext is
-    /// [`Sender::Member`] or [`Sender::NewMemberCommit`] as per MLS specification.
-    pub fn into_verifiable_content(
-        self,
-        serialized_context: &[u8],
-    ) -> VerifiableAuthenticatedContent {
+    /// Set the context for later verification. This should only be done for
+    /// messages with [`Sender::Member`] and [`Sender::NewMemberCommit`].
+    pub(super) fn set_context(&mut self, context: &[u8]) {
```
franziskuskiefer, line 231: It shouldn't only be done for these but it won't happen for any other sender.
You call it unconditionally and the function sets it correctly. Maybe change the comment to say that? What it currently says is a little confusing.

franziskuskiefer (changes requested):
Thanks, generally looks good to me. Just a few comments and questions.

openmls/src/framing/public_message.rs
kkohbrok: Because it carries metadata (WireFormat and ProtocolVersion) that we have to pass down the chain anyway as we unwrap the message (i.e. from MlsMessage to PublicMessage to AuthenticatedContent, etc.). We don't have to do it this way, but I thought it would be an elegant way to do it.

openmls/src/framing/public_message.rs
kkohbrok: Yes, but it's downstream of PublicMessage. After validating the membership tag, we have to turn the PublicMessage into a VerifiableAuthenticatedContent. Ideally, there would be distinct types for before and after MembershipTag validation, but I think that's something we want to do generically for MAC validation at some point.

openmls/src/framing/public_message.rs
kkohbrok: Good point, I'll change the comment. Again, we probably want to enforce this on a type level at some point, but maybe not in this PR.

openmls/src/framing/mls_content.rs
```diff
@@ -195,8 +196,9 @@ impl<'a> AuthenticatedContentTbm<'a> {
     }
 }
 
-#[derive(PartialEq, Debug, Clone)]
+#[derive(PartialEq, Debug, Clone, Serialize, Deserialize)]
 pub(crate) struct FramedContentTbs {
+    pub(super) version: ProtocolVersion,
```
franziskuskiefer, line 201: Looks like this isn't serialized?

openmls/src/framing/public_message.rs
franziskuskiefer: I see. That's of course a little confusing, esp. with the comment.
Maybe note it in the comment why we do this and that the (de)serialization accounts for this?
Also, we call `tls_serialize` on this in `set_membership_tag` and in `verify_membership`, which appears to want the TBS. The spec is a little weird here because the tbs contains the struct, not the other way. We have to make sure that we know when we need the tbs and when the content and don't be too confusing here.

kkohbrok: Good point. I added a comment to the doc comments as you suggest.

It might cause some confusion, but on the other hand the change is accounted for in the serialization/deserialization functions and the fields are private, so anything outside of the module doesn't need to worry about it.

Both set_membership_tag and verify_membership should use the TBS, which I believe they indeed do.

openmls/src/framing/public_message.rs
```diff
@@ -144,7 +150,12 @@ impl PublicMessage {
         }
     }
 
-    /// Returns a reference to the `content` field.
+    /// Returns a reference to the [`ProtocolVersion`].
```
franziskuskiefer, line 153:
```suggestion
    /// Returns the [`ProtocolVersion`].
```

franziskuskiefer (approved):
Thanks (with one suggestion)

openmls/src/framing/public_message.rs
franziskuskiefer: Oh, we don't enforce resolving conversations. Not too important 🤷🏻

PR #1168, opened 2022-12-30
Filtered direct path

### LINKED ISSUE #875: [MLS Spec change] Stop generating redundant PKE key pairs. (#587)
This PR changes the way that nodes are provisioned with PKE key pairs, as well as some tree mechanics such as truncation. See the PR description in the MLS spec repo for details.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/587

raphaelrobert:
Fixes #875.

Another preparatory PR for the full tree implementation.

### DIFF
--- a/openmls/src/treesync/diff.rs
+++ b/openmls/src/treesync/diff.rs
@@ -19,6 +19,7 @@
 //! [`LibraryError`](TreeSyncDiffError::LibraryError).
 use openmls_traits::{types::Ciphersuite, OpenMlsCryptoProvider};
 use serde::{Deserialize, Serialize};
+use tls_codec::VLBytes;
 
 use std::collections::HashSet;
 
@@ -27,7 +28,7 @@ use super::{
     node::{
         leaf_node::{LeafNode, OpenMlsLeafNode},
         parent_node::{ParentNode, PathDerivationResult, PlainUpdatePathNode},
-        Node,
+        Node, NodeReference,
     },
     treesync_node::{TreeSyncLeafNode, TreeSyncNode, TreeSyncParentNode},
     TreeSync, TreeSyncParentHashError, TreeSyncSetPathError,
@@ -35,7 +36,7 @@ use super::{
 
 use crate::{
     binary_tree::{
-        array_representation::{direct_path, LeafNodeIndex, ParentNodeIndex, TreeNodeIndex},
+        array_representation::{LeafNodeIndex, ParentNodeIndex, TreeNodeIndex},
         MlsBinaryTreeDiff, StagedMlsBinaryTreeDiff,
     },
     ciphersuite::{HpkePrivateKey, HpkePublicKey, Secret},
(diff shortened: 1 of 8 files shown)


openmls/src/treesync/diff.rs
```diff
@@ -525,32 +573,52 @@ impl<'a> TreeSyncDiff<'a> {
...
-        let mut direct_path = direct_path(leaf_index, self.diff.size());
-        if !direct_path.is_empty() {
-            // Remove root
-            direct_path.pop();
+        let mut copath_resolutions = Vec::new();
+        for node_index in self.diff.copath(leaf_index) {
+            let resolution = self.resolution(node_index, &HashSet::new());
+            copath_resolutions.push(resolution);
+        }
+        copath_resolutions
```
kkohbrok, line 587: Not crucial, but since we're using iterators everywhere, we might as well do it here, too. The same goes for the function below.

openmls/src/treesync/diff.rs
```diff
@@ -835,8 +913,29 @@ impl<'a> TreeSyncDiff<'a> {
             .collect()
     }
 
-    /// Get the length of the direct path of the given [`LeafNodeIndex`].
-    pub(super) fn direct_path_len(&self, leaf_index: LeafNodeIndex) -> usize {
-        self.diff.direct_path(leaf_index).len()
+    /// Returns the filtered common path two leaf nodes share. If the leaves are
+    /// identical, the common path is empty.
```
kkohbrok, line 917: When the leaves are identical, shouldn't the common path be the leaf's direct path?

openmls/src/treesync/node.rs
```diff
@@ -16,3 +16,9 @@ pub enum Node {
     LeafNode(OpenMlsLeafNode),
     ParentNode(ParentNode),
 }
+
+/// Container enum with reference to a node in a tree.
+pub(crate) enum NodeReference<'a> {
+    Leaf(&'a OpenMlsLeafNode),
+    Parent(&'a ParentNode),
+}
```
kkohbrok, line 24: It is a bit annoying that we have to have an extra struct to accommodate these references, but I don't see a way to make it work like `&Node`.

openmls/src/treesync/treekem.rs
```diff
@@ -137,19 +150,18 @@ impl<'a> TreeSyncDiff<'a> {
...
-
         let mut path: Vec<ParentNode> = params
             .update_path
             .into_iter()
             .take(path_position)
             .map(|update_path_node| update_path_node.public_key.into())
             .collect();
-        path.append(&mut derived_path);
-
-        debug_assert_eq!(path.len(), _update_path_len);
```
kkohbrok, line 152: Can't we leave this assert in here?

kkohbrok (changes requested):
Looks good! Just a few comments and nits.

openmls/src/treesync/node.rs
raphaelrobert: I find it quite practical tbh. And yes, I also don't see a way to combine it with Node (unless we do something super complicated).

kkohbrok (approved):

Issue #720, opened 2022-01-21
[MLS Spec change]  Require extensions to be unique in any given extensions vector (#535)


kkohbrok:
**Description of the changes**
See title.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/535

duesee:
The change is ...

> * Any field containing a list of extensions MUST NOT have more than one extension of any given type.

I would like to tackle this by introducing a Wrapper around `Vec` that ensures uniqueness and fixup all compilation problems afterwards.

duesee:
Before tackling this, we should introduce a `KeyPackageBundleBuilder`. This way, we can cleanup the API and avoid to touch all tests again when solving this issue here.

PR #1170, opened 2023-01-02
feat: Introduce `Extensions` type.

duesee:
This PR introduces an `Extensions` type that should eventually replace all instances of `Vec<Extension>` and generally unify extension handling. Specifically, `Extensions` ensures that there are no duplicate extensions. Thus, this PR is a step towards closing #720.

Some preliminary notes:

This PR only introduces the `Extensions` struct for now. The replacement of `Vec<Extension>` everywhere requires quite some mechanical changes. The last time I worked on this, I proposed to introduce a `KeyPackageBundleBuilder` first to get rid of many `Extensions::empty/default()` calls. This is why we decided to close the last PR.

The functions are a bit repetitive but I think this is worth it given that they are easy to maintain and will eventually simplify some more complex code.

As discussed offline, there doesn't seem to be anything straightforward to lookup an element by type and already "convert" it from `Extension` to, e.g., `ApplicationId`.

Edit: This PR doesn't handle unknown extensions. We can tackle this in some follow-up.

Alternative: Use a `BTreeSet`. Downside: This requires to introduce an `ExtensionWrapped` that implements `{Partial,}Ord, {Partial,}Eq` using `ExtensionType` only. Also, we would need `tls_codec::` pass-throughs, etc. We could do that but it will result in quite some more code.

### DIFF
--- a/openmls/src/extensions/mod.rs
+++ b/openmls/src/extensions/mod.rs
@@ -521,7 +521,7 @@ mod test {
             (vec![x.clone(), y.clone(), y.clone()], false),
             (vec![y.clone(), x.clone(), x.clone()], false),
             (vec![x.clone(), y.clone(), x.clone()], false),
-            (vec![y.clone(), x.clone(), y.clone()], false),
+            (vec![y.clone(), x, y], false),
         ];
 
         for (test, should_work) in tests.into_iter() {
--- a/openmls/src/extensions/mod.rs
+++ b/openmls/src/extensions/mod.rs
@@ -171,7 +171,7 @@ pub enum Extension {
 }
 
 /// A list of extensions with unique extension types.
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Default, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct Extensions {
     map: HashMap<ExtensionType, Extension>,
 }
@@ -207,20 +207,6 @@ impl Extensions {
         Self {
             map: HashMap::new(),
         }
-    }
-
-    /// Create an extension list that contains a single extension.
-    pub fn single(extension: Extension) -> Self {
-        Self {
-            map: HashMap::from([(extension.extension_type(), extension)]),
-        }
-    }
-
-    /// Create an extension list that contains multiple extensions.
-    ///
-    /// This function will fail when the list of extensions contains duplicate extension types.
-    pub fn multi(extensions: Vec<Extension>) -> Result<Self, InvalidExtensionError> {
-        extensions.try_into()
     }
 
     // ---------------------------------------------------------------------------------------------
@@ -265,12 +251,6 @@ impl Extensions {
     }
 }
 
-impl Default for Extensions {
-    fn default() -> Self {
-        Self::empty()
-    }
-}
-
 impl TryFrom<Vec<Extension>> for Extensions {
     type Error = InvalidExtensionError;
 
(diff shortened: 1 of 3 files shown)


openmls/src/extensions/mod.rs
```diff
@@ -20,7 +20,7 @@
 //! - [`ExternalPubExtension`] (GroupInfo extension)
 
 use serde::{Deserialize, Serialize};
-use std::{convert::TryFrom, fmt::Debug};
+use std::{collections::BTreeSet, convert::TryFrom, fmt::Debug};
```
franziskuskiefer, line 23: While you're here you can drop the `TryFrom`.

openmls/src/extensions/mod.rs
```diff
@@ -166,6 +166,200 @@ pub enum Extension {
...
+    pub fn single(extension: Extension) -> Self {
+        Self {
+            inner: vec![extension],
+        }
+    }
+
+    /// Create an extension list that contains multiple extensions.
+    ///
+    /// This function will fail when the list of extensions contains duplicate extension types.
+    pub fn multi(extensions: Vec<Extension>) -> Result<Self, InvalidExtensionError> {
```
franziskuskiefer, line 193: The name is a little weird. But since `TryInto` is implemented we don't need it.

openmls/src/extensions/mod.rs
```diff
@@ -166,6 +166,200 @@ pub enum Extension {
...
+        extensions.try_into()
+    }
+
+    // ---------------------------------------------------------------------------------------------
+
+    /// Add an extension to the extension list.
+    ///
+    /// Returns an error when there already is an extension with the same extension type.
+    pub fn add(&mut self, extension: Extension) -> Result<(), InvalidExtensionError> {
+        if !self.contains(extension.extension_type()) {
```
franziskuskiefer, line 203: Switch the cases around to have an early return.

openmls/src/extensions/mod.rs
```diff
@@ -166,6 +166,200 @@ pub enum Extension {
...
+        replaced
+    }
+
+    /// Remove an extension from the extension list.
+    ///
+    /// Returns the removed extension or an error when there is no extension with the given extension type.
+    pub fn remove(
+        &mut self,
+        extension_type: ExtensionType,
+    ) -> Result<Extension, InvalidExtensionError> {
```
franziskuskiefer, line 226: Data structures usually don't error when removin a non-existing element but return an `Option` and `None` in case the element is not removed. We should do the same here.

openmls/src/extensions/mod.rs
```diff
@@ -166,6 +166,200 @@ pub enum Extension {
...
+
+    /// Check that the candidate extension list is valid.
+    ///
+    /// Specifically, the candidate list must ...
+    ///
+    /// * ... not contain duplicate extension types (ValSem012) ...
+    ///
+    /// ... to be valid.
+    pub fn validate(candidate: &[Extension]) -> Result<(), InvalidExtensionError> {
+        let mut map = BTreeSet::new();
```
franziskuskiefer, line 274: The operations performed on the internal vector would all be more efficient when you'd always use a set. You also could get rid of this function. Is there a reason not to use a set for `inner`?

franziskuskiefer (changes requested):
Thanks a few comments and questions about the design.

openmls/src/extensions/mod.rs
duesee: Can you say more? I assume that a successful add will be the happy path?

openmls/src/extensions/mod.rs
duesee: As discussed privately, using a set would require a wrapper around `Extension` that implements a custom `Ord` and we could use a `HashMap` instead.

openmls/src/extensions/mod.rs
duesee: I do like the conversion traits a lot but its sometimes not clear that these are implemented. That's why I wanted to provide this constructor. Given that we have `empty` and `single`, I think it is reasonable for a user to expect another one for `multi`?

openmls/src/extensions/mod.rs
duesee: Ahh... Rust 2021 has `TryFrom` in the prelude? I didn't know :-)

duesee:
Okay, next try. Using `HashMap` requires us to provide a custom `tls_codec` implementation. But, thinking about it, this seems like the more appropriate solution anyway. If we would keep `Vec`, e.g., to spare us the `tls_codec` impls, it would be possible to miss the duplicate check. having this check so close to `Extensions` makes this really robust.

openmls/src/extensions/mod.rs
```diff
@@ -166,6 +170,180 @@ pub enum Extension {
...
+        }
+
+        Ok(Self { map })
+    }
+}
+
+impl Extensions {
+    /// Get a reference to the [`ApplicationIdExtension`] if there is any.
+    pub fn application_id(&self) -> Option<&ApplicationIdExtension> {
+        for extension in self.map.values() {
```
franziskuskiefer, line 293: There's no need to iterate here. We have the map. Something like this would work.
```rust
self.map.get(&ExtensionType::ApplicationId)
  .and_then(|e| match e {
    Extension::ApplicationId(e) => Some(e),
    _ => None,
})
```

franziskuskiefer (changes requested):
Thanks, a few more comments.

openmls/src/extensions/mod.rs
duesee: Thanks, havn't thought about readability here. Makes sense :+1:

openmls/src/extensions/mod.rs
duesee: Oh yeah, thanks! Got a bit lazy...

openmls/src/extensions/mod.rs
```diff
@@ -166,6 +170,180 @@ pub enum Extension {
...
+        let got = self.map.remove(&extension.extension_type());
+
+        match got {
+            Some(extension) => Ok(extension),
+            None => Err(InvalidExtensionError::NotFound),
+        }
+    }
+}
+
+impl Default for Extensions {
```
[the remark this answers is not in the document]
duesee: > Can't you derive this?

Good point!

> Also, having Default as well as empty() seems redundant (like multi).

Yes, it unfortunately is. But I would make the same argument as with `TryFrom` and `multi`. I think that users generally do expect to have constructors. Conversion traits are more difficult to discover.

From [Rust API guidelines](https://rust-lang.github.io/api-guidelines/predictability.html#c-ctor):

> "Note that it is common and expected for types to implement both Default and a new constructor. For types that have both, they should have the same behavior. Either one may be implemented in terms of the other."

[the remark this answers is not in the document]
duesee: I removed the methods for now. We can reintroduce as soon as we need them. (With the upcoming `KeyPackageBuilder` PR this will be less relevant.)

franziskuskiefer (approved):
Let's get this in.

PR #1171, opened 2023-01-02
ci: Run Clippy during pull_request event.

duesee:


franziskuskiefer (changes requested):
This shouldn't be touched before there's a good solution for enabling clippy for everyone.

duesee:
As discussed privately, we won't use the Clippy action for now and just run `cargo clippy` directly.

franziskuskiefer (approved):
thanks

.github/workflows/clippy.yml
```diff
@@ -1,14 +1,18 @@
...
   clippy_check:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - run: rustup component add clippy
-      - uses: actions-rs/clippy-check@v1
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          args: -p openmls --tests -- -D warnings
+      - run: cargo clippy -p openmls --tests -- -D warnings
```
raphaelrobert, line 18: I think you need to specify something like ` --message-format=json` so that GH can parse it for annotations.

duesee: Oh, good point. I'll make a follow-up.

PR #1172, opened 2023-01-03
Extend WireFormat as per spec

### LINKED ISSUE #979: [MLS Spec change] Move wire_format to a separate tagged-union structure MLSMessage #523

kkohbrok:
This PR finally fixes #979 by extending the WireFormat to GroupInfo, Welcome and KeyPackages. This caused some changes to the `MlsGroup` API.

* The `MlsMessageIn` and `MlsMessageOut` structs corresponds to the MLSMessage struct from the spec
* Before usage with the MlsGroup API, an `MlsMessageIn` first needs to be "extracted" into a `ProtocolMessage`s, i.e. a Public or PrivateMessage (feel free to propose a different name), a Welcome message, a GroupInfo etc.
* Instead of `MlsMessageIn`, MlsGroup functions for message processing now take as input the extracted struct, e.g. ProtocolMessages
* MlsGroup functions now generally return `MlsMessageOut` rather than Welcome, etc.
  * MlsMessageOut is its own struct now (rather than containing an MlsMessage), as it contains a GroupInfo rather than a VerifiableGroupInfo. If we extend type-based validation, more structs will differ between the two.
* There is still a boundary between MlsMessageIn and MlsMessageOut s.t. one cannot be converted to the other without serialization and subsequent deserialization
* The DS needs to both de-serialize and serialize, which is why it now uses the test-utils feature flag of OpenMLS, where a direct conversion is possible

This PR is best reviewed per-commit. The first one contains the structural changes to structs, the second the changes to the APIs, followed by two commits that deal with the resulting changes to DS and CLI, as well as the tests.


openmls/src/framing/mod.rs
```diff
@@ -90,6 +92,12 @@ pub enum WireFormat {
     PublicMessage = 1,
     /// Encrypted message
     PrivateMessage = 2,
+    /// Welcome message
```
franziskuskiefer, line 95: Unrelated to this PR but the IANA values in the spec still use plaintext and cihpertext.

openmls/src/framing/message_out.rs
```diff
@@ -0,0 +1,187 @@
...
+///             KeyPackage key_package;
+///     }
+/// } MLSMessage;
+/// ```
+#[derive(Debug, PartialEq, Clone, TlsSerialize, TlsSize)]
+#[cfg_attr(feature = "test-utils", derive(TlsDeserialize))]
+#[repr(u8)]
+pub(crate) enum MlsMessageOutBody {
+    /// Plaintext message
+    #[tls_codec(discriminant = 1)]
```
franziskuskiefer, line 57: It would be nice if we wouldn't have to define the wire format value twice. Maybe for a follow up.

openmls/src/framing/message_out.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub(crate) fn from_welcome(welcome: Welcome, version: ProtocolVersion) -> Self {
+        MlsMessageOut {
+            version,
+            body: MlsMessageOutBody::Welcome(welcome),
+        }
+    }
+
+    /// Turn an [`MlsMessageOut`] into a [`Welcome`].
+    #[cfg(any(feature = "test-utils", test))]
+    pub fn into_welcome(self) -> Option<Welcome> {
```
franziskuskiefer, line 120: Lets move the `#[cfg(any(feature = "test-utils", test))]` to a separate `impl` block to have nicer separation.

franziskuskiefer (approved):
Thanks, looks good to me.
One nit, but otherwise this is good to go in.

openmls/src/framing/mod.rs
kkohbrok: Yes, I noticed that, too. That's why I filed https://github.com/mlswg/mls-protocol/pull/844.

openmls/src/framing/message_out.rs
kkohbrok: That would be nice, but I think we have to if we want to include different message types for `In` and `Out` message. This is already the case for `GroupInfo` vs. `VerifiableGroupInfo` and there are going to be more values like that in the future.

franziskuskiefer (approved):

franziskuskiefer (approved):
:shipit:

PR #1173, opened 2023-01-03
feat: Introduce `KeyPackageBuilder`.

duesee:
This PR introduces a `KeyPackageBuilder`. This simplifies `KeyPackage` creation and reduces the times when we have to touch the tests. Specifically, changing all instances of `Vec<Extension>` to `Extensions` would require to touch all `KeyPackage::create` instances. After this PR is merged, this is not needed anymore. Thus, in a way, I see this as preliminary work towards closing #720.

### DIFF
--- a/openmls/src/key_packages/mod.rs
+++ b/openmls/src/key_packages/mod.rs
@@ -41,15 +41,13 @@
 //!     &backend,
 //! )
 //! .expect("Error creating credential.");
-//! let key_package = KeyPackage::create(
+//! let key_package = KeyPackage::builder().build(
 //!     CryptoConfig {
 //!         ciphersuite,
 //!         version: ProtocolVersion::default(),
 //!     },
 //!     &backend,
 //!     &credential_bundle,
-//!     vec![],
-//!     vec![],
 //! )
 //! .unwrap();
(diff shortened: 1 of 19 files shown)


openmls/src/key_packages/mod.rs
```diff
@@ -480,6 +485,69 @@ impl KeyPackage {
...
+    pub fn new() -> Self {
+        Self {
+            init_key: None,
+            key_package_extensions: None,
+            leaf_node_extensions: None,
+        }
+    }
+
+    /// Set the init key.
+    pub fn init_key(mut self, init_key: Vec<u8>) -> Self {
```
franziskuskiefer, line 507: It looks like this is never used? Can we just drop this version?

franziskuskiefer (changes requested):
Thanks, lgtm
I think we can skip the `init_key` though.

franziskuskiefer (approved):
Thanks

duesee:
Sorry, @franziskuskiefer. I forgot that new pushes dismiss reviews. Can you approve again?

franziskuskiefer (approved):

PR #1177, opened 2023-01-04
Ensure init_key and encryption_key are different

### LINKED ISSUE #885: [MLS Spec change] Require init key and leaf key to be different (#618)
Adds a validation check that ensures that the two key pairs in a key package are different.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/618

franziskuskiefer:
This PR separates the init_key from the key package and the encryption_key in the leaf node.

Changes
- Creating a new key package generates a fresh encryption key for the leaf node now. The key pair is stored in the key store with the labelled init_key as index.
- When processing a welcome the encryption key is read from the key store and used to build the tree.
- Remove ValSem105: init keys can't be checked anymore because they are not in the tree
- Introduce ValSem113: init key and encryption key must be different (checks only in add proposal for now)
- Introduce ValSem114: encryption keys must be unique in the tree (checks only in add proposal for now)

Fixes #885


openmls/src/group/core_group/new_from_welcome.rs
```diff
@@ -26,6 +26,19 @@ impl CoreGroup {
...
+            .key_store()
+            .read(&LeafNode::encryption_key_label(
+                key_package_bundle.key_package.hpke_init_key().as_slice(),
+            ))
+            .ok_or(WelcomeError::NoMatchingEncryptionKey)?;
+        backend
+            .key_store()
+            .delete(key_package_bundle.key_package.hpke_init_key().as_slice())
+            // This error really shouldn't happen. We just read the value.
+            .map_err(|_| WelcomeError::NoMatchingEncryptionKey)?;
```
kkohbrok, line 40: Would it be worth extending the key store with a read-and-delete operation?

openmls/src/group/core_group/validation.rs
```diff
@@ -137,14 +137,31 @@ impl CoreGroup {
...
+            // ValSem113
+            if proposal_init_key == proposal_encryption_key.as_ref() {
+                return Err(ProposalValidationError::InitEncryptionKeyCollision);
+            }
+
             // ValSem102
-            if !public_key_set.insert(public_key) {
+            if !init_key_set.insert(proposal_init_key) {
+                return Err(ProposalValidationError::DuplicatePublicKeyAddProposal);
+            }
```
kkohbrok, line 161: I don't think this check is strictly necessary in that it is not mandated by the spec. If I understand Section 10.1 correctly, it is technically okay for two KeyPackages to have the same init key. No need to remove it with this PR, though I guess.

openmls/src/group/core_group/validation.rs
```diff
@@ -137,14 +137,31 @@ impl CoreGroup {
...
+
+            // ValSem114
+            if !encryption_key_set.insert(proposal_encryption_key.as_slice().to_vec()) {
                 return Err(ProposalValidationError::DuplicatePublicKeyAddProposal);
             }
```
kkohbrok, line 166: Maybe note that this tests ValSem114 only partially with the rest of the test being a little ways below?

openmls/src/group/tests/test_proposal_validation.rs
```diff
@@ -1062,11 +1062,11 @@ fn test_valsem104(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider
...
 }
 
-/// ValSem105:
+/// ValSem113:
 /// Add Proposal:
-/// HPKE init key in proposals must be unique among existing group members
+/// HPKE init key and encryption key must be different
 #[apply(ciphersuites_and_backends)]
-fn test_valsem105(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider) {
+fn test_valsem113(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider) {
```
kkohbrok, line 1069: Is there also a test for the new ValSem114? Or did I miss it? It would be nice to have at least a basic test.

openmls/src/treesync/treekem.rs
```diff
@@ -97,6 +97,13 @@ impl<'a> TreeSyncDiff<'a> {
         // ValSem202: Path must be the right length
         let direct_path_length = self.filtered_direct_path(params.sender_leaf_index).len();
         if direct_path_length != params.update_path.len() {
+            // XXX: Rewrite tests to allow for debug asserts.
```
kkohbrok, line 100: What do you mean here? Is this meant to be for a follow-up?

kkohbrok (changes requested):
Thanks for the PR! A few comments, but nothing big.

openmls/src/treesync/treekem.rs
franziskuskiefer: Generally it would be nice to be able to write debug asserts. But we test a lot for errors and hence can't use debug asserts without rewriting the tests.
This one here I used for testing and just left in for future reference to use once we can. But if you prefer, I can remove it.

kkohbrok (approved):
Thanks for addressing my comments. Let's get this in and get started with the follow-ups!

PR #1180, opened 2023-01-05
test: Ensure commits not containing all enqueued proposals can be processed

### LINKED ISSUE #999: [MLS Spec change] Downgrade MUST to SHOULD for commit senders including all valid commits (#546)
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/546

> Addresses the comments in ( Resolves #542 )
 > 
 > Since a member performing a commit including all valid proposals within an epoch is best effort, it makes sense to downgrade the requirement from MUST to SHOULD. 
 > 
 > IMO, MUST should be reserved for instances where the receiver would throw an error in the condition the sender did not 100% comply with the rule. Since in this case it is not a hard error under certain conditions, such as the member performing the commit having not received a proposal in transit while generating the commit, SHOULD is a better choice.

augustocdias:
Closes #999 

So from my understanding OpenMLS already has the change implemented although not tested.

I'm adding a test to ensure that OpenMLS is capable of processing commits not containing all proposals in the store of a peer. I understand that this use case is not possible to implement with OpenMLS at the moment (as you can see in the test with the proposal store being manipulated) but it is valid according to the spec. Since OpenMLS should be interoperable with other implementations I believe it is worth to have such a test there. If OpenMLS eventually supports this use case, the test should be adapted accordingly.

### DIFF
--- a/openmls/src/group/tests/test_commit_validation.rs
+++ b/openmls/src/group/tests/test_commit_validation.rs
@@ -672,3 +672,65 @@ fn test_valsem205(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider
         .process_message(backend, MlsMessageIn::from(original_plaintext))
         .expect("Unexpected error.");
 }
+
+// this ensures that a member can process commits not containing all the stored proposals
+#[apply(ciphersuites_and_backends)]
+fn test_partial_proposal_commit(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider) {
+    // Test with PublicMessage
+    let CommitValidationTestSetup {
+        mut alice_group,
+        mut bob_group,
+        ..
+    } = validation_test_setup(PURE_PLAINTEXT_WIRE_FORMAT_POLICY, ciphersuite, backend);
+
+    let charlie_index = alice_group
+        .members()
+        .find(|m| m.identity == b"Charlie")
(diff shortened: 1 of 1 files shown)


openmls/src/group/tests/test_commit_validation.rs
```diff
@@ -672,3 +672,65 @@ fn test_valsem205(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider
...
+        ProcessedMessageContent::ProposalMessage(p) => bob_group.store_pending_proposal(*p),
+        _ => unreachable!(),
+    }
+
+    // Alice creates a commit with only a subset of the epoch's proposals. Bob should still be able to process it.
+    let remaining_proposal = alice_group
+        .proposal_store
+        .proposals()
+        .cloned()
+        .next()
```
kkohbrok, line 719: Clippy wants these two lines swapped.

kkohbrok (changes requested):
Thanks for the PR! More tests are always welcome. One small request by clippy. Otherwise this is good to go!

openmls/src/group/tests/test_commit_validation.rs
augustocdias: Done :)

kkohbrok (approved):
Thanks!

PR #1182, opened 2023-01-06
fix: Remove `GroupInfo`s `tls_codec::Deserialize` implementation. 

### LINKED ISSUE #1181: `GroupInfo` can be deserialized, thus skipping the verification step

duesee:
This implementation should not exist because it can be used to bypass signature verification.

This closes #1181.

### DIFF
--- a/openmls/src/messages/mod.rs
+++ b/openmls/src/messages/mod.rs
@@ -241,6 +241,7 @@ impl Signable for GroupInfoTBS {
 /// } GroupInfo;
 /// ```
 #[derive(Debug, PartialEq, Clone)]
+#[cfg_attr(feature = "test-utils", derive(TlsDeserialize))]
 pub struct GroupInfo {
     payload: GroupInfoTBS,
     signature: Signature,
@@ -260,11 +261,6 @@ impl GroupInfo {
     /// Returns the confirmation tag.
     pub(crate) fn confirmation_tag(&self) -> &ConfirmationTag {
         &self.payload.confirmation_tag
-    }
-
-    /// Returns the signer.
-    pub(crate) fn signer(&self) -> LeafNodeIndex {
-        self.payload.signer
     }
 
     #[cfg(any(feature = "test-utils", test))]
--- a/openmls/src/messages/tests/test_welcome.rs
+++ b/openmls/src/messages/tests/test_welcome.rs
@@ -6,7 +6,8 @@ use crate::{
     group::{config::CryptoConfig, errors::WelcomeError, GroupId, MlsGroup, MlsGroupConfigBuilder},
     key_packages::KeyPackage,
     messages::{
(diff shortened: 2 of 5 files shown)


openmls/src/messages/mod.rs
```diff
@@ -241,6 +241,7 @@ impl Signable for GroupInfoTBS {
 /// } GroupInfo;
 /// ```
 #[derive(Debug, PartialEq, Clone)]
+#[cfg_attr(feature = "test-utils", derive(TlsDeserialize))]
```
duesee, line 244: This is needed because `MlsMessageOutBody` derives `TlsDeserialize` under the same conditions and uses `GroupInfo`.

openmls/src/messages/mod.rs
```diff
@@ -262,11 +263,6 @@ impl GroupInfo {
         &self.payload.confirmation_tag
     }
 
-    /// Returns the signer.
```
duesee, line 265: This is not used anymore in `pub(crate)`. If we want to keep it, we should define a public API.

openmls/src/messages/tests/test_welcome.rs
```diff
@@ -162,26 +163,24 @@ fn test_welcome_ciphersuite_mismatch(
     let group_info_bytes = welcome_key
         .aead_open(backend, welcome.encrypted_group_info(), &[], &welcome_nonce)
         .expect("Could not decrypt GroupInfo.");
-    let mut group_info = GroupInfo::tls_deserialize(&mut group_info_bytes.as_slice())
-        .expect("Could not deserialize GroupInfo.");
```
duesee, line 166: Expects like this can easily get out-of-sync and the `expect` doesn't really add information here. Thus, I used `unwrap` instead.

kkohbrok (approved):
Looks good to me, thanks!

openmls/src/messages/tests/test_welcome.rs
```diff
@@ -162,26 +163,24 @@ fn test_welcome_ciphersuite_mismatch(
...
 
-    let encrypted_group_info = welcome_key
-        .aead_seal(backend, &group_info_bytes, &[], &welcome_nonce)
-        .expect("Could not encrypt GroupInfo.");
+    let encrypted_verifiable_group_info = welcome_key
+        .aead_seal(backend, &verifiable_group_info_bytes, &[], &welcome_nonce)
+        .unwrap();
 
-    welcome.encrypted_group_info = encrypted_group_info.into();
+    welcome.encrypted_group_info = encrypted_verifiable_group_info.into();
```
duesee, line 183: Here, we "coerce" a previously *unverified* `group info (seemingly) into a group info (`encrypted_group_info` vs `unverified_encrypted_group_info`). I'm not sure if this is something we should worry about right now.

kkohbrok: What we want would be something like: Encryption only for `GroupInfo`, decryption should always yield a `UnverifiedGroupInfo`. For tests, I think coercion is okay.

Issue #1187, opened 2023-01-09
ValSem103 & ValSem104 need partial rewrite


raphaelrobert:
After introducing the new parent hash with #995, some parts of the tests don't work anymore because the new parent hash is strictly stronger and makes manipulation of Commit messages harder.

PR #1188, opened 2023-01-09
Introduce new parent hash

### LINKED ISSUE #883: [MLS Spec change] Update parent hash validation (#603)
I believe this is only editorial, but we should make sure we're doing what's in the text. Filing this just to be safe.

**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/603

### LINKED ISSUE #995: [MLS Spec change] Stronger parent hashes for authenticated identities (#527)

raphaelrobert:
Fixes #995.
Fixes #883.

The changes should be straightforward. 

Since the new parent hash ist stronger, some parts of ValSem103 and ValSem104 no longer work. This is tracked in #1187, since it requires a larger rewrite that would be out-of-scope for this PR.


openmls/src/treesync/treesync_node.rs
```diff
@@ -145,21 +147,46 @@ impl TreeSyncParentNode {
...
+                Some(&new_node),
+                VLByteSlice(&left_hash),
+                VLByteSlice(&right_hash),
+            )
+            .hash(backend, ciphersuite)?
+        } else {
+            // If the node is blank
+            TreeHashInput::new_parent(None, VLByteSlice(&left_hash), VLByteSlice(&right_hash))
+                .hash(backend, ciphersuite)?
+        };
```
kkohbrok, line 189: As discussed OOB there are some minor refactoring opportunities here around clone, but the impact will probably be minimal. (Although this function will be executed a number of times during the loop in parent hash verification.)

kkohbrok (changes requested):
Looks good, thanks for finishing this PR series! Just a few nits and minor refactoring opportunities.

openmls/src/treesync/treesync_node.rs
raphaelrobert: I looked into it and the change would require going through (and refactoring) many more layers until we can isolate unmerged leaves. The benefit would be that we only clone unmerged leaves and not the HPKE public key of the node.
I decided that the added code complexity was not worth the benefit. We can revisit this kind of optimisation in the context of performance improvements, right now this is in line with other performance levels.

PR #1190, opened 2023-01-10
Fix AuthenticatedContent serialization

### LINKED ISSUE #1189: Wrong serialization of AuthenticatedContent
TLS serialization of `AuthenticatedContent` appears to be wrong. The `wire_format` is missing in the type and it's not serialized as the first value in `FramedContentTbs`.

cc @kkohbrok

kkohbrok:
This PR ensures that AuthenticatedContent is serialized as defined in the MLS spec. As pointed out in #1189, it currently serializes ProtocolVersion as its first field, which it shouldn't do.

Fixes  #1189.


franziskuskiefer (changes requested):
Thanks this fixed the files issue but it's still broken I'm afraid. The serialisation is really complicated now. So I'm not sure if using the `FramedContentTbs` is really worth it. But let's keep it like this for now.

The `FramedContentTbs` has the context that appears to get serialised here as well but shouldn't.

```rust
AuthenticatedContent {
    tbs: FramedContentTbs {
        version: Mls10,
        wire_format: WireFormat::KeyPackage,
        content: FramedContent {
            group_id: GroupId::from_slice(&[1, 1, 1]),
            epoch: GroupEpoch::from(66),
            sender: Sender::NewMemberCommit,
            authenticated_data: vec![7, 7, 7].into(),
            body: FramedContentBody::Application(vec![8, 8, 8].into()),
        },
        serialized_context: Some(vec![1, 2, 3]),
    },
    auth: FramedContentAuthData {
        signature: Signature::from_slice(&vec![2, 2, 2]),
        confirmation_tag: None,
    },
}
```

produces

```
05
  03
    010101
  0000000000000042
  04
  03
    070707
  01
  00000003080808
  010203
  03
    020202
```

This shows some other issues as well
* Looks like `WireFormat` is a `u8` in OpenMLS but it should be a `u16` per spec. Can you fix this here as well?
* `FramedContentBody::Application` is not variable length encoded yet. Let's do this in a follow up.

kkohbrok:
Yeah, you're right. I originally thought including the TBS would be a good fit, but I had overlooked the issue with the `ProtocolVersion`. I'll revert the change and fix the issues you point out.

kkohbrok:
Rolling back the original changes to serialization had a bit more impact than expected. But it should now again work as intended.

openmls/src/framing/private_message.rs
```diff
@@ -366,15 +366,15 @@ impl PrivateMessage {
         );
 
         let verifiable = VerifiableAuthenticatedContent::new(
-            FramedContentTbs::new(
-                WireFormat::PrivateMessage,
-                self.group_id.clone(),
-                self.epoch,
+            WireFormat::PrivateMessage,
+            FramedContent {
```
franziskuskiefer, line 370: Is there a reason not to use the `new` function here?

franziskuskiefer (changes requested):
Thanks, looks much better. Only a few nits.
And please update `WireFormat` to `u16` per spec.

openmls/src/framing/private_message.rs
kkohbrok: I don't think there is a `new` function for `FramedContent`. Would you prefer if I added one?

franziskuskiefer: Ah, it's only on the body and the tbs. Nevermind then.

franziskuskiefer (approved):
Thanks

PR #1192, opened 2023-01-11
Allow to pass `Capabilities` and `Extensions` when creating an `OpenMlsLeafNode`.

duesee:
This (rather) resolves openmls/annotations#33 and resolves an `xxx:`s on the way.

### DIFF
--- a/openmls/src/key_packages/mod.rs
+++ b/openmls/src/key_packages/mod.rs
@@ -89,7 +89,7 @@ use crate::{
     extensions::Extensions,
     group::config::CryptoConfig,
     treesync::{
-        node::leaf_node::{LeafNodeSource, Lifetime},
+        node::leaf_node::{Capabilities, LeafNodeSource, Lifetime},
         LeafNode,
     },
     versions::ProtocolVersion,
@@ -267,6 +267,7 @@ impl KeyPackage {
             config,
             credential, // FIXME
             LeafNodeSource::KeyPackage(Lifetime::default()),
+            Capabilities::default(),
             leaf_node_extensions,
             backend,
         )?;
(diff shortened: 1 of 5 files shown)


openmls/src/treesync/node/leaf_node.rs
```diff
@@ -718,12 +747,15 @@ impl OpenMlsLeafNode {
         leaf_node_source: LeafNodeSource,
         backend: &impl OpenMlsCryptoProvider,
         credential_bundle: &CredentialBundle,
+        capabilities: Capabilities,
```
franziskuskiefer, line 750: Looks like comments need to be updated here on the function (and maybe elswhere).

openmls/src/key_packages/mod.rs
```diff
@@ -267,6 +267,7 @@ impl KeyPackage {
             config,
             credential, // FIXME
             LeafNodeSource::KeyPackage(Lifetime::default()),
+            Capabilities::default(),
```
franziskuskiefer, line 270: So we still don't set them. Let's do a follow up to allow actually passing in values.

franziskuskiefer (changes requested):
Thanks. Some comment needs updating. Otherwise this is good to go.

I see at least one follow up already that we should either file or tackle right after.

openmls/src/key_packages/mod.rs
duesee: Yeah. I tried to not touch key packages here. Will do a follow-up :+1:

duesee: https://github.com/openmls/openmls/issues/1195

franziskuskiefer (dismissed):
Thanks

franziskuskiefer:
looks like `rust fmt` isn't happy

duesee:
Urgs... now it should :+1:

franziskuskiefer (approved):

PR #1193, opened 2023-01-12
update with leaf node

franziskuskiefer:
Another followup to #819.

Update proposals operate on leaf nodes now, not key packages. This PR changes the API that this is reflect this. For this to work I also introduce a public `generate` function on the `LeafNode` to generate leaf nodes without key packages.

### DIFF
--- a/openmls/src/group/mls_group/updates.rs
+++ b/openmls/src/group/mls_group/updates.rs
@@ -1,7 +1,7 @@
 use core_group::create_commit_params::CreateCommitParams;
 use tls_codec::Serialize;
 
-use crate::{ciphersuite::HpkePublicKey, versions::ProtocolVersion};
+use crate::{ciphersuite::HpkePublicKey, treesync::LeafNode, versions::ProtocolVersion};
 
 use super::*;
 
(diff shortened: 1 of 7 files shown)


openmls/src/group/mls_group/updates.rs
```diff
@@ -145,19 +145,21 @@ impl MlsGroup {
...
         let mut rekeyed_own_leaf = tree
             .own_leaf_node()
             .ok_or_else(|| LibraryError::custom("The tree is broken. Couldn't find own leaf."))?
             .clone();
-        if let Some(key_package) = key_package {
+        if let Some(leaf) = leaf_node {
             let private_key: Vec<u8> = backend
                 .key_store()
-                .read(key_package.hpke_init_key().as_slice())
+                .read(&LeafNode::encryption_key_label(
```
kkohbrok, line 156: Is there already documentation on what we store under which label? Since this is all somewhat hidden due to the "global" nature of the key store, it might make sense to create an overview somewhere.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -399,6 +400,41 @@ impl LeafNode {
...
     }
 
+    /// Generate a fresh leaf node.
+    ///
+    /// This includes generating a new encryption key pair that is stored in the
+    /// key store.
+    ///
+    /// This function can be used when generating an update. In most other cases
+    /// a leaf node should be generated as part of a new [`KeyPackage`].
+    pub fn generate(
```
kkohbrok, line 410: This function is meant to be part of the public API, right? Maybe it would be worth marking it somehow s.t. it won't be used internally in the future.

kkohbrok (commented):
Looks good! Just two questions.

kkohbrok (approved):

PR #1196, opened 2023-01-16
chore: Cleanup `LeafNode` a bit.

duesee:
This is a pure cleanup PR. Having the leaf node methods sorted a bit makes it easier to see what is implemented.


franziskuskiefer (changes requested):
What's the purpose of moving the functions around? I don't think I understand what you are trying to achieve here.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -522,11 +473,80 @@ impl LeafNode {
             .map_err(|_| LibraryError::custom("Signing failed"))
     }
 
+    /// Expose [`new_with_key`] for tests.
+    #[cfg(any(feature = "test-utils", test))]
+    pub(crate) fn create_new_with_key(
```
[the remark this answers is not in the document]
duesee: Moved in [e7c5a5a](https://github.com/openmls/openmls/pull/1196/commits/e7c5a5a952234b9a67f82797c2b0e60337e690d2).

duesee:
> What's the purpose of moving the functions around? I don't think I understand what you are trying to achieve here.

I want to have constructors, field access, and validation somehow in order to make this easier to read.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -537,6 +537,12 @@ impl LeafNode {
         &self.payload.credential
     }
 
+    /// Replace the credential in the KeyPackage.
+    #[cfg(any(feature = "test-utils", test))]
+    pub(crate) fn set_credential(&mut self, credential: Credential) {
```
franziskuskiefer, line 542: This should go to the test `impl` as well.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -545,26 +551,39 @@ impl LeafNode {
...
+    }
+
+    /// Return a reference to [`Capabilities`].
+    pub(crate) fn capabilities(&self) -> &Capabilities {
+        &self.payload.capabilities
+    }
+
+    /// Return a mutable reference to [`Capabilities`].
+    #[cfg(test)]
+    pub fn capabilities_mut(&mut self) -> &mut Capabilities {
```
franziskuskiefer, line 576: This could be moved to a different `impl` as well. If not, let's at least put it at the bottom of the getters or something.

franziskuskiefer (changes requested):
Two more nits

openmls/src/treesync/node/leaf_node.rs
duesee: Is the difference between `#[cfg(any(feature = "test-utils", test))]` and ` #[cfg(test)]` relevant? Can we unify that into the former?

franziskuskiefer: You will probably get dead code warnings because the function is not used outside of unit tests.

duesee: Right... I was thinking about making it public through the `test-utils` feature. But not all methods are currently well-suited for this as they would leak private types. Let's keep this annotations as they are for now.

franziskuskiefer (approved):
Thanks

PR #1201, opened 2023-01-16
spec: Add `RequiredCapabilitiesExtension.credential_types` field.

### LINKED ISSUE #1200: Sync `RequiredCapabilities` with draft-17.
```c
struct {
    ExtensionType extension_types<V>;
    ProposalType proposal_types<V>;
    CredentialType credential_types<V>; // This is missing currently.
} RequiredCapabilities;
```

I think, no matter to what degree we support credentials, we should align this with draft-17. This is also relevant for leaf node validation because we need to make sure that a leaf supports all credential types listed in the group context.

duesee:
This closes #1200.

Note: I used `CredentialType::Basic` to fixup the tests. However, this is not checked for now. I recommend to review commit-by-commit. The first commit just aligns the field names and renames the corresponding methods.

### DIFF
--- a/openmls/src/extensions/required_capabilities.rs
+++ b/openmls/src/extensions/required_capabilities.rs
@@ -18,6 +18,15 @@ use super::{Deserialize, ExtensionError, ExtensionType, Serialize};
 /// extensions can be updated, a GroupContextExtensions proposal is invalid if it
 /// contains a required capabilities extension that requires capabilities not
 /// supported by all current members.
+///
+/// ```c
+/// // draft-ietf-mls-protocol-17
+/// struct {
+///     ExtensionType extension_types<V>;
+///     ProposalType proposal_types<V>;
+///     CredentialType credential_types<V>;
+/// } RequiredCapabilities;
+/// ```
 #[derive(
     PartialEq,
     Eq,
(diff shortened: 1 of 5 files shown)


openmls/src/extensions/required_capabilities.rs
```diff
@@ -31,37 +40,49 @@ use super::{Deserialize, ExtensionError, ExtensionType, Serialize};
...
+    extension_types: Vec<ExtensionType>,
+    proposal_types: Vec<ProposalType>,
+    credential_types: Vec<CredentialType>,
 }
 
 impl RequiredCapabilitiesExtension {
     /// Creates a new [`RequiredCapabilitiesExtension`] from extension and proposal types.
-    pub fn new(extensions: &[ExtensionType], proposals: &[ProposalType]) -> Self {
+    pub fn new(
+        extension_types: &[ExtensionType],
```
franziskuskiefer, line 51: Should this take `Vec`s while we're here?

openmls/src/extensions/required_capabilities.rs
```diff
@@ -31,37 +40,49 @@ use super::{Deserialize, ExtensionError, ExtensionType, Serialize};
...
     /// Check if all extension and proposal types are supported.
     pub(crate) fn check_support(&self) -> Result<(), ExtensionError> {
-        for extension in self.extensions() {
+        for extension in self.extension_types() {
             if !extension.is_supported() {
                 return Err(ExtensionError::UnsupportedExtensionType);
             }
         }
-        for proposal in self.proposals() {
+        for proposal in self.proposal_types() {
```
franziskuskiefer, line 85: We should probably add this for credentials as well. But not in this PR.

openmls/src/group/core_group/test_proposals.rs
```diff
@@ -353,7 +357,9 @@ fn test_group_context_extensions(ciphersuite: Ciphersuite, backend: &impl OpenMl
         ProposalType::Remove,
         ProposalType::Update,
     ];
-    let required_capabilities = RequiredCapabilitiesExtension::new(extensions, proposals);
+    let credentials = &[CredentialType::Basic];
```
franziskuskiefer, line 360: Generally this shouldn't be necessary because "basic" is a "default" credential and should therefore always be supported. But let's file a follow up to handle the "default" credentials.

franziskuskiefer (approved):
There are some follow-ups needed. But I'm fine with getting this in as is.

openmls/src/extensions/required_capabilities.rs
duesee: Yeah. This is very common in OpenMLS. Maybe we should define a tracking issue "Revisit Rust API Guidelines" or so and add `TODO(C-CALLER-CONTROL)` whenever we see it?

openmls/src/group/core_group/test_proposals.rs
duesee: https://github.com/openmls/openmls/issues/1204

openmls/src/extensions/required_capabilities.rs
duesee: I filed https://github.com/openmls/openmls/issues/1205.

PR #1202, opened 2023-01-16
Introduce leaf node validation using builder pattern.

duesee:
This should be reviewed only after #1196 was merged.


openmls/src/treesync/node/leaf_node.rs
```diff
@@ -545,123 +571,200 @@ impl LeafNode {
...
+    /// Return a reference to the leaf node extensions.
+    pub fn extensions(&self) -> &Extensions {
+        &self.payload.extensions
+    }
+
+    // ----- Validation ----------------------------------------------------------------------------
+
+    /// Check that lifetime is acceptable.
+    ///
+    /// * Check that lifetime range is acceptable.
```
kkohbrok, line 609: Could you label this as `TODO`, since the corresponding code is commented out?

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -545,123 +571,200 @@ impl LeafNode {
...
+    /// * Check that current time is between `Lifetime.not_before` and `Lifetime.not_after`.
+    pub(crate) fn validate_lifetime(&self) -> Result<&Self, LeafNodeValidationError> {
+        // self.validate_lifetime_range()?; // TODO(openmls/annotations#36)
+        self.validate_lifetime_actuality()?;
+
+        Ok(self)
+    }
+
+    /// Check that lifetime range is acceptable.
+    #[allow(unused)] // TODO(openmls/annotations#36)
```
kkohbrok, line 619: Why are we not using this in the function above? I'm sure there is a good reason, so it would be great if you could briefly explain it either here or in openmls/annotations#36.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -545,123 +571,200 @@ impl LeafNode {
...
+            // ... proposals, ...
             for required_proposal in required_capabilities.proposals() {
                 if !self.supports_proposal(required_proposal) {
-                    return Err(TreeSyncError::UnsupportedProposal);
+                    return Err(LeafNodeValidationError::UnsupportedProposals);
                 }
             }
+
+            // ... and credential types MUST be listed in the LeafNode's capabilities field.
+            // TODO(#1200)
```
kkohbrok, line 673: Issue #1200 is closed. Am I missing something here?

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -545,123 +571,200 @@ impl LeafNode {
...
 
-    /// Returns the [`Lifetime`] if present.
-    /// `None` otherwise.
-    pub(crate) fn life_time(&self) -> Option<&Lifetime> {
-        if let LeafNodeSource::KeyPackage(life_time) = &self.payload.leaf_node_source {
-            Some(life_time)
-        } else {
-            None
+    /// Validate that the signature key is unique among the members of the group.
+    #[allow(unused)]
```
kkohbrok, line 715: Could you add a comment here as to why this is unused? The same goes for the function below.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -545,123 +571,200 @@ impl LeafNode {
...
+    pub(crate) fn validate_that_signature_key_is_unique(
+        &self,
+        signature_keys: &[SignaturePublicKey],
```
kkohbrok, line 718: It might make sense to take an iterator over `Member` here. But we won't know for sure until we use it, I suppose.

kkohbrok (changes requested):
Thanks for the PR! I really like the builder approach to validation. Generally looks good. Just a few questions and nits.

openmls/src/treesync/node/leaf_node.rs
duesee: I can try it out in a follow-up when we introduce the type safe leaf node validation. Currently, not every validation step is executed because I was unsure where to do it. Thus, it is difficult to see what works and what not.

duesee:
Thanks for the review! I had some changes lying around locally and rebasing resolved some comments, too. Sorry for the force-push but I will double-check to incorporate all suggestions.

Regarding the unused methods: There are three new methods here now, `validate_in_{key_package,update,commit}`, that (will) refer to the other `validate_` methods. Should I proceed here with introducing the "type safe" leaf node validation or annotate the validation functions first with `#[unused]`? What do you think?

openmls/src/treesync/node/leaf_node/lifetime.rs
```diff
@@ -59,10 +63,19 @@ impl Lifetime {
     pub(crate) fn is_valid(&self) -> bool {
         let now = SystemTime::now()
             .duration_since(UNIX_EPOCH)
+            // TODO(#1197): Use `Instant`, return `false` (+ log), or return `LibraryError`?
```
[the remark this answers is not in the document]
duesee: I did both, `return false` + `log::error`. Hope that's fine. ([a9c9086](https://github.com/openmls/openmls/pull/1202/commits/a9c9086bc841232237933ac99f2f3c83c89e2fa9))

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -522,11 +473,80 @@ impl LeafNode {
...
+    }
+
     /// Returns the `encryption_key`.
     pub fn encryption_key(&self) -> &HpkePublicKey {
         &self.payload.encryption_key
     }
 
+    /// Get the identifier to search for the encryption key pair in the key
+    /// store.
+    pub fn encryption_key_label(id: &[u8]) -> Vec<u8> {
```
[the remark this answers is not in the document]
duesee: This was resolved by the rebase.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -522,11 +473,80 @@ impl LeafNode {
...
+            config,
+            credential_bundle,
+            LeafNodeSource::Update,
+            capabilities,
+            extensions,
+            backend,
+        )?;
+
+        // Store the encryption key pair in the key store.
+        backend
```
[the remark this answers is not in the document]
duesee: This was resolved by the rebase.

openmls/src/treesync/node/leaf_node.rs
duesee: This was resolved by the rebase.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -562,45 +562,224 @@ impl LeafNode {
...
+    // ---------------------------------------------------------------------------------------------
+
     /// Returns `true` if the [`ExtensionType`] is supported by this leaf node.
     pub(crate) fn supports_extension(&self, extension_type: &ExtensionType) -> bool {
         self.payload
             .capabilities
             .extensions
-            .iter()
-            .any(|et| et == extension_type)
+            .contains(extension_type)
```
duesee, line 767: This is only a refactoring.

openmls/src/treesync/node/leaf_node/lifetime.rs
```diff
@@ -57,11 +61,24 @@ impl Lifetime {
...
+        {
+            Ok(elapsed) => self.not_before < elapsed && elapsed < self.not_after,
+            Err(_) => {
+                log::error!("SystemTime before UNIX EPOCH.");
+                false
+            }
+        }
+    }
+
+    /// ValSem(openmls/annotations#32):
```
duesee, line 76: Not sure to what issue to link this. After I gaines a bit of experience with the annotation tool, I think it might be best to stick to the current `ValSem` approach -- maybe in a way that can be reflected using GitHub issues? Sometimes, multiple annotations refer to the same validation procedure but are phrased differently in the standard.

kkohbrok: We should probably have a general discussion on how to enumerate/name validation steps. The annotation tool wasn't around when the ValSem notation was introduced. I'd say leave it as it is for now.

kkohbrok (approved):
Thanks for addressing all my concerns. Let's merge this and then deal with the type-based stuff in a follow-up.

openmls/src/treesync/node/leaf_node.rs
duesee: Resolved by rebase and adding of `TODO`.

Issue #1194, opened 2023-01-12
Take a GroupStore as input to MlsGroup API and take control of group state serialization and storage


kkohbrok:
If we store key material from within the API, there is a potential divergence between group state and key material in case the process crashes mid-commit merge. To make this more robust, we could take a group store (analogous to the key store) as input to make the transition safer.

keks:
Is this resolved by the new Storage trait? It sounds a little like it could be, but I'm not sure.

kkohbrok:
Yes, this is indeed resolved with the new storage trait.

PR #1203, opened 2023-01-16
Towards a public ratchet tree (without private values)

kkohbrok:
This works towards isolating a public ratchet tree that allows applications without private key material (e.g. a DS) to perform various message processing and validation operations. In particular, this PR does the following:

* Remove private encryption key material from the various Node structs
* Store the encryption keys instead in the key store
* Load the right key material from the key store just below MlsGroup or CoreGroupAPI level
* Various API endpoints now potentially return key store errors. To keep things generic, these endpoints (and their error enums) now depend on the keystore implementation in the backend
* Key material is updated and stale keys are deleted when merging commits. I filed an issue to track the synchronization of group state and key store state changes (#1194)


openmls/src/treesync/node/parent_node.rs
```diff
@@ -12,47 +12,39 @@ use tls_codec::{TlsSerialize, TlsSize, VLBytes};
...
     treesync::{hashes::ParentHashInput, treekem::UpdatePathNode},
 };
 
+use super::encryption_keys::{EncryptionKey, EncryptionKeyPair};
+
 /// This struct implements the MLS parent node. It contains its public key,
 /// parent hash and unmerged leaves. Additionally, it may contain the private
 /// key corresponding to the public key.
 #[derive(Debug, Eq, PartialEq, Clone, Serialize, Deserialize)]
 pub struct ParentNode {
```
franziskuskiefer, line 28: Can we derive the TLS codec now?

openmls/src/treesync/diff.rs
```diff
@@ -798,66 +804,40 @@ impl<'a> TreeSyncDiff<'a> {
...
-                    .node()
-                    .as_ref()
-                    .map(|node| (node.private_key(), node.public_key())),
-            } {
-                // If we do have the private key, check if the key is in the
-                // resolution.
-                if let Some(resolution_position) = sender_copath_resolution
+        if let Some((resolution_position, private_key)) = sender_copath_resolution
+            .iter()
+            .enumerate()
```
franziskuskiefer, line 830: This looks much nicer now! ☺️

openmls/src/treesync/mod.rs
```diff
@@ -505,4 +462,13 @@ impl TreeSync {
     pub(crate) fn is_leaf_in_tree(&self, leaf_index: LeafNodeIndex) -> bool {
         is_node_in_tree(leaf_index.into(), self.tree.size())
     }
+
+    /// Return a vector containing all [`HpkePublicKey`]s for which we should
+    /// have the corresponding private keys.
+    pub(crate) fn owned_encryption_keys(&self) -> Vec<EncryptionKey> {
+        self.empty_diff()
```
franziskuskiefer, line 469: If this weren't on an empty diff (temp value), we could avoid owning the keys and collecting. Not sure if that's easily doable though ...

openmls/src/group/mls_group/creation.rs
```diff
@@ -21,11 +20,11 @@ impl MlsGroup {
...
     /// Returns an error ([`NewGroupError::NoMatchingCredentialBundle`]) if the
     /// private key for the [`SignaturePublicKey`] can not be found.
-    pub fn new(
-        backend: &impl OpenMlsCryptoProvider,
+    pub fn new<KeyStore: OpenMlsKeyStore>(
+        backend: &impl OpenMlsCryptoProvider<KeyStoreProvider = KeyStore>,
         mls_group_config: &MlsGroupConfig,
         signature_key: &SignaturePublicKey,
-    ) -> Result<Self, NewGroupError> {
+    ) -> Result<Self, NewGroupError<KeyStore::Error>> {
```
franziskuskiefer, line 27: Another way of writing these would be the following, fully moving to generics.

```rust
pub fn new<T: OpenMlsCryptoProvider>(
        backend: &T,
        mls_group_config: &MlsGroupConfig,
        signature_key: &SignaturePublicKey,
    ) -> Result<Self, NewGroupError<<<T as openmls_traits::OpenMlsCryptoProvider>::KeyStoreProvider as OpenMlsKeyStore>::Error>>{
```

franziskuskiefer (changes requested):
The general approach looks good to me. I only looked at half of the changes so far. But some prelim comments and thoughts to get started.

openmls/src/treesync/mod.rs
kkohbrok: Yeah, the proper solution for this is to have a `Treelike` trait that is implemented by both tree and diff. Then we could more easily share code between the two. I'll file an issue for this so we don't forget. For now creating an empty diff is unfortunately the easiest way.

openmls/src/group/mls_group/creation.rs
kkohbrok: True, that would work as well. Would you prefer that style? I have no particular preference here.

traits/src/key_store.rs
```diff
@@ -13,9 +13,36 @@ pub trait ToKeyStoreValue {
 }
 
 /// The Key Store trait
-pub trait OpenMlsKeyStore: Send + Sync {
+pub trait OpenMlsKeyStore: Send + Sync + Debug {
     /// The error type returned by the [`OpenMlsKeyStore`].
-    type Error: std::error::Error + Debug;
+    type Error: std::error::Error + Debug + PartialEq + Clone;
```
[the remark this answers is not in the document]
kkohbrok: Since a lot of the errors in the MlsGroup API depend on `KeyStore::Error` now, we need these bounds (or at least `PartialEq`) to check in tests that we get the right errors. Not sure if there is a way around that, but I'm open to suggestions. (I did remove `Clone`, though.)

openmls/src/treesync/node/encryption_keys.rs
```diff
@@ -0,0 +1,189 @@
...
+use crate::ciphersuite::{HpkePrivateKey, HpkePublicKey, Secret};
+use crate::error::LibraryError;
+use crate::versions::ProtocolVersion;
+
+/// [`EncryptionKey`] contains an HPKE public key that allows the encryption of
+/// path secrets in MLS commits.
+#[derive(
+    Debug, Clone, Serialize, Deserialize, TlsSerialize, TlsDeserialize, TlsSize, PartialEq, Eq,
+)]
+pub struct EncryptionKey {
```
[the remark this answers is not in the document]
kkohbrok: There are functions like the `new_from_encryption_key` constructor for KeyPackages and the `self_update` function that technically take an `EncryptionKey` as input, but we could also use `HpkePublicKey` on those and then convert to an `EncryptionKey` internally if you prefer that.

openmls/src/treesync/node/encryption_keys.rs
```diff
@@ -0,0 +1,189 @@
...
+            private_key,
+        }
+    }
+}
+
+impl ToKeyStoreValue for EncryptionKeyPair {
+    type Error = LibraryError;
+
+    fn to_key_store_value(&self) -> Result<Vec<u8>, Self::Error> {
+        serde_json::to_vec(self)
```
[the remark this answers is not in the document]
kkohbrok: I tried refactoring to avoid deriving the serde traits, but we have `EncryptionKeyPair` in the `MemberStagedCommitState`, which we need to serialize :-(

kkohbrok:
Thanks for the review! Lots of good points. I have addressed as many as I could and have commented on the rest.

traits/src/key_store.rs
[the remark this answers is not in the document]
franziskuskiefer: The problem here is that this makes troube if the error doesn't implement these traits.
Generally the trait bounds should be kept as minimal as possible. The requirement is when calling the functions later and using the error. So ideally we'd be adding the requirements there to the type. But I think this would only work if we'd had a trait bound for  it here.
I'd be fine with adding `PartialEq` here again for now. But I expect that we should remove it again at some point.

openmls/src/group/core_group/create_commit.rs
```diff
@@ -160,29 +162,43 @@ impl CoreGroup {
...
                     )?;
-                }
+                    vec![encryption_keypair]
+                } else {
+                    // If we're not in the tree, we just load
+                    // KeyPackage we prepared earlier.
+                    let own_leaf = diff.own_leaf()
+                        // We should have a leaf in the diff
+                        .map_err(|_| LibraryError::custom("Unable to get own leaf from diff"))?;
+                    let encryption_keypair = EncryptionKeyPair::read_from_key_store(backend, own_leaf.encryption_key())
```
franziskuskiefer, line 184: Hm, this read  here isn't great.
It's not a public function and it's pretty deep in the code such that it will be difficult to split the function on the read.
If it's something we just created, we should be able to pass it around instead of storing it?
The key package creation above where this is coming from, has the same issue. Could we have the key package builder above in a way that it is returning the keys instead of storing them (at least as an option)? Then we wouldn't write there and wouldn't read here.

openmls/src/group/core_group/mod.rs
```diff
@@ -738,6 +747,56 @@ impl CoreGroup {
...
+    ///
+    /// Returns an error if access to the key store fails.
+    pub(super) fn store_epoch_keypairs<KeyStore: OpenMlsKeyStore>(
+        &self,
+        backend: &impl OpenMlsCryptoProvider<KeyStoreProvider = KeyStore>,
+        keypair_references: &[EncryptionKeyPair],
+    ) -> Result<(), KeyStore::Error> {
+        backend.key_store().store_epoch_keys(
+            // Retrieving our identity should not fail.
+            self.own_identity().unwrap_or(&[]),
```
franziskuskiefer, line 761: This looks like a footgun to me where we use the same default identity for different things if this fails.
Maybe `own_identity` shouldn't be an option. But we could at least add a `debug_assert` on this.

Also, you can't you use `unwrap_or_default`?

openmls/src/group/core_group/new_from_welcome.rs
```diff
@@ -136,21 +122,56 @@ impl CoreGroup {
...
         )
         .map_err(|e| match e {
             TreeSyncFromNodesError::LibraryError(e) => e.into(),
             TreeSyncFromNodesError::PublicTreeError(e) => WelcomeError::PublicTreeError(e),
         })?;
 
+        let diff = tree.empty_diff();
+
+        // Read the encryption key pair from the key store and delete it there.
+        let leaf_keypair = EncryptionKeyPair::read_from_key_store(
```
franziskuskiefer, line 142: Let's do keystore operation at the beginning of the function if possible. This will make it easier splitting them out later.

openmls/src/group/core_group/new_from_welcome.rs
```diff
@@ -136,21 +122,56 @@ impl CoreGroup {
...
+                .map_err(|e| match e {
+                    TreeSyncSetPathError::LibraryError(e) => e.into(),
+                    TreeSyncSetPathError::PublicKeyMismatch => {
+                        WelcomeError::PublicTreeError(PublicTreeError::PublicKeyMismatch)
+                    }
+                })?;
+            vec![leaf_keypair]
+                .into_iter()
+                .chain(path_keypairs.into_iter())
+                .collect()
```
franziskuskiefer, line 170: It's a little unfortunate that we have to collect these. But let's keep this for now and only improve if we see that it's an issue.

openmls/src/group/core_group/process.rs
```diff
@@ -321,22 +321,24 @@ impl CoreGroup {
...
-    ) {
+    ) -> Result<(), KeyStore::Error> {
         // Save the past epoch
         let past_epoch = self.context().epoch();
         // Get all the full leaves
         let leaves = self.treesync().full_leave_members().collect();
         // Merge the staged commit into the group state and store the secret tree from the
         // previous epoch in the message secrets store.
-        if let Some(message_secrets) = self.merge_commit(staged_commit) {
+        if let Some(message_secrets) = self.merge_commit(backend, staged_commit)? {
```
franziskuskiefer, line 336: So in here we're using the key store now again. Can we pull this further up?

openmls/src/group/mls_group/creation.rs
kkohbrok: I'd personally prefer the one that's in the code right now. If you don't have a strong opinion, I'd prefer to leave it as it is.

openmls/src/treesync/node/encryption_keys.rs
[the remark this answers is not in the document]
kkohbrok: Yes, I filed #1194 to make track how persistence is coordinated.

openmls/src/treesync/node/encryption_keys.rs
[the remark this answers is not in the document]
kkohbrok: It's crate-public now and I've filed #1208 to re-enable passing in custom things into `self_update`.

traits/src/key_store.rs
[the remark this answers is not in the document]
kkohbrok: I understand. I've removed `Clone`, but left in `PartialEq` for now.

kkohbrok:
We're getting there. As discussed OOB, I've filed #1207 to track the shifting of key store access further up the call chain. I think that makes sense for now, as CoreGroup, as well as the associated tests will change quite a bit in the near future, so it doesn't make that much sense to do this now.

Besides addressing your concerns, I fixed a bug that I had missed before: If a self-update (commited by someone else) is processed, the keypair from the leaf needs to be deleted explicitly upon merging. That should happen properly now.

openmls/src/messages/tests/test_welcome.rs
```diff
@@ -182,6 +182,9 @@ fn test_welcome_ciphersuite_mismatch(
 
     welcome.encrypted_group_info = encrypted_verifiable_group_info.into();
 
+    // Create backup of encryption keypair, s.t. we can process the welcome a second time after failing.
+    //let encryption_keypair = EncryptionKeyPair::read_from_key_store(backend, encryption_key);
```
franziskuskiefer, line 186: Is this needed?

franziskuskiefer (changes requested):
Thanks, only a few small things. Then this can go in.

traits/src/key_store.rs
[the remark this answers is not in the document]
OtaK: The biggest issue with `PartialEq` on errors is that any error that contains (deep in the error chain or not) a `Backtrace` cannot be used, as `std::backtrace::Backtrace` doesn't implement `PartialEq` (https://doc.rust-lang.org/std/backtrace/struct.Backtrace.html#trait-implementations) and never will. As stacktraces are always unique, you cannot compare them.

If you want to assert that you get a certain error in tests you could always use `assert!(matches!(err, MyError::MyVariant(_)))` or something along those lines.

openmls/src/group/core_group/staged_commit.rs
```diff
@@ -196,100 +199,152 @@ impl CoreGroup {
...
+                    &decryption_keypairs,
+                )?;
+
+                // Check if one of our update proposals was applied. If so, we
+                // need to store that keypair separately, because after merging
+                // it needs to be removed from the key store separately and in
+                // addition to the removal of the keypairs of the previous
+                // epoch.
+                let new_leaf_keypair_option = if let Ok(leaf) = diff.own_leaf() {
+                    own_keypairs
```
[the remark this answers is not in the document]
kkohbrok: Yeah, I had a feeling while I was writing it, but I didn't know about `find_map`. Thanks for the pointer!

traits/src/key_store.rs
```diff
@@ -15,7 +15,38 @@ pub trait ToKeyStoreValue {
 /// The Key Store trait
 pub trait OpenMlsKeyStore: Send + Sync {
     /// The error type returned by the [`OpenMlsKeyStore`].
-    type Error: std::error::Error + Debug;
+    type Error: std::error::Error + Debug + PartialEq;
+
+    /// Load all encryption keys associated with the given client identity (i.e.
+    /// a byte vector uniquely identifying the client), epoch and group ID from
+    /// the key store.
+    fn read_epoch_keys<V: FromKeyStoreValue>(
```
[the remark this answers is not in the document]
kkohbrok: Yeah, I never know when it's best to go `Option<Vec>` vs. `Vec` that is potentially empty. But you're probably correct in that the code yields the same result either way, so let's simplify this a little.

openmls/src/group/core_group/staged_commit.rs
```diff
@@ -427,7 +491,37 @@ impl CoreGroup {
...
+                let new_owned_encryption_keys = self.tree.owned_encryption_keys();
+                // From the old and new keys, keep the ones that are still relevant in the new epoch.
+                let epoch_keypairs: Vec<EncryptionKeyPair> = old_epoch_keypairs
+                    .into_iter()
+                    .chain(state.new_keypairs.into_iter())
+                    .chain(leaf_keypair.into_iter())
+                    .filter(|keypair| new_owned_encryption_keys.contains(keypair.public_key()))
+                    .collect();
+                // We should have private keys for all owned encryption keys.
+                debug_assert_eq!(new_owned_encryption_keys.len(), epoch_keypairs.len());
```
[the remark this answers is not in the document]
kkohbrok: We could throw an error here, but I would say it's a bit of a correctness thing. If the code is correct and the key store works as intended, then we should have all the keys we need. If we were to throw an error here, it should probably be a key store error? Alternatively maybe a LibraryError? But in the recent past, we have tried to avoid those where possible, so I'm not sure. What do you think?

openmls/src/messages/tests/test_welcome.rs
kkohbrok: Apparently not 🤷🏻‍♂️.

kkohbrok:
Thanks again for the next round of comments! I think I got everything (or replied where I wasn't sure what to do).

traits/src/key_store.rs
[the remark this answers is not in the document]
kkohbrok: Fair point. Using `matches` is a good idea for tests. I won't comb through all the tests in this PR, though, as it's already big enough. Do you want to create an issue for that? Should be easy enough to do.

franziskuskiefer (changes requested):
Thanks, I resolved all the comments that are addressed. Only small ones 2 left.

openmls/src/group/core_group/staged_commit.rs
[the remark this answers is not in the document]
kkohbrok: Fair enough. I have introduced a LibraryError there for the time being. We can still rip it out later when we move the key store access further up and when we're more confident in the correctness of our key store interactions.

franziskuskiefer (approved):
Thanks, lets get this in.

PR #1216, opened 2023-01-23
Pull own_index out of TreeSync

kkohbrok:
This PR pulls the `own_index` out of TreeSync and TreeSyncDiff and instead stores it in CoreGroup. Functions in TreeSync that rely on an `own_index` now get it as an additional input.

One thing I noticed during testing is that indexing encryption keypairs in the KeyStore with the `identity` (which is what was introduced in #1203) is not great, because MLS doesn't really have a notion of `identity` anyway and instead relies completely on the application to keep track of clients' identities. This PR changes the indexing in the key store such that instead of the `identity`, it uses the client's leaf index in the group. The index is thus a combination of group ID, group epoch and leaf index, which should uniquely identify a given group member and allow storage of key material for multiple members of the same group (which is mostly relevant for testing).

This PR is again in preparation of having a notion of a public group that can be used by a non-group member (such as a DS or even an external committer).

### DIFF
--- a/memory_keystore/src/lib.rs
+++ b/memory_keystore/src/lib.rs
@@ -1,7 +1,7 @@
 use openmls_traits::key_store::{FromKeyStoreValue, OpenMlsKeyStore, ToKeyStoreValue};
 use std::{collections::HashMap, sync::RwLock};
 
-type EpochStoreIndex = (Vec<u8>, u64, u32); // Identity, GroupId and GroupEpoch
+type EpochStoreIndex = (Vec<u8>, u64, u32); // GroupId, GroupEpoch and leaf index
 
 #[derive(Debug, Default)]
 pub struct MemoryKeyStore {
--- a/memory_keystore/src/lib.rs
+++ b/memory_keystore/src/lib.rs
@@ -1,14 +1,17 @@
 use openmls_traits::key_store::{FromKeyStoreValue, OpenMlsKeyStore, ToKeyStoreValue};
 use std::{collections::HashMap, sync::RwLock};
 
+type EpochStoreIndex = (Vec<u8>, Vec<u8>, u64); // Identity, GroupId and GroupEpoch
+
 #[derive(Debug, Default)]
 pub struct MemoryKeyStore {
     values: RwLock<HashMap<Vec<u8>, Vec<u8>>>,
+    epoch_values: RwLock<HashMap<EpochStoreIndex, Vec<Vec<u8>>>>,
 }
 
 impl OpenMlsKeyStore for MemoryKeyStore {
     /// The error type returned by the [`OpenMlsKeyStore`].
-    type Error = Error;
+    type Error = MemoryKeyStoreError;
 
     /// Store a value `v` that implements the [`KeyStoreValue`] trait for
     /// serialization for ID `k`.
@@ -17,7 +20,7 @@ impl OpenMlsKeyStore for MemoryKeyStore {
     fn store<V: ToKeyStoreValue>(&self, k: &[u8], v: &V) -> Result<(), Self::Error> {
         let value = v
             .to_key_store_value()
-            .map_err(|_| Error::SerializationError)?;
+            .map_err(|_| MemoryKeyStoreError::SerializationError)?;
         // We unwrap here, because this is the only function claiming a write
         // lock on `credential_bundles`. It only holds the lock very briefly and
         // should not panic during that period.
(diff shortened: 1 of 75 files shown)


openmls/src/group/core_group/create_commit.rs
```diff
@@ -38,28 +38,9 @@ impl CoreGroup {
...
-                // `remove` proposal with the index of our previous self in the
-                // group.
-                let leaf_index =
-                    self.free_leaf_index(params.inline_proposals().iter().map(Some))?;
-                (Sender::NewMemberCommit, leaf_index)
-            }
-            CommitType::Member => (
-                Sender::build_member(self.own_leaf_index()),
-                self.own_leaf_index(),
-            ),
```
kkohbrok, line 62: The leaf index calculation here didn't take the "resync" case into account. Index computation now happens one layer above, where the index is set properly in the `CoreGroup`.

openmls/src/treesync/mod.rs
```diff
@@ -193,68 +185,19 @@ impl TreeSync {
...
-            Err(PublicTreeError::MissingKeyPackage.into())
-        }
-    }
-
-    /// Create a [`TreeSync`] instance from a vector of nodes without expecting
-    /// there to be a [`KeyPackage`] that belongs to this particular MLS client.
-    /// WARNING: Some of the [`TreeSync`] invariants will not hold for this
-    /// tree, as the `own_leaf_index` does not point to a leaf with private key
-    /// material in it.
-    pub(crate) fn from_nodes_without_leaf(
```
kkohbrok, line 225: We don't need a separate constructor for `without_leaf` anymore, since `TreeSync` doesn't have an `own_leaf` anymore.

openmls/src/group/core_group/apply_proposals.rs
```diff
@@ -52,6 +52,7 @@ impl CoreGroup {
         backend: &impl OpenMlsCryptoProvider,
         proposal_queue: &ProposalQueue,
         leaf_nodes: &[OpenMlsLeafNode],
+        own_leaf_index: Option<LeafNodeIndex>,
```
franziskuskiefer, line 55: Can we have an `impl Into<...>` here? Then we don't need a `Some` when calling.
Actually, I don't see a call without an index. Doe we need this?

openmls/src/group/core_group/apply_proposals.rs
```diff
@@ -95,18 +96,19 @@ impl CoreGroup {
...
+                        {
+                            Some(leaf_node) => leaf_node,
+                            // We lost the LeafNode apparently
+                            None => return Err(ApplyProposalsError::MissingLeafNode),
+                        };
+                        own_leaf_node.clone()
+                    }
+                    _ => update_proposal.leaf_node().clone().into(),
                 };
                 diff.update_leaf(leaf_node, sender_index);
```
franziskuskiefer, line 113: `update_leaf` should probably get a `LeafNode` instead, then we safe us some unnecessary conversions.

openmls/src/treesync/mod.rs
```diff
@@ -159,14 +156,12 @@ impl TreeSync {
...
-        encryption_key: &EncryptionKey,
     ) -> Result<Self, TreeSyncFromNodesError> {
         // TODO #800: Unmerged leaves should be checked
         // Before we can instantiate the TreeSync instance, we have to figure
         // out what our leaf index is.
         let mut ts_nodes: Vec<TreeNode<TreeSyncLeafNode, TreeSyncParentNode>> =
             Vec::with_capacity(node_options.len());
-        let mut own_index_option = None;
 
         // Check that our own encryption key is in the tree.
```
franziskuskiefer, line 166: We don't do this anymore.

openmls/src/treesync/mod.rs
```diff
@@ -465,9 +410,12 @@ impl TreeSync {
 
     /// Return a vector containing all [`HpkePublicKey`]s for which we should
     /// have the corresponding private keys.
-    pub(crate) fn owned_encryption_keys(&self) -> Vec<EncryptionKey> {
+    pub(crate) fn owned_encryption_keys(
```
franziskuskiefer, line 413: Should we move all "own" functions out of the tree?

openmls/src/treesync/diff.rs
```diff
@@ -270,13 +260,14 @@ impl<'a> TreeSyncDiff<'a> {
         &self,
         backend: &impl OpenMlsCryptoProvider,
         ciphersuite: Ciphersuite,
+        own_leaf_index: LeafNodeIndex,
```
franziskuskiefer, line 263: Isn't this just deriving a path now, we could do this for any path.

openmls/src/treesync/diff.rs
```diff
@@ -295,16 +286,19 @@ impl<'a> TreeSyncDiff<'a> {
         ciphersuite: Ciphersuite,
         group_id: GroupId,
         credential_bundle: &CredentialBundle,
+        own_leaf_index: LeafNodeIndex,
```
franziskuskiefer, line 289: Similar to the module file, shall we tro to get rid of all the "own"s in here and have it general?
There are of course requirements on the input and available keys. But that's really independent of the "own" now.

memory_keystore/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 use openmls_traits::key_store::{FromKeyStoreValue, OpenMlsKeyStore, ToKeyStoreValue};
 use std::{collections::HashMap, sync::RwLock};
 
-type EpochStoreIndex = (Vec<u8>, Vec<u8>, u64); // Identity, GroupId and GroupEpoch
+type EpochStoreIndex = (Vec<u8>, u64, u32); // Identity, GroupId and GroupEpoch
```
franziskuskiefer, line 4: Update the comment here.

franziskuskiefer (changes requested):
Thanks, I really like where this is going. Only a few small issues.

openmls/src/group/core_group/apply_proposals.rs
kkohbrok: I'm not sure there is something we can save here until we have ripped out the `OpenMlsLeafNode` abstraction. That's probably possible (it only contains an optional `LeafNodeIndex` that we can probably get in some other way), but not quite yet.

openmls/src/treesync/mod.rs
kkohbrok: We could do that, but that would mean exposing more functions of `TreeSyncDiff` and `TreeSync` access such as `parent` and `leaf`, but also `filtered_direct_path`. You would also probably have to implement this twice, once for the tree and once for the diff. If it's okay with you, I'd like to punt this to the point where we have a `Treelike` trait that allows us to implement this function once for both tree and diff.

openmls/src/treesync/diff.rs
kkohbrok: Yes, I'll rename those input variables and change the docs to reflect the requirements.

kkohbrok:
Thanks for the review! I think I replied to or addressed most comments. Let me know if I missed something.

openmls/src/group/core_group/apply_proposals.rs
franziskuskiefer: But `update_proposal` isn't using the index from the `leaf_node`. So the `into` in `update_proposal` is unnecessary, which would allow you to drop the `into` in line 113.
But not too important for now.

franziskuskiefer (approved):
Thanks, one comment on the `update_proposal`. But I'm fine with leaving this as is.

openmls/src/group/core_group/apply_proposals.rs
kkohbrok: I'm confused. `update_leaf` puts the new leaf into the tree, which means that at some point whatever input it takes needs to become a `TreeSyncLeafNode` via an `OpenMlsLeafNode`. `update_proposal` contains a `LeafNode`. So those `into` calls have to happen at some point. If I remove them here, I have to add them inside of `update_leaf`. Is that what you mean?

PR #1218, opened 2023-01-24
Introduce PublicGroup

### LINKED ISSUE #931: [MLS Spec change] Use new tree for context in path secret encryption (#692)
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/692

> Closes #648 via option (3)

kkohbrok:
This PR introduces the notion of a `PublicGroup` and is follow-up based on #1216. The PublicGroup contains only public values associated with an MLS group and lives in a CoreGroup. Long-term, the goals is to provide an API (maybe under a feature flag) that allows public observers of MLS public messages (such as a delivery service) to use the public group to keep track of the group and its membership.

Concretely, this PR does the following:
* Introduce PublicGroup within CoreGroup (including PublicGroupDiffs).
* Move most parts of CoreGroup that deal with public group values into PublicGroup. For size reasons, this PR doesn't move all such values and functions into PublicGroup. In particular, validation is still part of CoreGroup. More in the next follow-up.
* Hide the tree from CoreGroup to a large extent. Again for size reasons, it doesn't hide it full yet (more in a separate follow-up). There are still myriad places where the treesync is exposed (e.g. via the `treesync` getter of CoreGroup), so this would add quite a few more changes to this PR.
* Fixes a small bug in the tree logic (lookup of nodes in a diff)
* Does some restructuring in create_commit and stage_commit. These are simplifications that are possible due to the recent PRs and overall improvements to OpenMLS :-)

Fixes #931.


openmls/tests/book_code.rs
```diff
@@ -324,12 +324,6 @@ fn book_operations(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvide
     if let ProcessedMessageContent::StagedCommitMessage(staged_commit) =
         alice_processed_message.into_content()
     {
-        let update = staged_commit
-            .commit_update_key_package()
-            .expect("Expected a KeyPackage.");
-        // Check that Bob updated
-        assert_eq!(update.credential(), &bob_credential.credential);
-
```
kkohbrok, line 332: I removed these checks, because there is no more `commit_update_key_package` in the staged commit. We could pull it out of the `StagedCommit`, but we'd have to pierce a number of layers to pull something out (`StagedCommit` contains `MemberStagedCommmitState` contains `StagedPublicGroupDiff` and none of those is meant to be accessed outside of tests). I can create all of those getters, but I thought it's probably not worth it for these extra checks.

openmls/src/binary_tree/array_representation/diff.rs
```diff
@@ -323,10 +312,16 @@ impl<'a, L: Clone + Debug + Default, P: Clone + Debug + Default> AbDiff<'a, L, P
     pub(crate) fn leaf(&self, leaf_index: LeafNodeIndex) -> &L {
         // Check if it's in the diff.
         if let Some(node) = self.leaf_diff.get(&leaf_index) {
-            return node;
+            node
+        // If it's not in the diff, it could be that it's outside of the
```
franziskuskiefer, line 316: Do we have a test for this?

openmls/src/group/public_group/diff/apply_proposals.rs
```diff
@@ -0,0 +1,164 @@
+use std::collections::HashSet;
```
franziskuskiefer, line 1: If only we could see moved code ... I'll assume that most of the code in public_group was moved and only look at the API.

openmls/src/group/public_group/mod.rs
```diff
@@ -0,0 +1,265 @@
...
+
+    pub fn extensions(&self) -> &Extensions {
+        self.group_context.extensions()
+    }
+
+    pub fn required_capabilities(&self) -> Option<&RequiredCapabilitiesExtension> {
+        self.group_context.required_capabilities()
+    }
+
+    pub(crate) fn treesync(&self) -> &TreeSync {
```
franziskuskiefer, line 250: I assume that this is supposed to go away in follow ups?

openmls/src/group/public_group/mod.rs
```diff
@@ -0,0 +1,265 @@
+use openmls_traits::{crypto::OpenMlsCrypto, types::Ciphersuite, OpenMlsCryptoProvider};
```
franziskuskiefer, line 1: Can you add some doc comments to the module here explaining what it's about and how it works?

openmls/src/group/public_group/diff.rs
```diff
@@ -0,0 +1,240 @@
...
+}
+
+impl<'a> PublicGroupDiff<'a> {
+    /// Create a new [`PublicGroupDiff`] based on the given [`PublicGroup`].
+    pub(super) fn new(public_group: &'a PublicGroup) -> PublicGroupDiff<'a> {
+        Self {
+            original_group: public_group,
+            diff: public_group.treesync().empty_diff(),
+            group_context: public_group.group_context().clone(),
+            interim_transcript_hash: public_group.interim_transcript_hash.to_vec(),
```
franziskuskiefer, line 46: Can we use a getter here as well? It makes it easier to understand where it's used.

openmls/src/group/public_group/diff.rs
```diff
@@ -0,0 +1,240 @@
...
+            confirmation_tag: public_group.confirmation_tag().clone(),
+        }
+    }
+
+    /// Turn this [`PublicGroupDiff`] into a [`StagedPublicGroupDiff`], thus
+    /// freezing it until it is merged with the original [`PublicGroup`].
+    pub(crate) fn into_staged_diff(
```
franziskuskiefer, line 53: With treesync and this public group we replicate some of the diff mechanisms. I wonder if this is too much complexity and whether there shold be only one diff.

openmls/src/group/core_group/new_from_external_init.rs
```diff
@@ -136,7 +88,7 @@ impl CoreGroup {
             index,
             signature_key,
             ..
-        } in treesync.full_leave_members()
+        } in public_group.treesync().full_leave_members()
```
franziskuskiefer, line 91: Shall we properly define the `Member` as part of the public as well and make use of it in publi APIs (like remove)? Not in this PR but maybe in a follow up?

franziskuskiefer (changes requested):
Thanks, generally looks good to me. A few thoughts inline.
Some of the code would benefit from some more unit tests. But let's not block on this and add some tests in a systematic way later.

openmls/src/group/public_group/diff/apply_proposals.rs
kkohbrok: Yes. The only difference is that instead of deriving the external pub keypair internally, it returns the external proposal s.t. the caller can do it.

openmls/src/group/public_group/mod.rs
kkohbrok: Exactly. I thought about doing it in this one, but treesync is used externally in quite a few places, so I decided on punting it.

openmls/src/group/public_group/diff.rs
kkohbrok: Good point. This pattern goes all the way up from `ABinaryTree` at this point.

franziskuskiefer (approved):
Thanks 🚢  it

PR #1219, opened 2023-01-24
test: Derive the commit secret from the end of the UpdatePath, not the root

### LINKED ISSUE #942: [MLS Spec change] Derive the commit secret from the end of the UpdatePath, not the root (#707)
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/707

> Fixes #690

augustocdias:
Closes #942

This adds a test to ensure that the commit secret is derived from the end of the update path and not from the root. The implementation already adheres to the spec.


kkohbrok (changes requested):
Thanks for picking this up! Good to see that we're already doing the right thing. Generally looks good, except that clippy is unhappy and that the tests are not building just yet.

openmls/src/treesync/tests_and_kats/tests.rs
```diff
@@ -1,2 +1,150 @@
...
+    prelude::*,
+};
+
 mod test_diff;
 mod test_unmerged_leaves;
+
+/// Pathological example taken from ...
+///   https://github.com/mlswg/mls-protocol/issues/690#issue-1244086547.
+#[allow(non_snake_case)]
+#[test]
```
kkohbrok, line 16: Instead of `#[test]`, could you use the `apply` macro that we use in essentially all other tests s.t. all backends and ciphersuites are tested? We might change the way we do tests in the near future (cf #1133), but let's stay consistent for the time being.

kkohbrok (changes requested):
Thanks for the fixes. Just one more nit, then this can go in.

openmls/src/treesync/tests_and_kats/tests.rs
augustocdias: I've just did, but I think I found a bug in another part... It seems the `own_leaf_index` is being evaluated incorrectly at some point. 

I'm getting randomly the error that I'm trying to remove self. I've added a function to get the correct index for removal but if you see on the failures what the group has in the `members()` iterator is different from the `own_leaf_index`

augustocdias: Check here for ex:

https://github.com/openmls/openmls/actions/runs/3998097907/jobs/6860287051#step:3:1390

Charlie's index in the `members()` iterator is 2 but its own leaf index is 1, so it is evaluating to the wrong index, causing the error.

augustocdias: And this is happening randomly if you see that run, it failed in just one ciphersuite but I had runs locally that failed all or even none.

kkohbrok: Yes, I think that is an issue with the way we do backends. I think @franziskuskiefer ran into the same issue today. For now, the easiest solution is probably to go back to your original one-backend-per client solution that you started out with. Sorry for sending you down that path, but I also only just learned about this problem. I already filed issue #1221. It would probably make sense to tackle this more generally with some sort of test framework. We already have one for integration tests, but I don't think we have one for unit tests at the MlsGroup or CoreGroup level yet.

augustocdias: Some jobs in the CI failed but I can't see any error when I click it. 😐

kkohbrok: Yeah, I think GH had some issues there. I re-ran the jobs and now it works 🙂

kkohbrok (approved):
Thanks for the PR and thanks dealing with those last issues there! Let's get this in.

PR #1222, opened 2023-01-24
update credentials & introduce signer

franziskuskiefer:
This PR removed the CredentialBundle and instead introduces a basic credential crate and signer trait.

* The `Signer` trait implements a `sign` function and is passed around for signing instead of the credential bundle.
* basic credential crate in /basic_credential.
  This is a simple key pair that can be stored/read in/from the key store. It implements the `Signer` trait. This is using RustCrypto for now. Another one with evecrypt could be implemented.
* The other main difference is how "credentials" are created now because the key is generated independently. This will of course change again slightly when we support x509.
  ```rust
  let credential = Credential::new("Identity".into(), CredentialType::Basic).unwrap();
  let signature_keys = BasicCredential::new(signature_algorithm, backend.crypto()).unwrap();
  ```
* OpenMLS itself only handles the public keys that are in the spec now.
* Because the signer is required in a lot of public APIs now there's _a lot_ of churn in tests.


openmls/src/key_packages/mod.rs
```diff
@@ -229,15 +231,12 @@ impl KeyPackage {
...
-        credential: &CredentialBundle, // FIXME: make credential
+        signer: &(impl ByteSigner + ?Sized),
+        credential_with_key: CredentialWithKey,
         extensions: Extensions,
         leaf_node_capabilities: Capabilities,
         leaf_node_extensions: Extensions,
     ) -> Result<KeyPackageCreationResult, KeyPackageNewError<KeyStore::Error>> {
-        if SignatureScheme::from(config.ciphersuite) != credential.credential().signature_scheme() {
-            return Err(KeyPackageNewError::CiphersuiteSignatureSchemeMismatch);
-        }
```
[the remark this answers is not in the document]
franziskuskiefer: There is no ciphersuite in the credential anymore so we can't check anything.

openmls/src/group/mls_group/updates.rs
```diff
@@ -76,38 +65,12 @@ impl MlsGroup {
...
-            .key_store()
-            .read(
-                &old_credential
-                    .signature_key()
-                    .tls_serialize_detached()
-                    .map_err(LibraryError::missing_bound_check)?,
-            )
-            .ok_or(ProposeSelfUpdateError::NoMatchingCredentialBundle)?;
-
+        // let old_credential = self.credential()?;
```
kkohbrok, line 73: snip?

openmls/src/key_packages/mod.rs
```diff
@@ -347,13 +347,11 @@ impl KeyPackage {
...
         // Verify the signature on this key package.
-        <Self as Verifiable>::verify_no_out(
-            self,
-            backend,
-            self.leaf_node().signature_key(),
-            self.leaf_node().credential().signature_scheme(),
-        )
-        .map_err(|_| {
+        let pk = OpenMlsSignaturePublicKey::from_signature_key(
+            self.leaf_node().signature_key().clone(),
```
kkohbrok, line 351: Maybe not for this PR, but we should really try to use OpenMlsSignaturePublicKey everywhere.

openmls/src/messages/mod.rs
```diff
@@ -334,8 +334,10 @@ impl GroupSecrets {
         self
     }
 
-    #[cfg(any(feature = "test-utils", test))]
-    pub fn random_encoded(
+    // XXX[KAT]: Used in kat_messages.
+    //           Get in when kat's are in again.
```
kkohbrok, line 338: Add link to #1051? There are a few others like this.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -458,26 +462,51 @@ impl LeafNode {
...
     }
 
+    /// Generate a fresh leaf node with a fresh encryption key but otherwise
+    /// the same properties as the current leaf node.
+    ///
+    /// The newly generated encryption key pair is stored in the key store.
+    ///
+    /// This function can be used when generating an update. In most other cases
+    /// a leaf node should be generated as part of a new [`KeyPackage`].
+    pub fn updated<KeyStore: OpenMlsKeyStore>(
```
kkohbrok, line 491: Do we actually need this function in the public API? Couldn't we instead call `self_update` and have it generate an updated leaf node for us?

kkohbrok (changes requested):
This is a big one. Thanks pushing through the whole thing! 

I made a general pass, but ignored a bunch of the tests. Let's get the main story straight, then I'll take a closer look at the tests, although I expect that they are mostly churn plus some refactoring.

openmls/src/group/tests/test_proposal_validation.rs
```diff
@@ -1833,52 +1932,42 @@ fn test_valsem110(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider
...
+        alice_credential_with_key_and_signer: _,
+        bob_group: _,
+        bob_credential_with_key_and_signer: _,
     } = validation_test_setup(PURE_PLAINTEXT_WIRE_FORMAT_POLICY, ciphersuite, backend);
 
+    // TODO[FK]: This must go in again when #819 is finished and the leaf node
+    //           uses an encryption key that's different from the init key in the
+    //           key package.
+    //           Right now we can't do this because we don't have Bob's private
+    //           key any more.
```
kkohbrok, line 1945: Can you create an issue for this, since this effectively disables this test?

kkohbrok (changes requested):
Almost there. Just a few more nits, then this is good to go.

kkohbrok (approved):
🚢 it!

PR #1224, opened 2023-01-26
ci: Fix coverage job.

duesee:
~I couldn't find a "timeout" in GitHub actions but assumed it should mean "seconds". Minutes would exceed the default `timeout-minutes: 360`, which is likely not what we want. Given that 500 seconds are roughly 10 minutes and the job is taking longer, I removed the timeout altogether. The added `version` should fix the rest. Not sure why there was a `cargo clean` but it doesn't seem necessary.~

This PR switches the coverage measurement from tarpaulin to grcov.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -18,15 +18,13 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
-
-      - name: Clean
-        run: cargo clean
+          override: true
 
       - name: Run cargo-tarpaulin
         uses: actions-rs/tarpaulin@v0.1
         with:
-          timeout: "500"
-          args: "--force-clean"
+          version: "0.15.0"
+          args: "-- --test-threads 1"
 
       - name: Upload to codecov.io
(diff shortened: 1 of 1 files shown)


duesee:
Despite the failing test here due to the `pull_request_target` context, measurement with grcov seems to work. See https://github.com/openmls/openmls/actions/runs/4014259323/jobs/6894556999, and https://app.codecov.io/github/openmls/openmls/commit/6798965a96ff858753fda4214f07555f4f5a26df.

franziskuskiefer (changes requested):
The actions you use here (`cargo` and `grcov`) don't appear to be maintained. So it's not really an improvement, other than that it works for now.
Let's do this properly.

duesee:
> So it's not really an improvement, other than that it works for now.

Unblocking the PR workflow sounds like an improvement to me... Anyway, we could execute the required grcov steps manually. Would that suffice or do you have something specific in mind?

franziskuskiefer (approved):
yeah, you're right. Let's get this in but leave the issue open to do the steps manually so we don't run into the same issues again when the actions break.

PR #1225, opened 2023-01-26
Feedback: Introduce a typestate to track leaf node validation.

duesee:
This draft shows roughly how I imagine to proceed with leaf node validation.

Instead of having a single `LeafNode`, we would have `LeafNode<T>` where `T` serves as a compile time marker to denote what validation steps were executed. In a first step, I just replaced every instance of `LeafNode` with `LeafNode<Unvalidated>`. This was a pure refactoring. In the next step, I replaced `LeafNode<Unvalidated>` with, e.g., `LeafNode<KeyPackage>` in places were I was sure that validation should have taken place. Because the compiler treats `LeafNode<Unvalidated>` and `LeafNode<KeyPackage>` as two distinct types, it was required to convert `LeafNode<Unvalidated>` to `LeafNode<KeyPackage>`. Now, the idea is that the only way to obtain a `LeafNode<KeyPackage>` is to call `validate_in_key_package`. This way, the compiler will guide us to every place in the code where this transition is required, i.e., where it is needed to validate the leaf node.

I left a lot `unwrap`s and am unsure in some places if I use the correct typestate. So, please see this draft PR only as an early example how this idea will look like.

Regarding TLS serialization: By deriving `Tls(De)serialize` only on some (but not all) typestates, we can control what concrete types will implement `Tls(De)serialize` in the end. This could be a bit subtle but seems to work out. I am a bit worried however what implications "perfect derive" will have. We do rely here on the fact that `#derive(Foo)` implies `T: Foo` which is, from my understanding, what perfect derive will lift. So... not sure about this.

### DIFF
--- a/openmls/src/group/core_group/apply_proposals.rs
+++ b/openmls/src/group/core_group/apply_proposals.rs
@@ -8,7 +8,9 @@ use crate::{
     group::errors::ApplyProposalsError,
     messages::proposals::{AddProposal, Proposal, ProposalType},
     schedule::InitSecret,
-    treesync::{diff::TreeSyncDiff, node::leaf_node::OpenMlsLeafNode},
+    treesync::{
+        diff::TreeSyncDiff, errors::LeafNodeValidationError, node::leaf_node::OpenMlsLeafNode,
+    },
 };
 
 use super::*;
@@ -100,13 +102,17 @@ impl CoreGroup {
                 };
                 let leaf_node: OpenMlsLeafNode = match own_leaf_index {
                     Some(leaf_index) if leaf_index == sender_index => {
-                        let own_leaf_node = match leaf_nodes
-                            .iter()
-                            .find(|&leaf_node| leaf_node.leaf_node() == update_proposal.leaf_node())
-                        {
+                        let own_leaf_node = match leaf_nodes.iter().find(|leaf_node| {
+                            leaf_node.leaf_node().clone()
+                                == update_proposal.leaf_node().clone().into()
+                        }) {
                             Some(leaf_node) => leaf_node,
                             // We lost the LeafNode apparently
-                            None => return Err(ApplyProposalsError::MissingLeafNode),
+                            None => {
+                                return Err(ApplyProposalsError::LeafNodeValidation(
+                                    LeafNodeValidationError::MissingLeafNode,
+                                ))
+                            }
                         };
                         own_leaf_node.clone()
                     }
--- a/openmls/src/key_packages/mod.rs
+++ b/openmls/src/key_packages/mod.rs
@@ -95,7 +95,7 @@ use crate::{
     treesync::{
         node::{
             encryption_keys::EncryptionKeyPair,
-            leaf_node::{Capabilities, LeafNodeSource, Lifetime},
+            leaf_node::{Capabilities, Lifetime},
         },
         LeafNode,
     },
@@ -111,10 +111,11 @@ use tls_codec::{
 };
 
 #[cfg(test)]
-use crate::treesync::node::encryption_keys::EncryptionKey;
+use crate::treesync::node::{encryption_keys::EncryptionKey, leaf_node::LeafNodeSource};
 
 // Private
 mod codec;
+use crate::treesync::node::leaf_node::{Unknown, ValidKeyPackage};
 use errors::*;
 
 // Public
@@ -144,7 +145,7 @@ struct KeyPackageTBS {
     protocol_version: ProtocolVersion,
     ciphersuite: Ciphersuite,
     init_key: HpkePublicKey,
-    leaf_node: LeafNode,
+    leaf_node: LeafNode<ValidKeyPackage>,
     extensions: Extensions,
 }
 
@@ -287,10 +288,10 @@ impl KeyPackage {
     ) -> Result<(Self, EncryptionKeyPair), KeyPackageNewError<KeyStore::Error>> {
         // We don't need the private key here. It's stored in the key store for
         // use later when creating a group with this key package.
-        let (leaf_node, encryption_key_pair) = LeafNode::new(
+        let (leaf_node, encryption_key_pair) = LeafNode::key_package(
             config,
             credential_with_key,
-            LeafNodeSource::KeyPackage(Lifetime::default()),
+            Lifetime::default(),
             leaf_node_capabilities,
             leaf_node_extensions,
             backend,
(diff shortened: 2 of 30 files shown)


openmls/src/group/core_group/apply_proposals.rs
```diff
@@ -96,17 +99,24 @@ impl CoreGroup {
...
+                    }) {
                         Some(leaf_node) => leaf_node,
                         // We lost the LeafNode apparently
                         None => return Err(ApplyProposalsError::MissingLeafNode),
                     };
                     own_leaf_node.clone()
                 } else {
-                    update_proposal.leaf_node().clone().into()
+                    // TODO
+                    LeafNode::<StateValidated>::from(update_proposal.leaf_node().clone()).into()
```
duesee, line 119: This is easy to simplify.

openmls/src/key_packages/mod.rs
```diff
@@ -323,29 +325,6 @@ impl KeyPackage {
         &self,
         backend: &impl OpenMlsCryptoProvider,
     ) -> Result<(), KeyPackageVerifyError> {
-        // Extension included in the extensions or leaf_node.extensions fields
```
duesee, line 326: These (and other) validations could all be deleted when there is already a (always validated) `LeafNode<KeyPackage>`.

openmls/src/key_packages/mod.rs
```diff
@@ -485,7 +464,7 @@ impl KeyPackage {
 
         // We don't need the private key here. It's stored in the key store for
         // use later when creating a group with this key package.
-        let leaf_node = LeafNode::create_new_with_key(
+        let leaf_node = LeafNode::<StateUnvalidated>::create_new_with_key(
```
duesee, line 467: This can be simplified.

openmls/src/treesync/tests_and_kats/tests/test_diff.rs
```diff
@@ -38,15 +41,15 @@ fn test_free_leaf_computation(ciphersuite: Ciphersuite, backend: &impl OpenMlsCr
     // Build a rudimentary tree with two populated and two empty leaf nodes.
     let nodes: Vec<Option<Node>> = vec![
         Some(Node::LeafNode(
-            kpb_0.key_package().leaf_node().clone().into(),
+            LeafNode::<StateValidated>::from(kpb_0.key_package().leaf_node().clone()).into(),
```
duesee, line 44: This can be simplified. The `StateValidated` serves as a umbrella type currently that can be used after any of `StateKeyPackage, StateAddUpdateCommit, StateRatchet`. I'm not happy with it and would like to improve this.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -382,6 +384,26 @@ impl TlsDeserializeTrait for LeafNodeTbs {
     }
 }
 
+// TODO: Name up to debate.
+#[derive(Debug, Clone, PartialEq, Eq, TlsSize, TlsDeserialize)]
+pub struct StateUnvalidated;
```
duesee, line 389: This is the initial leaf node state. No validations were done to this leaf node.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -382,6 +384,26 @@ impl TlsDeserializeTrait for LeafNodeTbs {
     }
 }
 
+// TODO: Name up to debate.
+#[derive(Debug, Clone, PartialEq, Eq, TlsSize, TlsDeserialize)]
```
duesee, line 388: Note that this only derives `TlsDeserialize` and not `TlsSerialize` because we don't want so serialize (and send) invalid leaf nodes.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -382,6 +384,26 @@ impl TlsDeserializeTrait for LeafNodeTbs {
...
+pub struct StateUnvalidated;
+
+// TODO: Name up to debate.
+#[derive(Debug, Clone, PartialEq, Eq, TlsSize, TlsSerialize)]
+pub struct StateKeyPackage;
```
duesee, line 393: This leaf node was validated in the context of a key package. The MLS draft lists the required checks. Notably, the leaf node must have a key package leaf node source and a valid lifetime (and lifetime range).

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -382,6 +384,26 @@ impl TlsDeserializeTrait for LeafNodeTbs {
...
+
+// TODO: Name up to debate.
+#[derive(Debug, Clone, PartialEq, Eq, TlsSize, TlsSerialize)]
+pub struct StateAddUpdateCommit;
```
duesee, line 397: Placeholder until we have a better name or more fitting state.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -382,6 +384,26 @@ impl TlsDeserializeTrait for LeafNodeTbs {
...
+
+// TODO: Name up to debate.
+#[derive(Debug, Clone, PartialEq, Eq, TlsSize, TlsSerialize)]
+pub struct StateRatchetTree;
```
duesee, line 401: Placeholder until we have a better name or more fitting state.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -382,6 +384,26 @@ impl TlsDeserializeTrait for LeafNodeTbs {
...
+
+// TODO: Name up to debate.
+#[derive(Debug, Clone, PartialEq, Eq, TlsSize, TlsSerialize)]
+pub struct StateValidated;
```
duesee, line 405: This state can be created from any of the three other "validated" states. The reason we have this is because we use the leaf node "generically" as soon as we have it in the ratchet tree. I'm not sure if this is tight enough or if we loose guarantees due to this. In any case, I don't like this and have to think again.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -410,20 +432,47 @@ impl TlsDeserializeTrait for LeafNodeTbs {
...
+        LeafNode {
+            payload: value.payload,
+            signature: value.signature,
+            phantom: Default::default(),
+        }
     }
 }
 
-impl LeafNode {
+impl<T> LeafNode<T> {
```
duesee, line 475: This was an easy way to make the code compile again. Later, we do want to have separate `impl`s for every typestate.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -438,7 +487,7 @@ impl LeafNode {
         capabilities: Capabilities,
         extensions: Extensions,
         backend: &impl OpenMlsCryptoProvider,
-    ) -> Result<(Self, EncryptionKeyPair), LibraryError> {
+    ) -> Result<(LeafNode<StateUnvalidated>, EncryptionKeyPair), LibraryError> {
```
duesee, line 490: Not sure about this. I think we should try to ensure that every leaf node we create is valid. Returning `LeafNode<Unvalidated>` here was a quick way to make the compiler happy. Validation will need to be done later (when more appropriate).

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -563,60 +602,64 @@ impl LeafNode {
...
+    pub(crate) fn unvalidate(self) -> LeafNode<StateUnvalidated> {
+        LeafNode {
+            payload: self.payload,
+            signature: self.signature,
+            phantom: Default::default(),
+        }
+    }
+
+    /// Validate the leaf node in the context of a key package.
+    pub(crate) fn validate_in_key_package(
```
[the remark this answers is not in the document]
duesee: Yes. However, there are a few instances in the code where it happens so that we need to go from `ValidKeyPackage` to `ValidUpdate`. Maybe these can be resolved in some other way. As an alternative, we would use `invalidate().validate_in_<other>`.

PR #1227, opened 2023-01-30
re-enable valsem110 test

### LINKED ISSUE #1226: Fix ValSem110 test

franziskuskiefer:
Fixes #1226.

This also fixes a bug in LeafNode.update_and_re_sign when updating with a leaf node.

### DIFF
--- a/openmls/src/treesync/node/leaf_node.rs
+++ b/openmls/src/treesync/node/leaf_node.rs
@@ -1026,13 +1026,12 @@ impl OpenMlsLeafNode {
         if let Some(leaf_node) = leaf_node.into() {
             leaf_node_tbs.payload.credential = leaf_node.credential().clone();
             leaf_node_tbs.payload.encryption_key = leaf_node.encryption_key().clone();
-        }
-
-        if let Some(new_encryption_key) = new_encryption_key.into() {
+        } else if let Some(new_encryption_key) = new_encryption_key.into() {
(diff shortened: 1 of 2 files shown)


openmls/src/treesync/node/leaf_node.rs
```diff
@@ -1026,13 +1026,12 @@ impl OpenMlsLeafNode {
         if let Some(leaf_node) = leaf_node.into() {
             leaf_node_tbs.payload.credential = leaf_node.credential().clone();
             leaf_node_tbs.payload.encryption_key = leaf_node.encryption_key().clone();
-        }
-
-        if let Some(new_encryption_key) = new_encryption_key.into() {
+        } else if let Some(new_encryption_key) = new_encryption_key.into() {
```
kkohbrok, line 1029: This might be a bit overkill, but can't wouldn't it be nicer to introduce an enum (e.g. `LeafOrEncryptionKey` here that is either a leaf node or an encryption key and that has a `From` implementation from both? We could then use an impl Into<LeafOrEncryptionKey> as input instead of the two options. That would mean that we wouldn't have this awkward double-option setup and the caller still wouldn't have to worry about constructing the enum externally.

kkohbrok (approved):
Looks good, thanks! One suggestion for the `update_and_re_sign` function, but no need to do that in this PR if you don't want to.

PR #1233, opened 2023-01-30
refactor/chore: Move `Capabilities` into own module + Improve readability of `leaf_node.rs`.

duesee:
This is a pure refactoring. Thus, I used `pub use capabilities::*` to re-export and `pub(crate)` for access control. Edit: Also added two `TODO`s. Edit: While at it, I also improved the readability of the `leaf_node.rs` file. I'll split the file into distinct typestate modules in a follow-up. (This PR is best reviewed commit-by-commit.)

### DIFF
--- a/openmls/src/group/core_group/staged_commit.rs
+++ b/openmls/src/group/core_group/staged_commit.rs
@@ -5,7 +5,7 @@ use crate::framing::mls_content::FramedContentBody;
 use crate::treesync::errors::TreeSyncAddLeaf;
 use crate::treesync::node::encryption_keys::EncryptionKeyPair;
 use crate::treesync::node::leaf_node::{
-    LeafNodeTbs, OpenMlsLeafNode, TreeInfoTbs, VerifiableLeafNodeTbs,
+    LeafNodeTbs, OpenMlsLeafNode, TreeInfo, VerifiableLeafNodeTbs,
 };
 use crate::treesync::{diff::StagedTreeSyncDiff, treekem::DecryptPathParams};
 
@@ -210,7 +210,7 @@ impl CoreGroup {
                 //       up.
                 let tbs = LeafNodeTbs::from(
                     leaf_node.clone(),
-                    TreeInfoTbs::commit(self.group_id().clone(), sender_index),
+                    TreeInfo::commit(self.group_id().clone(), sender_index),
                 );
                 let verifiable_leaf_node = VerifiableLeafNodeTbs {
                     tbs: &tbs,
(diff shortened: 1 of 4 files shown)


openmls/src/treesync/node/leaf_node.rs
```diff
@@ -797,13 +619,17 @@ impl LeafNode {
             .capabilities
             .extensions
             .contains(extension_type)
-            || default_extensions().iter().any(|et| et == extension_type)
+            || capabilities::default_extensions()
```
franziskuskiefer, line 441: There's a `pub use capabilities::*` so why this change?

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -690,49 +503,103 @@ impl LeafNode {
...
-impl<'a> Verifiable for VerifiableLeafNodeTbs<'a> {
-    fn unsigned_payload(&self) -> Result<Vec<u8>, tls_codec::Error> {
-        self.tbs.tls_serialize_detached()
-    }
+pub type ParentHash = VLBytes;
 
-    fn signature(&self) -> &Signature {
-        self.signature
-    }
+// -------------------------------------------------------------------------------------------------
```
franziskuskiefer, line 568: I'm not sure I understand what these are supposed to be.
If there's that much code here that something like a divider is needed, it should go somewhere else and not add even more content (even if it's a comment).

openmls/src/group/core_group/staged_commit.rs
```diff
@@ -5,7 +5,7 @@ use crate::framing::mls_content::FramedContentBody;
 use crate::treesync::errors::TreeSyncAddLeaf;
 use crate::treesync::node::encryption_keys::EncryptionKeyPair;
 use crate::treesync::node::leaf_node::{
-    LeafNodeTbs, OpenMlsLeafNode, TreeInfoTbs, VerifiableLeafNodeTbs,
+    LeafNodeTbs, OpenMlsLeafNode, TreeInfo, VerifiableLeafNodeTbs,
```
franziskuskiefer: This is an unnecessary change and unrelated to moving code around.

franziskuskiefer (changes requested):
Some unnecessary changes in here. Otherwise lgtm.

openmls/src/group/core_group/staged_commit.rs
duesee: Can we keep this change? The `TreeInfo` is not signed and this PR is about cleaning up.

openmls/src/treesync/node/leaf_node.rs
duesee: It separates `LeafNode`, `LeafNodeTbs`, etc. code. Not sure if we want own modules for this. I can delete them.

openmls/src/group/core_group/staged_commit.rs
duesee: We discussed this offline and I'll revert this commit for now.

franziskuskiefer (approved):

PR #1235, opened 2023-01-31
feat: review keystore

beltram:
The goal of this PR is to facilitate the usage of the keystore.
* add `MlsEntity` with serde bound to delegate the serialization format to the implementor
* add `MlsEntity::ID` for branching (when using a SQL database, one has to know to which table the value should be looked up which wasn't possible with a generic)
* simplify `OpenMlsKeyStore::*_epoch_keys` by introducing a struct for handling this composite key


beltram:
@kkohbrok I would love your feedback on this one. Especially the part where I removed the dedicated methods for handling keys in the keystore :)

openmls/src/ciphersuite/mod.rs
```diff
@@ -42,8 +43,36 @@ mod tests;
 /// A simple type for HPKE public keys using [`VLBytes`] for (de)serializing.
 pub type HpkePublicKey = VLBytes;
 
+// TODO: derive zeroize
```
kkohbrok, line 46: Is this something you still want to do here? If not, could you create an issue and link it here?

openmls/src/group/core_group/mod.rs
```diff
@@ -751,6 +751,47 @@ impl CoreGroup {
...
+                group_id.as_slice(),
+                &leaf_index.u32().to_be_bytes(),
+                &epoch.to_be_bytes(),
+            ]
+            .concat(),
+        )
+    }
+}
+
+impl CoreGroup {
```
kkohbrok, line 794: I'm a bit confused as to the structure here. Is there a reason you interrupt the CoreGroup impl at this point? Just to introduce the `EpochKeypairId`? Also, the test-utils block was further below. Is there a reason you moved it up here?

kkohbrok (changes requested):
Thanks for the PR! This certainly better than what we currently have. However, it would be even nicer if the caller had the option (in addition to serde) to just use TLS serialization. I don't see an immediate way of doing that, but you have thought about this more than me, so maybe you have an idea.

Otherwise just two nits.

openmls/src/ciphersuite/mod.rs
beltram: Right, will remove and link it in #1240

openmls/src/group/core_group/mod.rs
beltram: Reordered

kkohbrok (approved):
Looks good! Thanks for addressing my comments.

PR #1236, opened 2023-01-31
test vectors 1/n

franziskuskiefer:
This PR enables the encryption test vectors again. It also fixes some issues along the way.

Note that there are a number of bulk changes in here that we might want to pull out (clippy changes will go away when #1237 is merged).


openmls/src/tree/tests_and_kats/kats/kat_encryption.rs
```diff
@@ -220,15 +245,15 @@ fn build_handshake_messages(
...
-        MlsMessageHeader {
-            group_id: group.group_id().clone(),
-            epoch: group.context().epoch(),
-            sender: sender_index,
-        },
+        // MlsMessageHeader {
+        //     group_id: group.group_id().clone(),
+        //     epoch: group.context().epoch(),
+        //     sender: sender_index,
+        // },
```
kkohbrok, line 256: Snip?

kkohbrok (changes requested):
Just one nit, then we can get this in.

openmls/src/tree/tests_and_kats/kats/kat_encryption.rs
franziskuskiefer: Thanks, missed this one.

kkohbrok (approved):
Thanks for starting to breathe life back into the test vectors. I'm going to regret saying this, but it's good to have test vectors back in the test suite.

PR #1239, opened 2023-02-01
tls codec cleanup

franziskuskiefer:
This PR removes the remaining TlsVecs from the library.
The only noteworthy change here is the one in extensions, that required some additional fixes because the length of variable length encoded vectors is not known in advance.


kkohbrok (approved):
:shipit:

PR #1241, opened 2023-02-02
refactor: Derive `tls_codec` implementations for `Proposal`.

duesee:
This will help later with message validation.


kkohbrok (approved):
Looks straight-forward enough. Let's get this in!

openmls/src/messages/proposals.rs
```diff
@@ -131,19 +131,30 @@ impl TryFrom<u16> for ProposalType {
...
-#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+#[derive(
+    Debug, PartialEq, Clone, Serialize, Deserialize, TlsSize, TlsSerialize, TlsDeserialize,
+)]
 #[allow(missing_docs)]
+#[repr(u16)]
 pub enum Proposal {
+    #[tls_codec(discriminant = 1)]
     Add(AddProposal),
+    #[tls_codec(discriminant = 2)]
```
raphaelrobert, line 142: Do you need all of them? I thought the first one was enough (but I might remember it wrong)

franziskuskiefer: The first is indeed enough. This works too, it's just a little verbose ...

duesee: > Do you need all of them? I thought the first one was enough (but I might remember it wrong)

The first one would be enough, yes. But I decided to make it explicit so that we don't rearrange by accident and are already prepared to handle extensions that start with `0xff...`.

PR #1246, opened 2023-02-02
Implement leaf node validation.

duesee:


openmls/src/treesync/node/leaf_node.rs
```diff
@@ -263,53 +285,122 @@ impl LeafNode {
...
         signature_keys: &[SignaturePublicKey],
         encryption_keys: &[EncryptionKey],
         members_supported_credentials: &[&[CredentialType]],
         currently_in_use: &[CredentialType],
     ) -> Result<&Self, LeafNodeValidationError> {
-        self.validate_required_capabilities(required_capabilities)?
-            .validate_that_capabilities_contain_extension_types()?
+        // The client verifies the validity of a LeafNode using the following steps:
+        self
+            // Verify that the credential in the LeafNode is valid as described in Section 5.3.1.
```
franziskuskiefer, line 357: The AS verification has to happen on the outside. Maybe add a comment to the entry points to the validation that all the AS validation must have happened before.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -263,53 +285,122 @@ impl LeafNode {
...
+            // ...
+            // Verify that the signature on the LeafNode is valid using signature_key.
```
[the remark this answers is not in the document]
franziskuskiefer: Also, there's no signature verification in here 🤔

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -263,53 +285,122 @@ impl LeafNode {
...
+            // signature_key (ValSem104)
+            .validate_that_signature_key_is_unique(signature_keys)?
+            // encryption_key (ValSem114)
+            .validate_that_encryption_key_is_unique(encryption_keys)?;
 
         Ok(self)
     }
 
+    /// ValSem106: Check the required capabilities of the add proposals
+    // ~ - Are the two listed in the `Capabilities` Extension?
```
franziskuskiefer, line 384:
```suggestion
    /// - Are the two listed in the `Capabilities` Extension?
```

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -263,53 +285,122 @@ impl LeafNode {
...
+    pub(crate) fn validate_version_ciphersuite(
+        &self,
+        version: ProtocolVersion,
+        ciphersuite: Ciphersuite,
+    ) -> Result<&Self, LeafNodeValidationError> {
+        if !self.payload.capabilities.versions.contains(&version) {
+            Err(LeafNodeValidationError::UnsupportedVersion)
```
franziskuskiefer, line 391: If you use returns, the function becomes more readable because you don't need all the nesting.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -263,53 +285,122 @@ impl LeafNode {
...
+            // (ValSem106)
+            .validate_version_ciphersuite(version, ciphersuite)?
+            // If the GroupContext has a required_capabilities extension, then the required extensions, proposals, and credential types MUST be listed in the LeafNode's capabilities field.
+            // (ValSem106)
+            .validate_required_capabilities(required_capabilities)?
             .validate_that_capabilities_contain_credential_type()?
-            .validate_that_signature_key_is_unique(signature_keys)?
-            .validate_that_encryption_key_is_unique(encryption_keys)?
+            // Verify that the credential type is supported by all members of the group, as specified by the capabilities field of each member's LeafNode, ...
             .validate_against_group_credentials(members_supported_credentials)?
```
franziskuskiefer, line 369: This function name seems a little off. We use it to verify member credentials, but the leaf node doesn't know anything about other members. So what it's really doing is verifying that all the credential types it gets in the list are supported by this leaf node.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -263,53 +285,122 @@ impl LeafNode {
...
-            .validate_credential_in_use(currently_in_use)?;
+            // ... and that the capabilities field of this LeafNode indicates support for all the credential types currently in use by other members.
+            .validate_credential_in_use(currently_in_use)?
```
franziskuskiefer, line 371: I see that this is in the spec but I wonder if this is really necessary here. It sounds like unnecessary overhead to me. Other leaf node credentials are checked when i) added to the tree, and ii) when changed. So whatever is in there should be supported.

openmls/src/group/core_group/validation.rs
```diff
@@ -171,6 +165,59 @@ impl CoreGroup {
                 return Err(ProposalValidationError::InsufficientCapabilities);
             }
 
+            // Leaf node validation.
+            {
+                let leaf_node = add_proposal.add_proposal().key_package().leaf_node();
+
+                let filtered_group_signature_keys = {
```
franziskuskiefer, line 172: We talked about scoping like this. And this looks like a clear case to me where this should be a function. Then you don't need any of the additional scopes here.

openmls/src/group/core_group/validation.rs
```diff
@@ -171,6 +165,59 @@ impl CoreGroup {
...
+                    .validate_in_key_package(
+                        self.version(),
+                        self.ciphersuite(),
+                        self.group_context_extensions().required_capabilities(),
+                        &filtered_group_signature_keys,
+                        &self.group_encryption_keys(),
+                        &self.members_supported_credentials(),
+                        &self.currently_in_use(),
+                    )
+                    .map_err(|error| match error {
```
franziskuskiefer, line 204: Shouldn't all this be handled by the `into`?

openmls/src/group/core_group/validation.rs
```diff
@@ -208,10 +244,14 @@ impl CoreGroup {
                 .remove_proposals()
                 .any(|p| p.remove_proposal().removed == index);
             // ValSem104
+            // Note: This is different to the leaf node check above. It checks for duplicate
```
franziskuskiefer, line 247: For both of these: Do we need to check twice or would it be enough to check once, after the proposals are applied?

openmls/src/treesync/node/leaf_node.rs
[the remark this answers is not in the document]
duesee: That's one reason why I left this as a draft PR. I would have prefered for `LeafNode` to always have a valid signature. We even have a `VerifiableLeafNode` in place. But these things are not really in place now. Thus, I will implement this check manually.

openmls/src/group/core_group/validation.rs
duesee: In theorie yes. But there are differences in the errors that are returns regarding duplicate signature keys in a proposal list v.s. duplicate signature keys in a leaf node and I wanted to make sure that I don't break the tests when I can prevent it. We can hopefully remove all these things when things are settled and we unify error handling.

Issue #1253, opened 2023-02-05
Refactor secret tree


raphaelrobert:
The secret tree code is pretty old and needs some attention:

 - It is not panic-free
 - The structure should be revisited
 - It should be better aligned with the treemath from the binary tree, the interface between the two is suboptimal

PR #1254, opened 2023-02-05
Add new treemath to secret tree

### LINKED ISSUE #1252: [Bug] Secret tree still uses legacy treemath

raphaelrobert:
Fixes #1252.

This is a somewhat quick fix to get towards interop. The secret tree now uses the new full tree math from the binary tree.
Some more refactoring would be needed, I filed #1253 for that.


openmls/src/tree/mod.rs
```diff
@@ -6,9 +6,8 @@ use openmls_traits::OpenMlsCryptoProvider;
 pub mod sender_ratchet;
 
 // Crate
-pub(crate) mod index;
+//pub(crate) mod index;
```
kkohbrok, line 9: Snip!

kkohbrok (changes requested):
One nit, then this is good to go!

kkohbrok (approved):

Issue #1255, opened 2023-02-06
[Feature request] Allow creation of public groups and their use for message processing


kkohbrok:
Some applications of MLS require that a client be able to track a group through processing public messages. For example, a delivery service could use this feature to track group membership and delivery information.

There is already a PublicGroup component in CoreGroup that we could integrate more functionality into such as message validation. We could then (maybe behind a feature flag) expose an API that allows the creation of standalone PublicGroups, as well as message processing through them.

PR #1256, opened 2023-02-06
Move validation into public group

kkohbrok:
This PR moves commit validation that's currently part of `CoreGroup` into `PublicGroup` without changing any of the semantics. It is another step towards #1255.


raphaelrobert (approved):
Thanks, LGTM!

PR #1260, opened 2023-02-09
hpke encrypt with label

### LINKED ISSUE #1258: Add label and context to public-key encryption #842
https://github.com/mlswg/mls-protocol/pull/842

> As discussed in the virtual interim 2022-12-16, it is possible that extensions might be tempted to re-use HPKE public keys created for MLS. Even within MLS, we have two different public-key encryption cases, Welcome and Commit (though these currently use different keys).
> 
> Currently, for Commit encryption, the HPKE `info` parameter is set to the group context. The content of `info` is actually undefined for Welcome encryption (!) This PR makes the following changes:
> 
> * Define general `EncryptWithLabel` and `DecryptWithLabel` functions that populte the HPKE `info` parameter with an encoded label and context
> * Apply these to Commit and Welcome encryption:
>   
>   * Commit: label="UpdatePathNode", context=`group_context`
>   * Welcome: label="Welcome", context=`encrypted_group_info`
> * Add an IANA registry of public-key encryption labels (as for signing)
> 
> There is also a bit of reorganization of the Ciphersuites section to split it into three subsections.

franziskuskiefer:
This PR closes #1258 by implementing the labeled HPKE encryption.
It also fixes the signature label.

### DIFF
--- a/openmls/src/ciphersuite/hpke.rs
+++ b/openmls/src/ciphersuite/hpke.rs
@@ -1,0 +1,134 @@
+//! ### Public-Key Encryption
+//!
+//! As with signing, MLS includes a label and context in encryption operations to
+//! avoid confusion between ciphertexts produced for different purposes.  Encryption
+//! and decryption including this label and context are done as follows:
+//!
+//! ```text
+//! EncryptWithLabel(PublicKey, Label, Context, Plaintext) =
+//!   SealBase(PublicKey, EncryptContext, "", Plaintext)
+//!
+//! DecryptWithLabel(PrivateKey, Label, Context, KEMOutput, Ciphertext) =
+//!   OpenBase(KEMOutput, PrivateKey, EncryptContext, "", Ciphertext)
+//! ```
+//!
+//! Where EncryptContext is specified as:
+//!
+//! ```text
+//! struct {
+//!   opaque label<V>;
+//!   opaque context<V>;
+//! } EncryptContext;
+//! ```
+//!
+//! And its fields set to:
+//!
+//! ```text
+//! label = "MLS 1.0 " + Label;
+//! context = Context;
+//! ```
+//!
+//! Here, the functions `SealBase` and `OpenBase` are defined {{RFC9180}}, using the
+//! HPKE algorithms specified by the group's ciphersuite.  If MLS extensions
+//! require HPKE encryption operations, they should re-use the EncryptWithLabel
+//! construction, using a distinct label.  To avoid collisions in these labels, an
+//! IANA registry is defined in {{mls-public-key-encryption-labels}}.
+
+use openmls_traits::{
+    crypto::OpenMlsCrypto,
+    types::{Ciphersuite, CryptoError, HpkeCiphertext},
+};
+use thiserror::Error;
+use tls_codec::{Serialize, TlsDeserialize, TlsSerialize, TlsSize, VLBytes};
+
+/// HPKE labeled encryption errors.
+#[derive(Error, Debug, PartialEq, Clone)]
+pub(crate) enum Error {
+    /// Error while serializing content. This should only happen if a bounds check was missing.
+    #[error(
+        "Error while serializing content. This should only happen if a bounds check was missing."
+    )]
+    MissingBoundCheck,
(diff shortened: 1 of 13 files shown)


openmls/src/ciphersuite/hpke.rs
```diff
@@ -0,0 +1,134 @@
...
+//! ```text
+//! label = "MLS 1.0 " + Label;
+//! context = Context;
+//! ```
+//!
+//! Here, the functions `SealBase` and `OpenBase` are defined {{RFC9180}}, using the
+//! HPKE algorithms specified by the group's ciphersuite.  If MLS extensions
+//! require HPKE encryption operations, they should re-use the EncryptWithLabel
+//! construction, using a distinct label.  To avoid collisions in these labels, an
+//! IANA registry is defined in {{mls-public-key-encryption-labels}}.
```
kkohbrok, line 35: This won't point to anything in the rust docs, will it?

openmls/src/ciphersuite/signature.rs
```diff
@@ -36,7 +36,7 @@ pub struct SignContent {
     content: VLBytes,
 }
 
-const SIGN_LABEL_PREFIX: &str = "MLS 1.0";
+const SIGN_LABEL_PREFIX: &str = "MLS 1.0 ";
```
kkohbrok, line 39: Can't we have this label at one place? E.g. in `ciphersuite/mod.rs`. Right now it's the same in `hpke` and `signature`.

openmls/src/ciphersuite/hpke.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+impl From<(&str, &[u8])> for EncryptContext {
+    fn from((label, context): (&str, &[u8])) -> Self {
+        Self::new(label, context.into())
+    }
+}
+
+/// Encrypt to an HPKE key with a label.
+pub(crate) fn encrypt_with_label(
```
kkohbrok, line 93: Would it make sense to elevate `HpkePublicKey` into a struct and make this part of the implementation? The same goes for the decryption function below.

openmls/src/messages/mod.rs
```diff
@@ -236,14 +236,16 @@ impl PathSecret {
         ciphersuite: Ciphersuite,
         public_key: &HpkePublicKey,
         group_context: &[u8],
-    ) -> HpkeCiphertext {
-        backend.crypto().hpke_seal(
-            ciphersuite.hpke_config(),
+    ) -> Result<HpkeCiphertext, LibraryError> {
+        hpke::encrypt_with_label(
```
kkohbrok, line 240: I think if you follow the callers of `encrypt` here up until `encrypt_path`, you can actually take an `&EncryptionKey` as input here, where you can just call `encrypt` and the implementation of `EncryptionKey` will add the right label for you. This would mirror what `decrypt` below is already doing.

openmls/src/group/core_group/new_from_welcome.rs
```diff
@@ -58,16 +58,15 @@ impl CoreGroup {
...
-                ciphersuite.hpke_config(),
-                egs.encrypted_group_secrets(),
-                key_package_bundle.private_key.as_slice(),
-                &[],
-                &[],
-            )
-            .map_err(|_| WelcomeError::UnableToDecrypt)?;
+        let group_secrets_bytes = hpke::decrypt_with_label(
+            key_package_bundle.private_key.as_slice(),
+            "Welcome",
```
kkohbrok, line 63: Maybe not in this PR, but it would probably make sense to have a dedicated type for `InitKey` in the same way we already have for `Encryptionkey`.

kkohbrok (changes requested):
Looks good! Just a nit or two.

openmls/src/ciphersuite/hpke.rs
franziskuskiefer: We could, but this would produce quite some churn (every call to it would have to be changed to access the value).
I open an issue to do it later if we think it makes sense.

kkohbrok (approved):
Thanks for making the changes! Let's get this in.

PR #1261, opened 2023-02-09
chore: Towards leaf node validation.

duesee:
This is a first PR towards leaf node validation. It mostly consists of refactorings and the addition of documentation but also contains a required lifetime fix that will otherwise break the tests later when we validate leaf nodes.


openmls/src/group/core_group/mod.rs
```diff
@@ -31,20 +31,29 @@ mod test_past_secrets;
 mod test_proposals;
 
 #[cfg(test)]
-use super::errors::CreateGroupContextExtProposalError;
-use super::public_group::PublicGroup;
-use crate::binary_tree::array_representation::TreeSize;
+use std::io::{Error, Read, Write};
```
duesee, line 34: This was automatically cleaned-up using `rustfmt --config imports_granularity=Crate,group_imports=StdExternalCrate openmls/src/group/tests/test_external_commit_validation.rs`. Looking through the codebase, I think we pretty much agree on this style, thus, I configured my IDE to do this in changed files. I could also make a PR that does all this as a one-time pill but refrained from it to not introduce conflicts.

openmls/src/group/core_group/mod.rs
```diff
@@ -803,6 +800,56 @@ impl CoreGroup {
         );
         backend.key_store().delete::<Vec<EncryptionKeyPair>>(&k.0)
     }
+
+    /// Return supported credentials of all members.
```
duesee, line 804: Added methods. These are useful for leaf node validation later. As I see this... maybe I should already make them iterators (as Konrad suggessted in a previous review.)

openmls/src/group/core_group/new_from_external_init.rs
```diff
@@ -84,20 +85,15 @@ impl CoreGroup {
...
-            ..
-        } in public_group.members()
-        {
-            if signature_key == params_credential_with_key.signature_key.as_slice() {
-                let remove_proposal = Proposal::Remove(RemoveProposal { removed: index });
-                inline_proposals.push(remove_proposal);
-                break;
-            };
-        }
+        if let Some(us) = public_group.members().find(|member| {
```
duesee, line 88: This is only a refactoring. While debugging, I suspected that the leaf node index is not calculated correctly because we always used "free_leaf_index" even for "us". However, there are two `free_leaf_index` methods and the one used here is subtly different. Thus, I renamed it.

openmls/src/group/tests/test_external_commit_validation.rs
```diff
@@ -4,152 +4,83 @@
...
-
 // ValSem240: External Commit, inline Proposals: There MUST be at least one ExternalInit proposal.
 #[apply(ciphersuites_and_backends)]
 fn test_valsem240(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider) {
-    // Test with PublicMessage
     let ECValidationTestSetup {
         mut alice_group,
-        alice_credential: _,
         bob_credential,
-        mut plaintext,
```
duesee, line 114: Rename because there is no "plaintext" anymore in MLS.

openmls/src/group/tests/test_external_commit_validation.rs
```diff
@@ -4,152 +4,83 @@
...
-            ProposalOrRef::Proposal(proposal) => proposal.is_type(ProposalType::ExternalInit),
-            ProposalOrRef::Reference(_) => false,
-        })
-        .expect("Couldn't find external init proposal.");
-
-    content.proposals.remove(proposal_position);
-
-    plaintext.set_content(FramedContentBody::Commit(content));
+    // Setup
+    let public_message_commit_bad = {
```
duesee, line 43: Narrowed scoping. This makes it a lot easier to see what is happening in the test. The intention here was to better see which parts could be refactored out as functions.

openmls/src/group/tests/utils.rs
```diff
@@ -419,32 +416,19 @@ pub(crate) fn resign_message(
 #[cfg(test)]
 pub(crate) fn resign_external_commit(
     signer: &impl Signer,
-    plaintext: PublicMessage,
-    original_plaintext: &PublicMessage,
+    public_message: PublicMessage,
+    old_confirmation_tag: ConfirmationTag,
```
duesee, line 420: We do not use anything from the `original_plaintext` despite the confirmation tag. Thus, I made that more clear.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -136,7 +137,7 @@ impl LeafNode {
         backend: &impl OpenMlsCryptoProvider<KeyStoreProvider = KeyStore>,
         signer: &impl Signer,
     ) -> Result<Self, LeafNodeGenerationError<KeyStore::Error>> {
-        Self::generate(
+        Self::generate_update(
```
duesee, line 140: I renamed this because `generate_update` makes it more clear that the leaf node source is, in fact, update.

openmls/src/treesync/node/leaf_node/lifetime.rs
```diff
@@ -7,15 +7,16 @@ use tls_codec::{TlsDeserialize, TlsSerialize, TlsSize};
...
-/// The maximum total lifetime range that is acceptable for a leaf node.
-/// The value is in seconds and amounts to 3 * 28 Days, i.e., about 3 months.
-const DEFAULT_MAX_LEAF_NODE_LIFETIME_RANGE_SECONDS: u64 = 60 * 60 * 24 * 28 * 3;
-
 /// This value is used as the default amount of time (in seconds) the lifetime
 /// of a `KeyPackage` is extended into the past to allow for skewed clocks. The
 /// value is in seconds and amounts to 1h.
 const DEFAULT_KEY_PACKAGE_LIFETIME_MARGIN_SECONDS: u64 = 60 * 60;
 
+/// The maximum total lifetime range that is acceptable for a leaf node.
```
duesee, line 15: This is the lifetime fix mentioned in the PR description. It sets the maximum lifetime range to default + margin. One less and the tests will break (when we actually validate). I think this is what we want.

raphaelrobert (approved):
Thanks for the PR, LGTM!

openmls/src/group/core_group/mod.rs
duesee: Oh, sorry? Konrad mentioned this in a comment but we havn't opened an issue for that.

openmls/src/group/core_group/mod.rs
duesee: I hoped it would create less churn, but yeah, reformating "bit-by-bit" in PRs isn't the best idea. Sadly, the rustfmt options we would need to do it once and enforce in CI are still nightly-only :-(

PR #1266, opened 2023-02-09
Move `stage_commit` into `PublicGroup`

kkohbrok:
This PR moves the `stage_commit` function from `CoreGroup` into `PublicGroup`. The function now only optionally takes private key material as input. The functionality to process messages with only a `PublicGroup` will be added with the next follow-up PR.

`stage_commit` has been refactored into multiple individual functions and the individual steps have been re-ordered, but otherwise the function has not changed functionally.


openmls/src/group/public_group/diff/process_path.rs
```diff
@@ -27,23 +27,23 @@ use super::PublicGroupDiff;
 /// A helper struct which contains the values resulting from the preparation of
 /// a commit with path.
 #[derive(Default)]
-pub(crate) struct PathProcessingResult {
+pub(crate) struct PathComputationResult {
```
franziskuskiefer, line 30: If you want to rename this, the file itself should be renamed as well.

openmls/src/group/public_group/staged_commit.rs
```diff
@@ -0,0 +1,467 @@
...
+        psk::PskSecret, EpochSecrets, GroupEpochSecrets, InitSecret, JoinerSecret, KeySchedule,
+    },
+    treesync::node::{
+        encryption_keys::EncryptionKeyPair,
+        leaf_node::{LeafNodeTbs, TreeInfoTbs, VerifiableLeafNode},
+    },
+};
+use std::collections::HashSet;
+
+pub(crate) struct PrivateGroupParams<'a> {
```
franziskuskiefer, line 21: I'm not sure I understand this.
You are passing secrets into the public group. Isn't the idea of the public group to operate only on the public parts?

franziskuskiefer (changes requested):
I have a conceptial issue with this (see inline comment).

openmls/src/group/public_group/staged_commit.rs
```diff
@@ -0,0 +1,260 @@
...
+            .confirmation_tag()
+            .ok_or(StageCommitError::ConfirmationTagMissing)?;
+
+        // If we have private key material, derive the secrets for the next
+        // epoch and check the confirmation tag.
+        diff.update_interim_transcript_hash(
+            ciphersuite,
+            backend,
+            received_confirmation_tag.clone(),
+        )?;
```
franziskuskiefer, line 254: All of this could be moved to a separate function that's creating the diff to scope the `mut` and separate the operations on the diff from the commit staging.
Same for the `stage_commit` in the core group. This way it's also easier to see where the two differ and making sure we keep them in sync.
Can you also add comments on both functions that the other one exists?

franziskuskiefer (changes requested):
Thanks, let's pull out a little more, just make things cleaner and do the rename. Then this is good to go.

franziskuskiefer (approved):
Thanks, let's get this in.

franziskuskiefer (approved):

PR #1268, opened 2023-02-10
chore: Derive `tls_codec` for `Node`.

duesee:
An easy one :-)

### DIFF
--- a/openmls/src/treesync/node.rs
+++ b/openmls/src/treesync/node.rs
@@ -2,6 +2,7 @@
 //! variants of the enum are `LeafNode` and [`ParentNode`], both of which are
 //! defined in the respective [`leaf_node`] and [`parent_node`] submodules.
 use serde::{Deserialize, Serialize};
+use tls_codec::{TlsDeserialize, TlsSerialize, TlsSize};
 
 use self::{leaf_node::OpenMlsLeafNode, parent_node::ParentNode};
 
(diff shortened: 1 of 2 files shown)


openmls/src/treesync/node.rs
```diff
@@ -11,11 +12,27 @@ pub(crate) mod leaf_node;
...
+///         case leaf:   LeafNode leaf_node;
+///         case parent: ParentNode parent_node;
+///     };
+/// } Node;
+/// ```
+#[derive(
+    Debug, PartialEq, Eq, Clone, Serialize, Deserialize, TlsSize, TlsDeserialize, TlsSerialize,
+)]
 #[allow(missing_docs)]
 #[allow(clippy::large_enum_variant)]
```
kkohbrok, line 30: My local clippy is not complaining if I remove those two allows. While you're touching this code, could you check if we can get rid of them?

kkohbrok (approved):
Looks good with an optional nit since it's somewhat unrelated.

openmls/src/treesync/node.rs
duesee: Removing code is always good :-)

duesee:
Fixed-up and force-pushed to cleanup the `fixup!`. As this PR is easy enough I thought that force-pushing directly may spare use some time because it would invalidate your review later.

kkohbrok (approved):
Thanks!

PR #1270, opened 2023-02-10
refactor: Introduce `RatchetTreeExported` to unify `Vec<Option<Node>>`-handling.

duesee:
This PR introduces a `RatchetTreeExported` type. This is useful for (later) validation and improves readability. This PR also changes the pattern of "borrowing and cloning internally" to "moving" (and letting the caller decide when to clone.)


franziskuskiefer:
Note that the spec changes here to make this stricter in mlswg/mls-protocol#868

openmls/src/treesync/mod.rs
```diff
@@ -68,6 +71,102 @@ pub use node::{leaf_node::LeafNode, parent_node::ParentNode, Node};
...
+    Clone,
+    Debug,
+    Default,
+    Serialize,
+    Deserialize,
+    TlsSerialize,
+    TlsDeserialize,
+    TlsSize,
+)]
+// TODO(validation):
```
duesee, line 87: We could already open an issue for that.

raphaelrobert: Yes, please convert this into TODOs with a ticket number

openmls/src/treesync/mod.rs
```diff
@@ -68,6 +78,102 @@ pub use node::{leaf_node::LeafNode, parent_node::ParentNode, Node};
...
+//
+// * ValSemXXX: "If the tree has 2^d leaves, then it has 2^(d+1) - 1 nodes. The ratchet_tree vector logically has this number of entries, but the sender SHOULD NOT include blank nodes after the last non-blank node."
+// * ValSemXXX: "Regardless of how the client obtains the tree, the client MUST verify that the root hash of the ratchet tree matches the tree_hash of the GroupContext before using the tree for MLS operations."
+pub struct RatchetTreeExported(Vec<Option<Node>>);
```
raphaelrobert, line 98: Why not generically call this `RatchetTree`? That way we can have a `RatchetTreeIn` in the context of validation.

duesee: No good reason. Happy to rename it to `RatchetTree` when you don't see potential for confusion :+1:

openmls/src/treesync/mod.rs
duesee: I created #1305.

openmls/src/treesync/mod.rs
duesee: Ah... I renamed it to `RatchetTree` but could have already used `RatchetTreeIn`. Should we leave it that way anticipating this change or rename later? I'm leaning towards renaming later.

raphaelrobert (approved):

raphaelrobert:
`cargo doc` is unhappy, otherwise we are good to go

raphaelrobert (approved):

PR #1280, opened 2023-02-16
Introduce an API that allows message processing by a `PublicGroup`

kkohbrok:
This PR finally introduces proper message processing through `PublicGroup`. In particular, it exposes a `process` API endpoint that is analogous to the one exposed by `CoreGroup`, as well as a corresponding `merge_commit` endpoint. The PR also includes a few improvements to the PublicGroup and commit processing in general:

* Introduce builder pattern for PublicGroup (required to fully hide TreeSync)
* Introduce message processing for PublicGroup
  * move `parse_message` into PublicGroup (still called from CoreGroup)
  * Add a short test
  * Integrate the `PublicGroup` into the test framework. This leads to longer test times (since each message is not processed an additional time), but thoroughly tests the PublicGroup's message processing.
* push key store access out of stage_commit in CoreGroup
* Remove unnecessary (no-op) layer in the validation chain (UnverifiedContextMessage vs. UnverifiedMessage)

We should discuss if we really want to integrate PublicGroup into the test framework. It's a nice way to get test coverage for public group processing, but it also slows some of the test by a factor of 10 🙈.

For the purposes of review, I split this PR into individual commits that don't individually compile, but show the main changes and that don't overlap.


openmls/src/group/core_group/staged_commit.rs
```diff
@@ -352,6 +337,29 @@ impl CoreGroup {
...
+    pub(crate) fn read_keys_and_stage_commit(
+        &self,
+        mls_content: &AuthenticatedContent,
+        proposal_store: &ProposalStore,
+        own_leaf_nodes: &[OpenMlsLeafNode],
+        backend: &impl OpenMlsCryptoProvider,
+    ) -> Result<StagedCommit, StageCommitError> {
+        let (old_epoch_keypairs, leaf_node_keypairs) = self
+            .read_decryption_keypairs(backend, own_leaf_nodes)
+            .unwrap();
```
raphaelrobert, line 353: I know it's just a test function, but since you return a `Result`, why not use that instead of the `unwrap()`?

openmls/src/group/public_group/process.rs
```diff
@@ -0,0 +1,234 @@
...
+                ))
+            }
+            Sender::External(_) => {
+                // We don't support messages from external senders yet
+                // TODO #151/#106
+                todo!()
+            }
+        }
+    }
+}
```
raphaelrobert, line 234: Can you confirm this is just copied over from the `CoreGroup` without major changes?

openmls/src/test_utils/test_framework/mod.rs
```diff
@@ -322,18 +336,54 @@ impl MlsGroupTestSetup {
...
+        )
+        .unwrap();
+
+        // Update the public group. If it's a public message, just process it. Create a new group from the sender.
+        match message {
+            ProtocolMessage::PrivateMessage(_) => {
+                // If it's a ciphertext we can't track group membership with a
+                // PublicGroup, so we have to create a new one every time with
+                // the sender's state.
+                let (new_public_group, _extensions) = PublicGroup::from_external(
```
raphaelrobert, line 356: As noted offline, creating a group will involve validating all signatures going forward and this step will become really expensive. Maybe add a comment at least.

raphaelrobert (changes requested):
Thanks for the PR! I did a first pass and noted some nits.

openmls/src/group/core_group/staged_commit.rs
kkohbrok: Ah, my bad. I somehow thought it would return a `KeyStore::Error`, but since it's returning a `StageCommitError`, we can indeed just pass the error back up. Good catch!

raphaelrobert (approved):

PR #1282, opened 2023-02-16
Fix the key schedule KAT

raphaelrobert:
Re-introduces the PSK secret.


franziskuskiefer (approved):
Thanks, I added the exporter context and test as well.

PR #1285, opened 2023-02-20
spec: Use `encrypted_group_info` as context in decryption of welcome secrets.

duesee:
This PR makes it so that the decryption of the encrypted group secrets takes the encrypted group info as context. Notably, it was required to adapt a test because it can now not happen that the ciphersuite mismatches.

### DIFF
--- a/openmls/src/group/core_group/mod.rs
+++ b/openmls/src/group/core_group/mod.rs
@@ -933,23 +933,13 @@ impl CoreGroup {
         )
         .map_err(LibraryError::unexpected_crypto_error)?;
 
-        // Create group secrets for later use, so we can afterwards consume the
-        // `joiner_secret`.
-        let encrypted_secrets = diff.encrypt_group_secrets(
-            &joiner_secret,
-            apply_proposals_values.invitation_list,
-            path_computation_result.plain_path.as_deref(),
-            &apply_proposals_values.presharedkeys,
-            backend,
-            self.own_leaf_index(),
-        )?;
-
         // Prepare the PskSecret
         let psk_secret =
             PskSecret::new(ciphersuite, backend, &apply_proposals_values.presharedkeys)?;
 
         // Create key schedule
-        let mut key_schedule = KeySchedule::init(ciphersuite, backend, joiner_secret, psk_secret)?;
+        let mut key_schedule =
+            KeySchedule::init(ciphersuite, backend, joiner_secret.clone(), psk_secret)?;
 
         let serialized_provisional_group_context = diff
             .group_context()
@@ -978,7 +968,7 @@ impl CoreGroup {
         diff.update_interim_transcript_hash(ciphersuite, backend, confirmation_tag.clone())?;
 
         // only computes the group info if necessary
-        let group_info = if !encrypted_secrets.is_empty() || self.use_ratchet_tree_extension {
+        let group_info = {
             // Create the ratchet tree extension if necessary
             let external_pub = provisional_epoch_secrets
                 .external_secret()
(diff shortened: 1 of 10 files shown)


openmls/src/group/core_group/mod.rs
```diff
@@ -1006,12 +996,10 @@ impl CoreGroup {
             };
             // Sign to-be-signed group info.
             Some(group_info_tbs.sign(signer)?)
-        } else {
-            None
         };
 
         // Check if new members were added and, if so, create welcome messages
-        let welcome_option = if !encrypted_secrets.is_empty() {
```
kkohbrok, line 1014: Is there a reason you moved the condition to the end? It seems that if the condition is false, we're doing quite a few computations only to throw them away afterwards. Although it's of course possible that this is optimized away by LLVM.

openmls/src/group/core_group/new_from_welcome.rs
```diff
@@ -94,10 +94,6 @@ impl CoreGroup {
             VerifiableGroupInfo::tls_deserialize(&mut group_info_bytes.as_slice())
                 .map_err(|_| WelcomeError::MalformedWelcomeMessage)?;
 
-        if ciphersuite != verifiable_group_info.ciphersuite() {
-            return Err(WelcomeError::GroupInfoCiphersuiteMismatch);
```
kkohbrok, line 98: Is the error used anywhere else? If not, you can probably just remove it.

kkohbrok (changes requested):
Thanks for the PR! Just two nits/questions. Also clippy is unhappy about too many arguments.

openmls/src/group/core_group/mod.rs
duesee: Good catch! I thought there was a reason but `diff.encrypt_group_secrets` does way less than I thought. By replacing `encrypted_group_secrets.is_empty()` with `apply_proposals_values.invitation_list.is_empty()` we can spare us the work. Regarding the joiner_secret: I discussed that with Franziskus and we don't think it is possible to get this working without cloning as the `joiner_secret` is effectively required two times: once to derive the welcome_secret and once to derive the secret used to encrypt the group secrets (label "Welcome" vs "welcome"). I made `KeySchedule::init` take a reference now to spare us the clone. But I think we would need to refactor the key schedule to make this nicer. But I would propose to do it in a follow-up. What do you think?

openmls/src/group/core_group/mod.rs
```diff
@@ -933,23 +933,13 @@ impl CoreGroup {
...
-        )?;
-
         // Prepare the PskSecret
         let psk_secret =
             PskSecret::new(ciphersuite, backend, &apply_proposals_values.presharedkeys)?;
 
         // Create key schedule
-        let mut key_schedule = KeySchedule::init(ciphersuite, backend, joiner_secret, psk_secret)?;
+        let mut key_schedule =
+            KeySchedule::init(ciphersuite, backend, joiner_secret.clone(), psk_secret)?;
```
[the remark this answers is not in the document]
duesee: See comment above. We would need to refactor the key schedule to make this work.

openmls/src/group/public_group/diff.rs
```diff
@@ -73,12 +73,14 @@ impl<'a> PublicGroupDiff<'a> {
     ///  - the invited members are not part of the tree yet
     ///  - the leaf index of a new member is identical to the own leaf index
     ///  - the plain path does not contain the correct secrets
+    #[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
kkohbrok: I agree that this is annoying, but I think this is a very justified use of an `allow`. In the end, the only place this function is used is `create_commit`, which is in need of some refactoring anyway. So I think it's okay to leave it for now.

kkohbrok (approved):

PR #1287, opened 2023-02-20
feat: process External Remove Proposal

augustocdias:
This adds the possibility of pre configured external senders to issue external remove proposals. 

We've added docs to the book with examples of how to use. We've had to re-export the `ExternalSender` from the extensions in order to allow consumers to set the allowed parties to issue those external remove proposals. A constructor to initialise the struct was also created.

The function `new_external_proposal` in the `AuthenticatedContent` was refactored to specify the `Sender` in order to be reused also for external remove proposals instead of only add as it was before.

The `Sender` variant `External` was refactored to have a `u32` value instead of the previous `LeafNodeIndex` to be more clear about what it is about. The external sender is not really part of the ratchet tree, so we think that setting it to `u32` leaves that clear. 

In the `UnverifiedExternalMessage`, signature verification was added in order to validate the sender of the proposals.

In the configuration struct, an array was added to set the external senders in the group. If the array is empty the group will be created without the extension. 

Processing an external remove proposal will result in a `ProcessedMessage` with a `RemoveProposal` to be stored in the queue a committed by a client.

We've renamed the `JoinProposal` struct to `ExternalProposal` so it could be used for external removes as well as for external adds.


raphaelrobert:
Thanks for the PR! After glancing over this quickly I'm confused about the intention. The PR seems to remove the new joiner mechanism. These are however two different things, both semantically and syntactically. New joiners have the sender type `SenderType.new_member_proposal`, whereas external senders have the sender type `SenderType.external`. Can you elaborate a bit more on this perhaps?

augustocdias:
> Thanks for the PR! After glancing over this quickly I'm confused about the intention. The PR seems to remove the new joiner mechanism. These are however two different things, both semantically and syntactically. New joiners have the sender type `SenderType.new_member_proposal`, whereas external senders have the sender type `SenderType.external`. Can you elaborate a bit more on this perhaps?

I didn't remove the new joiner. The only thing about the new joiner that changed is that I renamed the struct `JoinProposal` to `ExternalProposal` and it does alongside the join proposal a remove proposal as well.

raphaelrobert:
As mentioned above, those are two different things. You confuse the join proposal with the external add proposal. Can you please point out what part of the spec you are trying to implement?

augustocdias:
> As mentioned above, those are two different things. You confuse the join proposal with the external add proposal. Can you please point out what part of the spec you are trying to implement?

I'll refactor and put it in another struct.
This implements part of the `ExternalProposals` [topic](https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#name-external-proposals). We're implementing just external removes here.

raphaelrobert:
There was quite a bit of discussion around how we want to separate join proposals & external proposals (and I think you were part of it?).
The spec is a bit vague there and categorizes both as external proposals. Join proposals are however quite different in nature since the sender intends to join the group. The other external proposals however are sent from senders that will never be part of the group by design. That's where the External Senders Extension comes in.
In OpenMLS this semantic distinction should really be reflected, that's why I was confused by the PR. The fact that you left out external add proposals finally made it look like everything became conflated.

augustocdias:
@raphaelrobert I've rolled back the change with the `JoinProposal`. Now it has a specific struct for the external remove proposal.

raphaelrobert:
It looks like #1280 changed how messages are processed, so this PR will unfortunately need some refactoring.

augustocdias:
@raphaelrobert I've fixed the conflicts. The commit list seems to have some duplicated commits. That was because I've made a mistake during the conflict resolution that rolled back them. I've just cherry-picked them back.

book/src/user_manual/remove_members.md
```diff
@@ -28,3 +28,18 @@ A member is removed from a group if another member commits to a remove proposal
 ```rust,no_run,noplayground
 {{#include ../../../openmls/tests/book_code.rs:getting_removed}}
 ```
+
+## External Proposal
+
+Parties outside the group can also make proposals to remove members as long as they are registered as part of the `ExternalSendersExtension` extension.
+Since those proposals are crafted by outsiders, they are always plaintext messages.
```
raphaelrobert, line 35:
```suggestion
Since those proposals are crafted by outsiders, they are always public messages.
```

book/src/user_manual/remove_members.md
```diff
@@ -28,3 +28,18 @@ A member is removed from a group if another member commits to a remove proposal
...
+
+```rust,no_run,noplayground
+{{#include ../../../openmls/tests/book_code.rs:external_remove_proposal}}
+```
+
+It is then up to one of the group member to process the proposal and commit it.
```
raphaelrobert, line 41:
```suggestion
It is then up to one of the group members to process the proposal and commit it.
```

openmls/src/framing/mls_auth_content.rs
```diff
@@ -170,21 +170,20 @@ impl AuthenticatedContent {
...
 
     /// This constructor builds an `PublicMessage` containing an External Proposal.
-    /// The sender is [Sender::NewMemberProposal].
-    // TODO #151/#106: We don't support preconfigured senders yet
     pub(crate) fn new_external_proposal(
         proposal: Proposal,
         group_id: GroupId,
         epoch: GroupEpoch,
         signer: &impl Signer,
+        sender: Sender,
```
raphaelrobert, line 178: I'd rather continue the separation here as well and keep the explicit sender type and introduce a `new_join_proposal` instead.

openmls/src/group/core_group/process.rs
```diff
@@ -97,9 +97,35 @@ impl CoreGroup {
                 ))
             }
             Sender::External(_) => {
-                // We don't support messages from external senders yet
-                // TODO #151/#106
```
raphaelrobert, line 101: Can you leave the TODO, since external adds are not done yet?

openmls/src/group/public_group/process.rs
```diff
@@ -225,9 +227,35 @@ impl PublicGroup {
                 ))
             }
             Sender::External(_) => {
-                // We don't support messages from external senders yet
-                // TODO #151/#106
```
raphaelrobert, line 229: Can you leave the TODO, since external adds are not done yet?

openmls/src/group/tests/external_remove_proposal.rs
```diff
@@ -0,0 +1,327 @@
...
+        backend,
+    );
+
+    // get Bob's index
+    let bob_index = alice_group
+        .members()
+        .find(|member| member.credential.identity() == b"Bob")
+        .map(|member| member.index)
+        .unwrap();
+    // Now Delivery Service wants to remove Bob with invalid sender index but there's no extension
```
raphaelrobert, line 308: It would be nice to test these two things separately: a) wrong sender index and b) there is no extension.
Ideally also c) existing index but wrong signature.

raphaelrobert (changes requested):
Thanks for the PR and thanks for making the requested changes! I left a few comments, but otherwise the PR is ok.

openmls/src/group/core_group/process.rs
augustocdias: I left them. They're just down in the code, after the processing of the external remove.

openmls/src/group/public_group/process.rs
augustocdias: I left them. They're just down in the code, after the processing of the external remove.

openmls/src/group/tests/external_remove_proposal.rs
augustocdias: They're already split:
1. external_remove_proposal_should_fail_when_invalid_external_senders_index
2. external_remove_proposal_should_fail_when_no_external_senders
3. external_remove_proposal_should_fail_when_invalid_signature
and external_remove_proposal_should_remove_member for the positive case

raphaelrobert: Sorry, I missed that completely.

openmls/src/extensions/external_sender_extension.rs
```diff
@@ -21,15 +23,48 @@ pub struct ExternalSender {
...
+pub struct SenderExtensionIndex(u32);
+
+impl SenderExtensionIndex {
+    /// Creates a new `SenderExtensionIndex` instance.
+    pub fn new(index: u32) -> Self {
+        Self(index)
+    }
+
+    /// Returns the internal index as usize
+    pub(crate) fn index(&self) -> Result<usize, std::num::TryFromIntError> {
```
raphaelrobert, line 64: We can cast here since we [enforce a 32bit minimum](https://github.com/openmls/openmls/blob/main/openmls/src/lib.rs#L143) for the platform. That way we avoid the `Result`.

raphaelrobert (changes requested):
Thanks for the changes, just a nit. And the CI is unhappy.

openmls/src/extensions/external_sender_extension.rs
augustocdias: Okay. I've changed it to casting

raphaelrobert (approved):
Thanks, let's get this in!

PR #1288, opened 2023-02-21
interop: Process Welcome test vectors.

duesee:
The first commits are work towards processing the test vectors as they are. For this, I needed to factor out a few methods in order to not "reimplement" the logic from `new_from_welcome`.


openmls/src/group/core_group/kat_welcome.rs
```diff
@@ -0,0 +1,259 @@
+use openmls_rust_crypto::OpenMlsRustCrypto;
```
raphaelrobert, line 1: Can you add a markdown header like in the other test vectors?

(from here: https://github.com/mlswg/mls-implementations/blob/main/test-vectors.md#welcome)

raphaelrobert (changes requested):
Thanks for the PR! Just some nits

raphaelrobert (approved):

duesee:
Ah, sorry for the force push. It seems I screwed up something during merge conflict resolution. On the good side we have a cleaner commit history now.

kkohbrok (dismissed):

kkohbrok (approved):

PR #1290, opened 2023-02-21
Tree validation KAT

raphaelrobert:
This PR introduces the tree validation KAT vector. All checks pass, but so far we don't check the leaf node signatures. I filed #1289 for that.


openmls/src/treesync/node.rs
```diff
@@ -41,3 +44,22 @@ pub(crate) enum NodeReference<'a> {
     Leaf(&'a OpenMlsLeafNode),
     Parent(&'a ParentNode),
 }
+
+#[derive(
+    Debug, PartialEq, Eq, Clone, Serialize, Deserialize, TlsSize, TlsDeserialize, TlsSerialize,
+)]
+#[repr(u8)]
+pub(crate) enum RawNode {
```
franziskuskiefer, line 52: This isn't needed. You need to implement the TLS codec manually on the `OpenMlsLeafNode` struct instead of deriving it. I did this fix for the treekem tv already.

```rust
impl Size for OpenMlsLeafNode {
    fn tls_serialized_len(&self) -> usize {
        self.leaf_node.tls_serialized_len()
    }
}

impl tls_codec::Serialize for OpenMlsLeafNode {
    fn tls_serialize<W: std::io::Write>(&self, writer: &mut W) -> Result<usize, tls_codec::Error> {
        self.leaf_node.tls_serialize(writer)
    }
}

impl tls_codec::Deserialize for OpenMlsLeafNode {
    fn tls_deserialize<R: std::io::Read>(bytes: &mut R) -> Result<Self, tls_codec::Error> {
        Ok(Self {
            leaf_node: LeafNode::tls_deserialize(bytes)?,
            leaf_index: None,
        })
    }
}
```

franziskuskiefer (changes requested):
Thanks, only one nit.

openmls/src/treesync/node.rs
raphaelrobert: I made that change so that we are aligned. In general, all of this is temporary anyway until we have established if we really need `OpenMlsLeafNode`. My gut feeling is we really don't.

franziskuskiefer (approved):
Thanks.
I'm pretty sure the openmlsleafnode will go away but let's not confuse future selfs with yer another node 🙃

Issue #1279, opened 2023-02-16
Tracking missing interop features.


duesee:
* [ ] Generate secret tree.

PR #1291, opened 2023-02-22
message protection kat

franziskuskiefer:
This adds the message protection kat.
It's not entirely finished bute generally passing. There are still a few todos in the kat that I added to #1279.

Note that this also makes the change from mlswg/mls-protocol#866.


openmls/src/group/core_group/mod.rs
```diff
@@ -909,7 +909,7 @@ impl CoreGroup {
         };
 
         // Build AuthenticatedContent
-        let mut commit = AuthenticatedContent::commit(
+        let mut authenticated_content = AuthenticatedContent::commit(
```
raphaelrobert, line 912: Just double-checking: is this renaming intended or some leftover from earlier?

raphaelrobert (approved):
Just one question. Feel free to merge otherwise.

openmls/src/group/core_group/mod.rs
franziskuskiefer: It's part of me debugging issues. And I found the name here confusing.
I'll click the auto-merge button. Please request change if you want it changed back.

franziskuskiefer:
Looks like #1280 broke some APIs. So this will have to wait.

PR #1294, opened 2023-02-23
Small refactor of CoreGroup message processing

kkohbrok:
This PR factors out the decryption process in `CoreGroup` message processing and adds two getters for test purposes.


franziskuskiefer (approved):
Thanks for the fixup. Loks like a generally useful refactor.

PR #1295, opened 2023-02-23
interop: Fix `ProposalRef` calculation and add "Passive Client Scenarios".

duesee:
This PR has two changes: It aligns the `ProposalRef` calculation with draft-17 and adds the passive client scenarios test. (I made it a single PR so that we can verify against the test vectors provided in mls-implementations during review but could drop the last commit to do this in two steps.)

I've added a convinience `Candidate` struct in test_utils that I intend to use through more interop testing. It kinda duplicates the `generate_credential/keypackage` functions but allows to store (or not to store) private key material and gives immediate access to keys.


franziskuskiefer (commented):
*(not reviewing just a drive-by for something I spotted)*

openmls/src/test_utils/mod.rs
```diff
@@ -72,6 +82,123 @@ pub fn hex_to_bytes_option(hex: Option<String>) -> Vec<u8> {
     }
 }
 
+// === Convenience functions ===
+
+#[cfg(test)]
+pub(crate) struct Candidate {
```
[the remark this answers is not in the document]
duesee: Yeah... I intend to use this function to create everything that is needed for a potential new group member. "Potential" is why I called it a `Candidate` but I see how the name is unclear without context. How about `GroupCandidate` or `PotentialNewMember`?

openmls/src/treesync/node/encryption_keys.rs
```diff
@@ -64,7 +64,7 @@ impl EncryptionKey {
 
 #[derive(Debug, Clone, Serialize, Deserialize, TlsDeserialize, TlsSerialize, TlsSize)]
 pub(crate) struct EncryptionPrivateKey {
-    key: HpkePrivateKey,
+    pub(crate) key: HpkePrivateKey,
```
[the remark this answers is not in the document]
duesee: Yeah, not sure why I did this. Could be a leftover from testing. Good catch, I'll take a look later.

openmls/src/test_utils/mod.rs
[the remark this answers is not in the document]
franziskuskiefer: Yep, just put the thing that it is supposed to be a candidate for in front or something.

openmls/src/group/tests/test_proposal_validation.rs
```diff
@@ -1105,12 +1145,12 @@ fn test_valsem106(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider
                     let result = alice_group.add_members(
                         backend,
                         &alice_credential_with_key_and_signer.signer,
-                        &[test_kp.clone()],
+                        &[test_kp_2.clone()],
```
franziskuskiefer, line 1148: I'm not sure I understand why we need this.
The only difference between this test and the one above is that it's by value instead of reference. But this shouldn't influence the test.
Can you ellaborate why you changed this?

openmls/src/group/tests/test_proposal_validation.rs
```diff
@@ -1214,7 +1253,7 @@ fn test_valsem106(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider
             // Positive case
             bob_group
                 .process_message(backend, original_update_plaintext)
-                .expect("Unexpected error.");
+                .unwrap();
```
franziskuskiefer, line 1256: There are a lot of these unrelated changes here...

openmls/src/group/core_group/kat_passive_client.rs
```diff
@@ -0,0 +1,680 @@
...
+    let epoch_authenticator = group.epoch_authenticator().as_slice().to_vec();
+
+    TestEpoch {
+        proposals,
+        commit,
+        epoch_authenticator,
+    }
+}
+
+pub fn run_test_vector(test_vector: PassiveClientWelcomeTestVector) -> Result<(), &'static str> {
```
franziskuskiefer, line 547: You don't need to return anything here.

openmls/src/schedule/psk.rs
```diff
@@ -141,16 +154,18 @@ pub enum Psk {
 /// in the key schedule.
 ///
 /// ```c
-/// // draft-ietf-mls-protocol-16
+/// // draft-ietf-mls-protocol-17
```
franziskuskiefer, line 157: I still don't think that these make sense.
But if you actually want to update and use them, you should use the current version -18.

franziskuskiefer (changes requested):
Thanks.
Generally lgtm, only a few nits.

openmls/src/group/tests/test_proposal_validation.rs
duesee: We iterate over `[ProposalInclusion::ByReference, ProposalInclusion::ByValue]` and thus call `propose_add_member` and `add_members` with the same key package(s). Thus, instead of testing what the test says it does, we got an error due to duplicate signature keys.

openmls/src/group/tests/test_proposal_validation.rs
duesee: I think we agreed that a bit of cleanup is fine during pull requests? Do you want me to revert?

openmls/src/schedule/psk.rs
duesee: Checked that `PSKType`, `ResumptionPSKUsage`, and `PreSharedKeyID` are still the same in -18. Updated.

openmls/src/group/tests/test_proposal_validation.rs
franziskuskiefer: It's fine around the code that's being touched but these changes are all over the file (this and others). That makes it harder to review. Let's keep it here. But try to keep these type of changes a little more localised to the changes you are making.

openmls/src/group/tests/test_proposal_validation.rs
franziskuskiefer: Ah ok. I guess the idea was to have these on fresh groups instead or something. But this works too.

franziskuskiefer (approved):
🛳️

PR #1297, opened 2023-02-24
feat: return ProposalRef on proposals and allow to remove proposal from the store

augustocdias:
This PR adds the possibility to remove a proposal from the store by its reference. In order to do that all functions that adds a proposal to the store, now are returning a proposal reference.

A test was added to ensure that the proposal is indeed being removed and not added in future commits.

This is useful if a client needs to rollback proposal operations.


raphaelrobert:
This looks like a useful mechanism, but things might change because of #1295. The spec has changed and the proposal reference now also needs more input than what's available in the proposal itself, at least for standalone proposals. I suggest we wait for #1295.

duesee:
Sorry, @augustocdias. #1295 took a bit longer. I hope it will not be too difficult for you to rebase. Your changes are very useful, also for testing.

book/src/user_manual/processing.md
```diff
@@ -45,6 +45,15 @@ Standalone proposals are returned as a `QueuedProposal`, indicating that they ar
 {{#include ../../../openmls/tests/book_code.rs:inspect_add_proposal}}
 ```
 
+#### Rolling back proposals
+
+Operations that add a proposal to the queue, will return the its reference. This reference can be used to remove
```
duesee, line 50:
```suggestion
Operations that add a proposal to the queue, will return its reference. This reference can be used to remove
```

openmls/src/group/core_group/proposals.rs
```diff
@@ -49,6 +49,14 @@ impl ProposalStore {
     pub(crate) fn empty(&mut self) {
         self.queued_proposals = Vec::new();
     }
+    pub(crate) fn remove(&mut self, proposal_ref: ProposalRef) -> Option<()> {
```
duesee, line 52:
```suggestion

    pub(crate) fn remove(&mut self, proposal_ref: ProposalRef) -> Option<()> {
```

openmls/src/group/mls_group/membership.rs
```diff
@@ -7,7 +7,7 @@ use openmls_traits::signatures::Signer;
 
 use crate::{
     binary_tree::array_representation::LeafNodeIndex, group::errors::CreateAddProposalError,
-    messages::group_info::GroupInfo, treesync::LeafNode,
+    messages::group_info::GroupInfo, prelude::hash_ref::ProposalRef, treesync::LeafNode,
```
duesee, line 10: Could you please use `ciphersuite::hash_ref::ProposalRef` instead?

openmls/src/group/mls_group/mod.rs
```diff
@@ -11,6 +11,7 @@ use crate::{
     group::*,
     key_packages::{KeyPackage, KeyPackageBundle},
     messages::{proposals::*, Welcome},
+    prelude::hash_ref::ProposalRef,
```
duesee, line 14:
```suggestion
    ciphersuite::hash_ref::ProposalRef,
```

openmls/src/group/mls_group/mod.rs
```diff
@@ -422,6 +423,16 @@ impl MlsGroup {
     pub(crate) fn clear_pending_proposals(&mut self) {
         self.proposal_store.empty()
     }
+
+    /// Removes a specific proposal from the store and return it if it exists.
+    pub fn clear_pending_proposal(
```
duesee, line 428: I think I would call this `remove` and not clear. "clear" is mostly used as in "clear everything" (see above).

duesee (commented):
Thanks! I made some suggestions already but let's wait for another approval.

openmls/src/group/core_group/proposals.rs
```diff
@@ -49,6 +49,14 @@ impl ProposalStore {
...
+        let index = self
+            .queued_proposals
+            .iter()
+            .position(|p| p.proposal_reference() == proposal_ref)?;
+        self.queued_proposals.swap_remove(index);
```
duesee, line 57: I'm not sure about this. `swap_remove` changes the order of elements. I think the more safe thing to do would be to just remove the proposal.

openmls/src/group/core_group/proposals.rs
duesee: (This is only about a newline.)

augustocdias:
> Thanks! I made some suggestions already but let's wait for another approval.

Thanks. I've addressed all comments.

openmls/tests/book_code.rs
```diff
@@ -828,9 +828,19 @@ fn book_operations(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvide
         unreachable!("Expected a QueuedProposal.");
     }
 
+    // Create AddProposal and remove it
+    // ANCHOR: remove_proposal_by_ref
```
raphaelrobert, line 832: Just a nit, but `remove_proposal_by_ref` sounds confusing since we have an add proposal. Maybe 'roll back'?

book/src/user_manual/processing.md
```diff
@@ -45,6 +45,15 @@ Standalone proposals are returned as a `QueuedProposal`, indicating that they ar
 {{#include ../../../openmls/tests/book_code.rs:inspect_add_proposal}}
 ```
 
+#### Rolling back proposals
+
+Operations that add a proposal to the queue, will return its reference. This reference can be used to remove
+a proposal from the store. This can be useful for example to rollback in case of errors.
```
raphaelrobert, line 51:
```suggestion
Operations that add a proposal to the proposal store, will return its reference. This reference can be used to remove
a proposal from the proposal store. This can be useful for example to roll back in case of errors.
```

openmls/src/group/core_group/proposals.rs
```diff
@@ -49,6 +49,15 @@ impl ProposalStore {
     pub(crate) fn empty(&mut self) {
         self.queued_proposals = Vec::new();
     }
+
+    pub(crate) fn remove(&mut self, proposal_ref: ProposalRef) -> Option<()> {
```
raphaelrobert, line 53: Can you add some docs?

openmls/src/group/mls_group/mod.rs
```diff
@@ -422,6 +423,16 @@ impl MlsGroup {
...
+
+    /// Removes a specific proposal from the store and return it if it exists.
+    pub fn remove_pending_proposal(
+        &mut self,
+        proposal_ref: ProposalRef,
+    ) -> Result<(), MlsGroupStateError> {
+        self.proposal_store
+            .remove(proposal_ref)
+            .ok_or(MlsGroupStateError::PendingProposalNotFound)
+    }
```
raphaelrobert, line 435: The comment says the proposal is returned, but it isn't

raphaelrobert (changes requested):
Thanks for the PR, the tests look thorough, just a few comments.

openmls/tests/book_code.rs
augustocdias: Would it work if I put the add proposal out of the anchor? Or you prefer to remove the doc completely regarding this?

augustocdias: I've just got your comment. You meant to rename the anchor, right?

raphaelrobert:
Can you merge main? I can't push to your branch, unfortunately.

augustocdias:
> Can you merge main? I can't push to your branch, unfortunately.

done

raphaelrobert (approved):

PR #1309, opened 2023-03-03
rename PrivateContentTbe to PrivateMessageContent

franziskuskiefer:
mlswg/mls-protocol#870


raphaelrobert:
While this is likely to get merged into the protocol, don't we want to wait for that? Unless of course this is relevant for interop.

franziskuskiefer:
> While this is likely to get merged into the protocol, don't we want to wait for that? Unless of course this is relevant for interop.

It's marked as ready to merge, so I think it's safe to merge. But it this doesn't have to be merged right away. But having to look at the change again takes longer than just opening this PR, that's why I did it right away 😉

franziskuskiefer:
@raphaelrobert please rebase and fix conflicts. It has been sitting here for a while now and broke with #1318.

raphaelrobert (approved):

PR #1312, opened 2023-03-06
interop: Update key schedule.

duesee:


openmls/src/schedule/mod.rs
```diff
@@ -709,7 +710,7 @@ impl ExporterSecret {
         backend: &impl OpenMlsCryptoProvider,
         epoch_secret: &EpochSecret,
     ) -> Result<Self, CryptoError> {
-        let secret = epoch_secret.secret.derive_secret(backend, "exported")?;
+        let secret = epoch_secret.secret.derive_secret(backend, "exporter")?;
```
duesee, line 713: Good that we do interop... 🙈

openmls/src/schedule/kat_key_schedule.rs
```diff
@@ -253,12 +252,18 @@ fn write_test_vectors() {
...
     let _ = pretty_env_logger::try_init();
-    let tests: Vec<KeyScheduleTestVector> = read("test_vectors/key-schedule.json");
 
-    for test_vector in tests {
-        match run_test_vector(test_vector, backend) {
-            Ok(_) => {}
-            Err(e) => panic!("Error while checking key schedule test vector.\n{e:?}"),
+    let test_files: Vec<Vec<KeyScheduleTestVector>> = vec![
+        read("test_vectors/key-schedule.json"),
+        read("test_vectors/mlspp/key-schedule.json"),
```
franziskuskiefer, line 258: Where is this coming from? There should be only one test vector by now.

franziskuskiefer (changes requested):
Thanks. The actual change looks good to me. But I'm not sure where that mlspp tv is coming from. There should be only one tv.

franziskuskiefer (approved):
Thanks

PR #1314, opened 2023-03-07
interop: Update "Transcript Hashes".

duesee:


openmls/src/framing/mls_auth_content.rs
```diff
@@ -281,16 +280,35 @@ impl AuthenticatedContent {
...
     pub fn test_signature(&self) -> &Signature {
         &self.auth.signature
     }
 
-    #[cfg(test)]
     pub(super) fn unset_confirmation_tag(&mut self) {
         self.auth.confirmation_tag = None;
     }
+
+    pub(crate) fn from_bytes(bytes: &[u8]) -> Result<Self, tls_codec::Error> {
```
duesee, line 295: Only this is new.

openmls/src/schedule/mod.rs
```diff
@@ -819,25 +819,27 @@ impl ConfirmationKey {
...
-    #[cfg(test)]
-    pub(crate) fn _from_secret(secret: Secret) -> Self {
+#[cfg(test)]
+impl ConfirmationKey {
+    pub(crate) fn from_secret(secret: Secret) -> Self {
         Self { secret }
     }
+}
 
-    #[cfg(any(feature = "test-utils", test))]
```
duesee, line 829: Only moved.

traits/src/types.rs
```diff
@@ -287,10 +287,10 @@ impl core::fmt::Display for Ciphersuite {
     }
 }
 
-impl From<&Ciphersuite> for u16 {
+impl From<Ciphersuite> for u16 {
```
duesee, line 290: We should remove all instances of `&Ciphersuite` as `Ciphersuite`s are `Copy`able. If we agree on this I can already open an issue for some later cleanup.

openmls/src/framing/public_message.rs
```diff
@@ -291,11 +295,53 @@ impl<'a> ConfirmedTranscriptHashInput<'a> {
     }
 }
 
+#[cfg(test)]
```
duesee, line 298: I think that {Confirmed,Interim}TranscriptHash{Input,} can be made very modular, and that we can remove all other `calculate_` methods everywhere. I can open an issue to do this in a later cleanup phase.

franziskuskiefer: Can you make use of these functions in the code already? Then we don't need to add the test cfg and would know we are testing the same thing.

openmls/src/framing/mls_auth_content.rs
franziskuskiefer: Why is this needed? Can't you just derive this for test builds?

openmls/src/group/tests/kat_transcript_hashes.rs
```diff
@@ -0,0 +1,284 @@
...
+
+    #[serde(with = "hex::serde")]
+    pub confirmed_transcript_hash_after: Vec<u8>,
+    #[serde(with = "hex::serde")]
+    pub interim_transcript_hash_after: Vec<u8>,
+}
+
+/// Transcript test vector error
+#[derive(Error, Debug, PartialEq, Eq, Clone)]
+pub enum TranscriptTestVectorError {
```
franziskuskiefer, line 66: We don't really need the errors anymore because the interop runner doesn't do test vectors anymore.

franziskuskiefer (commented):
Thanks, a few nits and questions but then lets get this in.

openmls/src/framing/mls_auth_content.rs
duesee: `derive` is not possible due to `FramedContentAuthData::deserialize` that requires the content type. But yes, I should already do this properly: I will change this to a test-gated custom tls_codec implementation and do the trailing data check in the test for now.

We should think about providing a blanket implementation that provides `from_bytes` for everything that is `tls_codec::deserialize`. Sometimes we want to ensure there is no trailing data after deserialize and this would be a very effective way to get that for everything. What do you think?

franziskuskiefer (changes requested):
Thanks, generally looks good to me with a few nits and potential follow ups.

openmls/src/framing/public_message.rs
```diff
@@ -279,10 +276,33 @@ pub(crate) struct ConfirmedTranscriptHashInput<'a> {
...
+            .hash(
+                ciphersuite.hash_algorithm(),
+                &[interim_transcript_hash, &serialized].concat(),
+            )
+            .map_err(LibraryError::unexpected_crypto_error)
+    }
+}
+
+impl<'a> TryFrom<&'a AuthenticatedContent> for ConfirmedTranscriptHashInput<'a> {
+    type Error = &'static str;
```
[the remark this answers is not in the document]
duesee: This was more of a refactoring so I don't wanted to add yet another change. I opened https://github.com/openmls/openmls/issues/1322.

openmls/src/messages/group_info.rs
```diff
@@ -158,21 +158,17 @@ impl GroupInfo {
...
-            )
-            .map_err(LibraryError::unexpected_crypto_error)
+        let interim_transcript_hash = {
+            let input = InterimTranscriptHashInput::from(self.confirmation_tag());
+
+            input.calculate_interim_transcript_hash(
+                crypto,
+                self.group_context().ciphersuite(),
+                self.group_context().confirmed_transcript_hash(),
+            )?
```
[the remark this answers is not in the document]
duesee: I adjusted it here but can't find another instance where this would work without making the code weird, i.e., relying on the return value of a assignment to be `()`. In the other cases the code is subtly different.

franziskuskiefer (dismissed):
Thanks, you'll have to ask for review again after merging main. But this is ready to go in.

duesee:
Sorry for the force push. But this just got too messy. I made sure to keep all suggestions.

openmls/src/framing/mls_auth_content_in.rs
```diff
@@ -158,6 +155,26 @@ impl AuthenticatedContentIn {
     }
 }
 
+/// Note: we can't `derive(tls_codec::Deserialize)` because [`FramedContentAuthData`] cannot
+///       implement the usual `tls_codec::Deserialize` as it requires the content type as parameter.
+impl tls_codec::Deserialize for AuthenticatedContentIn {
```
duesee, line 160: Previously this was `AuthenticatedContent`. Now, we use `AuthenticatedContentIn` and a test-gated conversion to `AuthenticatedContent`.

openmls/src/group/tests/kat_transcript_hashes.rs
```diff
@@ -0,0 +1,264 @@
...
+    let ciphersuite = Ciphersuite::try_from(test_vector.cipher_suite).unwrap();
+    if backend.crypto().supports(ciphersuite).is_err() {
+        log::debug!("Skipping unsupported ciphersuite `{ciphersuite:?}`.");
+        return;
+    }
+
+    // Verification:
+    //
+    // Verify that `authenticated_content` contains a `Commit`, ...
+    let authenticated_content = AuthenticatedContent::from(
```
duesee, line 86: Here, we use the conversion from above now.

franziskuskiefer:
> Sorry for the force push but this just got too messy. I made sure to keep all suggestions.

I don't see why a force push was needed.
Anyway, I'll have to re-review everything now as I don't know what changed.

franziskuskiefer (approved):

PR #1315, opened 2023-03-08
spec: Make the ratchet tree non-malleable.

### LINKED ISSUE #1305: Make the ratchet tree non-malleable.
Opening in advance, see https://github.com/mlswg/mls-protocol/pull/868.

duesee:
This closes #1305.


openmls/src/extensions/mod.rs
```diff
@@ -514,7 +515,7 @@ mod test {
         // that all permutations keep their order after being (de)serialized.
         // The extension content does not matter in this test.
         let ext_x = Extension::ApplicationId(ApplicationIdExtension::new(b"Test"));
-        let ext_y = Extension::RatchetTree(RatchetTreeExtension::default());
+        let ext_y = Extension::ExternalPub(ExternalPubExtension::new(HpkePublicKey::new(vec![])));
```
duesee, line 518: This change was required because `RatchetTree{,Extension}` should never have implemented `Default`.

openmls/src/treesync/mod.rs
```diff
@@ -79,24 +81,77 @@ pub use node::{leaf_node::LeafNode, parent_node::ParentNode, Node};
...
-    Serialize,
-    Deserialize,
-    TlsSerialize,
-    TlsDeserialize,
-    TlsSize,
-)]
-// TODO(#1305): Make the ratchet tree non-malleable.
-//
-// * ValSemXXX: "If the tree has 2^d leaves, then it has 2^(d+1) - 1 nodes. The ratchet_tree vector logically has this number of entries, but the sender SHOULD NOT include blank nodes after the last non-blank node."
-// * ValSemXXX: "Regardless of how the client obtains the tree, the client MUST verify that the root hash of the ratchet tree matches the tree_hash of the GroupContext before using the tree for MLS operations."
```
duesee, line 97: I removed the second ValSem for now but we should get to this during valiadtion.

openmls/src/treesync/mod.rs
```diff
@@ -79,24 +81,77 @@ pub use node::{leaf_node::LeafNode, parent_node::ParentNode, Node};
...
+    /// The ratchet tree has a trailing blank node.
+    #[error("The ratchet tree has trailing blank nodes.")]
+    TrailingBlankNodes,
+}
+
+impl RatchetTree {
+    /// Create a [`RatchetTree`] from a vector of nodes stripping all trailing blank nodes.
+    ///
+    /// Note: The caller must ensure to call this with a vector that is *not* empty after removing all trailing blank nodes.
+    fn from_nodes(mut nodes: Vec<Option<Node>>) -> Self {
```
duesee, line 102: It is a bit unfortunate that we have two ways to create a `RatchetTree` for now. `TryFrom` fails on non-normalized `Vec`s, but this function does the normalization to the extend that empty vectors are accepted.

openmls/src/treesync/mod.rs
```diff
@@ -79,24 +81,77 @@ pub use node::{leaf_node::LeafNode, parent_node::ParentNode, Node};
...
+}
+
+impl TryFrom<Vec<Option<Node>>> for RatchetTree {
+    type Error = RatchetTreeError;
+
+    fn try_from(value: Vec<Option<Node>>) -> Result<Self, Self::Error> {
+        // ValSem300: "Exported ratchet trees must not have trailing blank nodes."
+        //
+        // We can check this by only looking at the last node (if any).
+        match value.last() {
```
duesee, line 127: "No trailing blank nodes" can be reduced to "the last node is not blank".

openmls/src/treesync/mod.rs
```diff
@@ -79,24 +81,77 @@ pub use node::{leaf_node::LeafNode, parent_node::ParentNode, Node};
...
+            Some(Some(_)) => {
+                // The ratchet tree is not empty, i.e., has a last node, and the last node is not blank.
+                Ok(Self(value))
+            }
+            Some(None) => {
+                // The ratchet tree is not empty, i.e., has a last node, *but* the last node *is* blank.
```
duesee, line 133: Arguably, `TryFrom` could already do the trimming.

franziskuskiefer (changes requested):
Thanks, one main question inline.

openmls/src/treesync/mod.rs
duesee: None of the functions accepts trailing blank nodes. The difference lies in the handling of *empty* vectors. `TryFrom` rejects them. `from_nodes` accepts them (because doing otherwise would lead to too much churn).

In any case, i think we do want to have both functions:

* During deserialization we always want to error-out when given trailing blank nodes. Thus, we have `TryFrom`.
* Internally, we want to have a constructor that accepts a `Vec<Option<Node>>`, does the trimming, and returns a valid `RatchetTree`. Thus, we have `RatchetTree::from_nodes`. (That I should probably rename to `from_nodes_trim`.)

duesee: Discussed off-GitHub.

openmls/src/treesync/mod.rs
```diff
@@ -79,24 +81,77 @@ pub use node::{leaf_node::LeafNode, parent_node::ParentNode, Node};
...
+                // We need to add 1 to `rightmost_nonempty_position` to keep the rightmost node.
+                nodes.resize(rightmost_nonempty_position + 1, None);
+            }
+            None => {
+                // If there is no rightmost non-blank node, the vector consist of blank nodes only.
+                nodes.clear();
+            }
+        }
+
+        debug_assert!(!nodes.is_empty(), "Caller should have ensured that `RatchetTree::from_nodes` is not called with a vector that is empty after removing all trailing blank nodes.");
```
franziskuskiefer, line 115:
```suggestion
        debug_assert!(!nodes.is_empty(), "Caller should have ensured that `RatchetTree::trimmed` is not called with a vector that is empty after removing all trailing blank nodes.");
```

franziskuskiefer (approved):
Thanks, let's get this in (one suggestion inline)

duesee:
Sorry, the applied suggestion dismissed the review.

franziskuskiefer (approved):

PR #1317, opened 2023-03-10
interop runner

franziskuskiefer:
This updates the interop harness to work with again.

It's not implementing all functions yet, but the implemented ones are working.

Note that I moved the protobuf definitions into mls-implementations crate to avoid getting out of sync (see `Cargo.toml`).


openmls/src/group/mls_group/creation.rs
```diff
@@ -97,6 +97,21 @@ impl MlsGroup {
...
+        backend: &impl OpenMlsCryptoProvider<KeyStoreProvider = KeyStore>,
+        mls_group_config: &MlsGroupConfig,
+        welcome: MlsMessageIn,
+        ratchet_tree: Option<RatchetTree>,
+    ) -> Result<Self, WelcomeError<KeyStore::Error>> {
+        let welcome = welcome
+            .into_welcome()
+            .ok_or(WelcomeError::MalformedWelcomeMessage)?;
+        Self::new_from_welcome(backend, mls_group_config, welcome, ratchet_tree)
+    }
```
raphaelrobert, line 113: Why do we need this in the public API?

openmls/src/treesync/node/leaf_node/capabilities.rs
```diff
@@ -2,6 +2,8 @@ use openmls_traits::types::Ciphersuite;
 use serde::{Deserialize, Serialize};
 use tls_codec::{TlsDeserialize, TlsSerialize, TlsSize};
 
+#[cfg(doc)]
+use super::LeafNode;
```
raphaelrobert, line 6: Given this is the only change in the file, is it really needed?

raphaelrobert (changes requested):
I only reviewed the parts that are not part of testing/interop, it looks like there are some leftovers.

openmls/src/treesync/node/leaf_node/capabilities.rs
franziskuskiefer: Building docs is broken right now. This fixes it. `LeafNode` is used in the comment in line 14.

openmls/src/group/mls_group/creation.rs
franziskuskiefer: Because the API as is isn't really usable. All of the API should operate on `MlsMessage`, something like `Welcome`.
I'm fine putting this behind `test-utils` until all of the API is rewritten. But this where it should go.

raphaelrobert: I agree the API needs changes with the new framing and I created #1326 to track that. Let's look at that separately though, because there might be alternatives to this, e.g. by remodeling `MlsMessageIn`.
So you could either put it behind a test flag or use `welcome.into_welcome().unwrap()` in the interop client. In either case please leave a TOFDO for #1326 somewhere.

Issue #1186, opened 2023-01-09
Implement `LeafNode` validation.


duesee:
7.3.  Leaf Node Validation
    
       The validity of a LeafNode needs to be verified at a few stages:
    
       *  When a LeafNode is downloaded in a KeyPackage, before it is used
          to add the client to the group
    
       *  When a LeafNode is received by a group member in an Add, Update,
          or Commit message
    
       *  When a client validates a ratchet tree, e.g., when joining a group
          or after processing a commit

* [x] openmls/annotations#8
  * [x] openmls/annotations#20
* [x] openmls/annotations#9
* [x] openmls/annotations#10
* [x] openmls/annotations#11
* [x] openmls/annotations#12
* [x] openmls/annotations#13
* [x] openmls/annotations#16
* [x] openmls/annotations#17
* [x] openmls/annotations#18
* [ ] openmls/annotations#30
* [x] openmls/annotations#31
* [ ] openmls/annotations#32
* [x] openmls/annotations#33

PR #1318, opened 2023-03-12
Prepare LeafNode verification

raphaelrobert:
This PR prepares `LeafNode` verification by introducing separate types for incoming messages. A few notes:
 - The PR is large, but the changes are merely refactoring. No logic changes were made.
 - The new types all have the `-In` suffix, just like `MessageIn`. Ultimately this trickles down to `AddProposalIn`, `UpdateProposalIn` and `UpdatePathIn` where the `LeafNode` verification will happen.
 - `TlsDeserialize` has been removed for all message structs, except for the new types introduced in this PR.
 - The new types are effectively a copy of the spec structs. Wrapping the existing structs would mean that `TlsSerialize` would have to be derived again, which I think is a bigger risk than keeping two copies of the wire format for select messages (especially since the spec is pretty frozen now).
  - In this PR, the new types are still converted to the original types. The actual verification will follow in another PR. It will be easier to review when this PR is in.


openmls/src/framing/mls_auth_content_in.rs
```diff
@@ -0,0 +1,356 @@
...
+            confirmation_tag.tls_serialize(writer)?
+        } else {
+            0
+        };
+        Ok(written)
+    }
+}
+
+// The following two `From` implementations break abstraction layers and MUST
+// NOT be made available outside of tests or "test-utils".
```
kkohbrok, line 324: Then there should probably be cfg statements for both of them, right? If not, can you move the TODO below up here?

openmls/src/framing/mls_content_in.rs
```diff
@@ -0,0 +1,353 @@
...
+            &self.content,
+            self.serialized_context.as_deref(),
+        )
+    }
+}
+
+// The following two `From` implementations break abstraction layers and MUST
+// NOT be made available outside of tests or "test-utils".
+
+// TODO: re-enable #[cfg(any(feature = "test-utils", test))]
```
kkohbrok, line 290: Can you add in issue number here if you don't want to re-enable it in this PR?

openmls/src/framing/public_message_in.rs
```diff
@@ -0,0 +1,325 @@
...
+use tls_codec::{
+    Deserialize as TlsDeserializeTrait, Serialize as TlsSerializeTrait, TlsDeserialize,
+    TlsSerialize, TlsSize,
+};
+
+/// Wrapper around a `Mac` used for type safety.
+#[derive(
+    Debug, PartialEq, Clone, Serialize, Deserialize, TlsSerialize, TlsDeserialize, TlsSize,
+)]
+pub(crate) struct MembershipTagIn(pub(crate) Mac);
```
kkohbrok, line 30: Do we have to have an -In type for MembershupTag? It doesn't seem to be used anywhere as far as I can tell.

openmls/src/framing/public_message_in.rs
```diff
@@ -0,0 +1,325 @@
...
+            0
+        };
+        Ok(written)
+    }
+}
+
+// The following two `From` implementations break abstraction layers and MUST
+// NOT be made available outside of tests or "test-utils".
+
+// TODO: Re-enable this #[cfg(any(feature = "test-utils", test))]
```
kkohbrok, line 305: Same as for the other instance of this comment.

openmls/src/messages/proposals_in.rs
```diff
@@ -0,0 +1,324 @@
...
+impl From<ProposalOrRefIn> for crate::messages::proposals::ProposalOrRef {
+    fn from(proposal: ProposalOrRefIn) -> Self {
+        match proposal {
+            ProposalOrRefIn::Proposal(proposal) => Self::Proposal(proposal.into()),
+            ProposalOrRefIn::Reference(reference) => Self::Reference(reference),
+        }
+    }
+}
+
+impl From<crate::messages::proposals::ProposalOrRef> for ProposalOrRefIn {
```
kkohbrok, line 313: Is this From implementation meant to go away at some point? I thought this was the direction where we don't wan to allow conversion.

openmls/src/treesync/treekem.rs
```diff
@@ -363,3 +361,72 @@ impl UpdatePath {
         self.nodes.push(last_node)
     }
 }
+
+#[derive(
+    Debug, PartialEq, Eq, Clone, Serialize, Deserialize, TlsSerialize, TlsDeserialize, TlsSize,
```
kkohbrok, line 366: This should not implement TlsSerialize, should it?

openmls/src/framing/mls_auth_content_in.rs
```diff
@@ -0,0 +1,356 @@
...
+/// ```c
+/// // draft-ietf-mls-protocol-17
+///
+/// struct {
+///     WireFormat wire_format;
+///     FramedContent content;
+///     FramedContentAuthData auth;
+/// } AuthenticatedContent;
+/// ```
+#[derive(PartialEq, Debug, Clone, TlsSerialize, TlsSize)]
```
kkohbrok, line 98: This should not implement TlsSerialize, should it?

openmls/src/framing/mls_content_in.rs
```diff
@@ -0,0 +1,353 @@
...
+///     opaque group_id<V>;
+///     uint64 epoch;
+///     Sender sender;
+///     opaque authenticated_data<V>;
+///
+///     // ... continued in [FramedContentBody] ...
+/// } FramedContent;
+/// ```
+#[derive(
+    Debug, PartialEq, Clone, Serialize, Deserialize, TlsSerialize, TlsDeserialize, TlsSize,
```
kkohbrok, line 38: Same as above. Since this has an -In prefix, I assume it shouldn't implement TlsSerialize.

kkohbrok (changes requested):
Thanks for pushing through this. This is a lot of new code and I hope we won't have too many more changes to the spec, or otherwise this will duplicate whatever work we have to make on the structs. But hopefully it's worth it for the type safety we get in return. Anyway. A few nits and questions. In particular, I noted on some -In suffixed structs that they probably shouldn't implement TlsSerialize, although that probably doesn't pose any particular danger.

openmls/src/framing/mls_content_in.rs
raphaelrobert: That's needed for `*Tbs` to be serialised. There's no issue with deriving `TlsSerialize`, except that it might be unnecessary in some cases perhaps.

openmls/src/treesync/treekem.rs
raphaelrobert: This one is needed again in a -Tbs somewhere else

openmls/src/messages/proposals_in.rs
raphaelrobert: This is the "safe" direction. I guess the goal will still be to get rid of this as well in #1186.

openmls/src/framing/public_message_in.rs
raphaelrobert: Not sure what the other instance is...

openmls/src/framing/mls_auth_content_in.rs
raphaelrobert: See below

kkohbrok (approved):

PR #1324, opened 2023-03-14
Cleanup after #1318

raphaelrobert:
Cleaning up after the code duplication of #1318.


duesee (dismissed):
Thanks for digging through this so fast! I think it would also help to remove the file-based split and have the types close by each other. We could also remove the documentation of one type and only refer to the other. But let's do one piece at a time.

raphaelrobert:
> Thanks for digging through this so fast! I think it would also help to remove the file-based split and have the types close by each other. We could also remove the documentation of one type and only refer to the other. But let's do one piece at a time.

We can certainly revisit how things are split once we have better validation. We might want to add more types (e.g. `RemoveProposalIn`) and then we need another structure anyway.

kkohbrok (approved):

PR #1328, opened 2023-03-20
refactor: Store PSKs independently from nonce.

duesee:
While it is certainly not the last PR that reworks PSK handling, it should unblock interoperability work. I made it a draft because there are a few `TODO`s left.

### DIFF
--- a/openmls/src/group/core_group/kat_passive_client.rs
+++ b/openmls/src/group/core_group/kat_passive_client.rs
@@ -7,10 +7,9 @@ use crate::{
     framing::{
         MlsMessageIn, MlsMessageInBody, MlsMessageOut, ProcessedMessageContent, TlsFromBytes,
     },
-    group::{config::CryptoConfig, core_group::GroupSecrets, *},
+    group::{config::CryptoConfig, *},
     key_packages::*,
-    prelude::ProtocolVersion,
-    schedule::psk::{PreSharedKeyId, PskBundle},
+    schedule::psk::PreSharedKeyId,
     test_utils::*,
     treesync::{
         node::encryption_keys::{EncryptionKeyPair, EncryptionPrivateKey},
@@ -141,35 +140,17 @@ struct PassiveClient {
 }
 
 impl PassiveClient {
-    fn new(
-        group_config: MlsGroupConfig,
-        psk_ids: Vec<PreSharedKeyId>,
-        psks: Vec<ExternalPskTest>,
-    ) -> Self {
+    fn new(group_config: MlsGroupConfig, psks: Vec<ExternalPskTest>) -> Self {
         let backend = OpenMlsRustCrypto::default();
 
         // Load all PSKs into key store.
-        for psk_id in psk_ids.into_iter() {
-            let psk_bundle = {
-                let secret = {
-                    let entry = psks
-                        .iter()
-                        .find(|s| s.psk_id == psk_id.external_id().unwrap())
-                        .unwrap();
-
-                    Secret::from_slice(
-                        &entry.psk,
-                        ProtocolVersion::default(),
-                        group_config.crypto_config.ciphersuite,
-                    )
-                };
-
-                PskBundle::new(secret).unwrap()
-            };
-
-            backend
-                .key_store()
-                .store(&psk_id.tls_serialize_detached().unwrap(), &psk_bundle)
+        for psk in psks.into_iter() {
+            // TODO: Better API?
+            // We only construct this to easily save the PSK in the keystore.
+            // The nonce is not saved, so it can be empty...
+            let psk_id = PreSharedKeyId::external(psk.psk_id, vec![]);
+            psk_id
+                .write_to_key_store(&backend, group_config.crypto_config.ciphersuite, &psk.psk)
                 .unwrap();
         }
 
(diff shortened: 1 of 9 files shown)


openmls/src/group/core_group/kat_passive_client.rs
```diff
@@ -141,35 +140,17 @@ struct PassiveClient {
...
-
-                PskBundle::new(secret).unwrap()
-            };
-
-            backend
-                .key_store()
-                .store(&psk_id.tls_serialize_detached().unwrap(), &psk_bundle)
+        for psk in psks.into_iter() {
+            // TODO: Better API?
+            // We only construct this to easily save the PSK in the keystore.
```
duesee, line 149: Feedback welcome.

openmls/src/schedule/psk.rs
```diff
@@ -178,30 +179,88 @@ pub struct PreSharedKeyId {
...
 
-    /// Return the PSK nonce
+    /// Return the PSK nonce.
     pub fn psk_nonce(&self) -> &[u8] {
         self.psk_nonce.as_slice()
     }
+
+    /// Save this `PreSharedKeyId` in the keystore.
+    ///
+    /// Note: The nonce is not saved as it must be present in the messages when processing PSKs.
```
franziskuskiefer, line 233: I'd say something along the line of... the reason is that it must be unique for each time it's being applied.

openmls/src/group/tests/test_commit_validation.rs
```diff
@@ -281,8 +279,9 @@ fn test_valsem201(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryptoProvider
             Psk::External(ExternalPsk::new(rand)),
         )
         .unwrap();
-        let psk_key = psk_id.tls_serialize_detached().unwrap();
-        backend.key_store().store(&psk_key, &psk_bundle).unwrap();
+        psk_id
+            .write_to_key_store(backend, ciphersuite, psk_bundle.secret().as_slice())
```
duesee, line 283: Note to myself: Remove PskBundle.

openmls/src/schedule/psk.rs
```diff
@@ -178,30 +179,88 @@ pub struct PreSharedKeyId {
...
+        let psk_id_with_empty_nonce = {
+            let mut psk_id = self.clone();
+            psk_id.psk_nonce = VLBytes::new(vec![]);
+            psk_id
+        };
+
+        psk_id_with_empty_nonce
+            // TODO: Just rely on tls-codec not being broken?
+            .tls_serialize_detached()
+            .map_err(|_| LibraryError::custom("tls-codec failed."))
```
franziskuskiefer, line 262: tls encoding errors are mostly wrapped as missing length checks by now, because it should be the only reason this can happen. `LibraryError::missing_bound_check`

openmls/src/schedule/psk.rs
```diff
@@ -178,30 +179,88 @@ pub struct PreSharedKeyId {
...
+    /// Save this `PreSharedKeyId` in the keystore.
+    ///
+    /// Note: The nonce is not saved as it must be present in the messages when processing PSKs.
+    pub fn write_to_key_store<KeyStore: OpenMlsKeyStore>(
+        &self,
+        backend: &impl OpenMlsCryptoProvider<KeyStoreProvider = KeyStore>,
+        ciphersuite: Ciphersuite,
+        psk: &[u8],
+    ) -> Result<(), KeyStore::Error> {
+        // TODO: Introduce PskKeyStoreError? -> Just rely on tls-codec not being broken?
```
duesee, line 240: Feedback welcome.

openmls/src/schedule/psk.rs
```diff
@@ -178,30 +179,88 @@ pub struct PreSharedKeyId {
...
+
+    /// Save this `PreSharedKeyId` in the keystore.
+    ///
+    /// Note: The nonce is not saved as it must be present in the messages when processing PSKs.
+    pub fn write_to_key_store<KeyStore: OpenMlsKeyStore>(
+        &self,
+        backend: &impl OpenMlsCryptoProvider<KeyStoreProvider = KeyStore>,
+        ciphersuite: Ciphersuite,
+        psk: &[u8],
+    ) -> Result<(), KeyStore::Error> {
```
franziskuskiefer, line 239: This should probably retunr a `PskError`, which is either a key store error or a library error due to the serialization.

openmls/src/schedule/psk.rs
```diff
@@ -178,30 +179,88 @@ pub struct PreSharedKeyId {
...
+
+            PskBundle { secret }
+        };
+
+        backend.key_store().store(&keystore_id, &psk_bundle)
+    }
+
+    pub(crate) fn keystore_id(&self) -> Result<Vec<u8>, LibraryError> {
+        let psk_id_with_empty_nonce = {
+            let mut psk_id = self.clone();
```
franziskuskiefer, line 254: Creating a new `PreSharedKeyId` and only `clone` the `psk` field requires fewer allocations. But it's probably not critical.

openmls/src/schedule/psk.rs
```diff
@@ -147,6 +147,7 @@ impl ResumptionPsk {
 pub enum Psk {
     #[tls_codec(discriminant = 1)]
     External(ExternalPsk),
+    #[tls_codec(discriminant = 2)]
```
franziskuskiefer, line 150: Let's not introduce unnecessary code.

franziskuskiefer (changes requested):
I only looked at `pskr.rs` for now.

openmls/src/group/core_group/kat_passive_client.rs
franziskuskiefer: The only thin here I dislike is that we need to provide an empty nonce. What we really do need here, is to serialize the `Psk`, which is the `PskId` without the nonce. Maybe we shold operate on the `Psk` struct for this type of API instead.
We can do (something like) this in a follow up as well.

franziskuskiefer (approved):
Thanks. I left a comment for a potential follow up.

PR #1333, opened 2023-03-22
Ensure to fully consume TLS input (in tests).

### LINKED ISSUE #1332: Ensure to fully consume TLS messages in tests.
Generally looks good to me. I left some comments inline

Let's skip 45a78063f469f40d47861c58a82e78a0e12b9399. If that's something we want to do, it should happen in a separate PR.

_Originally posted by @franziskuskiefer in https://github.com/openmls/openmls/pull/1331#pullrequestreview-1352161813_

duesee:
This closes #1332.


openmls/src/framing/mod.rs
```diff
@@ -62,17 +65,14 @@ pub(crate) mod public_message_in;
 pub(crate) mod sender;
 pub(crate) mod validation;
 pub(crate) use errors::*;
-
```
franziskuskiefer, line 65: Why did you remove the empty lines here? This makes this way less readable.

openmls/src/framing/mod.rs
```diff
@@ -200,14 +200,14 @@ where
     ///
     /// Note: This function makes sure that the input is fully consumed, i.e., that there is no
     ///       trailing data.
-    fn try_from_bytes(bytes: impl AsRef<[u8]>) -> Result<Self, TlsFromBytesError>;
+    fn tls_deserialize_complete(bytes: impl AsRef<[u8]>) -> Result<Self, TlsFromBytesError>;
```
franziskuskiefer, line 203: Why the rename? I don't think `tls_deserialize_complete` is what we want. Staying in the Rust terminology we could call it `_exact`.

franziskuskiefer (changes requested):
Thanks, a few nits and there's a conflict.

franziskuskiefer (dismissed):
Thanks!

openmls/src/group/core_group/kat_passive_client.rs
```diff
@@ -155,10 +155,10 @@ pub fn run_test_vector(test_vector: PassiveClientWelcomeTestVector) {
     let ratchet_tree: Option<RatchetTree> = test_vector
         .ratchet_tree
         .as_ref()
-        .map(|bytes| RatchetTree::tls_deserialize(&mut bytes.0.as_slice()).unwrap());
+        .map(|bytes| RatchetTree::tls_deserialize_exact(&mut bytes.0.as_slice()).unwrap());
```
franziskuskiefer, line 158: clippy was unhappy about this.

```suggestion
        .map(|bytes| RatchetTree::tls_deserialize_exact(bytes.0.as_slice()).unwrap());
```

franziskuskiefer (approved):

PR #1340, opened 2023-04-05
Duesee/make interop work 1

duesee:
This PR introduces some fixes in OpenMLS to interop with other implementations. There is one test failing and I can't really trace down what the cause is. At some point Bob and the Public Group disagree about the confirmed_transcript_hash (`b''`) for public group, something for Bob. But I'm a bit puzzled about this because it appears to me that `b''` is correct. But this would mean Bob or Alice do something wrong, which would mean that interop shouldn't work. I'll need some more time (and better tracing) to figure this out.

### DIFF
--- a/interop_client/Cargo.toml
+++ b/interop_client/Cargo.toml
@@ -6,6 +6,10 @@ edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[features]
+# Forward feature to openmls.
+crypto-debug = ["openmls/crypto-debug"]
+
 [dependencies]
 openmls = { path = "../openmls", features = ["test-utils"]}
 openmls_rust_crypto = { path = "../openmls_rust_crypto" }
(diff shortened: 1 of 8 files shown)


franziskuskiefer:
> At some point Bob and the Public Group disagree about the confirmed_transcript_hash (`b''`) for public group, something for Bob. But I'm a bit puzzled about this because it appears to me that `b''` is correct.

The transcript hash is only empty in epoch 0. The error appears to occur later and the public group is not updating it's group context correctly.

duesee:
> The transcript hash is only empty in epoch 0. The error appears to occur later and the public group is not updating it's group context correctly.

Do you have time to take a fresh look on it? Maybe I'm just overlooking somehting. In the meantime I could continue to work on `duesee/make_interop_work_2`. When we get these things merged, OpenMLS would almost interoperate on all scenarios.

franziskuskiefer (commented):
Looks like we need for this.
Reverting the changes in `staged_commit.rs` and `group_info.rs` makes thing work again and I'm not sure these changes are correct.
So we need a test that requires these changes here to go through to figure out why/if they are needed.

franziskuskiefer (approved):
Thanks, let's get this in

Issue #1337, opened 2023-03-30
[MLS Spec change] Implement GREASE support (1/2, Receiving).


duesee:
In order to interoperate with other implementations, OpenMLS needs to GREASE. (See https://www.ietf.org/archive/id/draft-ietf-mls-protocol-20.html#section-13.5.)

# TODO

* [ ] LeafNode.capabilities.ciphersuites
* [ ] LeafNode.capabilities.extensions
* [ ] LeafNode.capabilities.proposals
* [ ] LeafNode.capabilities.credentials
* [ ] LeafNode.extensions
* [ ] KeyPackage.extensions
* [ ] GroupInfo.extensions

# Related

* [ ] #1311 

It might be worth it to already implement this in `tls_codec`.

duesee:
I already implemented support for unknown types and extensions during interop. However, in order to be a good Internet citizen, OpenMLS should not only accept unknown values but GREASE itself. I propose to split this issue to track GREASE-receive and GREASE-send.

PR #1342, opened 2023-04-06
feat/fix: Handle unknown values + fix extension reordering.

duesee:
This PR makes it so that OpenMLS can handle unknown values as described in #1337. However, it has the issue that it calls `panic!()` in certain functions. The correct thing to do would be to return a `CryptoError`. However, this change has a huge impact on the rest of the codebase because we must `unwrap()` in almost all tests, and introduce many `map_err()?`s. I experimented with this a bit and it bloats the code by a lot.

I would propose some alternative solutions but we better discuss them personally before I start with the implementation.


traits/src/types.rs
```diff
@@ -378,6 +392,8 @@ impl Ciphersuite {
             Ciphersuite::MLS_256_DHKEMX448_AES256GCM_SHA512_Ed448
             | Ciphersuite::MLS_256_DHKEMP521_AES256GCM_SHA512_P521
             | Ciphersuite::MLS_256_DHKEMX448_CHACHA20POLY1305_SHA512_Ed448 => HashType::Sha2_512,
+            // TODO
+            Ciphersuite::UNKNOWN(_) => panic!("This should not be called."),
```
duesee, line 396: Do not `panic!`.

duesee:
I'll close this PR so that we can tackle the ciphersuite separately.

Issue #1311, opened 2023-03-06
[Feature request] Allow opaque extensions in the custom extension range


kkohbrok:
We currently require extensions to be part of an enum that lives within OpenMLS. It would be great if we would support opaque extensions that are opaque to OpenMLS, but that are accessible to the application. We should restrict this to the area of extension ids that is reserved for private use.

duesee:
Closed by #1343.

PR #1343, opened 2023-04-06
feat/fix: Handle unknown extensions and do not reorder them.

duesee:
This PR makes it so that unknown extensions are accepted and kept as opaque bytes. I'm not sure if it fully captures the idea behind https://github.com/openmls/openmls/issues/1311 but it probably closes it.


OtaK:
Looking at the changes, this is not making sure that the `Unknown` extensions are within the "private reserved" range `0xFF00..=0xFFFF` (or GREASE range) so it seems to allow out-of-spec values. I might have missed something but I couldn't see a range check.

duesee:
> Looking at the changes, this is not making sure that the `Unknown` extensions are within the "private reserved" range `0xFF00..=0xFFFF` (or GREASE range) so it seems to allow out-of-spec values. I might have missed something but I couldn't see a range check.

Yes, you are right. These changes make it so that OpenMLS will accept any (unknown) extension that is well-formed regarding to the TLS presentation language no matter what extension type is used as `Unknown`. But I think this is the correct thing to do.

We want to make sure that neither an unknown value in the "public range", nor an unknown value in the "private range" will lead to deserialization errors. That's why the MLS draft defines GREASE values in both ranges. Handling GREASE values separately is no option because "since this negates their utility for detecting extensibility failures".

We *could* handle private extensions differently to `Unknown` but I'm not sure why we would want to do that. Maybe I'm overlooking something?

duesee:
A reason to already separate `Unknown` and `Private` could be to provide cleaner APIs that allow to construct private extensions but not unknown ones. But given that nothing of this is in place currently, maybe it's something we should tackle later?

openmls/src/extensions/codec.rs
```diff
@@ -115,6 +111,9 @@ impl Deserialize for Extension {
             ExtensionType::ExternalSenders => Extension::ExternalSenders(
                 ExternalSendersExtension::tls_deserialize(&mut extension_data)?,
             ),
+            ExtensionType::Unknown(unknown) => {
+                Extension::Unknown(unknown, UnknownExtension(extension_data.to_vec()))
+            }
         })
     }
 }
```
franziskuskiefer, line 119: Let's add at least one short test here to make sure (de)serializing of unknown extensions works.

franziskuskiefer (changes requested):
Thanks, lgtm with a small test to make sure it is working as intended.

franziskuskiefer (approved):
Thanks

PR #1345, opened 2023-04-11
interop: Implement support for unknown `{Credential,Proposal}Type`s.

duesee:
This PR adds the `{Credential,Proposal}Type::Unknown` variants. I think this is all that is needed here. It must not happen that we receive an actual proposal that we don't understand. If it would, there would be no way to parse it. This is different for extensions as they are designed to have an easily "skippable" payload.


franziskuskiefer (approved):
thanks

Issue #1210, opened 2023-01-18
Implement `KeyPackage` validation.


duesee:
```
The validity of a KeyPackage needs to be verified at a few stages:

* When a KeyPackage is downloaded by a group member, before it is used to add the client to the group
* When a KeyPackage is received by a group member in an Add message
```

* [x] openmls/annotations#48
  * [x] openmls/annotations#39

---

* [x] openmls/annotations#49
  * [x] openmls/annotations#45

---

* [x] openmls/annotations#40
* [x] openmls/annotations#41
* [x] openmls/annotations#42
* [x] openmls/annotations#43
* [x] openmls/annotations#44
* [x] openmls/annotations#46
* [x] openmls/annotations#47

PR #1346, opened 2023-04-11
Introduce KeyPackageIn & LeafNodeIn

### LINKED ISSUE #1242: validation: Do not implement `TlsDeserialize` for `LeafNode`.
This can be used to bypass validation but is not easy to fix right now. This issue serves as a reminder.

### LINKED ISSUE #1289: Add leaf node signature verification to tree validation KAT

raphaelrobert:
- Works towards #1186 & #1210 and introduces signature verification for KeyPackages and LeafNodes.
 - Key store no longer requires TLS serialization
 - Removes OpenMlsLeafNode
 - Fixed a bug where the LeafNode in the UpdatePath was not correctly signed
 - Fixed a bug where the leaf node source was not set correctly

 
Fixes #1242.
Fixes #1289.


openmls/src/framing/mls_content_in.rs
```diff
@@ -41,10 +44,29 @@ pub(crate) struct FramedContentIn {
     pub(super) epoch: GroupEpoch,
     pub(super) sender: Sender,
     pub(super) authenticated_data: VLBytes,
-
     pub(super) body: FramedContentBodyIn,
 }
 
+impl FramedContentIn {
+    /// Returns a [`FramedContent`] after successful validation.
+    pub(crate) fn into_validated(
```
kkohbrok, line 52: `into_validated` generally sounds a bit clunky. Maybe just `validate` in the same vein as `verify`?

openmls/src/key_packages/key_package_in.rs
```diff
@@ -0,0 +1,196 @@
...
+            credential,
+            signature_key,
+        }
+    }
+
+    /// Verify that this key package is valid:
+    /// * verify that the signature on this key package is valid
+    /// * verify that the signature on the leaf node is valid
+    /// * verify that all extensions are supported by the leaf node
+    /// * make sure that the lifetime is valid Returns a [`KeyPackage`] after
```
kkohbrok, line 65: There's a line break missing after `valid`.

openmls/src/key_packages/key_package_in.rs
```diff
@@ -0,0 +1,196 @@
...
+        let key_package = KeyPackage {
+            payload: KeyPackageTBS {
+                protocol_version: self.payload.protocol_version,
+                ciphersuite: self.payload.ciphersuite,
+                init_key: self.payload.init_key,
+                leaf_node,
+                extensions: self.payload.extensions,
+            },
+            signature: self.signature,
+        };
```
kkohbrok, line 106: I'm trying to figure out what a good pattern would be here. I would expect `KeyPackageTbsIn` to implement `into_validated` as well, which then validates its own fields (extensions, leaf node). I'm not sure what should happen next, but it would be great if we could call `verify` instead of `verify_no_out` on the KeyPackage in the end and just return the result.

One way to do this would be to build a `VerifiableKeyPackage` from the (validated) KeyPackageTbs and then validate that. That would be a little cleaner, but would also require another struct.

openmls/src/messages/mod.rs
```diff
@@ -181,11 +186,62 @@ impl CommitIn {
             }
         })
     }
-}
 
-// TODO #1186: The following must be removed once the validation refactoring is
-// complete.
+    /// Returns a [`FramedCommitContentBody`] after successful validation.
```
kkohbrok, line 190: The function signature indicates that instead returns a `Commit`.

openmls/src/messages/proposals_in.rs
```diff
@@ -100,24 +131,29 @@ impl ProposalIn {
...
+        let key_package = self.key_package.into_validated(crypto, ciphersuite)?;
+        Ok(AddProposal { key_package })
     }
 }
 
 /// Update Proposal.
 ///
 /// An Update proposal is a similar mechanism to [`AddProposalIn`] with the distinction that it
-/// replaces the sender's [`LeafNode`] in the tree instead of adding a new leaf to the tree.
+/// replaces the sender's LeafNodeIn in the tree instead of adding a new leaf to the tree.
```
kkohbrok, line 156:
```suggestion
/// replaces the sender's leaf node instead of adding a new leaf to the tree.
```

openmls/src/treesync/mod.rs
```diff
@@ -627,26 +712,4 @@ mod test {
     fn test_ratchet_tree_internal_empty_after_trim() {
         RatchetTree::trimmed(vec![None]);
     }
-
-    #[test]
-    fn test_ratchet_tree_trailing_blank_nodes() {
```
kkohbrok, line 632: Why was this removed? I remember there was a discussion on trailing blank nodes, but I can't remember the outcome. Are they allowed now?

kkohbrok (changes requested):
Congrats on finishing up this one. Just a few comments, then we can it in.

openmls/src/treesync/mod.rs
raphaelrobert: It's back now below.

openmls/src/framing/mls_content_in.rs
raphaelrobert: I'm following the convention here: https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

openmls/src/key_packages/key_package_in.rs
raphaelrobert: The current pattern just doesn't work very well for this case. I figured it was ok to do it like that since all of it happens within the `into_validated()` function.

openmls/src/framing/mls_content_in.rs
kkohbrok: I'm aware of those conventions, but I still think it's clunky and is not consistent with the existing use of `verify`. What I'm getting at is there in this particular instance, the point to is not to highlight the ownership, but the fact that we validate the given struct. In that sense, I believe that `validate` gives the reader a better idea of what's going on.

kkohbrok (approved):
Thanks!

PR #1347, opened 2023-04-12
interop/fix: Introduce (minimal) `VerifiableCiphersuite` + ratchet tree fix in interop runner.

duesee:
This PR makes it so that we have `VerifiableCiphersuite`s in `Capabilities` only. We may want to change that later but for now it unblocks interoperability with implementations that support GREASE. It also introduces a fix in the interop runner (ratchet tree from request was not processed) and improves readability a bit. I would recommend to review the commits separately.


franziskuskiefer (approved):
:shipit:

PR #1348, opened 2023-04-13
interop: Add some fixes to interop runner + chore.

duesee:
This PR improves the readability of the interop runner, improves the logging, adjusts AD, and adjusts the config so that the interop scenarios can pass. It is unfortunately a bit difficult to review due to some renamings. I think these are an improvement but could understand if we want to revert. The motivation behind the changes is that we can now see very well where a specifc log item is coming from. This proved to be very helpful in tracing down the cause of various mismatches.

This PR is a basis for the next ones that implement support for PSK, external join, and fix a transcript hash error. So, I hope it is one of the last PRs in this series.

### DIFF
--- a/interop_client/Cargo.toml
+++ b/interop_client/Cargo.toml
@@ -24,6 +24,6 @@ clap_derive = "4.1"
 serde = { version = "^1.0", features = ["derive"] }
 serde_json = "^1.0"
 tls_codec = { workspace = true }
-pretty_env_logger = "0.4"
 openmls_basic_credential = { version = "0.1.0", path = "../basic_credential" }
-log = { version = "0.4", features = ["std"] }
+tracing = "0.1.37"
+tracing-subscriber = { version = "0.3.16", features = ["env-filter"] }
--- a/interop_client/Cargo.toml
+++ b/interop_client/Cargo.toml
@@ -26,4 +26,4 @@ serde_json = "^1.0"
 tls_codec = { workspace = true }
 openmls_basic_credential = { version = "0.1.0", path = "../basic_credential" }
 tracing = "0.1"
-tracing-subscriber = { version = "0.3.16", features = ["env-filter"] }
+tracing-subscriber = { version = "0.3", features = ["env-filter"] }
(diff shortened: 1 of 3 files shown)


interop_client/Cargo.toml
```diff
@@ -24,6 +24,6 @@ clap_derive = "4.1"
 serde = { version = "^1.0", features = ["derive"] }
 serde_json = "^1.0"
 tls_codec = { workspace = true }
-pretty_env_logger = "0.4"
 openmls_basic_credential = { version = "0.1.0", path = "../basic_credential" }
-log = { version = "0.4", features = ["std"] }
+tracing = "0.1.37"
```
franziskuskiefer, line 28:
```suggestion
tracing = "0.1"
```

interop_client/src/main.rs
```diff
@@ -626,30 +706,36 @@ impl MlsClient for MlsClientImpl {
         // Store the proposal for potential future use.
         interop_group.messages_out.push(proposal.clone().into());
 
-        // log::trace!("   proposal: {proposal:#x?}");
+        // trace!("   proposal: {proposal:#x?}");
```
franziskuskiefer, line 709: remove commented code.

interop_client/src/main.rs
```diff
@@ -779,14 +877,12 @@ impl MlsClient for MlsClientImpl {
 
         // TODO #692: The interop client cannot process these proposals yet.
```
franziskuskiefer, line 878: Can this be resolved?

franziskuskiefer (changes requested):
Thanks, the trait changed. This needs the `free` function to compile again.

interop_client/src/main.rs
duesee: It will in the next PR.

duesee:
The last commit makes it very obvious that we probably want to rethink the default config values (or refactor the test runner). But I think we should not delay this more and refactor later.

franziskuskiefer (approved):
Thanks, lets get this in

PR #1350, opened 2023-04-17
Fix build break and warnings in interop_client

brianathere:
Fix build break in how log is being used (introduced by merge conflict). 
Fix warning about unused variable by logging it like other methods do.

### DIFF
--- a/interop_client/src/main.rs
+++ b/interop_client/src/main.rs
@@ -174,6 +174,7 @@ impl MlsClient for MlsClientImpl {
         request: Request<SupportedCiphersuitesRequest>,
     ) -> Result<Response<SupportedCiphersuitesResponse>, Status> {
         let request = request.get_ref();
+        info!(?request, "Request");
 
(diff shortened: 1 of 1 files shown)


interop_client/src/main.rs
```diff
@@ -174,6 +174,7 @@ impl MlsClient for MlsClientImpl {
         request: Request<SupportedCiphersuitesRequest>,
     ) -> Result<Response<SupportedCiphersuitesResponse>, Status> {
         let request = request.get_ref();
+        info!(?request, "Request");
```
franziskuskiefer, line 177: Can you instead remove the line above and make the argument a `_request`? This request is always empty so there's nothing to log here.

franziskuskiefer (changes requested):
Thanks! lgtm with a small nit.
We will get this on CI soon to avoid this type of issue.

franziskuskiefer (approved):
Thanks

PR #1351, opened 2023-04-18
Work towards interop.

duesee:
This PR has almost everything that we need to interop with the current set of live scenarios. However, there is a confirmation tag mismatch that we need to debug. I do have a "fix" for that but we were not sure it is correct given that it breaks a test. Let's do that in a final follow-up.


openmls/src/group/core_group/new_from_external_init.rs
```diff
@@ -108,6 +108,8 @@ impl CoreGroup {
             group_epoch_secrets,
             message_secrets_store,
             own_leaf_index,
+            // TODO
```
franziskuskiefer, line 111: TODO what? Either add an issue for the todo so everyone knows what's needed, or remove.

openmls/src/schedule/psk.rs
```diff
@@ -458,3 +464,83 @@ impl From<Secret> for PskSecret {
...
+) -> Result<Vec<(&'p PreSharedKeyId, Secret)>, PskError> {
+    let mut psk_bundles = Vec::new();
+
+    for psk_id in psk_ids.iter() {
+        log_crypto!(trace, "PSK store {:?}", resumption_psk_store);
+
+        match &psk_id.psk {
+            Psk::Resumption(resumption) => {
+                if let Some(psk_bundle) = resumption_psk_store.get(resumption.psk_epoch()) {
+                    // TODO
```
franziskuskiefer, line 481: TODO what?

openmls/src/schedule/psk.rs
```diff
@@ -458,3 +464,83 @@ impl From<Secret> for PskSecret {
...
+                    return Err(PskError::KeyNotFound);
+                }
+            }
+        }
+    }
+
+    Ok(psk_bundles)
+}
+
+// ----- ResumptionPskStore ------------------------------------------------------------------------
```
franziskuskiefer, line 500: Please move the `ResumptionPskStore` in it's own module instead.

openmls/src/schedule/psk.rs
```diff
@@ -458,3 +464,83 @@ impl From<Secret> for PskSecret {
...
+
+/// Resumption psks store. This is where the resumption psks are kept in a
+/// rollover list.
+#[derive(Debug, Serialize, Deserialize)]
```
franziskuskiefer, line 504: As far as I can tell this code was only moved, so nothing to touch here, but this store needs to be updated.

I don't think that this should be serialised. This way we end up writing secrets to the disk in an unprotected way. But the secrets should be in the key store anyway such that there should be no need to serialise these within the group itself.

The behaviour of the `add` function is pretty difficult to understand for callers because they doesn't know whether they removed a value or not. You could solve this by returning the optional old value.

openmls/src/group/tests/test_proposal_validation.rs
```diff
@@ -1918,63 +1918,64 @@ fn test_valsem401_valsem402(ciphersuite: Ciphersuite, backend: &impl OpenMlsCryp
...
-            )],
-            ProcessMessageError::InvalidCommit(StageCommitError::ProposalValidationError(
-                ProposalValidationError::Psk(PskError::UsageMismatch {
-                    allowed: vec![ResumptionPskUsage::Application],
-                    got: ResumptionPskUsage::Branch,
-                }),
-            )),
-        ),
+    let bad_psks: [(Vec<PreSharedKeyId>, ProcessMessageError); 0] = [
+        // TODO: This is currently not tested because we can't easily create invalid commits.
```
franziskuskiefer, line 1922: Can you add an issue?

openmls/src/utils.rs
```diff
@@ -20,10 +20,12 @@ pub fn random_u8() -> u8 {
     b[0]
 }
 
-#[cfg(test)]
-pub(crate) fn zero(length: usize) -> Vec<u8> {
-    vec![0u8; length]
-}
+// TODO: This is currently only used in deactivated test code.
```
franziskuskiefer, line 23: Lets just delete it.

franziskuskiefer (commented):
I had a first look and it generally looks good to me with some nits.
But there's some code from me in here as well. @kkohbrok can you do a full review?

I'm not requesting changes to not block an r+ from @kkohbrok.

openmls/src/group/tests/test_proposal_validation.rs
duesee: I opened #1354.

openmls/src/schedule/psk.rs
duesee: This was a leftover. Deleted.

openmls/src/schedule/psk.rs
duesee: I filed #1355 and #1356.

openmls/src/schedule/psk.rs
```diff
@@ -458,3 +464,83 @@ impl From<Secret> for PskSecret {
         Self { secret }
     }
 }
+
+pub(crate) fn load_psks<'p>(
```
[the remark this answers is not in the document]
duesee: This function couples two (currently rather unrelated) things: external PSKs (that are "sideloaded") and resumption PSKs (that are handled internally). So, I think the `ResumptionPskStore` is not the correct type that should implement `load_psks`. We could make it so that the `PskSecret` has a `load` method, similar to the previous `new`, but I would prefer to make this change in a follow-up.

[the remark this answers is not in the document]
franziskuskiefer: Ok, let's leave this to a follow-up.

franziskuskiefer (approved):
Let's get this in.
There is cleanup for the API to be done, but that's not for this PR.

PR #1358, opened 2023-04-19
ci: Build workspace.

duesee:
This changes the CI so that we build all members in the workspace with `--all-targets` (i.e., `--lib --bins --tests --benches --examples`) on ubuntu-latest. 

Previously we build `cli`, `ds-lib`, and `mls-ds` but, for example, not the `interop_client` (see https://github.com/openmls/openmls/pull/1350#pullrequestreview-1389406828) on ubuntu-latest and macos-latest.


.github/workflows/build_workspace.yml
```diff
@@ -16,18 +16,11 @@ env:
...
 jobs:
   tests:
-    strategy:
-      fail-fast: false
-      matrix:
-        os:
-          - macos-latest
-          - ubuntu-latest
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
```
franziskuskiefer, line 19: Why did you remove macos? We should keep building on at least ubuntu and macos.

.github/workflows/ds.yml
```diff
@@ -1,36 +0,0 @@
...
-          - windows-latest
-    runs-on: ${{ matrix.os }}
-    steps:
-      - uses: actions/checkout@v3
-        with:
-          ref: ${{ github.event.pull_request.head.sha }}
-      - uses: dtolnay/rust-toolchain@stable
-      - uses: Swatinem/rust-cache@v2
-      - name: Test DS lib
-        run: cargo test --verbose -p ds-lib
```
franziskuskiefer, line 34: We don't run these tests anymore. Not sure if that's a big deal for DS though.

franziskuskiefer (changes requested):
Thanks, one nit and looks like the CI is failing.

.github/workflows/ds.yml
duesee: Good point. As discussed off-GitHub I'll add a `cargo test` afterwards.

franziskuskiefer (commented):
Thanks, what are the crash files in `fuzz`?

duesee:
> Thanks, what are the crash files in fuzz?

Forgot to add `--exclude` and libFuzzer doesn't like to be called through `cargo test`. Removed the files.

franziskuskiefer (approved):
Thanks!

PR #1359, opened 2023-04-20
Make interop work.

duesee:
This is the final PR that makes interop (modulo some unsupported features) work.

### DIFF
--- a/openmls/src/group/core_group/mod.rs
+++ b/openmls/src/group/core_group/mod.rs
@@ -32,15 +32,6 @@ mod test_past_secrets;
 #[cfg(test)]
 mod test_proposals;
 
-use super::builder::TempBuilderPG1;
-use super::errors::CreateCommitError;
-
-use self::create_commit_params::{CommitType, CreateCommitParams};
-#[cfg(test)]
-use super::errors::CreateGroupContextExtProposalError;
-use super::public_group::diff::compute_path::PathComputationResult;
-use super::public_group::PublicGroup;
-use crate::binary_tree::array_representation::TreeSize;
 #[cfg(test)]
 use std::io::{Error, Read, Write};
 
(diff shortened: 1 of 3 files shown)


openmls/src/group/core_group/mod.rs
```diff
@@ -5,42 +5,6 @@
 //! This means that some functions that are not expected to fail and throw an
 //! error, will still return a `Result` since they may throw a `LibraryError`.
 
-// Private
```
duesee, line 8: This is only a reformat + an addition of `use log::`.

openmls/src/group/public_group/mod.rs
```diff
@@ -79,7 +79,20 @@ impl PublicGroup {
         group_context: GroupContext,
         initial_confirmation_tag: ConfirmationTag,
     ) -> Self {
-        let interim_transcript_hash = vec![];
+        let backend = OpenMlsRustCrypto::default();
```
duesee, line 82: This is the important change.

openmls/src/group/public_group/mod.rs
```diff
@@ -79,7 +79,20 @@ impl PublicGroup {
         group_context: GroupContext,
         initial_confirmation_tag: ConfirmationTag,
     ) -> Self {
-        let interim_transcript_hash = vec![];
```
duesee, line 82: This is the important change.

openmls/src/group/public_group/mod.rs
```diff
@@ -131,9 +144,7 @@ impl PublicGroup {
 
         let group_context = GroupContext::from(group_info.clone());
 
-        let interim_transcript_hash = if group_context.epoch() == GroupEpoch::from(0) {
-            vec![]
```
duesee, line 135: This is the important change.

openmls/src/group/public_group/mod.rs
```diff
@@ -79,7 +79,20 @@ impl PublicGroup {
...
+        let interim_transcript_hash = {
+            let input = InterimTranscriptHashInput::from(&initial_confirmation_tag);
+
+            input
+                .calculate_interim_transcript_hash(
+                    crypto,
+                    group_context.ciphersuite(),
+                    group_context.confirmed_transcript_hash(),
+                )
+                .unwrap()
```
duesee, line 94: We will need to handle this properly. As soon as we agree that this is what we want, I can make this method take a `backend` and return a `Result`.

franziskuskiefer (approved):
Thanks, gltm

PR #1361, opened 2023-04-21
feat: Implement Group Context Extensions proposal support

### LINKED ISSUE #1238: Apply GroupContextExtension (GCE) proposals
As pointed out by @augustocdias, we currently don't actually apply GCE proposals. As @beltram pointed out, he implemented this already in https://github.com/wireapp/openmls/pull/7.

augustocdias:
Closes #1238, #566

Adding support to make GroupContextExtensions proposals.

As discussed in Zulip, there's the leaf of the group creator has no Capability, so I've created a new config to set extensions to the leaf of the creator in the group builder. It uses the `RequiredCapabilitiesExtension` to fill the capabilities of the leaf. As for draft 20 it seems that leaves should have `Capabilities` and the only extension should be the `ApplicationId`. I believe this should be refactored later to set only the capabilities and the application id. Since there's no way to set the capabilities at the moment, I thought of using the  `RequiredCapabilitiesExtension`. 

The function to create the proposal is now not only for tests and I've added a parameter to ignore remove proposals, since removed members don't have to have their capabilities checked. 

I've also created a function to inline the proposal and make the commit in the same fashion as the `add_members`, `leave_group`, etc.


openmls/src/group/mls_group/extension.rs
```diff
@@ -0,0 +1,103 @@
+//! MLS group context extensions
+//!
+//! Contains all the method related to modifying a group's extensions
```
duesee, line 3:
```suggestion
//! Contains all the methods related to modifying a group's extensions.
```

openmls/src/group/mls_group/extension.rs
```diff
@@ -0,0 +1,103 @@
...
+    prelude::{
+        create_commit_params::CreateCommitParams, hash_ref::ProposalRef,
+        ProposeGroupContextExtensionError,
+    },
+};
+
+use super::*;
+
+impl MlsGroup {
+    /// Creates proposals to update extensions of the group. This replaces the existing extensions
```
duesee, line 20:
```suggestion
    /// Propose to update the group context extensions. This replaces the existing extensions
```

openmls/src/group/mls_group/extension.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// of a group and does not merge them.
```
duesee, line 21:
```suggestion
    /// of the group but does not merge them yet.
```

openmls/src/group/mls_group/extension.rs
```diff
@@ -0,0 +1,103 @@
...
+
+        Ok((mls_message, reference))
+    }
+
+    /// Updates the extensions of the group
+    ///
+    /// This operation results in a Commit with a `path`, i.e. it includes an
+    /// update of the committer's leaf [KeyPackage].
+    ///
+    /// If successful, it returns a tupple where the first element
```
duesee, line 61:
```suggestion
    /// If successful, it returns a tuple where the first element
```

openmls/src/group/mls_group/extension.rs
```diff
@@ -0,0 +1,103 @@
...
+        self.group_state = MlsGroupState::PendingCommit(Box::new(PendingCommitState::Member(
+            create_commit_result.staged_commit,
+        )));
+
+        // Since the state of the group might be changed, arm the state flag
+        self.flag_state_change();
+        Ok((mls_messages, create_commit_result.group_info))
+    }
+
+    /// Get a group's [`Extension`].
```
duesee, line 99:
```suggestion
    /// Get the group's [`Extensions`].
```

openmls/src/group/public_group/diff.rs
```diff
@@ -200,13 +201,17 @@ impl<'a> PublicGroupDiff<'a> {
     pub(crate) fn update_group_context(
         &mut self,
         backend: &impl OpenMlsCryptoProvider,
+        extensions: Option<&Extensions>,
```
duesee, line 204: I suggest to move `Extensions` here and let the caller `clone`. See https://rust-lang.github.io/api-guidelines/flexibility.html#c-caller-control

openmls/src/group/public_group/validation.rs
```diff
@@ -420,15 +419,39 @@ impl PublicGroup {
         Ok(())
     }
 
+    /// Validate GroupContextExtensions proposals. There must not be more than 1
```
duesee, line 422:
```suggestion
    /// Validate GroupContextExtensions proposals.
    /// 
    /// There must be at most one GroupContextExtensions proposal.
```

openmls/src/group/public_group/validation.rs
```diff
@@ -420,15 +419,39 @@ impl PublicGroup {
...
+    pub(crate) fn validate_group_context_extensions_proposals(
+        &self,
+        proposal_queue: &ProposalQueue,
+    ) -> Result<(), ProposalValidationError> {
+        let nb_gce = proposal_queue
```
duesee, line 427: nit(optional): `nb` means number? :-) Maybe call it `gce_count`?

openmls/src/group/public_group/validation.rs
```diff
@@ -420,15 +419,39 @@ impl PublicGroup {
...
+            ));
+        }
+
+        Ok(())
+    }
+
     /// Returns a [`LeafNodeValidationError`] if an [`ExtensionType`]
     /// in `extensions` is not supported by a leaf in this tree.
-    #[cfg(test)]
+    /// A list leaves proposed to be removed must be provided as they
```
duesee, line 443:
```suggestion
    /// A list of leaves proposed to be removed must be provided as they
```

openmls/src/group/tests/test_gce_proposals.rs
```diff
@@ -0,0 +1,744 @@
...
+    group::core_group::test_core_group::setup_client,
+    group::{
+        config::CryptoConfig, errors::*, test_core_group::setup_client_with_extensions,
+        PURE_PLAINTEXT_WIRE_FORMAT_POLICY,
+    },
+    messages::{
+        external_proposals::JoinProposal,
+        proposals::{GroupContextExtensionProposal, Proposal, ProposalOrRef},
+    },
+    prelude::{
```
duesee, line 17: Could we not use `prelude` here?

openmls/src/group/tests/test_update_extensions.rs
```diff
@@ -0,0 +1,90 @@
+use openmls_traits::{types::Ciphersuite, OpenMlsCryptoProvider};
+
+use crate::{
+    extensions::{
+        ApplicationIdExtension, Extension, ExtensionType, Extensions, RequiredCapabilitiesExtension,
+    },
+    group::errors::UpdateExtensionsError,
+    prelude::{MlsMessageIn, ProcessedMessageContent},
```
duesee, line 8: Could we not use `prelude` here?

openmls/src/group/errors.rs
```diff
@@ -369,6 +370,9 @@ pub enum ProposalValidationError {
     /// See [`PskError`] for more details.
     #[error(transparent)]
     Psk(#[from] PskError),
+    /// There cannot be more than 1 GroupContextExtensions proposal in a commit
+    #[error("Expected 1 GroupContextExtensions proposal per commit found {0}")]
```
duesee, line 374:
```suggestion
    #[error("Expected at most 1 GroupContextExtensions proposal, found {0}")]
```

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -598,6 +598,20 @@ impl LeafNode {
             .credentials
             .contains(credential_type)
     }
+
+    /// Check whether the this leaf node supports all the required extensions
```
duesee, line 602: I know it's only a move, but we could already fix the typo.

```suggestion
    /// Check whether this leaf node supports all the required extensions
```

duesee (commented):
Thanks! I left a preliminary review until we get the interop straight. Happy to have this in soon!

openmls/src/treesync/node/leaf_node/capabilities.rs
```diff
@@ -152,7 +152,7 @@ impl Capabilities {
     }
 
     /// Check if these [`Capabilities`] contain all the extensions.
-    pub(crate) fn contain_extensions(&self, extension: &Extensions) -> bool {
+    pub(crate) fn are_extensions_supported(&self, extension: &Extensions) -> bool {
```
augustocdias, line 155: I have renamed this method because I believe the previous naming leads to confusion. It does not check if the capabilities contain all the extensions but rather if the extensions are in the capabilities.

openmls/src/group/mls_group/config.rs
```diff
@@ -60,6 +60,10 @@ pub struct MlsGroupConfig {
     pub(crate) lifetime: Lifetime,
     /// Ciphersuite and protocol version
     pub(crate) crypto_config: CryptoConfig,
+    /// Extensions to be added to the own leaf node
+    pub(crate) leaf_extensions: Extensions,
+    /// Capabilities of the own leaf node
+    pub(crate) leaf_capabilities: Option<Capabilities>,
```
kkohbrok, line 66: Is there a reason we need these two to live in `MlsGroupConfig`? `MlsGroupConfig` is not just used to build the group, but is a fixed part of `MlsGroup`, where both extensions and capabilities are already stored in the owner's leaf.

kkohbrok (changes requested):
Thanks for the PR! Getting GCEs working would be a nice step forward for OpenMLS. I'm still investigating if interop is failing only due to MLSpp's handling of GREASE, or if there is something we can do on our side. Until then, here's already a review with a few comments that are unrelated to interop.

kkohbrok:
The (hopefully) last PR on the MLS spec is relevant here as well, as it clarifies whether GREASE values should be included in capabilities or not, which is at least in part what breaks interop right now. I played around with the PR a bit, but I can't figure out why the `application.json` interop scenario keeps failing even though it works on `main`. `commit.json` also doesn't work right now, but that is to be expected because of the difference in how MLSpp and OpenMLS handle GREASE.

franziskuskiefer:
We added handling of group context extension proposals in #1238. I'm closing this. There seem to be more changes in here though. @augustocdias if there's anything that you still want to get merged, please re-open or move it to a new PR.

PR #1363, opened 2023-04-25
docs: Describe interop workflow.

duesee:
This is a first attempt to make the interop workflow easy to reproduce before we put it into CI. The idea is that we provide `Dockerfile`s for MLS++ and the test-runner so that these can easily be build. These can be adapted but the idea now is that we start an interactive session in these containers so that we can more easily debug or add logging.

@augustocdias, could you try this out and review? :-)


franziskuskiefer (changes requested):
Thanks, generally lgtm with a few nits.

franziskuskiefer (approved):
Leaving an r+ here but you can wait for @augustocdias before merging.

augustocdias:
looks good to me too

PR #1367, opened 2023-04-25
fix: EncryptionKeyPair has the wrong MlsEntityId

OtaK:
This fixes a small bug that is unlikely to cause any real-world issues.
The EncryptionKeyPair struct has the wrong `MlsEntityId`. There's a small possibility (probably well under 0.1%) that a HPKE keypair generates the same public key as one of the signature public keys and ends up erasing it, breaking everything.

### DIFF
--- a/openmls/src/treesync/node/encryption_keys.rs
+++ b/openmls/src/treesync/node/encryption_keys.rs
@@ -281,5 +281,5 @@ impl From<(EncryptionKey, EncryptionPrivateKey)> for EncryptionKeyPair {
 }
 
 impl MlsEntity for EncryptionKeyPair {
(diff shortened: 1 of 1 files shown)


kkohbrok (dismissed):
Good catch! Thanks for the fix.

kkohbrok (dismissed):

kkohbrok (approved):

franziskuskiefer:
> @franziskuskiefer I see that you added that the PR is waiting on me; It seems the only blocker is the code coverage & drone CI tasks before auto-merge can trigger.

The PR needed a merge from main.
The coverage job isn't working for external PRs unfortunately. I'm going to merge this as is.
Thanks!

OtaK:
@franziskuskiefer Just rebased on top of `main`!

Cheers!

PR #1368, opened 2023-04-25
Remove identity from the credential

franziskuskiefer:
To support other credentials like X.509 the identity function needs to be removed.

To achieve this I moved the credential to the traits crate and added the `OpenMlsCredential` trait to be used for all credentials passed into OpenMls.


openmls/src/treesync/node/leaf_node.rs
```diff
@@ -339,11 +350,24 @@ impl LeafNode {
...
     }
 
-    /// Returns the `signature_key` as byte slice.
+    /// Returns the credential reference.
     pub fn credential(&self) -> &Credential {
         &self.payload.credential
     }
 
+    /// Returns the full credential.
+    pub fn full_credential<T>(&self, ciphersuite: Ciphersuite) -> Result<T, u8>
```
franziskuskiefer, line 359: Return proper Error.

traits/src/credential.rs
```diff
@@ -0,0 +1,29 @@
+//! # OpenMLS Credentials
+//!
+//! The MLS spec does not define credentials very well.
```
franziskuskiefer, line 3: Make clear that this here is not the credential as defined in the spec.

openmls/src/group/mls_group/creation.rs
```diff
@@ -42,43 +41,41 @@ impl MlsGroup {
         signer: &impl Signer,
```
franziskuskiefer, line 41: Add function to get `VerificationCredential` from `Signer`.

kkohbrok (commented):
Thanks for the PR! As discussed OOB, here's at least a partial review for now.

basic_credential/src/lib.rs
```diff
@@ -75,14 +130,14 @@ fn id(public_key: &[u8], signature_scheme: SignatureScheme) -> Vec<u8> {
     id
 }
 
-impl MlsEntity for SignatureKeyPair {
-    const ID: MlsEntityId = MlsEntityId::SignatureKeyPair;
+impl MlsEntity for OpenMlsBasicCredential {
+    const ID: MlsEntityId = MlsEntityId::OpenMlsBasicCredential;
```
OtaK, line 134: Small thing to keep in mind; 

The current state of `main` decorrelates Credentials (whether they're from a Basic credential or a x509 credential) from SignatureKeyPairs (whether again, they're from a basic credential or part of a x509 chain).

This seems to be a step back in the sense that Basic Credentials are again bound to their signature keypair (reminder, there can be N signature keypairs for the same credential, whether it's for ciphersuite reasons, or for the simple fact that the credential itself is a x509 credential and can contain many signature keypairs, up to the user) which seems incorrect in the context of ciphersuite agility.

PR #1369, opened 2023-04-26
feat: wrap sensitive types in order to zero the memory on drop (#1240)

### LINKED ISSUE #1240: zeroize private key material
It is a good security practice to erase from memory private/sensitive data (e.g. private/secret keys) to avoid side-channel attacks. While Rust will drop those memory regions when the struct goes out of scope, it does not guarantee the OS will erase its content ; it just won't be accessible anymore.

See [why zeroing memory securely is hard](http://www.daemonology.net/blog/2014-09-04-how-to-zero-a-buffer.html)

Thanks to the [zeroize](https://crates.io/crates/zeroize) crate, one can do this easily. It should be derived for all structs holding private key materials

augustocdias:
Closes #1240 

This wraps sensitve variables into new types that will implement `Zeroize` and will zero the memory on drop. Unfortunately we can't use the `VLBytes` in that case and I had to switch to the regular `Vec<u8>`.

I've moved the `HpkePrivateKey` struct to the `traits` crate and since for testing `PartialEq` and `Eq` are required I had to introduce the `test-utils` feature in it.

### DIFF
--- a/openmls/src/ciphersuite/mac.rs
+++ b/openmls/src/ciphersuite/mac.rs
@@ -1,17 +1,44 @@
 use super::*;
 
+/// A wrapper to contain MAC values that will be zeroed on drop.
+#[derive(
+    Debug,
+    Clone,
+    Serialize,
+    Deserialize,
+    TlsDeserialize,
+    TlsSerialize,
+    TlsSize,
+    zeroize::ZeroizeOnDrop,
+)]
+pub(crate) struct MacValue(Vec<u8>);
+
+impl From<Vec<u8>> for MacValue {
+    fn from(value: Vec<u8>) -> Self {
+        Self(value)
+    }
+}
+
+impl std::ops::Deref for MacValue {
+    type Target = [u8];
+
+    fn deref(&self) -> &Self::Target {
+        self.0.as_slice()
+    }
+}
+
 /// 7.1 Content Authentication
 ///
 /// opaque MAC<V>;
 #[derive(Debug, Clone, Serialize, Deserialize, TlsDeserialize, TlsSerialize, TlsSize)]
 pub(crate) struct Mac {
-    pub(crate) mac_value: VLBytes,
+    pub(crate) mac_value: MacValue,
 }
 
 impl PartialEq for Mac {
     // Constant time comparison.
     fn eq(&self, other: &Mac) -> bool {
-        equal_ct(self.mac_value.as_slice(), other.mac_value.as_slice())
+        equal_ct(&self.mac_value, &other.mac_value)
     }
 }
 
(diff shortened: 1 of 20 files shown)


openmls/src/ciphersuite/mac.rs
```diff
@@ -1,17 +1,44 @@
...
+    Debug,
+    Clone,
+    Serialize,
+    Deserialize,
+    TlsDeserialize,
+    TlsSerialize,
+    TlsSize,
+    zeroize::ZeroizeOnDrop,
+)]
+pub(crate) struct MacValue(Vec<u8>);
```
kkohbrok, line 14: MACs are generally considered public information. Is there a reason you want to derive ZeroizeOnDrop for them?

kkohbrok (commented):
Thanks for the PR! I talked about your proposal with @franziskuskiefer and @raphaelrobert and we came to the conclusion that deriving zeroize on the level of tls_codec might be a better option. We can then change the relevant structs you highlight in this PR to use the corresponding type from tls_codec.

I filed a draft PR in tls_codec as basis for further discussion and I would suggest we wait with this PR until that discussion is settled. In case you want to chime in: https://github.com/RustCrypto/formats/pull/1029

openmls/src/ciphersuite/mac.rs
augustocdias: I didn't know. I thought it was secret as it can be constructed from a secret.

kkohbrok: Fair point. In fact I just realized that we're being a bit sloppy by computing a MAC by calling HKDF extract. That works, but only as long as the HPKE ciphersuite uses an HKDF as its key derivation function. I filed #1375 to fix this (although it's functionally correct for now).

That being said, MACs are constructed in a similar way that ciphertexts are constructed from an encryption key, so I think we can do away with zeroize here.

kkohbrok:
With #1376 merged, you should now be able to just use `SecretVLBytes` from `tls_codec` instead of replacing `VLBytes` with `Vec<u8>`.

augustocdias:
> With #1376 merged, you should now be able to just use `SecretVLBytes` from `tls_codec` instead of replacing `VLBytes` with `Vec<u8>`. Do you want to give that a try?

In the types that were using `Vec<u8>` before, should I use `SecretVLBytes` or zeroize on them? For ex: the `value` attribute in the `AeadKey` is/was a `Vec` and not `VLBytes`

kkohbrok:
Good question. I guess it's going to be functionally the same since VLBytes is just a wrapper around Vec<u8>. Then I guess if it lets us avoid depending on zeroize explicitly, just use the `SecretVLBytes`. If it turns out to be a problem, it should be easy to reverse.

augustocdias:
> Good question. I guess it's going to be functionally the same since VLBytes is just a wrapper around Vec. Then I guess if it lets us avoid depending on zeroize explicitly, just use the `SecretVLBytes`. If it turns out to be a problem, it should be easy to reverse.

Another question. Should we also change the `OpenMlsCrypto` trait to return `SecretVLBytes` too (in the hkdf_expand and maybe the extract too)?

kkohbrok (approved):
Looks good! Thanks for making the additional changes.

PR #1370, opened 2023-04-26
interop/ci: Test interop between OpenMLS and MLS++.

duesee:
I do expect the new "interop" job to fail because we don't have support for group context extensions proposals for now. Also, the job will take quite long the first time it is run (~35 minutes). Afterwards, all VCPKG's should be cached and it will take roughly a few minutes.

Edit: I should add another commit that explicitly filters the errors due to group context extensions with a `TODO`. This way, CI will be green and @augustocdias could rebase #1361 and remove the `TODO` from this job.


franziskuskiefer (approved):
:shipit:

PR #1381, opened 2023-05-02
Improve interop

augustocdias:
This adds a docker compose and some improvements in the images in order to make them compatible with non Linux environments. I haven't tested it on Windows, but I expect no issue there. 

I've also updated the README with instructions of how to use it.

### DIFF
--- a/interop_client/README.md
+++ b/interop_client/README.md
@@ -15,7 +15,33 @@ through Docker.
 As an example, we want to test if OpenMLS can interoperate with MLS++.
 We need three components: the OpenMLS gRPC server, the MLS++ gRPC server, and the test-runner.
 
-### Build & start the OpenMLS gRPC server
+There are 2 ways to run the services: running each image separately or through docker compose.
+
+### Docker compose
+
+Simply navigate to the `./interop/docker` folder and run docker compose up. This will build the images start the services and run the test.
+
+This will run the `welcome_join.json` config file in the test-runner. To choose a different one, set the environment variable `CONFIG_RUN` prior to starting the services to the desired file name. The variable is set during container creation, so in order to change, it needs to be recreated.
+
+```
+CONFIG_RUN=commit.json docker compose up
+```
+
+Alternatively, you can run the test runner directly through this command:
+
+```
+docker compose run test-runner -client openmls:50051 -client mlspp:50052 -config=../configs/commit.json
+```
+
+All the parameters after the `test-runner` will be passed to the executable. The configs are in the `../configs` folder and the default hostnames and ports for the openmls implementation and mlspp are `openmls:50051` and `mlspp:50052`.
+
+The ports are open to the host so one can manually call each of the services. To change the port that it will be run, set the environment variables `OPENMLS_PORT` and `MLSPP_PORT` prior to the container creation.
+
+### Each service separately
+
+Be aware that the network flag used here for docker only works on linux. Consider using the docker compose in other platforms.
+
+#### Build & start the OpenMLS gRPC server
 
 The OpenMLS gRPC server can be started with ...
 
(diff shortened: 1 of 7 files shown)


interop_client/src/main.rs
```diff
@@ -43,6 +45,22 @@ use tracing::{debug, error, info, instrument, trace, Span};
 use tracing_subscriber::EnvFilter;
 
 const IMPLEMENTATION_NAME: &str = "OpenMLS";
+const PROPOSAL_TYPES: [ProposalType; 6] = [
```
franziskuskiefer, line 48: This shouldn't be necessary here.
These are all default types and are assumed to be present.

interop_client/src/main.rs
```diff
@@ -43,6 +45,22 @@ use tracing::{debug, error, info, instrument, trace, Span};
...
+    ProposalType::Add,
+    ProposalType::Update,
+    ProposalType::Remove,
+    ProposalType::PreSharedKey,
+    ProposalType::Reinit,
+    ProposalType::GroupContextExtensions,
+];
+const CREDENTIAL_TYPES: [CredentialType; 2] = [CredentialType::X509, CredentialType::Basic];
```
franziskuskiefer, line 56: x509 is not supported.

interop_client/src/main.rs
```diff
@@ -284,7 +302,21 @@ impl MlsClient for MlsClientImpl {
...
 
         let key_package = KeyPackage::builder()
-            .leaf_node_capabilities(Capabilities::default())
+            .leaf_node_capabilities(Capabilities::new(
+                Some(&[ProtocolVersion::Mls10, ProtocolVersion::Mls10Draft11]),
+                Some(&[
+                    Ciphersuite::MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519,
+                    Ciphersuite::MLS_128_DHKEMP256_AES128GCM_SHA256_P256,
+                    Ciphersuite::MLS_128_DHKEMX25519_CHACHA20POLY1305_SHA256_Ed25519,
+                    Ciphersuite::MLS_256_DHKEMP521_AES256GCM_SHA512_P521,
```
franziskuskiefer, line 311: The ciphersuites starting here are not supported.

interop_client/src/main.rs
```diff
@@ -1477,7 +1509,7 @@ impl MlsClient for MlsClientImpl {
 
 #[derive(Parser)]
 struct Opts {
-    #[clap(long, default_value = "[::1]")]
+    #[clap(long, default_value = "0.0.0.0")]
```
franziskuskiefer, line 1512: Why did you change the default to v4?

interop_client/README.md
```diff
@@ -28,41 +54,34 @@ RUST_LOG=interop=info cargo run
...
 
 ### Notes on interop testing
 
-* Each "step" in the config files is translated to one (or more) gRPC calls. For example, `"action": "externalJoin"` will request a group info, request an external commit from the joiner, and request all members to process the commit.
-* References such as `"byReference": [5, 6]` in the config files refer to the **index** of a step in the scenario.
-* Currently, the supported ciphersuites were fixated by a patch to `[1, 2, 3]` in the test-runner.
-* In order to pinpoint discrepancies, it might help to add (more) logging to OpenMLS or MLS++. Use a Docker volume to persist your changes.
+- Each "step" in the config files is translated to one (or more) gRPC calls. For example, `"action": "externalJoin"` will request a group info, request an external commit from the joiner, and request all members to process the commit.
+- References such as `"byReference": [5, 6]` in the config files refer to the **index** of a step in the scenario.
+- Currently, the supported ciphersuites were fixated by a patch to `[1, 2, 3]` in the test-runner.
```
franziskuskiefer, line 83:
```suggestion
- Currently, the supported ciphersuites are fixed by a patch to `[1, 2, 3]` in the test-runner.
```

interop_client/README.md
```diff
@@ -15,7 +15,33 @@ through Docker.
...
+docker compose run test-runner -client openmls:50051 -client mlspp:50052 -config=../configs/commit.json
+```
+
+All the parameters after the `test-runner` will be passed to the executable. The configs are in the `../configs` folder and the default hostnames and ports for the openmls implementation and mlspp are `openmls:50051` and `mlspp:50052`.
+
+The ports are open to the host so one can manually call each of the services. To change the port that it will be run, set the environment variables `OPENMLS_PORT` and `MLSPP_PORT` prior to the container creation.
+
+### Each service separately
+
+Be aware that the network flag used here for docker only works on linux. Consider using the docker compose in other platforms.
```
franziskuskiefer, line 42:
```suggestion
Be aware that the network flag used here for docker only works on Linux. Consider using the docker compose in other platforms.
```

interop_client/README.md
```diff
@@ -15,7 +15,33 @@ through Docker.
...
+
+Alternatively, you can run the test runner directly through this command:
+
+```
+docker compose run test-runner -client openmls:50051 -client mlspp:50052 -config=../configs/commit.json
+```
+
+All the parameters after the `test-runner` will be passed to the executable. The configs are in the `../configs` folder and the default hostnames and ports for the openmls implementation and mlspp are `openmls:50051` and `mlspp:50052`.
+
+The ports are open to the host so one can manually call each of the services. To change the port that it will be run, set the environment variables `OPENMLS_PORT` and `MLSPP_PORT` prior to the container creation.
```
franziskuskiefer, line 38:
```suggestion
The ports are open to the host so one can manually call each of the services. To change the port, set the environment variables `OPENMLS_PORT` and `MLSPP_PORT` prior to the container creation.
```

franziskuskiefer (changes requested):
Thanks, a few comments.

interop_client/src/main.rs
augustocdias: That was a change because of my PR for GCE support. As members are added, they are checked if they have all the required capabilities from the group. If they don't the add proposal would fail... For this PR itself this is not relevant...

franziskuskiefer: These are all default capabilities that do not need to be added. They should be added when used, even though the spec isn't entirely clear here. But they are not needed in general.

interop_client/src/main.rs
franziskuskiefer: Oh well, docker... But since docker isn't the main entry point here this should rather be solved in docker.

franziskuskiefer (commented):
If you want a review, please re-request it.

interop_client/src/main.rs
augustocdias: While I agree, this makes it unusable on Mac and Windows...

franziskuskiefer: How is this related to the OS? I can run this without problems on Mac.

franziskuskiefer (approved):
The ipv4 change is unnecessary but also doesn't really hurt right now, so lets get this in.

franziskuskiefer:
@augustocdias you need to rebase

interop_client/src/main.rs
augustocdias: Inside docker? I had issues with it...

raphaelrobert (approved):

PR #1382, opened 2023-05-04
Add credential validation section to book

### LINKED ISSUE #938: [MLS Spec change] Be more precise about credential validation (#703)
**Link to the exact changes**
https://github.com/mlswg/mls-protocol/pull/703

> Closes #678 
 > Replaces #660 
 > 
 > This PR attempts to add precision to the credential validation requirements in the document, along the lines raphaelrobert suggests in #678, in the spirit of what Bren2010 wrote up in #660 and drawing inspiration from what RFC 6125 does in the context of TLS.  It seems like this gives a clearer definition of (a) what the AS's job is and (b) when it needs to be done.

raphaelrobert:
Fixes #938.

Since OpenMLS doesn't handle credentials, this is not applicable directly. However, I added a section to the book.

### DIFF
--- a/book/src/user_manual/credential_validation.md
+++ b/book/src/user_manual/credential_validation.md
@@ -1,0 +1,12 @@
+# Credential validation
+
+Credential validation is a process that allows a member to verify the validity
+of the credentials of other members in the group.  The process is described in
+detail in the [MLS protocol
+specification](https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#name-credential-validation).
+
+In practice, the application should check the validity of the credentials of
+other members in two instances:
+
+ - When joining a new group (by looking at the ratchet tree)
+ - When [processing messages](./processing.md) (by looking at a StagedCommit)
--- a/book/src/user_manual/credential_validation.md
+++ b/book/src/user_manual/credential_validation.md
@@ -9,4 +9,4 @@ In practice, the application should check the validity of the credentials of
 other members in two instances:
 
  - When joining a new group (by looking at the ratchet tree)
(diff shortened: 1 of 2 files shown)


book/src/user_manual/credential_validation.md
```diff
@@ -0,0 +1,12 @@
+# Credential validation
+
+Credential validation is a process that allows a member to verify the validity
+of the credentials of other members in the group.  The process is described in
+detail in the [MLS protocol
+specification](https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#name-credential-validation).
```
kkohbrok, line 6: We should probably link to the datatracker version: https://www.ietf.org/archive/id/draft-ietf-mls-protocol-20.html#name-credential-validation

The MLS spec already refers to the architecture doc, but it might be helpful to mention the AS specifically and include the following link: https://www.ietf.org/archive/id/draft-ietf-mls-architecture-10.html#name-authentication-service

book/src/user_manual/credential_validation.md
```diff
@@ -0,0 +1,12 @@
...
+
+In practice, the application should check the validity of the credentials of
+other members in two instances:
+
+ - When joining a new group (by looking at the ratchet tree)
+ - When [processing messages](./processing.md) (by looking at a StagedCommit)
```
kkohbrok, line 12: Validating credentials is really only necessary in case of additions and updates (both via proposal and via path), isn't it? No need to validate credentials every time.

kkohbrok (changes requested):
Thanks for the PR! Just two comments.

kkohbrok (approved):

PR #1408, opened 2023-06-27
Remove empty `else` branch

mgeisler:


raphaelrobert (approved):

mgeisler:
By the way, I found this with Clippy — it was the only thing it found, so the code should now be Clippy warning free 🙂

PR #1353, opened 2023-04-18
replace evercrypt with libcrux

franziskuskiefer:
Use https://github.com/cryspen/libcrux instead of evercrypt.

Fixes #1319, #1250

PR #1411, opened 2023-06-29
Doc/build

Firefly35:
Add details about openmls building process on linux, either through bare metal installation or using a docker container.


kkohbrok:
Hi @Firefly35 and thanks for the PR! We have recently removed evercrypt support and it will only be added back (in a way) with #1353. Until then, I don't believe any additional build instructions beyond a basic rust setup are required. I'll leave the PR open and we can decide if we will need it once the libcrux PR has landed.

kkohbrok:
Sorry for the long wait @Firefly35. 

You're right. While the extra packages aren't needed for OpenMLS in itself, they are still necessary to compile the interop client and the other crates in the repo. 

Do you still want to contribute this to OpenMLS? If so, one thing that would be important to us is that building the dockerfile (and using it to compile OpenMLS) is tested on CI s.t. it doesn't grow stale over time. Similarly, we want to make sure that the instructions in the readme match those in the docker file (and are thus tested on CI). One way to do that would be by keeping them as a list in a separate file that the readme points to and that is also read by the docker file, but there may be more elegant ways to go about it.

franziskuskiefer:
Closing this for now. @Firefly35 feel free to re-open when you want to work on this again.

PR #1414, opened 2023-07-03
Fix[cli]: mismatched epoch proposal - invert proposal sending and group changes commit

Firefly35:
Once a user is registered and its group created, he invites another user. In the invited user cli, calling "update" led to an epoch validation error on a PrivateMessage. The group welcome message is processed correctly. 
The fix consists in sending the proposal prior to merge pending commits on the group, preventing the proposal to be sent to the new user


cli/src/user.rs
```diff
@@ -305,27 +306,29 @@ impl User {
             )
             .map_err(|e| format!("Failed to add member to group - {e}"))?;
 
-        // First, process the invitation on our end.
+        /* First, send the MlsMessage proposal to the group.
+        This must be done before the member invitation is committed.
+        It avoids the invited member to receive the proposal (which is in the previous group epoch).*/
+        log::trace!("Sending proposal");
```
raphaelrobert, line 312: This is really the Commit, not just the proposal.

raphaelrobert (approved):

PR #1415, opened 2023-07-11
Client removal and keypackage creation/publication

Firefly35:
This pull request provides the following additional features to the cli/DS:
- client removal from a group with the command remove [client name]
- key package management with the [create kp] command. The key package is created and added to the local client, and published to the delivery service for the client.


raphaelrobert:
@Firefly35 You requested a review, but there is a conflict

Firefly35:
@raphaelrobert Conflict resolved

raphaelrobert (changes requested):
Thanks, a few nits regarding naming & comments

raphaelrobert:
Question: What is the purpose of reserving KPs on the DS? Why not just delete them whenever a client has requested the? That is at least what messengers typically do. Unless of course the DS still needs them to route the Welcome (I'm not sure).

Firefly35:
Indeed the DS previously used the KP to match the client with the incoming welcome message to distribute the welcome message to the adequate client. I just made the evolutions to remove the KP once reserved through consume_key_package and remove its corresponding hash once the welcome message is distributed (previously, the KP remained on the DS, inviting the user twice led to errors as the user don't have the KP locally anymore once the welcome message is processed as it is removed from the keystore).

raphaelrobert (approved):
Let's get this in

PR #1416, opened 2023-07-13
Follow up CI fix

raphaelrobert:
Follow up for #1415 because it broke the CI somehow.
Also cleans up clippy warnings.

### DIFF
--- a/delivery-service/ds/src/main.rs
+++ b/delivery-service/ds/src/main.rs
@@ -160,6 +160,11 @@ async fn publish_key_packages(
     mut body: Payload,
     data: web::Data<DsData>,
 ) -> impl Responder {
+    let mut bytes = web::BytesMut::new();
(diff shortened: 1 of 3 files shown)


delivery-service/ds/src/main.rs
```diff
@@ -173,10 +178,6 @@ async fn publish_key_packages(
         None => return actix_web::HttpResponse::NotFound().finish(),
     };
 
-    let mut bytes = web::BytesMut::new();
-    while let Some(item) = body.next().await {
-        bytes.extend_from_slice(&unwrap_item!(item));
-    }
```
kkohbrok, line 179: Is there a particular reason you are moving this up to the beginning of the function?

raphaelrobert: Yes, there was a clippy warning that we were holding a MutexGuard across an .await

PR #1417, opened 2023-07-14
Ensure Add proposal order is maintained; add test

kkohbrok:
When adding group members via the MlsGroup API, the new members do not necessarily end up in the tree in the same order as they are in the slice of KeyPackages that is the input of the API call. This PR fixes that and adds a test.

### DIFF
--- a/openmls/src/group/core_group/proposals.rs
+++ b/openmls/src/group/core_group/proposals.rs
@@ -166,6 +166,38 @@ impl QueuedProposal {
     }
 }
 
+/// Helper struct to collect proposals such that they are unique and can be read
+/// out in the order in that they were added.
+struct OrderedProposalRefs {
+    proposal_refs: HashSet<ProposalRef>,
+    ordered_proposal_refs: Vec<ProposalRef>,
+}
+
+impl OrderedProposalRefs {
+    fn new() -> Self {
+        Self {
+            proposal_refs: HashSet::new(),
+            ordered_proposal_refs: Vec::new(),
+        }
+    }
+
+    /// Adds a proposal reference to the queue. If the proposal reference is
+    /// already in the queue, it ignores it.
+    fn insert(&mut self, proposal_ref: ProposalRef) {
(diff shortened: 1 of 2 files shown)


openmls/src/group/core_group/proposals.rs
```diff
@@ -166,6 +166,38 @@ impl QueuedProposal {
...
+    fn new() -> Self {
+        Self {
+            proposal_refs: HashSet::new(),
+            ordered_proposal_refs: Vec::new(),
+        }
+    }
+
+    /// Adds a proposal reference to the queue. If the proposal reference is
+    /// already in the queue, it ignores it.
+    fn insert(&mut self, proposal_ref: ProposalRef) {
```
raphaelrobert, line 186: `insert` is the wrong verb here, because it suggests new values can be inserted in the middle, like on a `HashMap`. Maybe something like `add_if_unique`?

openmls/src/group/core_group/proposals.rs
```diff
@@ -430,6 +464,8 @@ impl ProposalQueue {
         for queued_proposal in queued_proposal_list {
             match queued_proposal.proposal {
                 Proposal::Add(_) => {
+                    // `insert` returns true only if there was not entry present
+                    // with the same hash, i.e. if the entry is unique.
```
raphaelrobert, line 468: That comment seems unnecessary because the return value is not used.

raphaelrobert (changes requested):
Thanks for tackling the problem, just some nits.

openmls/src/group/core_group/proposals.rs
kkohbrok: Ah, sorry. That's a leftover.

openmls/src/group/core_group/proposals.rs
kkohbrok: I just called it `add` now. The uniqueness part should be clear from the nature of the type.

raphaelrobert (approved):
Thanks

PR #1420, opened 2023-07-25
Update tree operations tv from mlswg/mls-implementations#168

franziskuskiefer:
This PR also updates the tree hash in the group context after merging a diff. This step was missing before.


openmls/src/group/public_group/mod.rs
```diff
@@ -231,6 +231,8 @@ impl PublicGroup {
     pub(crate) fn merge_diff(&mut self, diff: StagedPublicGroupDiff) {
         self.treesync.merge_diff(diff.staged_diff);
         self.group_context = diff.group_context;
+        self.group_context
+            .update_tree_hash(self.treesync.tree_hash().to_vec());
```
kkohbrok, line 235: As far as the non-test code goes, this operation is mostly redundant. It only really does something in the tree operations test vector processing and in case the client was removed from the group, where in the latter case the tree hash isn't read afterwards anyway. In all other cases, the tree hash in the diff of the group context is already updated during the creation of the diff. It's not updated in the end, because we have to update it in the commit creation and processing process in the middle, because it's needed for the computation of the confirmation hash.

Since we're essentially processing the test vector manually already, I would propose updating the tree hash in the group context manually there, too. The way I see it is that all we're really doing by adding it here is make the test go through.

franziskuskiefer: Yes, this is due to the bad structure of the code here.
The main issue is that this function leaves the group context in a bad state. So either there needs to be clear instructions here on how to call the function, i.e. which other functions must be called after this to fix the broken/unfinished state, or it should be cleaned up in some other way.

kkohbrok (approved):

PR #1430, opened 2023-08-15
upgrade rust crypto libraries

kkohbrok:
This PR updates the ed25519-dalek crate to version 2.0 in the rust crypto provider and basic credential crates. Due to the updated API of ed25519-dalek, we no longer need to encode the private key as a concatenation of public and private key.

This PR also updates the aes-gcm and the chacha20poly1305 crates to their newest versions.

### DIFF
--- a/openmls/src/ciphersuite/tests/kat_crypto_basics.rs
+++ b/openmls/src/ciphersuite/tests/kat_crypto_basics.rs
@@ -280,21 +280,7 @@ pub fn run_test_vector(
         let mut parsed = ParsedSignWithLabel {
             key: SignatureKeyPair::from_raw(
                 ciphersuite.signature_algorithm(),
-                {
-                    if matches!(
-                        ciphersuite,
-                        Ciphersuite::MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519
-                            | Ciphersuite::MLS_128_DHKEMX25519_CHACHA20POLY1305_SHA256_Ed25519
-                            | Ciphersuite::MLS_256_DHKEMX448_CHACHA20POLY1305_SHA512_Ed448
-                    ) {
-                        let mut private: Vec<u8> = private;
-                        // For the RC crypto provider we need to have the public key in here.
-                        private.append(&mut public.clone());
-                        private
-                    } else {
-                        private
-                    }
-                },
+                private,
                 public.clone(),
             ),
             content,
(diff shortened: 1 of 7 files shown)


openmls/src/ciphersuite/tests/kat_crypto_basics.rs
```diff
@@ -280,21 +280,7 @@ pub fn run_test_vector(
         let mut parsed = ParsedSignWithLabel {
             key: SignatureKeyPair::from_raw(
                 ciphersuite.signature_algorithm(),
-                {
```
franziskuskiefer, line 283: Yay, this makes the code much nicer. That was a weird API in the first place.

openmls_rust_crypto/Cargo.toml
```diff
@@ -17,8 +17,7 @@ sha2 = { version = "0.10" }
 aes-gcm = { version = "0.9" }
```
franziskuskiefer, line 17: Aes and chacha could be updated as well.
But that will need a few more changes. Not sure if you want to do that here.

openmls_rust_crypto/src/provider.rs
```diff
@@ -235,12 +232,13 @@ impl OpenMlsCrypto for RustCrypto {
...
-                let k = ed25519_dalek::Keypair::generate(&mut rand_07::rngs::OsRng).to_bytes();
-                let pk = k[ed25519_dalek::SECRET_KEY_LENGTH..].to_vec();
-                // full key here because we need it to sign...
-                let sk_pk = k.into();
-                Ok((sk_pk, pk))
+                let mut rng = self
+                    .rng
+                    .write()
+                    .map_err(|_| CryptoError::InsufficientRandomness)?;
+                let sk = ed25519_dalek::SigningKey::generate(&mut *rng);
```
franziskuskiefer, line 239: I'm getting an error here. You need to enable the `rand_core` feature for this.
```text
   --> openmls_rust_crypto/src/provider.rs:239:53
    |
239 |                 let sk = ed25519_dalek::SigningKey::generate(&mut *rng);
    |                                                     ^^^^^^^^ function or associated item not found in `SigningKey`
```

openmls_rust_crypto/Cargo.toml
```diff
@@ -17,8 +17,7 @@ sha2 = { version = "0.10" }
...
 chacha20poly1305 = { version = "0.9" }
 hmac = { version = "0.12" }
-ed25519-dalek = { version = "1.0" }
-rand-07 = {version = "0.7", package = "rand" } # only needed because of ed25519-dalek
+ed25519-dalek = { version = "2.0" }
```
franziskuskiefer, line 20:
```suggestion
ed25519-dalek = { version = "2.0", features = ["rand_core"] }
```

franziskuskiefer (changes requested):
Thanks, this is so much nicer.
Just a few comments. I wonder why the error isn't caught by the CI.

franziskuskiefer (approved):
Thanks! :shipit:

PR #1434, opened 2023-09-02
Make the secret tree panic free

### LINKED ISSUE #1253: Refactor secret tree

raphaelrobert:
Removes potentially panicking indexing from the secret tree. Fixes #1253, since the other points mentioned there have been previously fixed (#1252).

### DIFF
--- a/openmls/src/tree/secret_tree.rs
+++ b/openmls/src/tree/secret_tree.rs
@@ -127,49 +127,33 @@ impl SecretTree {
         size: TreeSize,
         own_index: LeafNodeIndex,
     ) -> Self {
-        let mut leaf_nodes = std::iter::repeat_with(|| None)
-            .take(size.leaf_count() as usize)
-            .collect::<Vec<_>>();
+        let leaf_count = size.leaf_count() as usize;
+        let leaf_nodes = std::iter::repeat_with(|| None).take(leaf_count).collect();
+        let parent_nodes = std::iter::repeat_with(|| None).take(leaf_count).collect();
+        let handshake_sender_ratchets = std::iter::repeat_with(|| None).take(leaf_count).collect();
+        let application_sender_ratchets =
+            std::iter::repeat_with(|| None).take(leaf_count).collect();
 
-        let mut parent_nodes = std::iter::repeat_with(|| None)
-            .take(size.parent_count() as usize)
-            .collect::<Vec<_>>();
-
-        match root(size) {
-            TreeNodeIndex::Leaf(leaf_index) => {
-                leaf_nodes[leaf_index.usize()] = Some(SecretTreeNode {
-                    secret: encryption_secret.consume_secret(),
-                });
-            }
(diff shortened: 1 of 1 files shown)


openmls/src/tree/secret_tree.rs
```diff
@@ -127,49 +127,33 @@ impl SecretTree {
...
-
-        log::trace!(
-            "Created secret tree with {} leaves and {} nodes.",
-            leaf_nodes.len(),
-            parent_nodes.len()
-        );
-
-        SecretTree {
+        let leaf_count = size.leaf_count() as usize;
+        let leaf_nodes = std::iter::repeat_with(|| None).take(leaf_count).collect();
```
kkohbrok, line 131: Why not just `vec![None; leaf_count]`? No strong opinion, but it would make the code a little easier to read.

openmls/src/tree/secret_tree.rs
```diff
@@ -352,16 +350,18 @@ impl SecretTree {
 
     /// Returns a mutable reference to a specific SenderRatchet. The
     /// SenderRatchet needs to be initialized.
-    fn ratchet_mut(&mut self, index: LeafNodeIndex, secret_type: SecretType) -> &mut SenderRatchet {
+    fn ratchet_mut(
+        &mut self,
+        index: LeafNodeIndex,
+        secret_type: SecretType,
+    ) -> Result<&mut Option<SenderRatchet>, SecretTreeError> {
```
kkohbrok, line 357: Just a nit, but if you return `Option<&mut SenderRatchet>` here instead, you save the `.as_mut()` in the places where you call this. 

I know you don't want to refactor the secret tree generally, but there are a few places you touch where we get the secrets "manually" via `get_mut` instead of using this function. Might be worth just using this function everywhere instead.

openmls/src/tree/secret_tree.rs
```diff
@@ -394,7 +394,11 @@ impl SecretTree {
             ciphersuite
         );
         let hash_len = ciphersuite.hash_length();
-        let node_secret = match &self.parent_nodes[index_in_tree.usize()] {
+        let node_secret = match &self
+            .parent_nodes
+            .get(index_in_tree.usize())
+            .ok_or(SecretTreeError::IndexOutOfBounds)?
```
kkohbrok, line 400: Maybe use get_node instead?

kkohbrok (changes requested):
Thanks for picking this up! Looks good with just a few nits.

openmls/src/tree/secret_tree.rs
raphaelrobert: Because `vec![]` requires the element type to implement `Clone` (even for `None`), which we don't want for `EcryptionSecret`.

kkohbrok (approved):

PR #1437, opened 2023-09-07
feat[sender info]: add message sender information in conversation

Firefly35:
Handle message sender information to display the sender while receiving messages.

### DIFF
--- a/cli/src/user.rs
+++ b/cli/src/user.rs
@@ -643,7 +643,7 @@ impl User {
             .mls_group
             .borrow_mut()
             .remove_members(&self.crypto, &self.identity.borrow().signer, &[leaf_index])
-            .map_err(|e| format!("Failed to add member to group - {e}"))?;
+            .map_err(|e| format!("Failed to remove member from group - {e}"))?;
 
         // First, send the MlsMessage remove commit to the group.
         log::trace!("Sending commit");
--- a/cli/src/user.rs
+++ b/cli/src/user.rs
@@ -405,38 +405,28 @@ impl User {
                         Some(c) => c.username.clone(),
                         None => {
                             // Contact list is not updated right now, get the identity from the mls_group member
-                            let mut user_id: String = String::new();
-                            for Member {
-                                index: _,
-                                encryption_key: _,
-                                signature_key,
-                                credential,
-                            } in mls_group.members()
-                            {
-                                if self
-                                    .identity
-                                    .borrow()
-                                    .credential_with_key
-                                    .signature_key
-                                    .as_slice()
-                                    != signature_key.as_slice()
+                            let user_id = mls_group.members().find_map(|m| {
+                                if m.credential.identity()
+                                    == processed_message_credential.identity()
+                                    && (self
+                                        .identity
+                                        .borrow()
+                                        .credential_with_key
+                                        .signature_key
+                                        .as_slice()
+                                        != m.signature_key.as_slice())
                                 {
-                                    if processed_message_credential.identity()
-                                        == credential.identity()
-                                    {
-                                        log::debug!("update::Processing ApplicationMessage read sender name from credential identity for group {} ", group.group_name);
-                                        user_id = str::from_utf8(credential.identity())
-                                            .unwrap()
-                                            .to_owned()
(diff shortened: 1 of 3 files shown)


cli/src/user.rs
```diff
@@ -340,14 +351,26 @@ impl User {
...
+            let processed_message_credential: Credential = processed_message.credential().clone();
+
             match processed_message.into_content() {
                 ProcessedMessageContent::ApplicationMessage(application_message) => {
-                    let application_message =
-                        String::from_utf8(application_message.into_bytes()).unwrap();
+                    let sender_name =
+                        match self.contacts.get(processed_message_credential.identity()) {
+                            Some(c) => &c.username,
+                            None => panic!("There's a member in the group we don't know."),
```
kkohbrok, line 362: This implies that we expect to have all group members in our contact list, which seems a bit restricting to me. The cli is only meant as a proof of concept, but since most modern messaging applications allow groups where not everyone knows everyone else, we might want to keep that possibility open here, too.

I don't have a super strong opinion on this, but at the very least we don't want to panic here.

kkohbrok (changes requested):
Looks good to me in general. Just one comment.

cli/src/user.rs
Firefly35: Our proposal would be to split update in two different methods: 
- one to update contacts
- the other one to update messages and process commits...
Currently,  the contact list contains all registered clients of the DS at update time. 
So in the cli PoC context, every group member is known by the DS and is in the clients list.
When a message is received by a client which doesn't already knows the contact, it means the client needs to update to at least process the commit corresponding to the group invitation and to update the contact list from the DS.
Splitting the methods would allow to update the contact list when we reach this case.
Your opinion ?

kkohbrok: Ah, right. I made a wrong assumption there, sorry. Wouldn't it then be simpler to just always synchronize the contacts before processing the messages?

Firefly35: I thought about it but not sure it will work in all use case:
- we update the contact before processing the messages
- a new person register on the DS and is invited in a group, this person sends an application message
- on the client side we process the messages after the contact processing: the contact doesn't exists (although it is an extreme scenario, as the protocol is asynchronous it can happen)

kkohbrok: Since this is a PoC, I wouldn't be too worried about such edge cases. However, we could prevent this scenario from happening by proceeding as follows:
* we fetch the messages,
* we fetch the contacts,
* we update the contacts and finally
* we process the messages.
What do you think?

Firefly35: what about updating the contacts when we meet the special "missing contact" case as I described ?

kkohbrok: Fair enough. Let's see what that looks like. As long as it doesn't add a lot of complexity, it's fine by me.

cli/src/user.rs
```diff
@@ -340,14 +393,60 @@ impl User {
...
+                                            .to_owned()
+                                    }
+                                }
+                            }
+                            user_id
+                        }
+                    };
+                    if sender_name.is_empty() {
+                        panic!("We received a message from an unknown member group !");
+                    }
```
kkohbrok, line 439: If a member chooses an empty string as their identity, then that should be okay, so no need to panic here.

kkohbrok (changes requested):
Just two more nits, then this should be okay.

cli/src/user.rs
```diff
@@ -340,14 +393,60 @@ impl User {
...
+                                        == credential.identity()
+                                    {
+                                        log::debug!("update::Processing ApplicationMessage read sender name from credential identity for group {} ", group.group_name);
+                                        user_id = str::from_utf8(credential.identity())
+                                            .unwrap()
+                                            .to_owned()
+                                    }
+                                }
+                            }
+                            user_id
```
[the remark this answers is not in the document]
Firefly35: done

cli/src/user.rs
Firefly35: done

kkohbrok (approved):
Thanks!

PR #1442, opened 2023-09-27
Derive TlsDeserializeBytes for publicly exposed structs

kkohbrok:
This PR derives or implements the `tls_codec::DeserializeBytes` trait for a number of public structs and enums. Where possible, it uses the `TlsDeserializeBytes` derive macro. In a few cases, it implements the trait by hand to avoid deriving the struct for a large number of "inner" structs.

Implementing vs. deriving for more structs is a bit of trade off. For now, I've implemented it in a few places to avoid derivation in many more places, which in turn helps keep this PR small-ish. Let me know if you think we should instead use the derive macro.


openmls/src/group/core_group/kat_passive_client.rs
```diff
@@ -137,13 +137,14 @@ pub fn run_test_vector(test_vector: PassiveClientWelcomeTestVector) {
         test_vector.init_priv,
     );
 
-    let ratchet_tree: Option<RatchetTreeIn> = test_vector
-        .ratchet_tree
-        .as_ref()
-        .map(|bytes| RatchetTreeIn::tls_deserialize_exact(bytes.0.as_slice()).unwrap());
+    let ratchet_tree: Option<RatchetTreeIn> = test_vector.ratchet_tree.as_ref().map(|bytes| {
+        <RatchetTreeIn as tls_codec::Deserialize>::tls_deserialize_exact(bytes.0.as_slice())
```
franziskuskiefer, line 141: I don't think these changes are needed ...

openmls/src/group/core_group/kat_welcome.rs
```diff
@@ -126,7 +127,10 @@ pub fn run_test_vector(test_vector: WelcomeTestVector) -> Result<(), &'static st
 
     let key_package: KeyPackage = {
         let mls_message_key_package =
-            MlsMessageIn::tls_deserialize_exact(test_vector.key_package).unwrap();
+            <MlsMessageIn as tls_codec::Deserialize>::tls_deserialize_exact(
```
franziskuskiefer, line 130: Same here, these should be unnecessary.

franziskuskiefer (changes requested):
I didn't look at everything yet, but all the `as tls_codec::Deserialize` casts seem pretty ugly. You could change the return value instead to tell the compiler which version you want.
But I think that's the other question. Why do you want to use the Bytes version for `MlsMessageIn`?

kkohbrok:
Closing in favor of #1468.

PR #1443, opened 2023-09-28
Make backtrace crate an optional dependency.

anhph0:
backtrace has > 10 dependencies (transitively) which might make it difficult for applications to import it, and consequently using openmls.


kkohbrok (changes requested):
Hi @anhph0, thanks for the PR! Putting backtrace behind a feature flag sounds reasonable. Could you run `cargo fmt`? Then we can get this in.

kkohbrok (approved):

kkohbrok (approved):
Thanks!

PR #1451, opened 2023-11-13
Relax Send + Sync from OpenMlsProvider

insipx:
Send + Sync was removed from `OpenMlsKeyStore` but not `OpenMlsProvider` in #1440. This PR brings the change over to OpenMlsProvider.

`OpenMlsCrypto` is still Send + Sync. Can change that in this PR too if that's desired


franziskuskiefer (approved):
Thanks!

PR #1453, opened 2023-11-18
bump tls_codec and hpke-rs (pre-release)

franziskuskiefer:
Updating hpke and tls_codec to the pre-release version. `std` is explicitly required now as a feature.
@kkohbrok can you check that the things you want to do in #1442 work with this?


kkohbrok:
Thanks for preparing the bump! Give me a bit to do some preliminary testing. I'll report back here when I have results.

kkohbrok:
As discussed OOB, everything should work with tls_codec 4.0.

You've already bumped hpke to a new version in #1456. Do you want to update this PR accordingly? Alternatively we could wait with bumping tls_codec until we actually use the new features.

kkohbrok (approved):

PR #1461, opened 2024-01-03
Add `update_path_leaf_node` method to `StagedCommit`.

### LINKED ISSUE #1460: Add ability to get commit leaf node again
#1218 removed the ability to verify the credential in the update path.
Let's re-add this ability in the same way for now. Before redesigning the APIs to make it easier getting all leaf nodes in the diff.

keks:
In order to validate the credentials of the leaf node in the optional update path of a staged commit, the caller needs to be able to access it. They can now do so using the new method `update_path_leaf_node`.

Unfortunately the change on line 156 made the line so wide that the entire `if let`-block afterwards shifted indentation.

I have not yet written any tests because I am unfamiliar with the testing infrastructure. Any help with that is appreciated.

Fixes #1460

### DIFF
--- a/openmls/src/group/mls_group/test_mls_group.rs
+++ b/openmls/src/group/mls_group/test_mls_group.rs
@@ -430,7 +430,7 @@ fn test_commit_with_update_path_leaf_node(
     assert!(_welcome_option.is_none());
     assert!(_group_info.is_none());
 
-    // There should be a pending commit after issueing a self-update commit.
+    // There should be a pending commit after issuing a self-update commit.
     let alice_pending_commit = alice_group
         .pending_commit()
         .expect("alice should have the self-update as pending commit");
@@ -452,7 +452,7 @@ fn test_commit_with_update_path_leaf_node(
 
     // === transfer message to bob and process it ===
 
-    // this requires serializing and deserizing
+    // this requires serializing and deserializing
     let mut wire_msg = Vec::<u8>::new();
     commit_msg
         .tls_serialize(&mut wire_msg)
--- a/openmls/src/group/mls_group/test_mls_group.rs
+++ b/openmls/src/group/mls_group/test_mls_group.rs
@@ -1,6 +1,7 @@
 use core_group::test_core_group::setup_client;
 use openmls_rust_crypto::OpenMlsRustCrypto;
 use openmls_traits::{key_store::OpenMlsKeyStore, OpenMlsProvider};
+use tls_codec::{Deserialize, Serialize};
 
 use crate::{
     binary_tree::LeafNodeIndex,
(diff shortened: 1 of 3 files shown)


raphaelrobert:
Thanks for your first PR @keks!

I *think* that leaf node is exposed as an update proposal on a `StagedCommit`. This would be easy to verify with a test that's needed here anyway.

franziskuskiefer (changes requested):
Thanks. Generally looks good to me.

> I think that leaf node is exposed as an update proposal on a StagedCommit. This would be easy to verify with a test that's needed here anyway.

That would be pretty weird (and wrong). There's no proposal for that leaf node.

But a test is needed indeed.

openmls/src/group/mls_group/test_mls_group.rs
```diff
@@ -334,6 +335,185 @@ fn test_invalid_plaintext(ciphersuite: Ciphersuite, provider: &impl OpenMlsProvi
...
+        .self_update(provider, &alice_signer)
+        .expect("error creating self-update commit");
+    println!("Done creating commit.");
+
+    // empty commits should only produce a single message
+    assert!(_welcome_option.is_none());
+    assert!(_group_info.is_none());
+
+    // There should be a pending commit after issueing a self-update commit.
+    let alice_pending_commit = alice_group
```
franziskuskiefer, line 434:
```suggestion
    // There should be a pending commit after issuing a self-update commit.
```

openmls/src/group/mls_group/test_mls_group.rs
```diff
@@ -334,6 +335,185 @@ fn test_invalid_plaintext(ciphersuite: Ciphersuite, provider: &impl OpenMlsProvi
...
+
+    // great, they match! now commit
+    alice_group
+        .merge_pending_commit(provider)
+        .expect("alice failed to merge the pending empty commit");
+
+    // === transfer message to bob and process it ===
+
+    // this requires serializing and deserizing
+    let mut wire_msg = Vec::<u8>::new();
```
franziskuskiefer, line 456:
```suggestion
    // this requires serializing and deserializing
```

franziskuskiefer (approved):
Thanks!

The test looks good to me too.
Lets get this in to fix the issue and then design a better way.
As part of that we should write down all the possible cases we should test for this as well. The test right now only tests one way of creating an update path and we should make sure this always works.

franziskuskiefer (approved):

franziskuskiefer:
@raphaelrobert is there still something you want to look at here? Otherwise I'll go ahead and merge it.

openmls/src/group/mls_group/test_mls_group.rs
```diff
@@ -334,6 +335,185 @@ fn test_invalid_plaintext(ciphersuite: Ciphersuite, provider: &impl OpenMlsProvi
...
+
+    if let ProcessedMessageContent::ProposalMessage(staged_proposal) =
+        alice_processed_message.into_content()
+    {
+        alice_group.store_pending_proposal(*staged_proposal);
+    } else {
+        unreachable!("Expected a StagedCommit.");
+    }
+
+    println!("\nCreating commit with add proposal.");
```
raphaelrobert, line 386: I guess that's leftovers from debugging

raphaelrobert (changes requested):
Thanks, just a nit.

I checked and it seems I misremembered about the update path leaf node being added to the update proposals, so we are good.

raphaelrobert (approved):

openmls/src/group/mls_group/test_mls_group.rs
keks: I copied that from the other test, but I can remove these from all the tests in that file in a later PR.

PR #1464, opened 2024-01-03
MlsGroup: Builder, Pattern and Config

### LINKED ISSUE #1462: Split MlsGroupConfig into MlsGroupConfig and MlsGroupPattern

### LINKED ISSUE #1463: Add builder pattern for MlsGroup

kkohbrok:
This PR introduces a builder pattern for MlsGroup (i.e. fixes #1463). It also splits the existing MlsGroupConfig into two: MlsGroupJoinConfig and MlsGroupCreateConfig (i.e. fixes #1462).

Unfortunately this has grown quite large, but that is mostly to renaming of variables.

The MlsGroupJoinConfig contains the per-client "runtime-relevant" configuration options for an MlsGroup. It's needed when joining a group either via Welcome or via External commit.

The MlsGroupCreateConfig contains an MlsGroupJoinConfig, as well as additional configuration options that are relevant only when creating a fresh MLS group, such as an ExternalSenders extension, or a RequiredCapabilities.

Until now, group configuration was somewhat confusing for users, as it was possible, for example, to input an ExternalSendersExtension to a group when joining it via the MlsGroupConfig (although it was ignored by OpenMLS anyway).

### DIFF
--- a/openmls/src/group/mls_group/config.rs
+++ b/openmls/src/group/mls_group/config.rs
@@ -1,18 +1,18 @@
 //! Configuration module for [`MlsGroup`] configurations.
 //!
-//! ## Building an MlsGroupConfig
-//! The [`MlsGroupConfigBuilder`] makes it easy to build configurations for the
+//! ## Building an MlsGroupPattern
+//! The [`MlsGroupPatternBuilder`] makes it easy to build configurations for the
 //! [`MlsGroup`].
 //!
 //! ```
 //! use openmls::prelude::*;
 //!
-//! let group_config = MlsGroupConfig::builder()
+//! let group_config = MlsGroupPattern::builder()
 //!     .use_ratchet_tree_extension(true)
 //!     .build();
 //! ```
 //!
-//! See [`MlsGroupConfigBuilder`](MlsGroupConfigBuilder#implementations) for
+//! See [`MlsGroupPatternBuilder`](MlsGroupPatternBuilder#implementations) for
 //! all options that can be configured.
 //!
 //! ### Wire format policies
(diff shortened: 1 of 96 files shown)


keks:
I like the structure, but I'm wondering if we can find more telling names for the two types. Maybe we can rename `MlsGroupPattern` to `MlsGroupCreateConfig` or `MlsGroupInitConfig`, and `MlsGroupConfig` to `MlsGroupJoinConfig`? I expect that would make the usage of the values somewhat more clear. Do you agree?

openmls/src/group/mls_group/config.rs
```diff
@@ -50,65 +52,155 @@ pub struct MlsGroupConfig {
...
+    pub fn sender_ratchet_configuration(&self) -> &SenderRatchetConfiguration {
+        &self.sender_ratchet_configuration
+    }
+}
+
+/// Specifies configuration pattern for an [`MlsGroup`]. Refer to the [User
+/// Manual](https://openmls.tech/book/user_manual/group_config.html) for more
+/// information about the different configuration values.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct MlsGroupPattern {
```
franziskuskiefer, line 85: I agree with Jan that the naming may not be the best.
But I also wonder if we need both of these.
My understanding is that `Config` is for runtime parameters and `Pattern` is for creating a new group.
How about having one config with different ways of instantiating it depending on whether it's used for join or creation?

franziskuskiefer (commented):
Thanks!
Generally lgtm with some thoughts on the naming, but also the design where it's coming from.

kkohbrok:
Thanks for the reviews! I've added docs and renamed the structs, as well as variable names where they still referred to "pattern".

keks (commented):
This looks good!

One last nit: The join config getter on the create config struct has the somewhat unspecific name `mls_group_config`. I think when reading code that uses it, it would be helpful to be able to immediately see which config is being retrieved, without requiring the context of what type the call is used on. And for consistency it would be nice if the getter matched the field name.

franziskuskiefer (approved):
Let's get it in with the nits addressed.

keks (dismissed):

franziskuskiefer (approved):

PR #1467, opened 2024-01-04
perf: small perf improvements

beltram:
Hey, I stumbled upon those small improvements while using the lib, I guess they deserve to be merged upstream 😄


raphaelrobert (approved):
Thanks!

PR #1468, opened 2024-01-05
Derive DeserializeBytes alongside Deserialize

kkohbrok:
This PR derives `tls_codec::DeserializeBytes` for all structs and enums that already derive `tls_codec::Deserialize`. It also implements manually `DeserializeBytes` for those that implement `Deserialize`. The "manual" implementations all follow the same pattern in that they internally rely on the `Deserialize` implementation of that struct/enum.


franziskuskiefer (approved):
Thanks lgtm.
We may want to put some of this behind feature flags at some point to avoid generating too much code. But let's do that when we get there.

PR #1470, opened 2024-01-09
Add credentials_to_verify method to StagedCommit

### LINKED ISSUE #1469: API: Make credential validation more convenient
Difficult to get credential validation right. API needs to be improved here

keks:
A StagedCommit can carry credentials in several different places,
all of which have to be verified. To make this easier on the caller,
we add the `credentials_to_verify` method on StagedCommit, which returns
an iterator over all the credentials in a StagedCommit that need to
be verified.

Also we add infrastructure in test_framework for letting the test code
decide which credentials would pass validation, by passing in a closure
that takes a &Credential and returns a bool, as well as a simple test
that serves as an example for how to use the function.

Fixes #1469

### DIFF
--- a/openmls/src/group/core_group/staged_commit.rs
+++ b/openmls/src/group/core_group/staged_commit.rs
@@ -458,41 +458,43 @@ impl StagedCommit {
             vec![]
         };
 
-        update_path_leaf_node_cred.into_iter().chain(
-            self.queued_proposals()
-                .map(|proposal: &QueuedProposal| match proposal.proposal() {
-                    Proposal::Update(update_proposal) => {
-                        vec![update_proposal.leaf_node().credential()].into_iter()
-                    }
-                    Proposal::Add(add_proposal) => {
-                        vec![add_proposal.key_package().leaf_node().credential()].into_iter()
-                    }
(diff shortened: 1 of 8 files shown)


franziskuskiefer (commented):
Thanks, lgtm. Only a few nits, then this can go in.

keks:
@franziskuskiefer wait, don't we need to do similar verification for Add and Update proposals?

franziskuskiefer (approved):
I'm fine with this.
@kkohbrok did you want to have a look at this as well?

kkohbrok (approved):
Generally looks good to me! In terms of credential verification, the spec also says, the application needs to verify that a credential update is valid, which probably involves comparing the old and the new credential in some way. I think that might be useful to add, but we don't have to do that in this PR. (And I think it would only be relevant for Credentials in Update and Path credentials.)

franziskuskiefer:
> which probably involves comparing the old and the new credential in some way

Good point. That information can be retrieved with the `members` function on the group. But having a getter for the ones that got updated would make life for applications easier.

franziskuskiefer:
@keks do you want to squash and merge?

PR #1472, opened 2024-01-09
Refactor VerifiedStruct and Verifiable

### LINKED ISSUE #1465: [Bug] `VerifiedStruct` doesn't achieve what it appears to be attempting
[`VerifiedStruct`](https://github.com/openmls/openmls/blob/main/openmls/src/ciphersuite/signable.rs#L58-L80) has  an associated type  `SealingType` that the caller should not be able to construct, so they are not able to call the `from_verifiable` function. However, since `SealingType` has a constraint to implement `Default` (which it needs; it is used by the `verify` method of [`Verifiable`](https://github.com/openmls/openmls/blob/main/openmls/src/ciphersuite/signable.rs#L129-L170)), the user can trivially construct it, too.

I briefly discussed this with @franziskuskiefer. We believe the best way to fix this is to just make the `VerifiedStruct` trait private. Since this is mostly an internal tool for limiting access, this should be fine. It seems the only other way would be to do it (edit: i mean the conversion between the source type and target struct here) inside the `verify` method, which would require the implementation of the `Verifiable` trait to have full visibility into the target struct.

keks:
The current version of the two traits is the way it is because it attempted to prevent the caller from being able to convert to the type implementing `VerifiedStruct` without checking the signature. However, the approach that was taken does not work.

This commit refactors the two traits and performs the conversion in the same method as the verification. Unfortunately, this means that the `verify` method has to be defined per `impl` instead of being able to rely on the default one. There probably is no straightforward way around that, so this simple solution is preferred.

Fixes #1465

### DIFF
--- a/openmls/src/ciphersuite/signable.rs
+++ b/openmls/src/ciphersuite/signable.rs
@@ -53,32 +53,6 @@ pub trait SignedStruct<T> {
     fn from_payload(payload: T, signature: Signature) -> Self;
 }
 
-/// This trait must be implemented by all structs that contain a verified
-/// self-signature.
-pub trait VerifiedStruct<T> {
-    /// This type is used to prevent users of the trait from bypassing `verify`
-    /// by simply calling `from_verifiable`. `Seal` should be a dummy type
-    /// defined in a private module as follows:
-    /// ```
-    /// mod private_mod {
-    ///     pub struct Seal;
-    ///
-    ///     impl Default for Seal {
-    ///         fn default() -> Self {
-    ///             Seal {}
-    ///         }
-    ///     }
-    /// }
-    /// ```
-    type SealingType: Default;
-
-    /// Build a verified struct version from the payload struct. This function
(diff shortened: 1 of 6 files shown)


openmls/src/ciphersuite/signable.rs
```diff
@@ -137,22 +119,17 @@ pub trait Verifiable: Sized {
     fn label(&self) -> &str;
 
     /// Verifies the payload against the given `credential`.
-    /// The signature is fetched via the [`Verifiable::signature()`] function and
-    /// the payload via [`Verifiable::unsigned_payload()`].
+    /// Usually this is implemented by first checking that `self.verify_no_out()`
+    /// does not returnl an error, and then converting the value into
```
kkohbrok, line 123:
```suggestion
    /// does not return an error, and then converting the value into
```

kkohbrok (changes requested):
Thanks for taking care of this! One small nit, otherwise looks good!

kkohbrok (approved):
Thanks!

PR #1473, opened 2024-01-11
Add `with_group_context_extensions` method to group builders

keks:
Currently, only two kinds of extensions can be set for the initial group context, and they both have their own method on the builder to set them. In order to be able to add arbitrary extensions, this PR adds a method `with_group_context_extensions` to the group builders, that sets an arbitrary set of extensions. The set is of type `Extensions`, so we know that each extension is only set once.

Any RequiredCapability extension mentioned in that set will be overwritten, and if ExternalSenders is specified using the dedicated method, that one is preferred and the one provided using `with_group_context_extensions` is dropped.

### DIFF
--- a/openmls/src/group/core_group/mod.rs
+++ b/openmls/src/group/core_group/mod.rs
@@ -211,15 +211,9 @@ impl CoreGroupBuilder {
         mut self,
         extensions: Extensions,
     ) -> Result<Self, InvalidExtensionError> {
-        let is_valid_in_group_context = extensions.application_id().is_none()
-            && extensions.ratchet_tree().is_none()
-            && extensions.external_pub().is_none();
-        if !is_valid_in_group_context {
-            return Err(InvalidExtensionError::IllegalInGroupContext);
-        }
         self.public_group_builder = self
             .public_group_builder
-            .with_group_context_extensions(extensions);
+            .with_group_context_extensions(extensions)?;
         Ok(self)
     }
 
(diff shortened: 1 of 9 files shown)


openmls/src/group/mls_group/builder.rs
```diff
@@ -198,4 +199,12 @@ impl MlsGroupBuilder {
             .external_senders(external_senders);
         self
     }
+
+    /// Sets the initial group context extensions
+    pub fn with_group_context_extensions(mut self, extensions: Extensions) -> Self {
```
franziskuskiefer, line 204: Do we check that `extensions` are only valid group context extensions?
This is of course a little difficult for extensions that OpenMLS doesn't know about, and the spec doesn't give us much to help here. But we should at least make sure that we don't allow extensions in here from the RFC that are not group context extensions. And we should have a test case for that.

openmls/src/group/core_group/mod.rs
```diff
@@ -200,6 +200,19 @@ impl CoreGroupBuilder {
         }
         self
     }
+
+    /// Set initial group context extensions. Note that RequiredCapabilities
+    /// extensions will be overwritten, and should be set using a call to
+    /// `required_capabilities`. If `ExternalSenders` extensions are provided
+    /// both in this call, and a call to `external_senders`, only the one from
```
franziskuskiefer, line 207: Is this something that should be mentioned on the public API as well?

franziskuskiefer (commented):
Thanks, generally lgtm. 
One main comment.

keks:
@kkohbrok Good points, I added everything.

kkohbrok (approved):
Thanks, much appreciated!

PR #1475, opened 2024-01-17
Process GroupContextExtensions Proposals

### LINKED ISSUE #1238: Apply GroupContextExtension (GCE) proposals

keks:
This PR adds support for GroupContextExtensions proposals. The relevant validation has been added both to the code and the documentation.

Note that we can't build illegal messages, so there are no tests of the validation of incoming messages, yet.

Fixes #1238, #566

### DIFF
--- a/openmls/src/group/core_group/test_proposals.rs
+++ b/openmls/src/group/core_group/test_proposals.rs
@@ -22,7 +22,6 @@ use crate::{
     messages::proposals::{AddProposal, Proposal, ProposalOrRef, ProposalType},
     schedule::psk::store::ResumptionPskStore,
     test_utils::*,
-    treesync::errors::LeafNodeValidationError,
     versions::ProtocolVersion,
 };
 
@@ -308,58 +307,6 @@ fn test_required_unsupported_proposals(ciphersuite: Ciphersuite, provider: &impl
         e,
         CoreGroupBuildError::PublicGroupBuildError(PublicGroupBuildError::UnsupportedProposalType)
     )
-}
-
-#[apply(ciphersuites_and_providers)]
-fn test_required_extension_key_package_mismatch(
-    ciphersuite: Ciphersuite,
-    provider: &impl OpenMlsProvider,
-) {
-    // Basic group setup.
-    let group_aad = b"Alice's test group";
-    let framing_parameters = FramingParameters::new(group_aad, WireFormat::PublicMessage);
-
-    let (alice_credential, _, alice_signer, _alice_pk) =
-        setup_client("Alice", ciphersuite, provider);
-    let (_bob_credential_with_key, bob_key_package_bundle, _, _) =
-        setup_client("Bob", ciphersuite, provider);
-    let bob_key_package = bob_key_package_bundle.key_package();
-
-    // Set required capabilities
-    let extensions = &[
-        ExtensionType::RequiredCapabilities,
-        ExtensionType::ApplicationId,
-    ];
-    let proposals = &[
-        ProposalType::GroupContextExtensions,
-        ProposalType::Add,
-        ProposalType::Remove,
-        ProposalType::Update,
-    ];
-    let credentials = &[CredentialType::Basic];
-    let required_capabilities =
-        RequiredCapabilitiesExtension::new(extensions, proposals, credentials);
-
-    let alice_group = CoreGroup::builder(
-        GroupId::random(provider.rand()),
-        CryptoConfig::with_default_version(ciphersuite),
-        alice_credential,
-    )
-    .with_required_capabilities(required_capabilities)
-    .build(provider, &alice_signer)
-    .expect("Error creating CoreGroup.");
-
-    let e = alice_group
-        .create_add_proposal(
-            framing_parameters,
-            bob_key_package.clone(),
-            &alice_signer,
-        )
-        .expect_err("Proposal was created even though the key package didn't support the required extensions.");
-    assert_eq!(
-        e,
-        CreateAddProposalError::LeafNodeValidation(LeafNodeValidationError::UnsupportedExtensions)
-    );
 }
 
 #[apply(ciphersuites_and_providers)]
--- a/openmls/src/group/mls_group/test_mls_group.rs
+++ b/openmls/src/group/mls_group/test_mls_group.rs
@@ -1043,7 +1043,7 @@ fn group_context_extensions_proposal(ciphersuite: Ciphersuite, provider: &impl O
         .expect("couldn't get required_capabilities");
 
     // has required_capabilities as required capability
-    assert!(required_capabilities.extension_types() == &[ExtensionType::RequiredCapabilities]);
+    assert!(required_capabilities.extension_types() == [ExtensionType::RequiredCapabilities]);
 }
 
 // Test that the builder pattern accurately configures the new group.
--- a/openmls/src/group/mls_group/test_mls_group.rs
+++ b/openmls/src/group/mls_group/test_mls_group.rs
@@ -1059,7 +1059,7 @@ fn group_context_extensions_proposal(ciphersuite: Ciphersuite, provider: &impl O
     // === committing to two group context extensions should fail
 
     alice_group
-        .propose_group_context_extensions(provider, new_extensions.clone(), &alice_signer)
+        .propose_group_context_extensions(provider, new_extensions, &alice_signer)
         .expect("failed to build group context extensions proposal");
 
     // the proposals need to be different or they will be deduplicated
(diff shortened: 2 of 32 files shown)


keks:
The test that now fails is unrelated and (I believe) broken:

https://github.com/openmls/openmls/blob/main/openmls/src/group/core_group/test_proposals.rs#L314-L363

It adds regular extensions to the required capabilities and then asserts that this should fail because the extension types are not explicitly listed in the key package. However, the RFC says that the extensions defined therein must always be supported and do not need to be specified.

As long as we don't let the group types support more capabilities than the default ones, I don't think that test can be fixed. Should I just delete it? Comment it out?

@franziskuskiefer  @kkohbrok what do you think?

keks:
BTW the reason this fails now is that I fixed the supported capabilities test functions: 

https://github.com/openmls/openmls/pull/1475/files#diff-222e1633c67450da2c5b1fe61b546f6dc7d1d5f83ab0e8e9a28513009d798a1eL140-R156

openmls/src/group/mls_group/test_mls_group.rs
```diff
@@ -1002,6 +1002,96 @@ fn remove_prosposal_by_ref(ciphersuite: Ciphersuite, provider: &impl OpenMlsProv
...
+    let required_capabilities = alice_group
+        .group()
+        .group_context_extensions()
+        .required_capabilities()
+        .expect("couldn't get required_capabilities");
+
+    // no required extensions
+    assert!(required_capabilities.extension_types().is_empty());
+
+    let new_extensions = Extensions::from_vec(vec![Extension::RequiredCapabilities(
```
kkohbrok, line 1026: Just a small nit: I believe `Extensions` has a `single` constructor that allows you to create an instance of `Extensions` from a single extension. So you don't have to create a vector first. Same goes for two lines below.

openmls/src/group/mls_group/test_mls_group.rs
```diff
@@ -1002,6 +1002,96 @@ fn remove_prosposal_by_ref(ciphersuite: Ciphersuite, provider: &impl OpenMlsProv
...
+        .expect("error merging pending commit");
+
+    let required_capabilities = alice_group
+        .group()
+        .group_context_extensions()
+        .required_capabilities()
+        .expect("couldn't get required_capabilities");
+
+    // has required_capabilities as required capability
+    assert!(required_capabilities.extension_types() == [ExtensionType::RequiredCapabilities]);
```
kkohbrok, line 1057: For a little more coverage, you could compare the content of the extension in addition to just the type here.

openmls/src/group/mls_group/test_mls_group.rs
```diff
@@ -1002,6 +1002,96 @@ fn remove_prosposal_by_ref(ciphersuite: Ciphersuite, provider: &impl OpenMlsProv
...
+    // === committing to two group context extensions should fail
+
+    alice_group
+        .propose_group_context_extensions(provider, new_extensions, &alice_signer)
+        .expect("failed to build group context extensions proposal");
+
+    // the proposals need to be different or they will be deduplicated
+    alice_group
+        .propose_group_context_extensions(provider, new_extensions_2, &alice_signer)
+        .expect("failed to build group context extensions proposal");
```
kkohbrok, line 1068: In the future, this will hopefully start failing at some point because a RatchetTreeExtension is not actually a group context extension.

openmls/src/group/mls_group/test_mls_group.rs
```diff
@@ -1002,6 +1002,96 @@ fn remove_prosposal_by_ref(ciphersuite: Ciphersuite, provider: &impl OpenMlsProv
...
+    )])
+    .expect("failed to build new extensions list");
+
+    alice_group
+        .propose_group_context_extensions(provider, new_extensions, &alice_signer)
+        .expect_err("expected an error building GCE proposal with bad required_capabilities");
+
+    // TODO: we need to test that processing a commit with multiple group context extensions
+    //       proposal also fails. however, we can't generate this commit, because our functions for
+    //       constructing commits does not permit it.
```
kkohbrok, line 1093: If you don't want to do this in this PR (which is totally fine), can you create an issue and reference the issue number in this TODO? Just so we don't lose track of it.

openmls/src/group/public_group/diff/apply_proposals.rs
```diff
@@ -140,6 +142,15 @@ impl<'a> PublicGroupDiff<'a> {
             })
             .collect();
 
+        // apply group context extension proposal
+        let extensions = proposal_queue
+            .filtered_by_type(ProposalType::GroupContextExtensions)
+            .map(|queued_proposal| match queued_proposal.proposal() {
+                Proposal::GroupContextExtensions(extensions) => extensions.extensions().clone(),
+                _ => unreachable!(),
```
kkohbrok, line 150: Is there a reason why you're not using `filter_map` and an `if let` construction here? It would avoid the `unreachable`.

openmls/src/group/public_group/validation.rs
```diff
@@ -515,7 +516,42 @@ impl PublicGroup {
...
+                    let ext_type_list: Vec<_> = extensions
+                        .extensions()
+                        .iter()
+                        .map(|ext| ext.extension_type())
+                        .collect();
+
+                    self.check_extension_support(&ext_type_list).map_err(|_| GroupContextExtensionsProposalValidationError::ExtensionNotSupportedByAllMembers)?
+                }
+                _ => {
+                    return Err(GroupContextExtensionsProposalValidationError::LibraryError(
```
kkohbrok, line 537: I think you can get rid of this library error by using an `if let` construction here.

openmls/src/group/core_group/test_proposals.rs
```diff
@@ -310,58 +309,6 @@ fn test_required_unsupported_proposals(ciphersuite: Ciphersuite, provider: &impl
     )
 }
 
-#[apply(ciphersuites_and_providers)]
```
kkohbrok, line 313: Instead of deleting this test, can't you replace the default extensions used here with extensions of type `Unknown`?

kkohbrok (changes requested):
Thanks for tackling this! A few questions and nits, but this generally looks good to me.

openmls/src/group/core_group/test_proposals.rs
kkohbrok: I see. Then I think it's okay to delete it for now. I'll restore it once I get around to enabling Unknown extensions.

openmls/src/group/mls_group/test_mls_group.rs
keks: I don't think it should, because we are only using the ratchet tree extension type in the RequiredCapabilities extension, not the actual extension.

openmls/src/group/mls_group/test_mls_group.rs
keks: This is a bit wonky -- I am using a RequireCapability extension and use it to require the capability to handle "RequiredCapabilities" extensions. So I am checking that the extension has the correct extension types in it's contents.

kkohbrok: Ah, I didn't see that. Fair enough.

openmls/src/group/mls_group/test_mls_group.rs
kkohbrok: Good point. Let's leave it as it is, then.

openmls/src/group/public_group/validation.rs
keks: Hm, I could use `if let` but then what if I do find a non-gce proposal in there? I could ignore it or return a LibraryError -- it would just live in the `else` branch of the `if let`.

kkohbrok: The policy we agreed on at some point was that we ignore libraryerrors where possible. For example in `apply_proposals`, we filter for proposals quite a lot and always ignore "unexpected" results. Do I remember this correctly @franziskuskiefer? We should really document this somewhere if we haven't already.

franziskuskiefer: There's a tradeoff between readability, soundness, and correctness. If soundness and correctness isn't impacted, it's ok to ommit library errors. There were some refactorings that unfortunately introduced soundness issues by removing library errors (something that still needs fixing).
Since this function has to return an error anyway because it's checking for multiple gce proposals, I think we should keep the library error here.

kkohbrok: Fair enough. Have we written all of this down somewhere? If not, maybe the wiki would be a good place? Or the documentation of `LibraryError`?

kkohbrok (approved):

PR #1477, opened 2024-01-18
Add leaf node configuration options to MlsGroup builder

kkohbrok:
This PR adds (and tests) a way to add leaf node extensions and capabilities when creating a new group using the MlsGroup builder pattern.

### DIFF
--- a/openmls/src/group/core_group/mod.rs
+++ b/openmls/src/group/core_group/mod.rs
@@ -190,7 +190,7 @@ impl CoreGroupBuilder {
             .with_required_capabilities(required_capabilities);
         self
     }
-    /// Set the [`Capabilities`] of the [`CoreGroup`].
+    /// Set the [`Capabilities`] of the group's creator.
     pub(crate) fn with_capabilities(mut self, capabilities: Capabilities) -> Self {
         self.public_group_builder = self.public_group_builder.with_capabilities(capabilities);
         self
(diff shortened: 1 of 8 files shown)


openmls/src/group/public_group/builder.rs
```diff
@@ -64,7 +70,23 @@ impl TempBuilderPG1 {
...
+        mut self,
+        extensions: Extensions,
+    ) -> Result<Self, InvalidExtensionError> {
+        // None of the default extensions are leaf node extensions, so only
+        // unknown extensions can be leaf node extensions.
+        let is_valid_in_leaf_node = extensions
+            .iter()
+            .all(|e| matches!(e.extension_type(), ExtensionType::Unknown(_)));
+        if !is_valid_in_leaf_node {
+            return Err(InvalidExtensionError::IllegalInGroupContext);
```
keks, line 87: I believe this is the wrong error, or am I missing something?

openmls/src/group/core_group/mod.rs
```diff
@@ -189,6 +190,11 @@ impl CoreGroupBuilder {
             .with_required_capabilities(required_capabilities);
         self
     }
+    /// Set the [`Capabilities`] of the [`CoreGroup`].
```
keks, line 193: This confused me for a second because this sounds like the capabilities are a property of the CoreGroup and not the LeafNode. Maybe "Capabilities of the group creator's LeafNode" or something like that would be more accurate?

keks (changes requested):
Just two small nits, otherwise looks good!

kkohbrok:
Thanks for the feedback! I hope I addressed everything.

keks (approved):
Looks good!

franziskuskiefer (approved):
Thanks

PR #1478, opened 2024-01-18
Clean up `MlsGroup*` builders

kkohbrok:
This PR removes the dedicated `required_capabilities` and `external_senders` fields (along with their getters and setters) in favor of the `group_context_extension` field. This makes processing during group creation easier.

### DIFF
--- a/openmls/src/group/mls_group/builder.rs
+++ b/openmls/src/group/mls_group/builder.rs
@@ -2,7 +2,7 @@ use openmls_traits::{key_store::OpenMlsKeyStore, signatures::Signer, OpenMlsProv
 
 use crate::{
     credentials::CredentialWithKey,
-    extensions::{errors::InvalidExtensionError, Extensions, ExternalSendersExtension},
+    extensions::{errors::InvalidExtensionError, Extensions},
     group::{
         config::CryptoConfig, public_group::errors::PublicGroupBuildError, CoreGroup,
         CoreGroupBuildError, CoreGroupConfig, GroupId, MlsGroupCreateConfig,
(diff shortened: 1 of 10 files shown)


openmls/src/group/mls_group/builder.rs
```diff
@@ -194,22 +195,6 @@ impl MlsGroupBuilder {
         self
     }
 
-    /// Sets the `external_senders` property of the MlsGroup.
-    pub fn external_senders(mut self, external_senders: ExternalSendersExtension) -> Self {
```
franziskuskiefer, line 198: Remmoving these public functions should go into the changelog.
We should add the others we added as well.

openmls/src/group/tests/external_remove_proposal.rs
```diff
@@ -30,7 +30,10 @@ fn new_test_group(
     let mls_group_config = MlsGroupCreateConfig::builder()
         .wire_format_policy(wire_format_policy)
         .crypto_config(CryptoConfig::with_default_version(ciphersuite))
-        .external_senders(external_senders)
+        .with_group_context_extensions(Extensions::single(Extension::ExternalSenders(
```
franziskuskiefer, line 33: Nothing to change here. But I noticed that we're not consistent using `with_property` or only `property` in builders.

franziskuskiefer (approved):
Thanks, generally looks good. Let's try to keep the changelog a little up to date.

openmls/src/group/tests/external_remove_proposal.rs
kkohbrok: True, but that's a lot of additional changes that I'd rather have in a separate PR. We'll also have to agree on a style. Maybe generally something we should do once we're done with refactoring the builders.

franziskuskiefer (approved):
Thanks

PR #1479, opened 2024-01-18
Enable the use of unknown/opaque group context extensions.

kkohbrok:
This PR enables OpenMLS to handle unknown group context and leaf node extensions. Extensions that have to be instantiated using the `Unknown` extensiontype are handled  transparently by OpenMLS, which means that they can be used to encapsulate data, but they won't alter OpenMLS' behaviour. Extensions that have specific additional semantics need to be supported explicitly by OpenMLS.

This PR also re-adds the test removed in #1475.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#1475](https://github.com/openmls/openmls/pull/1475): Fully process GroupContextExtension proposals
 - [#1477](https://github.com/openmls/openmls/pull/1477): Allow setting leaf node extensions and capabilities of the group creator when creating an MlsGroup(Config)
 - [#1478](https://github.com/openmls/openmls/pull/1478): Remove explicit functions to set `RequiredCapabilitiesExtension` and `ExternalSendersExtension` when building an MlsGroup(Config) in favor of the more general function to set group context extensions
-- [#1479](https://github.com/openmls/openmls/pull/1479): Allow the use of extensions with `ExtensionType::Unknown` in group context and leaf nodes
+- [#1479](https://github.com/openmls/openmls/pull/1479): Allow the use of extensions with `ExtensionType::Unknown` in group context, key packages and leaf nodes
 
 ## 0.5.0 (XXXX-XX-XX)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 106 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#1475](https://github.com/openmls/openmls/pull/1475): Fully process GroupContextExtension proposals
 - [#1477](https://github.com/openmls/openmls/pull/1477): Allow setting leaf node extensions and capabilities of the group creator when creating an MlsGroup(Config)
 - [#1478](https://github.com/openmls/openmls/pull/1478): Remove explicit functions to set `RequiredCapabilitiesExtension` and `ExternalSendersExtension` when building an MlsGroup(Config) in favor of the more general function to set group context extensions
+- [#1479](https://github.com/openmls/openmls/pull/1479): Allow the use of extensions with `ExtensionType::Unknown` in group context and leaf nodes
```
franziskuskiefer, line 17: Is there a reason to restrict to group context and leaf nodes?

openmls/src/group/mls_group/test_mls_group.rs
```diff
@@ -1195,3 +1208,86 @@ fn builder_pattern(ciphersuite: Ciphersuite, provider: &impl OpenMlsProvider) {
...
     assert_eq!(builder_err, InvalidExtensionError::IllegalInLeafNodes);
 }
+
+// Test that unknown group context and leaf node extensions can be used in groups
+#[apply(ciphersuites_and_providers)]
+fn unknown_extensions(ciphersuite: Ciphersuite, provider: &impl OpenMlsProvider) {
+    let (alice_credential_with_key, _alice_kpb, alice_signer, _alice_pk) =
+        setup_client("Alice", ciphersuite, provider);
+
+    let unknown_gc_extension = Extension::Unknown(0xff00, UnknownExtension(vec![0, 1, 2, 3]));
```
franziskuskiefer, line 1218: You only use private IDs here. Which is good.
I wonder if we should enforce that, or allow any ID for unknown extensions.

franziskuskiefer (commented):
Thanks, generally looks good.
Only two questions.

CHANGELOG.md
kkohbrok: Now that I think about it, I could include KeyPackage extensions. It probably also works for GroupInfo extensions, but we don't have a way to send specific GroupInfo extensions.

kkohbrok: I've changed the changelog accordingly and added a test, as well as documentation in the book w.r.t. unknown KP extensions. I suspect that GroupInfo extensions also work, but we'd have to expose them at the API level first.

openmls/src/group/mls_group/test_mls_group.rs
kkohbrok: Turns out this is a bit trickier than I expected and I'd rather not do it in this PR. I've added an issue, though: #1480.

franziskuskiefer (approved):
Thanks, let's get it in.

PR #1483, opened 2024-01-23
Enable Wasm Tests

keks:
This PR enables most (but not all!) tests for Wasm. We don't run the tests using wasm-pack, instead we use [this approach](https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/usage.html#appendix-using-wasm-bindgen-test-without-wasm-pack). The reason is that using wasm_pack would require the tests to be public, including the entire module path leading to it. That did not seem feasible.

It also makes some changes so that this actually works:
- disable parallelism for wasm
  - rayon: inline conditional compilation chooses whether to use `(into_)iter` or `(into_)par_iter` based on `target_arch`
  - disable default features for criterion
- don't use the file system at runtime in tests
  - added the `read_json` macro that reads the test vectors at compile time
    - somehow it thinks the macro is unused, so I told it to ignore unused_macros here
  - in one test we write to a file and then reopen it. I replaced that by writing to a buffer (`Vec<u8>`) and then reeading it
- use `fluvio_wasm_timer` instead of `std::time`

This PR reuses ~300 tests that all pass when used with Wasm. We should add more later, but it's a good start I think.

### DIFF
--- a/openmls/Cargo.toml
+++ b/openmls/Cargo.toml
@@ -30,6 +30,9 @@ openmls_basic_credential = { version = "0.2.0", path = "../basic_credential", op
 ] }
 rstest = { version = "^0.16", optional = true }
 rstest_reuse = { version = "0.4", optional = true }
+wasm-bindgen-test = {version = "0.3.40", optional = true}
+getrandom = {version = "0.2.12", optional = true, features = [ "js" ]}
+fluvio-wasm-timer = "0.2.5"
 
 [features]
 default = ["backtrace"]
@@ -41,14 +44,16 @@ test-utils = [
     "dep:rand",
     "dep:rstest",
     "dep:rstest_reuse",
+    "dep:wasm-bindgen-test",
     "dep:openmls_basic_credential",
 ]
 crypto-debug = [] # ☣️ Enable logging of sensitive cryptographic information
 content-debug = [] # ☣️ Enable logging of sensitive message content
+js = ["dep:getrandom"] # enable js randomness source for provider
 
 [dev-dependencies]
 backtrace = "0.3"
-criterion = "^0.5"
+criterion = {version = "^0.5", default-features = false} # need to disable default features for wasm
 hex = { version = "0.4", features = ["serde"] }
 itertools = "0.10"
 lazy_static = "1.4"
@@ -60,6 +65,8 @@ pretty_env_logger = "0.5"
 rstest = "^0.16"
 rstest_reuse = "0.4"
 tempfile = "3"
+wasm-bindgen = "0.2.90"
+wasm-bindgen-test = "0.3.40"
 
 [[bench]]
 name = "benchmark"
--- a/openmls/Cargo.toml
+++ b/openmls/Cargo.toml
@@ -32,7 +32,7 @@ rstest = { version = "^0.16", optional = true }
 rstest_reuse = { version = "0.4", optional = true }
 wasm-bindgen-test = {version = "0.3.40", optional = true}
 getrandom = {version = "0.2.12", optional = true, features = [ "js" ]}
-fluvio-wasm-timer = "0.2.5"
+fluvio-wasm-timer = {version = "0.2.5", optional = true}
 
 [features]
 default = ["backtrace"]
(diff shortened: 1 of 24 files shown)


keks:
I have prevailed in my duel with the github actions yaml file, and this is now ready for review

kkohbrok (commented):
Thanks for tackling this! Looks pretty straight-forward. Just a few general questions, mostly due to my ignorance w.r.t. the limitations of WASM:

* Would it make sense to have some documentation (in the book and/or in the wiki) on the implications of using the WASM build (no parallelization, etc.)?
* As I understand it, WASM currently comes with two workarounds: the rstest one and the fact that we're using fluvio-wasm-timer, which I understand to also be a workaround (although I might be wrong on that one). Should we add an issue to track the eventual removal of those workarounds, or document them somewhere else?
* You say that not all tests are covered when using wasm. It doesn't seem obvious to me which ones are not covered, so could you give me an idea (at least generally) which ones are not covered and why not? This might also be worth including in documentation somewhere (an issue or the wiki maybe).

keks:
Good questions!

> Would it make sense to have some documentation (in the book and/or in the wiki) on the implications of using the WASM build (no parallelization, etc.)?

Yeah that's a good point, I'll write something up.

> As I understand it, WASM currently comes with two workarounds: the rstest one and the fact that we're using fluvio-wasm-timer, which I understand to also be a workaround (although I might be wrong on that one). Should we add an issue to track the eventual removal of those workarounds, or document them somewhere else?

I'll create an issue for the wasm_bindgen_test module. Regarding `fluvio-wasm-timer`: I am not sure I'd call it a workaround, it's more of a shim. On wasm it gets the time using `web_sys`, and otherwise it just reexports the stdlib types. I suppose I should hide it behind the `js` feature though, which signals whether the wasm environment provides the expected JS APIs. That I should definitely document!

> You say that not all tests are covered when using wasm. It doesn't seem obvious to me which ones are not covered, so could you give me an idea (at least generally) which ones are not covered and why not? This might also be worth including in documentation somewhere (an issue or the wiki maybe).

Unfortunately, running tests for wasm requires a different macro attribute than `#[test]`. I initially the wasm test macro manually to see if they worked, and then changed the macro templates that many tests use (the ones with e.g. `#[apply(ciphersuites)]`) to also add it when building for wasm. This means that the tests that only have the plain `#[test]` attribute do not yet run on wasm. It seems these are 49. Most of them should only require adding the attribute. Some will require easy fixes, and others may not be super feasible now (mostly those that work with the file system).

I can document it somewhere -- do you think the wiki is a better place than the book? I could just create a Wasm page in the book that explains all the caveats.

keks:
- I added a page in the User Manual part of the book that focuses on what the user needs to know, including the js feature
- I added a page in the wiki that captures the current state of wasm support and its caveats
- I made sure you can't compile without the `js` feature set, in order to avoid unexpected runtime crashes

Should I create the issue for the workaround now or after the PR is merged?

openmls/Cargo.toml
```diff
@@ -49,7 +49,7 @@ test-utils = [
 ]
 crypto-debug = [] # ☣️ Enable logging of sensitive cryptographic information
 content-debug = [] # ☣️ Enable logging of sensitive message content
-js = ["dep:getrandom"] # enable js randomness source for provider
+js = ["dep:getrandom", "dep:fluvio-wasm-timer"] # enable js randomness source for provider
```
franziskuskiefer, line 52: Should these really go here or are they only needed for tests?
Generally we should get rid of the `getrandom` dependency like this here.
See https://docs.rs/getrandom/latest/getrandom/#webassembly-support

But this is for a follow-up.

franziskuskiefer (approved):
I'd be fine getting this in and filing follow-ups.

kkohbrok (approved):
Thanks for answering my questions and for writing up the section in the book!

I agree that we can merge this now and deal with the rest in follow-ups. Once we're happy with the test coverage, we can add webassembly as officially supported in the README as well.

openmls/Cargo.toml
keks: So yes, `getrandom` does it nicely by requiring you to set the js flag, but `fluvio-wasm-timer` just assumes that the JS environment is available when you are running in `wasm32-unknown-unknown`. In order to gracefully fail in the situation that someone tries to use OpenMLS on `wasm32-unknown-unknown` without a JS env, I used the same approach as `getrandom`, because that seems the sanest. Otherwise, the caller only knows they need a JS env for randomness, not timing.

Though I suppose that if randomness works, timing likely will as well. Maybe I was overthinking here, but I wasn't super confident with just using the JS env when we are not sure that it is available.

keks:
I created issues #1484 and #1485

PR #1488, opened 2024-01-31
Allow more credentials

### LINKED ISSUE #1459: Support for custom credential types

MLS has a private-use space for types of credentials that are application-specific. OpenMLS doesn't support letting applications use this space at all.

OpenMLS should implement support for custom credential types. These credentials would be implemented by the application, and follow a trait implementation of serialization / verification / etc. OpenMLS would ensure that the verification method is called where necessary.

Currently when trying to implement an application-specific credential, the best option is to just shove a bunch of data into a BasicCredential and try to do all of the necessary verification in the application. This introduces a substantial amount of complexity and potential for bugs.

**Additional context**

franziskuskiefer:
This PR changes OpenMLS such that it can handle arbitrary credentials, _as long as they are encoded as a variable-length vector_.

Due to limitations in the spec we do not handle credentials that may be encoded differently. However, this shouldn't be a big deal because the credential payload can always be encoded as a byte vector, which is probably a good idea anyway when it is a custom credential.

Note that this PR requires two small changes to the tls_codec to expose functions for reading and writing the variable length bytes. If we go ahead with this approach, I'll make the necessary changes.

Fixes #1459


openmls/src/group/public_group/validation.rs
```diff
@@ -488,29 +484,12 @@ impl PublicGroup {
...
-                        }
-                    };
-                }
-            }
-        }
+        // If a Remove proposal is present,
+        // the credential in the LeafNode MUST present a set of
+        // identifiers that is acceptable to the application for
+        // the removed participant.
+        // This MUST be checked by the application.
```
kkohbrok, line 491: Is there an issue that we can link here? Or maybe just add this to the book?

openmls/src/credentials/mod.rs
```diff
@@ -187,7 +180,61 @@ pub enum MlsCredentialType {
...
+        self.credential_type.tls_serialize(writer)?;
+        writer.write_all(self.serialized_credential_content.as_slice())?;
+        Ok(self.tls_serialized_len())
+    }
+}
+
+impl tls_codec::Deserialize for Credential {
+    fn tls_deserialize<R: Read>(bytes: &mut R) -> Result<Self, Error> {
+        // We can not deserialize arbitrary credentials because we don't know
+        // the structure of it. While we don't care, we still need to parse it
```
kkohbrok, line 204:
```suggestion
        // their structure. While we don't care, we still need to parse it
```

openmls/src/credentials/mod.rs
```diff
@@ -197,47 +244,22 @@ impl Credential {
...
-                MlsCredentialType::Basic(_) => CredentialType::Basic,
-                MlsCredentialType::X509(_) => CredentialType::X509,
-            },
-            credential: mls_credential_type,
-        }
+    /// Get this serialized credential content.
+    ///
+    /// This is the content of the `select` statement. It is a TLS serialized
+    /// vector.
+    pub fn serialized_content(&self) -> &[u8] {
```
kkohbrok, line 261: It would be nice if we could make this a bit more ergonomic. Would it be possible to make this function generic in the type of credential (which has to support `tls_codec::Deserialize`) and then try to deserialize internally? That would return a Result, but it would spare the caller the additional explicit deserialization step. Does that make sense?

kkohbrok (commented):
Wow, this looks much more straight-forward than I would have expected. Sure, lots of small changes all over the place, but the core seems relatively simple.

Just a few comments.

openmls/src/credentials/mod.rs
franziskuskiefer: I added a second function below to get it deserialized. Is that what you mean?
I agree, that's a nice one to have. But I'd keep this too.

openmls/src/group/public_group/validation.rs
franziskuskiefer: Right, so this is a case where the application is responsible for the check and we can't do it (anymore). I added a sentence to the book and started a section. We'll want to fill that section more.

openmls/src/credentials/mod.rs
kkohbrok: Yes, that's what I meant, thanks!

book/src/app_validation.md
```diff
@@ -0,0 +1,26 @@
+# App Validation
+
+> **NOTE:** This chapter described the validation steps an application, using OpenMLS, has to perform for safe operation of the MLS protocol.
+>
+> **⚠️** This chapter is work in progress.
```
kkohbrok, line 5: Can you file an issue for this, or add a remark to an existing issue that we need to take care of this before the next release?

kkohbrok (changes requested):
Generally looks good to me, thanks! Just a two comments and two more requests:

- In addition to the book page you added, could you also note somewhere that OpenMLS only accepts credentials that we can deserialize? Probably in the book and/or in the struct docs.
- Could you add a changelog entry for this PR?

book/src/app_validation.md
franziskuskiefer: 👍🏻 Done

Cargo.toml
```diff
@@ -15,4 +15,8 @@ resolver = "2"
 
 # Central dependency management for some crates
 [workspace.dependencies]
-tls_codec = { version = "0.4.0", features = ["derive", "serde", "mls"] }
+tls_codec = { version = "0.4.2-pre.1", features = [
+    "derive",
+    "serde",
+    "mls",
+], git = "https://github.com/rustcrypto/formats" }
```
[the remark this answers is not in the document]
franziskuskiefer: We can't, even if wanted 😉 I filed #1505.

raphaelrobert (commented):
Thanks, this looks pretty straightforward in general. Just two questions.

openmls/src/credentials/mod.rs
```diff
@@ -187,7 +180,62 @@ pub enum MlsCredentialType {
...
+        // after this credential.
+
+        // The credential type is important, so we read that.
+        let credential_type = CredentialType::tls_deserialize(bytes)?;
+
+        // Now we don't know what we get unfortunately.
+        // We assume that it is a variable-sized vector. This works for the
+        // currently specified credentials and any other credential MUST be
+        // encoded in a vector as well. Otherwise this implementation will
+        // either crash or exhibit unexpected behaviour.
```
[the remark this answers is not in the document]
kkohbrok: My understanding is that while parsing can fail with an error, it doesn't have to. Depending on how the credential is encoded, the credential parser could terminate thinking that it has finished and the rest of the bytes that are actually part of the credential are interpreted as something else by the parser of the next field in the message. The behaviour of OpenMLS then depends on what all that looks like. It shouldn't be a security issue (because we still have signatures), but it could lead to errors in places that have nothing to do with the credential.

We might want to be more precise w.r.t. what we mean by crashes. I think the worst case is that OpenMLS will throw an error that is otherwise unrelated to the credential.

kkohbrok (dismissed):
Thanks for making the changes! Looks good to merge from my side.

openmls/src/group/tests/external_remove_proposal.rs
```diff
@@ -126,7 +127,11 @@ fn external_remove_proposal_should_remove_member(
     // get Bob's index
     let bob_index = alice_group
         .members()
-        .find(|member| member.credential.identity() == b"Bob")
+        .find(|member| {
+            let identity =
+                VLBytes::tls_deserialize_exact(member.credential.serialized_content()).unwrap();
+            identity.as_slice() == b"Bob"
+        })
```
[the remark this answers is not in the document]
franziskuskiefer: This could be a `member.credential.deserialized()`, but since this all tests it not worh rewriting everything.
But the function on `BasicCredential` was actually broken. I changed the test here to use it and fixed it.
I don't think it's worth rewriting all tests like this (also it's slower).

openmls/src/credentials/mod.rs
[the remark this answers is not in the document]
franziskuskiefer: Yep, what Konrad said. I rephrased slightly.
There is undefined behaviour here that we can not avoid. This is the trade-off for allowing other credentials and not refactoring everything for using generic credentials throughout (and figuring out a way to handle different credentials in a group).

franziskuskiefer:
I made another small fix (see inline comments).

kkohbrok (approved):

Issue #1319, opened 2023-03-13
chore: Update `openmls_evercrypt` chain.


duesee:
```
$ cargo clippy
...
warning: the following packages contain code that will be rejected by a future version of Rust: nom v5.1.2
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 6`
Process finished with exit code 0
```

```
$ cargo tree
...
├── openmls_evercrypt v0.1.0 (/home/duesee/repos/cryspen/mls/openmls/evercrypt_backend)
│   ├── evercrypt v0.0.11
│   │   ├── evercrypt-sys v0.0.9
│   │   │   └── libc v0.2.139
│   │   │   [build-dependencies]
│   │   │   └── bindgen v0.58.1
│   │   │       ├── bitflags v1.3.2
│   │   │       ├── cexpr v0.4.0
│   │   │       │   └── nom v5.1.2
│   │   │       │       └── memchr v2.5.0
...
```

franziskuskiefer:
Evercrypt shold be replaced. See #1250.

Issue #1250, opened 2023-02-03
[Feature request] libcrux backend


franziskuskiefer:
Replace the evercrypt crypto backend with the libcrux crate from [cryspen/libcrux](https://github.com/cryspen/libcrux).

keks:
This is being worked in on https://github.com/openmls/openmls/pull/1490

PR #1490, opened 2024-02-05
Add provider backed by libcrux

keks:
This draft PR adds a provider backed by libcrux. Note that libcrux is not supported on all platforms at the moment, so the tests do not run everywhere.

Fixes https://github.com/openmls/openmls/issues/1319, https://github.com/openmls/openmls/issues/1250


keks:
The failing coverage build looks like [this issue](https://github.com/dalek-cryptography/curve25519-dalek/pull/619) and the spot fuzz fails to resolve `api.github.com`. I've restarted it, maybe it works this time.

libcrux_crypto/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "openmls_libcrux_crypto"
+version = "0.1.0"
+edition = "2021"
```
kkohbrok, line 4: Could you add some metadata here (authors license, description, etc.) like we have in the other Cargo.tomls?

libcrux_crypto/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
kkohbrok, line 6: This line can probably go.

libcrux_crypto/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+[dependencies]
+getrandom = "0.2.12"
+libcrux = { git = "https://github.com/cryspen/libcrux", features = ["rand"] }
```
kkohbrok, line 10: Any chance we can pin this on a version?

libcrux_crypto/src/crypto.rs
```diff
@@ -0,0 +1,578 @@
...
+use openmls_traits::crypto::OpenMlsCrypto;
+use openmls_traits::types::{
+    AeadType, Ciphersuite, CryptoError, ExporterSecret, HashType, HpkeAeadType, HpkeCiphertext,
+    HpkeConfig, HpkeKdfType, HpkeKemType, HpkeKeyPair, KemOutput, SignatureScheme,
+};
+
+use rand::CryptoRng;
+use tls_codec::SecretVLBytes;
+
+const MAX_DATA_LEN: usize = 0x10000000;
```
kkohbrok, line 14: Where is this number coming from? Could you add a code comment?

libcrux_crypto/src/crypto.rs
```diff
@@ -0,0 +1,578 @@
...
+    }
+
+    fn hpke_seal(
+        &self,
+        config: HpkeConfig,
+        pk_r: &[u8],
+        info: &[u8],
+        aad: &[u8],
+        ptxt: &[u8],
+    ) -> HpkeCiphertext {
```
kkohbrok, line 216: Would it make sense to patch the crypto provider trait to return a `Result` here? There are quite a few unwraps in this function.

libcrux_crypto/src/crypto.rs
```diff
@@ -0,0 +1,578 @@
...
+    ) -> Result<(KemOutput, ExporterSecret), CryptoError> {
+        let config = hpke_config(config);
+        let randomness = self
+            .drbg
+            .lock()
+            .map_err(|_| CryptoError::CryptoLibraryError)?
+            .generate_vec(libcrux::hpke::kem::Nsk(config.1))
+            .map_err(|_| CryptoError::CryptoLibraryError)?;
+
+        let pk_r = libcrux::hpke::kem::DeserializePublicKey(config.1, pk_r).unwrap();
```
kkohbrok, line 280: Can't this be turned into a `CryptoError`?

libcrux_crypto/src/crypto.rs
```diff
@@ -0,0 +1,578 @@
...
+
+const MAX_DATA_LEN: usize = 0x10000000;
+
+/// The libcrux-backed cryptography provider for OpenMLS
+pub struct CryptoProvider {
+    drbg: Mutex<Drbg>,
+}
+
+impl Default for CryptoProvider {
+    fn default() -> Self {
```
kkohbrok, line 22: Is it that important for the provider to implement `Default` given that there are two unwraps here?

libcrux_crypto/src/crypto.rs
```diff
@@ -0,0 +1,578 @@
...
+
+        libcrux::hpke::Context_Export(config, &ctx, exporter_context.to_vec(), exporter_length)
+            .map_err(|_| CryptoError::ExporterError)
+            .map(ExporterSecret::from)
+    }
+
+    fn derive_hpke_keypair(&self, config: HpkeConfig, ikm: &[u8]) -> HpkeKeyPair {
+        let config = hpke_config(config);
+        let HPKEConfig(_, alg, _, _) = config;
+        let (sk, pk) = hpke::kem::DeriveKeyPair(alg, ikm).unwrap(); // XXX return err
```
kkohbrok, line 312: ☝🏻

kkohbrok (changes requested):
Thanks for the PR! Generally looks good to me. Some comments inline and two observations:

- There is quite a lot of unwrapping going on that we are generally trying to avoid. I commented on some instances. I'm not sure how to handle the mutex access, but it might make sense to have an error for that as well.
- Libcrux doesn't support all OSs and architectures and some ciphersuites only if AES-NI is available. That's all fine, but we should probably document that somewhere. Maybe on the `CryptoProvider` struct? Or a page in the book?

libcrux_crypto/src/crypto.rs
keks: Don't we use default all over the tests in the OpenMLS tests? If we don't implement it, I don't think we would be able to run the tests against this provider.

kkohbrok: Fair enough, in which case the unwraps are okay, as long as the implementation is qualified with the right config flags.

libcrux_crypto/src/crypto.rs
keks: noticed that the existing provider does the same unwraps, and that it might have been better to do this in a separate PR, but now it's done

libcrux_crypto/src/crypto.rs
keks: I was sure I got that from the called code, but I couldn't find it. Reading the AEAD RFC told me that this number is different for each ciphersuite, so I just removed it and let libcrux take care of it. The errors will be less specific but such is life.

libcrux_crypto/Cargo.toml
keks: I pinned it to a commit; we don't have a recent release yet

keks:
Thanks! Regarding your two points: I removed all the unwraps except the ones in `default` and replaced them with errors. I also added a section in the book that mentions the limitations.

kkohbrok (approved):
Thanks for addressing my concerns! Let's get this in and then deal with the (potentially) panicking`Default` implementations in a follow-up.

Issue #1130, opened 2022-11-30
Missing Test


franziskuskiefer:
This issue collects missing test cases

- [ ] `self_update` (with and without key pair) but don't merge the commit

PR #1491, opened 2024-02-05
Fix validation bug

### LINKED ISSUE #1487: Validation bug around Group Context Extension proposals
There is a subtle bug in our validation of Group Context Extension propposals. The validation happens here: https://github.com/openmls/openmls/blob/5052cbef2731214911ba69ea5a12e7a64481f45c/openmls/src/group/public_group/validation.rs#L517-L564

What we do is that for every extension, we check that all members support it.

What we should do is
- when setting RequiredCapabilites, check that all members support the listed extension types
- when setting any extension, check that it is in required capabilties (not necessarily explicit, the basic extensions are always supported)

keks:
This PR addresses issue #1487. The problem is that currently the list of extensions in a `GroupContextExtensions` proposal is compared against the capabilities of the current members. Instead, it needs to

1. check that the `RequiredCapabilities` of the group are satisfied by all members of the group
2. check that the types of all extensions in the proposal are in the `RequiredCapabilities`.

This PR also adds a method on the `RequiredCapabilitiesExtension` for checking whether an extension type is supported. This is needed because the extension types defined in the RFC are supported implicitly, so we also have to compare against the defaults. The defaults are defined close to the `Capabilities` struct and used to be private to these modules. I made the one for extension types `pub(crate)` so I can check against it.

Finally, I haven't written tests for this, yet. It seems this behaviour was barely tested before, as [the tests](https://github.com/openmls/openmls/blob/5052cbef2731214911ba69ea5a12e7a64481f45c/openmls/src/group/core_group/test_proposals.rs#L524-L540) seem to be commented out already. This is already tracked in #1130.

How do you all feel about merging this as-is and fixing the tests later (and opening an issue for it, of course)?

Fixes #1487

### DIFF
--- a/openmls/src/group/public_group/validation.rs
+++ b/openmls/src/group/public_group/validation.rs
@@ -524,7 +524,7 @@ impl PublicGroup {
         let iter = proposal_queue.filtered_by_type(ProposalType::GroupContextExtensions);
 
         for (i, queued_proposal) in iter.enumerate() {
-            // There may only be one group context extionsion proposal. Return an error if there are more
+            // There must at most be one group context extionsion proposal. Return an error if there are more
             if i > 0 {
                 return Err(GroupContextExtensionsProposalValidationError::TooManyGCEProposals);
             }
--- a/openmls/src/group/public_group/validation.rs
+++ b/openmls/src/group/public_group/validation.rs
@@ -536,12 +536,11 @@ impl PublicGroup {
                         Some(required_capabilities_new) => {
                             // a required capabilities extension in the group context extensions is
                             // only valid if all member support it.
-                            self.check_extension_support(&required_capabilities_new.extension_types()).map_err(|_| GroupContextExtensionsProposalValidationError::RequiredExtensionNotSupportedByAllMembers)?;
+                            self.check_extension_support(required_capabilities_new.extension_types()).map_err(|_| GroupContextExtensionsProposalValidationError::RequiredExtensionNotSupportedByAllMembers)?;
                             required_capabilities_new
                         }
                         None => self
                             .required_capabilities()
-                            .map(|required_capabilities| required_capabilities)
                             .unwrap_or(&default_required_capabilities),
                     };
 
--- a/openmls/src/group/public_group/validation.rs
+++ b/openmls/src/group/public_group/validation.rs
@@ -521,36 +521,47 @@ impl PublicGroup {
         &self,
         proposal_queue: &ProposalQueue,
     ) -> Result<(), GroupContextExtensionsProposalValidationError> {
-        let mut iter = proposal_queue.filtered_by_type(ProposalType::GroupContextExtensions);
+        let iter = proposal_queue.filtered_by_type(ProposalType::GroupContextExtensions);
 
-        match iter.next() {
-            Some(queued_proposal) => match queued_proposal.proposal() {
+        for (i, queued_proposal) in iter.enumerate() {
+            // There may only be one group context extionsion proposal. Return an error if there are more
+            if i > 0 {
+                return Err(GroupContextExtensionsProposalValidationError::TooManyGCEProposals);
+            }
+
+            match queued_proposal.proposal() {
                 Proposal::GroupContextExtensions(extensions) => {
                     let required_capabilities_in_proposal =
                         extensions.extensions().required_capabilities();
 
+                    // Prepare the empty required capabilities in case there is no
+                    // RequiredCapabilitiesExtension in the proposal
                     let default_required_capabilities =
                         RequiredCapabilitiesExtension::new(&[], &[], &[]);
 
+                    // If there is a RequiredCapabilitiesExtension in the proposal, validate it and
+                    // use that. Otherwise, use the empty default one.
                     let required_capabilities = match required_capabilities_in_proposal {
                         Some(required_capabilities_new) => {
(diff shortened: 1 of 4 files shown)


raphaelrobert:
What is the rationale behind skipping the tests?

Unless this is a blocker for something else or the test cannot be re-enabled without major refactoring (which then should have its own issue) there should be tests for validation checks.

keks:
The problem I am facing is that the validation bug was around the capabilities of the parties, which the current test functions do not really allow me to set. I would somehow get Alice's and Bob's capabilities into `setup_client`, which then pass it into the `KeyPackageBundle::new` which then would have to add it to the `KeyPackageBuilder`. As far as I can tell, only the last piece of that chain is implemented currently.

I think this does require some refactoring around how we can set up clients in tests. But right now I am not sure what the best way would be - `KeyPackageBundle::new` is a test helper function, and writing a builder for it seems somewhat overkill. And it's still not clear how to expose that functionality to `setup_client` -- maybe it could take a closure that performs operations on the builder. However, I think think these are changes that warrant their own issue and PR.

raphaelrobert (changes requested):
Thanks for catching this. According to the spec there's also another place where this should be checked: when a new member is added.
>  For new members, support for required capabilities is enforced by existing members during the application of Add commits. 

Did you check that we do the right thing there?

Otherwise we need more comments.

keks:
Thank you for the review @raphaelrobert! I have added some more comments and tried making the code more readable in general, by not handling the iterator directly, but instead looping over it.

> Did you check that we do the right thing there?

Yes, that is checked here: https://github.com/openmls/openmls/blob/main/openmls/src/group/public_group/validation.rs#L307-L316

openmls/src/group/public_group/validation.rs
```diff
@@ -520,18 +521,49 @@ impl PublicGroup {
         &self,
         proposal_queue: &ProposalQueue,
     ) -> Result<(), GroupContextExtensionsProposalValidationError> {
-        let mut iter = proposal_queue.filtered_by_type(ProposalType::GroupContextExtensions);
+        let iter = proposal_queue.filtered_by_type(ProposalType::GroupContextExtensions);
 
-        match iter.next() {
-            Some(queued_proposal) => match queued_proposal.proposal() {
+        for (i, queued_proposal) in iter.enumerate() {
+            // There may only be one group context extionsion proposal. Return an error if there are more
```
raphaelrobert, line 527:
```suggestion
            // There must at most be one group context extionsion proposal. Return an error if there are more
```

raphaelrobert:
```suggestion
            // There may only be one group context extension proposal. Return an error if there are more
```

openmls/src/group/public_group/validation.rs
```diff
@@ -520,18 +521,49 @@ impl PublicGroup {
...
+                    // RequiredCapabilitiesExtension in the proposal
+                    let default_required_capabilities =
+                        RequiredCapabilitiesExtension::new(&[], &[], &[]);
+
+                    // If there is a RequiredCapabilitiesExtension in the proposal, validate it and
+                    // use that. Otherwise, use the empty default one.
+                    let required_capabilities = match required_capabilities_in_proposal {
+                        Some(required_capabilities_new) => {
+                            // If a group context extensions proposal updates the required capabilities, we
+                            // need to chec that these are satisfied for all existing members of the group.
```
raphaelrobert, line 547:
```suggestion
                            // need to check that these are satisfied for all existing members of the group.
```

openmls/src/group/public_group/validation.rs
```diff
@@ -520,18 +521,49 @@ impl PublicGroup {
...
+                            self.check_extension_support(required_capabilities_new.extension_types()).map_err(|_| GroupContextExtensionsProposalValidationError::RequiredExtensionNotSupportedByAllMembers)?;
+                            required_capabilities_new
+                        }
+                        None => &default_required_capabilities,
+                    };
+
+                    // Make sure that all other extensions are know to be supported, by checking
```
raphaelrobert, line 554:
```suggestion
                    // Make sure that all other extensions are known to be supported, by checking
```

openmls/src/group/public_group/validation.rs
```diff
@@ -520,18 +521,49 @@ impl PublicGroup {
...
+                    // that they are required to be supported by the required capabilities.
```
raphaelrobert, line 555:
```suggestion
                    // that they are included in the required capabilities.
```

openmls/src/group/public_group/validation.rs
```diff
@@ -540,14 +572,17 @@ impl PublicGroup {
...
-            None => Ok(()),
-        }
+        Ok(())
+
+        // There may only be one group context extionsion proposal. Return an error if there are
+        // more
+        // match iter.next() {
+        //     Some(_) => Err(GroupContextExtensionsProposalValidationError::TooManyGCEProposals),
+        //     None => Ok(()),
+        //}
```
raphaelrobert, line 585: Leftover?

raphaelrobert (changes requested):
Thanks, just some nits in the comments

openmls/src/group/public_group/validation.rs
keks: Oops!

raphaelrobert (approved):

PR #1495, opened 2024-02-07
Improve processing

### LINKED ISSUE #1494: Wrong sample code in processing section of book
`impl From<MlsMessageIn> for ProtocolMessage` is guarded by test flags but is used in the sample code in the book.

raphaelrobert:
Fixes #1494.

This PR does the following:

 - Fixes and expands the section about message processing in the book
 - Introduces a `TryFrom` conversion from `MlsMessageIn` into `ProtocolMessage`
 - Renames `MlsMessage[In/Out]Body` to `MlsMessageBody[In/Out]` to harmonize it with the naming scheme used by other structs
 - Refactors the tests because the old way to convert `MlsMessageIn` into `ProtocolMessage` is no longer possible

### DIFF
--- a/book/src/user_manual/processing.md
+++ b/book/src/user_manual/processing.md
@@ -19,11 +19,11 @@ corresponding group.
 
 `MlsMessageIn` can carry all MLS messages, but only `PrivateMessageIn` and
 `PublicMessageIn` are processed in the context of a group. In OpenMLS these two
-message types are combined into a `ProtocolMessage` `enum`. There 3 ways to
+message types are combined into a `ProtocolMessage` `enum`. There are 3 ways to
 extract the messages from an `MlsMessageIn`:
 
 1. `MlsMessageIn.try_into_protocol_message()` returns a `Result<ProtocolMessage, ProtocolMessageError>`
-2. `ProtocolMessage::try_from(m: MlsMessageIn)` retruns a `Result<ProtocolMessage, ProtocolMessageError>`
+2. `ProtocolMessage::try_from(m: MlsMessageIn)` returns a `Result<ProtocolMessage, ProtocolMessageError>`
 3. `MlsMessageIn.extract()` returns an `MlsMessageBodyIn` `enum` that has two
    variants for `PrivateMessageIn` and `PublicMessageIn`
 
--- a/book/src/user_manual/processing.md
+++ b/book/src/user_manual/processing.md
@@ -12,11 +12,30 @@ Incoming messages can be deserialized from byte slices into an `MlsMessageIn`:
 
 If the message is malformed, the function will fail with an error.
 
-## Processing messages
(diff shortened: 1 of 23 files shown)


book/src/user_manual/processing.md
```diff
@@ -12,11 +12,30 @@ Incoming messages can be deserialized from byte slices into an `MlsMessageIn`:
...
-encrypted, it will be decrypted automatically. This step performs all syntactic
-and semantic validation checks and verifies the message's signature:
+## Processing messages in groups
+
+In the next step, the message needs to be processed in the context of the
+corresponding group.
+
+`MlsMessageIn` can carry all MLS messages, but only `PrivateMessageIn` and
+`PublicMessageIn` are processed in the context of a group. In OpenMLS these two
+message types are combined into a `ProtocolMessage` `enum`. There 3 ways to
```
JonasVautherin, line 22:
```suggestion
message types are combined into a `ProtocolMessage` `enum`. There are 3 ways to
```

JonasVautherin (commented):
Just two typos, otherwise this looks good to me :+1:.

openmls/src/framing/message_in.rs
```diff
@@ -97,24 +97,29 @@ impl MlsMessageIn {
...
     /// with the [`MlsGroup`] API.
-    pub fn extract(self) -> MlsMessageInBody {
+    pub fn extract(self) -> MlsMessageBodyIn {
         self.body
     }
 
+    /// Try to convert the message into a [`ProtocolMessage`].
+    pub fn try_into_protocol_message(self) -> Result<ProtocolMessage, ProtocolMessageError> {
+        self.try_into()
+    }
```
kkohbrok, line 117: Do we really need this in addition to the `TryFrom` implementation? I would think some documentation on the `ProtocolMessage` and `MlsMessageIn` structs should be enough.

openmls/src/framing/message_in.rs
```diff
@@ -243,13 +248,14 @@ impl From<PublicMessageIn> for ProtocolMessage {
     }
 }
 
-#[cfg(any(feature = "test-utils", test))]
-impl From<MlsMessageIn> for ProtocolMessage {
-    fn from(msg: MlsMessageIn) -> Self {
+impl TryFrom<MlsMessageIn> for ProtocolMessage {
```
kkohbrok, line 251: If we change this implementation and make it available for non-test purposes, wouldn't it make sense to change the `impl Into` input in `fn process_message` to `impl TryFrom`? I tried it and it's a bit annoying because we need to put constraints on the Error type in TryFrom, but I think it makes for a nicer API.

kkohbrok (commented):
Thanks! Looks good to me in general. Just two comments/suggestions.

kkohbrok (approved):

PR #1499, opened 2024-02-11
Cleanup unused CoreGroup methods

### LINKED ISSUE #645: Cleanup unused CoreGroup methods

raphaelrobert:
Fixes #645.

This PR does the following:

 - Remove test functions from the CoreGroup that are not useful along with the corresponding tests.
 - Moves code in `core_group/mod.rs`
 - Refactor unnecessary errors
 - Minor renaming

### DIFF
--- a/openmls/src/framing/private_message.rs
+++ b/openmls/src/framing/private_message.rs
@@ -129,7 +129,8 @@ impl PrivateMessage {
         )
     }
 
-    /// Internal function to encrypt content.
+    /// Internal function to encrypt content. The extra message header is only used
+    /// for tests. Otherwise, the data from the given `AuthenticatedContent` is used.
     fn encrypt_content(
         test_header: Option<MlsMessageHeader>,
         public_message: &AuthenticatedContent,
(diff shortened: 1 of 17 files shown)


openmls/src/framing/private_message.rs
```diff
@@ -129,8 +129,7 @@ impl PrivateMessage {
         )
     }
 
-    /// Internal function to encrypt content. The extra message header is only used
-    /// for tests. Otherwise, the data from the given `AuthenticatedContent` is used.
+    /// Internal function to encrypt content.
```
kkohbrok, line 132: Is there a particular reason you're removing the remark about the extra header here?

kkohbrok (commented):
Thanks! One question, otherwise this looks good!

openmls/src/framing/private_message.rs
raphaelrobert: It seemed outdated and no longer necessary.

kkohbrok: It does (still) describe what's going on in the code and I would say it's useful for anyone who is looking to understand what this function does.

kkohbrok (approved):

PR #1500, opened 2024-02-11
Introduce InitKey type for KeyPackages

### LINKED ISSUE #1264: Introduce type for `InitKey`
Maybe not in this PR, but it would probably make sense to have a dedicated type for `InitKey` in the same way we already have for `Encryptionkey`.

_Originally posted by @kkohbrok in https://github.com/openmls/openmls/pull/1260#discussion_r1101275451_

raphaelrobert:
Fixes #1264.

This PR introduces as distinct `InitKey` type for KeyPackages. This change makes it less likely to confuse the init key with other HPKE public keys.

### DIFF
--- a/openmls/src/key_packages/mod.rs
+++ b/openmls/src/key_packages/mod.rs
@@ -123,7 +123,9 @@ use openmls_traits::{
     OpenMlsProvider,
 };
 use serde::{Deserialize, Serialize};
-use tls_codec::{Serialize as TlsSerializeTrait, TlsSerialize, TlsSize};
+use tls_codec::{
+    Serialize as TlsSerializeTrait, TlsDeserialize, TlsDeserializeBytes, TlsSerialize, TlsSize,
+};
 
 // Private
 use errors::*;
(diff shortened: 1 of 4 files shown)


openmls/src/key_packages/mod.rs
```diff
@@ -216,6 +218,48 @@ pub(crate) struct KeyPackageCreationResult {
...
+        &self.key
+    }
+
+    /// Return the internal [`HpkePublicKey`] as a slice.
+    pub fn as_slice(&self) -> &[u8] {
+        self.key.as_slice()
+    }
+}
+
+impl From<HpkePublicKey> for InitKey {
```
kkohbrok, line 249: I think this implementation is not actually needed.

kkohbrok (changes requested):
Thanks for the PR! Just one comment.

It's a bit annoying that we need the `From<Vec<u8>>` and the `as_slice`. I think with a bit more refactoring we could make those go away and introduce a bit more type safety in some places. But that doesn't have to happen in this PR.

kkohbrok (approved):

PR #1506, opened 2024-02-21
Add StagedCoreWelcome and StagedWelcome

keks:
This commit adds the above types to make it possible to inspect the content of a Welcome message before joining a group.

### DIFF
--- a/openmls/src/group/core_group/new_from_welcome.rs
+++ b/openmls/src/group/core_group/new_from_welcome.rs
@@ -219,7 +219,7 @@ impl StagedCoreWelcome {
                 .derive_path_secrets(
                     provider.crypto(),
                     ciphersuite,
-                    path_secret.clone(),
+                    path_secret,
                     welcome_sender_index,
                     own_leaf_index,
                 )


keks:
Sorry, I should have been more explicit in the commit/PR message, not just the comments. The main point right now is to be able to see who signed the welcome. This information gets thrown away when building the group, so it's not possible to extract that afterwards.

On 21 February 2024 18:01:33 CET, raphaelrobert ***@***.***> wrote:
>Thanks! Can you describe what the advantage is when using a `StagedMlsGroup`? Or rather, what is the disadvantage when joining a group right away? Since there is no need to persist the group, you could join it, inspect it, and finally discard it.
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/openmls/openmls/pull/1506#issuecomment-1957313319
>You are receiving this because you authored the thread.
>
>Message ID: ***@***.***>
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

openmls/src/group/core_group/new_from_welcome.rs
```diff
@@ -11,6 +11,292 @@ use crate::{
...
+    /// Note: calling this function will consume the key material for decrypting the [`Welcome`]
+    /// message, even if the caller does not turn the [`StagedCoreGroup`] into a [`CoreGroup`].
+    pub fn new_from_welcome<KeyStore: OpenMlsKeyStore>(
+        welcome: Welcome,
+        ratchet_tree: Option<RatchetTreeIn>,
+        key_package_bundle: KeyPackageBundle,
+        provider: &impl OpenMlsProvider<KeyStoreProvider = KeyStore>,
+        mut resumption_psk_store: ResumptionPskStore,
+    ) -> Result<Self, WelcomeError<KeyStore::Error>> {
+        log::debug!("CoreGroup::new_from_welcome_internal");
```
franziskuskiefer, line 27: This duplicates quite a bit of code. Can we find a way to share the code? For example calling this in `new_from_welcome` or factoring out the common code into a helper function.

openmls/src/group/mls_group/creation.rs
```diff
@@ -189,3 +189,100 @@ impl MlsGroup {
...
     }
 }
+
+impl StagedMlsGroup {
+    /// Creates a new staged group from a [`Welcome`] message. Returns an error
+    /// ([`WelcomeError::NoMatchingKeyPackage`]) if no [`KeyPackage`]
+    /// can be found.
+    /// Note: calling this function will consume the key material for decrypting the [`Welcome`]
+    /// message, even if the caller does not turn the [`StagedMlsGroup`] into an [`MlsGroup`].
+    // TODO: #1326 This should take an MlsMessage rather than a Welcome message.
```
franziskuskiefer, line 199: Should we do that here? Since it's a new funtion we can make it better from the start 🙂

openmls/src/group/mls_group/mod.rs
```diff
@@ -433,3 +433,14 @@ pub enum InnerState {
...
     Persisted,
 }
+
+/// A [`StagedMlsGroup`] can be inspected and then turned into a [`MlsGroup`].
+/// This allows checking who authored the Welcome message.
+#[derive(Debug)]
+pub struct StagedMlsGroup {
+    // The group configuration. See [`MlsGroupJoinConfig`] for more information.
+    mls_group_config: MlsGroupJoinConfig,
+    // the internal `CoreGroup` used for lower level operations. See `CoreGroup` for more
```
franziskuskiefer, line 443:
```suggestion
    // The internal `CoreGroup` used for lower level operations. See `CoreGroup` for more
```

franziskuskiefer (changes requested):
Thanks, generally looks good to me with a couple nits.

raphaelrobert (changes requested):
Thanks, this is generally a valid thing we need. A few change requests:

 - We already have a similar concept with `Commit` and `StagedCommit` where the new group state is effectively hidden behind the `StagedCommit`. I think we should align the logic here and have `Welcome` and `StagedWelcome` (instead of `StagedMlsGroup`).
 - We should drop `MlsGroup::new_from_welcome()` and force applications to use a `StagedWelcome`
 - This should be explained in the book.
 - It's debatable why we shouldn't consume private keys when there is no LastResort extension.
 - Optional: a `StagedWelcome` could expose the public tree and maybe other things as well besides the signer. This could be a follow-up PR though.

keks:
@franziskuskiefer this is ready for review round two now

franziskuskiefer (changes requested):
Thanks. Looks good to me with a few nits.

Also, please add this change to the changelog with all the public facing functions that changed.

openmls/src/group/mls_group/mod.rs
```diff
@@ -11,12 +11,16 @@ use crate::{
...
     group::*,
     key_packages::{KeyPackage, KeyPackageBundle},
-    messages::{proposals::*, Welcome},
+    messages::proposals::*,
     schedule::ResumptionPskSecret,
     treesync::{node::leaf_node::LeafNode, RatchetTree},
 };
 use openmls_traits::{key_store::OpenMlsKeyStore, types::Ciphersuite, OpenMlsProvider};
 
+// this import is needed for doc comments
```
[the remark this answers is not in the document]
keks: I am not a fan of markdown inline links because it hurts readability, but apparently reference-style links work!

franziskuskiefer (dismissed):
Thanks

openmls/src/extensions/test_extensions.rs
```diff
@@ -104,18 +104,18 @@ fn ratchet_tree_extension(ciphersuite: Ciphersuite, provider: &impl OpenMlsProvi
...
         None,
         bob_key_package_bundle,
         provider,
         ResumptionPskStore::new(1024),
-    ) {
-        Ok(g) => g,
-        Err(e) => panic!("Could not join group with ratchet tree extension {e}"),
-    };
+    )
+    .unwrap_or_else(|e| panic!("Could not stage group join with ratchet tree extension: {e}"))
```
raphaelrobert, line 116: Just a nit, but `expect()` does exactly that and that's what we use elsewhere.

openmls/src/group/mls_group/creation.rs
```diff
@@ -105,87 +178,54 @@ impl MlsGroup {
...
             welcome,
             ratchet_tree,
             key_package_bundle,
             provider,
             resumption_psk_store,
         )?;
-        group.set_max_past_epochs(mls_group_config.max_past_epochs);
 
-        let mls_group = MlsGroup {
+        let mls_group = StagedWelcome {
```
raphaelrobert, line 189: THat variable name is now misleading

openmls/src/group/mls_group/creation.rs
```diff
@@ -60,27 +60,100 @@ impl MlsGroup {
...
+        Ok((
+            mls_group,
+            public_message.into(),
+            create_commit_result.group_info,
+        ))
+    }
+}
+
+impl StagedWelcome {
+    /// Creates a new staged group from a [`Welcome`] message. Returns an error
```
raphaelrobert, line 127:
```suggestion
    /// Creates a new staged welcome from a [`Welcome`] message. Returns an error
```

openmls/src/group/core_group/new_from_welcome.rs
```diff
@@ -238,21 +213,100 @@ impl CoreGroup {
...
+        let sender_index = self.welcome_sender_index();
+        self.public_group
+            .leaf(sender_index)
+            .ok_or(LibraryError::custom(
+                "no leaf with given welcome sender index exists",
+            ))
+    }
+
+    /// Consumes the [`StagedCoreWelcome`] and returns the respective [`CoreGroup`].
+    pub fn into_core_group<KeyStore: OpenMlsKeyStore>(
```
raphaelrobert, line 247: I suppose you only want to derive the path secrets and generate the keypairs in this step because you assume the application has done some inspection before and might abort the process, in which case we would have done something expensive for no reason.

I see how that makes sense from an efficiency perspective, but it means that `StagedWelcome::into_group()` can now fail because the Welcome message was invalid. This is weird from an application's point of view, because usually all possible checks are done before the application does the inspection.

I don't have a super strong opinion about this, but I think the deviation from the current norm should be strongly motivated (e.g. explain how important the efficiency benefit is) and properly documented. Otherwise I'd say we align this with the rest of the processing and do this step in `new_from_welcome()`.

raphaelrobert (changes requested):
Thank you for making the changes! I have some more comments and questions.

book/src/user_manual/join_from_welcome.md
```diff
@@ -1,10 +1,19 @@
 # Join a group from a Welcome message
 
-To join a group from a `Welcome` message, a new `MlsGroup` can be instantiated directly from the `Welcome` message and an `MlsGroupJoinConfig` (see [Group configuration](./group_config.md) for more details).  If the group configuration does not use the ratchet tree extension, the ratchet tree needs to be provided.
+To join a group from a `Welcome` message, a new `MlsGroup` can be instantiated from
+the `MlsMessageIn` message containing the `Welcome` and an `MlsGroupJoinConfig`
+(see [Group configuration](./group_config.md) for more details).  This is a
+two-step process: a `StagingMlsJoinFromWelcome` is constructed from the `Welcome`
```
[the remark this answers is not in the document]
keks: I just found that and was fixing this :sweat_smile: thanks!

openmls/src/extensions/test_extensions.rs
keks: `expect` does not support formating, does it?

keks: Oh, but it prints the error, right.

openmls/src/group/core_group/new_from_welcome.rs
keks: Good point. I was able to move the operations that might fail based for malformed messages into `new_from_welcome`. We still have to return a `Result` because storing the key might fail, though.

raphaelrobert (approved):
Thanks, let's get this in!

PR #1516, opened 2024-02-29
Make `MlsGroup::clear_pending_proposals` public

### LINKED ISSUE #1515: [Feature request] `MlsGroup::clear_pending_proposals`
I have a PR and would like to contribute.  Thanks!

Not exactly -- I've encountered a situation where I want to completely clear a group's `ProposalStore` in the event that a call to `MlsGroup::commit_to_pending_proposals` fails.  There is no easy way to do that, as best I can tell.  

A method for `MlsGroup` in the public API that clears the internal `ProposalStore` of all `QueuedProposal` objects.

Alternative course of action would be to generate a `ProposalRef` for each `QueuedProposal` to remove and then remove each one individually.  However, I cannot figure out how to do that using the public API.

**Additional context**
Here's an example of how I would use the requested method:
```rust
fn process_proposal(
        &mut self,
        proposal_in: openmls::group::QueuedProposal,
        group_id: &openmls::group::GroupId,
    ) -> SimpleMlsProcessingOutput {
        let group = self.groups.get_mut(group_id).unwrap();
        log::debug!("((( process proposal )))");
        group.store_pending_proposal(proposal_in);
        if group_id == &self.my_group_id {
            log::debug!("((( proposal is for my group )))");
            let (commit, welcome_opt_in, _) =
                match group.commit_to_pending_proposals(&self.backend, &self.signature_keypair) {
                    Ok(tuple) => tuple,
                    Err(e) => {
                        log::warn!("{e:?}");
                        group.clear_pending_proposals();
                        group.commit_to_pending_proposals(&self.backend, &self.signature_keypair)?
                    }
                };
            group.merge_pending_commit(&self.backend)?;
            log::debug!("{:?}", commit);
            use openmls::prelude::TlsSerializeTrait;
            let welcome_opt = match welcome_opt_in {
                Some(welcome) => {
                    log::debug!("{:?}", welcome);
                    Some(welcome.tls_serialize_detached()?)
                }
                None => None,
            };
            Ok((Some(commit.tls_serialize_detached()?), welcome_opt, None))
        } else {
            Ok((None, None, None))
        }
    }
```

josephlukefahr:
Closes #1515 

This PR provides a public method for `MlsGroup` that clears the internal `ProposalStore`.  Since this method has already been implemented in the codebase, this PR makes it public.


josephlukefahr:
Sure.  Two questions:
* I noticed that some of the other proposal processing methods (e.g. `MlsGroup::store_pending_proposal`) include the following statement: `self.flag_state_change()`; do I need to make that call also?
* I don't have unit tests in this PR; do I need any?  If so, where should they go?

raphaelrobert:
>     * I noticed that some of the other proposal processing methods (e.g. `MlsGroup::store_pending_proposal`) include the following statement: `self.flag_state_change()`; do I need to make that call also?

Yes, anything that changes the group state should arm that flag. You could make this fancy and check if the proposal store is actually non-empty before arming the flag.

>     * I don't have unit tests in this PR; do I need any?  If so, where should they go?

That would be nice, you could create a file in `src/group/tests/` for that. You can look at `test_group.rs` in that folder for inspiration.

openmls/src/group/mls_group/mod.rs
```diff
@@ -407,8 +407,7 @@ impl MlsGroup {
     }
 
     /// Clear the pending proposals.
-    #[cfg(test)]
-    pub(crate) fn clear_pending_proposals(&mut self) {
+    pub fn clear_pending_proposals(&mut self) {
```
raphaelrobert, line 410:
```suggestion
    ///
    /// Warning: Once the pending proposals are cleared it will be impossible to process
    /// a Commit message that references proposals those proposals. Only use this
    /// function as a last resort, e.g. when a call to
    /// `MlsGroup::commit_to_pending_proposals` fails.
    pub fn clear_pending_proposals(&mut self) {
```

josephlukefahr: 7cc8b08

josephlukefahr:
> Yes, anything that changes the group state should arm that flag. You could make this fancy and check if the proposal store is actually non-empty before arming the flag.

34f17d0

openmls/src/group/mls_group/mod.rs
```diff
@@ -406,10 +406,20 @@ impl MlsGroup {
...
     }
 
-    /// Clear the pending proposals.
-    #[cfg(test)]
-    pub(crate) fn clear_pending_proposals(&mut self) {
-        self.proposal_store.empty()
+    /// Clear the pending proposals, if the proposal store is not empty.
+    ///
+    /// Warning: Once the pending proposals are cleared it will be impossible to process
+    /// a Commit message that references proposals those proposals. Only use this
```
raphaelrobert, line 412:
```suggestion
    /// a Commit message that references those proposals. Only use this
```

openmls/src/group/mls_group/mod.rs
```diff
@@ -406,10 +406,20 @@ impl MlsGroup {
...
+    /// function as a last resort, e.g. when a call to
+    /// `MlsGroup::commit_to_pending_proposals` fails.
+    pub fn clear_pending_proposals(&mut self) {
+        // If the proposal store is not empty...
+        if !self.proposal_store.is_empty() {
+            // Empty the proposal store
+            self.proposal_store.empty();
+            // Since the state of the group might be changed, arm the state flag
```
raphaelrobert, line 420:
```suggestion
            // Since the state of the group is changed, arm the state flag
```

openmls/src/group/mls_group/mod.rs
```diff
@@ -406,10 +406,20 @@ impl MlsGroup {
...
-    #[cfg(test)]
-    pub(crate) fn clear_pending_proposals(&mut self) {
-        self.proposal_store.empty()
+    /// Clear the pending proposals, if the proposal store is not empty.
+    ///
+    /// Warning: Once the pending proposals are cleared it will be impossible to process
+    /// a Commit message that references proposals those proposals. Only use this
+    /// function as a last resort, e.g. when a call to
+    /// `MlsGroup::commit_to_pending_proposals` fails.
+    pub fn clear_pending_proposals(&mut self) {
```
raphaelrobert, line 415: Can you please move this up to the block of public functions? It is still in the block of test functions

raphaelrobert (changes requested):
Thanks, just a few nits and we are good to go

openmls/src/group/mls_group/mod.rs
josephlukefahr: 8b9cc80

openmls/src/group/mls_group/mod.rs
josephlukefahr: 8b9cc80

openmls/src/group/mls_group/mod.rs
josephlukefahr: 8b9cc80

josephlukefahr:
From stdout for the failing test:
```
                                            leaf_node_source: KeyPackage(
                                                Lifetime {
                                                    not_before: 1677837321,
                                                    not_after: 1709373321,
                                                },
                                            ),
```
Unix epoch time as of this comment is 1709613931 > 1709373321.  Could that be the culprit?

raphaelrobert (dismissed):
Thanks, let's get this in!

raphaelrobert (approved):

PR #1524, opened 2024-03-05
Some DS improvements

kkohbrok:


franziskuskiefer (approved):
Thanks, let's get this in before it gets stale.

PR #1525, opened 2024-03-05
Add Basic Wasm Bindings and add CI Workflow to Measure Size

keks:
This PR adds minimal Wasm bindings and example code, and a github action workflow that measures the size of the built bundle (currently around 1.2MB uncompressed, 450kB gzip'd). It fails when it gets significantly larger.

This is a draft. One test started failing after I merged in the current main, and I am not sure what the issue is. I think it might be related to #1495, but I'm not sure. Can you take a look and see if this is familiar @raphaelrobert?


openmls-wasm/src/lib.rs
```diff
@@ -0,0 +1,389 @@
...
+extern "C" {
+    fn alert(s: &str);
+
+    // Use `js_namespace` here to bind `console.log(..)` instead of just
+    // `log(..)`
+    #[wasm_bindgen(js_namespace = console)]
+    fn log(s: &str);
+}
+
+static CIPHERSUITE: Ciphersuite = Ciphersuite::MLS_128_DHKEMX25519_CHACHA20POLY1305_SHA256_Ed25519;
```
franziskuskiefer, line 29: Please add some comments to all the things.

openmls-wasm/tests/web.rs
```diff
@@ -0,0 +1,13 @@
...
+
+#![cfg(target_arch = "wasm32")]
+
+extern crate wasm_bindgen_test;
+use wasm_bindgen_test::*;
+
+wasm_bindgen_test_configure!(run_in_browser);
+
+#[wasm_bindgen_test]
+fn pass() {
```
franziskuskiefer, line 11: Drop?

openmls/src/key_packages/mod.rs
```diff
@@ -157,7 +157,9 @@ pub use lifetime::Lifetime;
 ///     Extension extensions<V>;
 /// } KeyPackageTBS;
 /// ```
-#[derive(Debug, Clone, PartialEq, TlsSize, TlsSerialize, Serialize, Deserialize)]
+#[derive(
+    Debug, Clone, PartialEq, TlsSize, TlsSerialize, Serialize, Deserialize, TlsDeserialize,
```
franziskuskiefer, line 161: These structs don't implement `deserialize` on purpose. Why do you need these?
The `VerifiableKeyPackage` is handling the deserialization to enforce validation and signature checks.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -71,7 +71,9 @@ pub(crate) struct NewLeafNodeParams {
 /// } LeafNode;
 /// ```
 // TODO(#1242): Do not derive `TlsDeserialize`.
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, TlsSerialize, TlsSize)]
+#[derive(
+    Debug, Clone, PartialEq, Eq, Serialize, Deserialize, TlsSerialize, TlsSize, TlsDeserialize,
```
franziskuskiefer, line 75: Same here. This shouldn't implement deserialize. Instead, there's a verifable version.

franziskuskiefer (changes requested):
Thanks. Generally looks good except for the openmls lib changes :)

openmls/src/key_packages/mod.rs
keks: Good catch. I initially had these because I had planned to serialize everything before returning to js, but (a) it looks like i was doing that wrong and (b) we use opaque objects now instead of serializing so it doesn't really matter.

franziskuskiefer (approved):
Thanks!

keks:
I can't merge this because of CI failing due to the expired key package. Can you override this @franziskuskiefer?

franziskuskiefer:
> I can't merge this because of CI failing due to the expired key package. Can you override this @franziskuskiefer?

Lets not have a failing CI. Can you disable the failing test for now and file an issue to re-enable it as soon as the kats were updated?

franziskuskiefer:
I disabled the test in #1530. @keks can you merge when the CI is through?

PR #1527, opened 2024-03-06
Improve BasicCredential

### LINKED ISSUE #1526: Make the BasicCredential code panic free
Get rid of the following `unwrap()` functions:

 - https://github.com/openmls/openmls/blob/main/openmls/src/credentials/mod.rs#L320
 - https://github.com/openmls/openmls/blob/main/openmls/src/credentials/mod.rs#L330

raphaelrobert:
Fixes #1526 

This PR makes `BasicCredential` more ergonomic, because it can now be converted from a `Credential` without having to serialize and deserialize the content of the credential. It also checks the input size of the indentity.

### DIFF
--- a/openmls/src/credentials/mod.rs
+++ b/openmls/src/credentials/mod.rs
@@ -27,13 +27,14 @@ use std::io::{Read, Write};
 use serde::{Deserialize, Serialize};
 use tls_codec::{
     Deserialize as TlsDeserializeTrait, DeserializeBytes, Error, Serialize as TlsSerializeTrait,
-    Size, TlsDeserialize, TlsDeserializeBytes, TlsSerialize, TlsSize, VLBytes,
+    Size, VLBytes,
 };
 
 #[cfg(test)]
 mod tests;
 
 use crate::ciphersuite::SignaturePublicKey;
+use errors::*;
 
 // Public
 pub mod errors;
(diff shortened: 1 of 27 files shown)


openmls/src/credentials/mod.rs
```diff
@@ -77,8 +78,8 @@ pub enum CredentialType {
     Basic = 1,
     /// An X.509 [`Certificate`]
     X509 = 2,
-    /// A currently unknown credential.
-    Unknown(u16),
+    /// Another type of credential that is not in the MLS protocol spec.
+    Other(u16),
```
franziskuskiefer, line 82: That's a breaking change that needs to go into the changelog (unless we didn't have the unknown in the last release)

openmls/src/credentials/mod.rs
```diff
@@ -287,48 +286,53 @@ impl Credential {
...
-    Deserialize,
-    TlsSerialize,
-    TlsDeserialize,
-    TlsDeserializeBytes,
-    TlsSize,
-)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct BasicCredential {
-    credential: Credential,
+    header: Vec<u8>,
```
franziskuskiefer, line 291: Is there a reason not use `VLBytes` for these two? That would be more efficient.

franziskuskiefer (approved):
Thanks, generally lgtm with two nits.

openmls/src/credentials/mod.rs
raphaelrobert: Yes, the reason is that I want to do the serialization in the constructor, otherwise we have to suppress a potential error in `impl From<BasicCredential> for Credential` and the whole point of this PR was to get rid of unwraps.

openmls/src/credentials/mod.rs
raphaelrobert: I added that to the changelog

franziskuskiefer (approved):

Issue #1536, opened 2024-03-18
Protocol version validation


raphaelrobert:
The protocol version for incoming messages is not fully verified currently.

keks:
I am not sure if this is a bug in the spec, but as far as I can tell, the only place where the `version` field of `MlsMessage` is every used is to compare that it is the same as that of a `KeyPackage`. It looks like we don't do that, but it seems like that is the only check we need to do.

keks:
I am tracking that missing check here: https://github.com/openmls/validation.dhall/issues/23

keks:
Oh, actually that is being checked: https://github.com/openmls/openmls/blob/7d436363b3e322084a720ad9e2e36c61e62d7fcc/openmls/src/framing/codec.rs#L41-L48. I am adding it to the dashboard and annotating it.

I think that would fully address this issue and we can close it once the documentation is in, do you all agree?

kkohbrok:
I think at some point we wanted to do this more thoroughly and, for example, tag key material with protocol version. But that was all taken out. What we could do is check that the version in a message aligns with the version of the group in the context of which it's processed. I'm not sure what else we'd want to do.

raphaelrobert:
I'm not sure we need a lot of complexity right now, since only one version exists. We could just hard-code version 1 and abort if we ever receive anything else.

keks:
I wonder whether we really should do that, given that the RFC doesn't ask us to. Technically we might be rejecting legal messages if we do that

raphaelrobert:
I guess it depends how you interpret the RFC. You could argue that RFC9420 specifies MLS 1.0 and if we want the library to implement exactly that, we would reject other version numbers, or at least inform the application about it.
Otherwise we end up with an inconsistent situation where the app always emits MLS 1.0 messages, but happily accepts MLS 99.0 messages, as long as they look like MLS 1.0 messages beyond the version number.

raphaelrobert:
I'd treat this as an omission in the RFC , given that in the IETF/IANA logic any other version number would now require another document.

raphaelrobert:
I think it boils down to two options:

a) We only aim to support RFC9420 for now. The library only emits MLS 1.0 messages (already the case) and it only accepts incoming MLS 1.0 messages to avoid any confusion.

b) We aim to support other versions beyond 1.0. The library can emit messages with different version numbers. When it processes incoming messages it clearly exposes the version number to the app for inspection/validation.

[...]

Issue #1513, opened 2024-02-28
Roll back protocol version support


raphaelrobert:
Protocol version support is partially implemented as defined in #34. Rolling it back requires the following changes:

 - [ ] Remove the `Mls10Draft11` variant from `ProtocolVersion`, which depends on:
 - [ ] https://github.com/openmls/openmls/issues/1512 (for tests)
 - [ ] https://github.com/openmls/openmls/issues/1511

Issue #1512, opened 2024-02-28
Extend the testing framework with flexible constructors


raphaelrobert:
Currently, it is very hard to maliciously manipulate structs that are defined in the spec. We need an easy way to set all fields in those structs to arbitrary values and to regenerate signatures and MACs (membership tag, confirmation tag).

keks:
This seems to be addressed by the frankenstein framework. Should we close this and open individual issues for missing functionality?

PR #1537, opened 2024-03-18
Introduce frankenstein framework

### LINKED ISSUE #1065: ProtocolVersion review

### LINKED ISSUE #1476: Find a way to generate illegal messages so we can test validation
Currently, we often can not create illegal messages, because the functions building them validate the input data. However, this means it is difficult for us to test the validation of incoming illegal messages. We need a general approach to this.

One approach that has been discussed is to have a second protocol implementation that can generate these messages.

Another approach would be to introduce a way to disable validation. This could be done either using boolean flags in the control flow (which seems fragile and prone to errors) or something like a ValidationProvider. One problem that might occur here is that the sender might break their local state when trying to perform illegal operations, but this might be okay in test scenarios.

raphaelrobert:
Fixes #1065, #1476, and #1513 and works towards #1512.

This PR introduces a framework to build illegal or broken messages. For now, it focuses on KeyPackages, LeafNodes and partially on framing. 
The framework is used in the relevant tests (mostly validation) and modifier functions that were behind test flags are dropped from the library.
This should make future validation tests much easier to write.

This PR also changes the `ProtocolVersion` to now accept any version number. The (also previously) missing validation is tracked in #1536.


kkohbrok (approved):
Looks good, thanks!

kkohbrok (approved):

PR #1539, opened 2024-04-02
Further improve credentials

raphaelrobert:
This PR further improves credential handling. Since [we assume](https://github.com/openmls/openmls/blob/main/openmls/src/credentials/mod.rs#L219) that all credentials are variable-size vectors we might as well simplify things.
This way handling basic credentials and custom credentials alike becomes easier.

### DIFF
--- a/openmls/src/credentials/mod.rs
+++ b/openmls/src/credentials/mod.rs
@@ -279,9 +279,9 @@ impl TryFrom<Credential> for BasicCredential {
     type Error = BasicCredentialError;
 
     fn try_from(credential: Credential) -> Result<Self, Self::Error> {
-        match credential.credential_type() {
+        match credential.credential_type {
             CredentialType::Basic => Ok(BasicCredential::new(
-                credential.serialized_credential_content.clone().into(),
+                credential.serialized_credential_content.into(),
             )),
             _ => Err(errors::BasicCredentialError::WrongCredentialType),
         }
@@ -344,7 +344,6 @@ pub mod test_utils {
 mod unit_tests {
     use tls_codec::{
         DeserializeBytes, Serialize, TlsDeserialize, TlsDeserializeBytes, TlsSerialize, TlsSize,
-        VLBytes,
     };
 
     use super::{BasicCredential, Credential, CredentialType};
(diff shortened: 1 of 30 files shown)


openmls/src/credentials/mod.rs
```diff
@@ -299,38 +254,35 @@ impl BasicCredential {
...
 
-    fn try_from(credential: &Credential) -> Result<Self, Self::Error> {
+    fn try_from(credential: Credential) -> Result<Self, Self::Error> {
         match credential.credential_type() {
-            CredentialType::Basic => {
-                let identity: VLBytes = credential.deserialized().unwrap();
-                Ok(BasicCredential::new(identity.into())?)
-            }
+            CredentialType::Basic => Ok(BasicCredential::new(
+                credential.serialized_credential_content.clone().into(),
```
franziskuskiefer, line 284: This clone should be unnecessary since we own the credential. You can own the `credential_type` in the `match`, that's cheaper.

openmls/src/credentials/mod.rs
```diff
@@ -412,11 +367,42 @@ mod unit_tests {
...
     }
+
+    /// Test the [`Credential`] with a custom credential.
+    #[test]
+    fn custom_credential() {
+        #[derive(
+            Debug, Clone, PartialEq, Eq, TlsSize, TlsSerialize, TlsDeserialize, TlsDeserializeBytes,
+        )]
+        struct CustomCredential {
+            inner: VLBytes,
```
franziskuskiefer, line 385: Can you add another field here to make sure it's working for other things that are not just bytes?

openmls/src/credentials/mod.rs
```diff
@@ -184,67 +184,23 @@ pub struct Certificate {
...
 /// ```
-#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+#[derive(
+    Debug,
+    PartialEq,
+    Eq,
+    Clone,
+    Serialize,
+    Deserialize,
+    TlsSize,
```
franziskuskiefer, line 194: I thought I had a reason not to do it this way. But I can't recall and maybe it was only necessary in a previous iteration. Anyway, I don't see a reason not to do it this way right now.

franziskuskiefer (approved):
lgtm with those nits addressed.

franziskuskiefer (approved):
Thanks!

PR #1542, opened 2024-04-04
Enable handling of other/unknown proposals

kkohbrok:
This PR allows OpenMLS to handle unknown/other proposals analogous to how extensions are handled.


openmls/src/group/core_group/test_proposals.rs
```diff
@@ -664,3 +637,158 @@ fn test_group_context_extension_proposal(
             .expect("Error exporting secret.")
     )
 }
+
```
raphaelrobert, line 640: Can you add a brief description?

raphaelrobert (changes requested):
There's one nit from a first pass, but I'm generally wondering how we deal with unsupported proposal types. Obviously, this now becomes the duty of the application. Have you checked what that entails and could you add something to the book in that case?

kkohbrok:
It's pretty much the same as for extensions, where OpenMLS doesn't fail if there is an extension it doesn't recognize. It just accepts that it's there, but doesn't do anything as a consequence. The application can then inspect the proposals, take whatever data is in the proposal and act according whatever the semantic of the proposal is.

Good point w.r.t. docs. I'll add something to the book and I'll also add a changelog entry.

openmls/src/group/core_group/test_proposals.rs
```diff
@@ -663,3 +636,145 @@ fn test_group_context_extension_proposal(
...
+    let custom_proposal_capabilities =
+        Capabilities::new(None, None, None, Some(&[custom_proposal_type]), None);
+    // Create Bob's key package with support for the custom proposal type
+    let bob_key_package = KeyPackageBuilder::new()
+        .leaf_node_capabilities(custom_proposal_capabilities.clone())
+        .build(ciphersuite, provider, &bob_signer, bob_credential_with_key)
+        .unwrap();
+
+    // Create group with custom proposal as required capability
+    let mut alice_group = CoreGroup::builder(
```
raphaelrobert, line 663: Since the `CoreGroup` will disappear soon, we shouldn't write new tests for it and instead base them on `MlsGroup`.

openmls/src/group/core_group/test_proposals.rs
raphaelrobert: That is still missing?

openmls/src/messages/proposals.rs
```diff
@@ -200,29 +182,22 @@ impl From<ProposalType> for u16 {
...
-    #[tls_codec(discriminant = 6)]
     ExternalInit(ExternalInitProposal),
-    #[tls_codec(discriminant = 7)]
     GroupContextExtensions(GroupContextExtensionProposal),
     // # Extensions
     // TODO(#916): `AppAck` is not in draft-ietf-mls-protocol-17 but
     //             was moved to `draft-ietf-mls-extensions-00`.
-    #[tls_codec(discriminant = 8)]
     AppAck(AppAckProposal),
+    Other((u16, OtherProposal)),
```
raphaelrobert, line 200: Why wouldn't we align this with `CustomProposal`? And the other question is whether we want a wrapper struct instead of a tuple. I don't have a strong opinion but we have not been using tuples a lot and they are less expressive in general.

openmls/tests/book_code.rs
```diff
@@ -1391,3 +1391,107 @@ fn test_empty_input_errors(ciphersuite: Ciphersuite, provider: &impl OpenMlsProv
...
+    let expected_proposal = OtherProposal(custom_proposal_payload.into());
+    assert!(staged_commit.queued_proposals().any(|qp| {
+        let Proposal::Other(custom_proposal) = qp.proposal() else {
+            return false;
+        };
+        custom_proposal.0 == custom_proposal_type && custom_proposal.1 == expected_proposal
+    }));
+
+    // ANCHOR_END: custom_proposal_usage
+}
```
raphaelrobert, line 1497: Just a nit, but the code uses both `expect()` and `unwrap()` arbitrarily. I think `unwrap()` is fine when it is self-explanatory, we have used it in the past.

openmls/src/group/public_group/validation.rs
```diff
@@ -121,6 +121,43 @@ impl PublicGroup {
...
+            let proposal_type = proposal.proposal().proposal_type();
+            if matches!(proposal_type, ProposalType::Other(_))
+                && !capabilities_intersection.contains(&proposal_type)
+            {
+                return Err(ProposalValidationError::UnsupportedProposalType);
+            }
+        }
+        Ok(())
+    }
+
```
raphaelrobert, line 160: Do we have both a positive and negative test for this? If not, please open an issue (or add the tests).

raphaelrobert (changes requested):
Thanks for the PR! A few comments and and questions.

raphaelrobert (approved):
Thanks for making the changes! We should still come up with a harmonized naming convention for other/custom/unknown credentials/proposals/extensions at some point.

PR #1545, opened 2024-04-04
Remove cipher suite and version from Secret

### LINKED ISSUE #1511: Remove cipher suite and version from Secret
`Secret` currently also stores the cipher suite and the protocol version:

```rust
pub(crate) struct Secret {
    pub(in crate::ciphersuite) ciphersuite: Ciphersuite,
    pub(in crate::ciphersuite) value: SecretVLBytes,
    pub(in crate::ciphersuite) mls_version: ProtocolVersion,
}
```

 Since the library won't handle different protocol versions we should remove this.
This likely affects the storage format and therefore depends on #672.

raphaelrobert:
Fixes #1511.

This PR removes the ciphersuite and protocol version from the `Secret` struct:

```rust
pub(crate) struct Secret {
    pub(in crate::ciphersuite) ciphersuite: Ciphersuite,
    pub(in crate::ciphersuite) value: SecretVLBytes,
    pub(in crate::ciphersuite) mls_version: ProtocolVersion,
}
```
↓

```rust
pub(crate) struct Secret {
    pub(in crate::ciphersuite) value: SecretVLBytes,
}
```

To make the review easy, this is the only change. The rest of the PR is churn. There's quite a bit of it, since the secrets were used to carry the ciphersuite which now has to be passed manually. Conversely, the protocol version no longer has to passed in quite a few functions.


openmls/src/ciphersuite/tests/test_secrets.rs
```diff
@@ -31,3 +31,4 @@ fn secret_incompatible(ciphersuite: Ciphersuite, provider: &impl OpenMlsProvider
     // This must panic because the two secrets have incompatible MLS versions.
     let _default_extracted = default_secret.hkdf_extract(provider.crypto(), &draft_secret);
 }
+ */
```
kkohbrok, line 34: Remove this test entirely?

kkohbrok (changes requested):
Looks good to me with one nit.

PR #1553, opened 2024-04-09
fixup build after #1548

franziskuskiefer:
This fixes the build issues introduced in #1548.

It also adds a `cargo hack check` command to ensure all features are built on ci.


raphaelrobert (approved):
Thanks for fixing that and also amending the CI!

PR #1554, opened 2024-04-09
expose propose_group_context_extensions and add a function

### LINKED ISSUE #1549: Expose GroupContextExtensionProposal API
For some reason, the API for proposing Group Context Extensions is currently `#[cfg(test)]`: https://github.com/openmls/openmls/blob/033f4e63d9b8d1bfe4377a341b8e9696f94712a2/openmls/src/group/mls_group/proposal.rs#L322-L353

This is wrong. Also, we should maybe add an integration test for group context extensions. That would have caught this accident.

franziskuskiefer:
Make propose_group_context_extensions not `test` only and add an integration test.

fixes #1549

### DIFF
--- a/openmls/tests/test_mls_group.rs
+++ b/openmls/tests/test_mls_group.rs
@@ -1194,3 +1194,84 @@ fn mls_group_ratchet_tree_extension(ciphersuite: Ciphersuite, provider: &impl Op
         assert_eq!(error, WelcomeError::MissingRatchetTree);
     }
 }
+
+//
+// Test that the builder pattern accurately configures the new group.
+#[apply(ciphersuites_and_providers)]
+fn group_context_extensions_proposal(ciphersuite: Ciphersuite, provider: &impl OpenMlsProvider) {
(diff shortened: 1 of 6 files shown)


openmls/tests/test_mls_group.rs
```diff
@@ -1194,3 +1194,84 @@ fn mls_group_ratchet_tree_extension(ciphersuite: Ciphersuite, provider: &impl Op
         assert_eq!(error, WelcomeError::MissingRatchetTree);
     }
 }
+
+//
+// Test that the builder pattern accurately configures the new group.
```
kkohbrok, line 1199: This description is not correct, is it?

kkohbrok (changes requested):
One small nit, otherwise looks good!

openmls/tests/test_mls_group.rs
franziskuskiefer: yeah... c&p

kkohbrok (approved):
Thanks!

PR #1560, opened 2024-04-09
Adds test for updating Unknown Group Context Extension

cameronvoell:
Related to https://github.com/openmls/openmls/issues/1549

 The idea of the test is basically to show that we can  save and update data to Group Context Extensions that all group members can read via Unknown Group Context Extension and Group Context Extension proposals.

### DIFF
--- a/openmls/src/group/mls_group/test_mls_group.rs
+++ b/openmls/src/group/mls_group/test_mls_group.rs
@@ -1318,21 +1318,23 @@ fn builder_pattern(ciphersuite: Ciphersuite, provider: &impl OpenMlsProvider) {
 
 // Test the successful update of Group Context Extension with type Extension::Unknown(0xff11)
 #[apply(ciphersuites_and_providers)]
-fn update_group_context_with_unknown_extension(ciphersuite: Ciphersuite, provider: &impl OpenMlsProvider) {
+fn update_group_context_with_unknown_extension(
+    ciphersuite: Ciphersuite,
+    provider: &impl OpenMlsProvider,
+) {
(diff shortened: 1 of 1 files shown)


openmls/src/group/mls_group/test_mls_group.rs
```diff
@@ -1316,6 +1316,169 @@ fn builder_pattern(ciphersuite: Ciphersuite, provider: &impl OpenMlsProvider) {
...
+fn update_group_context_with_unknown_extension(
+    ciphersuite: Ciphersuite,
+    provider: &impl OpenMlsProvider,
+) {
+    let (alice_credential_with_key, _alice_kpb, alice_signer, _alice_pk) =
+        setup_client("Alice", ciphersuite, provider);
+
+    // === Define the unknown group context extension and initial data ===
+    let unknown_extension_data = vec![1, 2];
+    let unknown_gc_extension = Extension::Unknown(0xff11, UnknownExtension(unknown_extension_data));
```
kkohbrok, line 1330: It's just a test, but it would be good to store `0xff11` in a variable and use that throughout the test instead of the literal number.

kkohbrok (commented):
Thanks for adding this test! Just a minor nit. Also you'll have to adapt the test to the removal of CryptoConfig to make the CI happy.

openmls/src/group/mls_group/test_mls_group.rs
cameronvoell: 👍 updated 7172d638294fb6c65deb65eaf6ce6045820de982

kkohbrok (approved):
Thanks! Let's get this in, then.

PR #1562, opened 2024-04-15
First Steps Towards Stable Storage

keks:
This PR adds the first steps towards a stable storage system. The code changes fall roughly into the following categories:

- Add storage provider interface -  `traits/src/storage`
  Define the interface of the storage provider
  - enums
    - GetErrorKind, UpdateErrorKind: describes the rough category of error, so the user doesn't have to know the specific type of the the error returned by the storage implementation for basic error handling
    - Update: describes a single change to the stored data
  - traits
    - Keys and Entities: one per type. That way, trait implementations can use generic types that look like the real types.
    - Types: A trait collecting all the Key and Entity types
    - GetError, UpdateError: traits to be implemented for the errors returned by the functions. Can give some insight into the error by returning an ErrorKind
    - StorageProvider: The main interface. Mostly getters, and the two functions `apply_update` and `apply_updates`
- Add implementation of interface based on KV-Stores - `storage-kv/`
  - checks version when instantiated with `open`, returns error if it doesn't match
- `openmls::storage`
  - implements `Types` trait with OpenMLS types
  - provides traits to refine the traits in `openmls_traits` to use the specific `Types` from `openmls`.
- store first things in the storage interface
  - when proposals are queued, and when a diff in a PublicGroup is merged, write them to storage
- "the rest"
  - fix all the small compile errors that this change requires, especially around testing

I think it probably makes sense to discuss this tomorrow. This still is a bit rough around the edges, but I think it's a good first step towards stability, and once this is in, we should have a good base for adding more and more to storage, as well as reading it for loading the state.


traits/src/storage.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    // update functions, single and batched
+    fn apply_update(&self, update: Update<VERSION, Self::Types>) -> Result<(), Self::UpdateError>;
+    fn apply_updates(
+        &self,
+        update: Vec<Update<VERSION, Self::Types>>,
+    ) -> Result<(), Self::UpdateError>;
+
+    // getter
+    fn get_queued_proposal_refs(
```
franziskuskiefer, line 35: Rust naming conventionsare not to use `get_` prefixes.

openmls/src/storage.rs
```diff
@@ -0,0 +1,64 @@
...
+
+use crate::ciphersuite::hash_ref::ProposalRef;
+use crate::group::GroupContext;
+use crate::group::GroupId;
+use crate::group::InterimTranscriptHash;
+use crate::group::QueuedProposal;
+use crate::messages::ConfirmationTag;
+use crate::treesync::TreeSync;
+
+impl Entity<1> for QueuedProposal {}
```
franziskuskiefer, line 14: Let's not use literals here for the versions but constants to make it harder to mess up.

openmls/src/storage.rs
```diff
@@ -0,0 +1,64 @@
+#[derive(Debug, Clone, Default)]
```
franziskuskiefer, line 1: Can this get some comments and description?

traits/src/storage.rs
```diff
@@ -0,0 +1,110 @@
...
+        &self,
+        group_id: &<Self::Types as Types<VERSION>>::GroupId,
+    ) -> Result<Vec<<Self::Types as Types<VERSION>>::ProposalRef>, Self::GetError>;
+
+    fn get_queued_proposals(
+        &self,
+        group_id: &<Self::Types as Types<VERSION>>::GroupId,
+    ) -> Result<Vec<<Self::Types as Types<VERSION>>::QueuedProposal>, Self::GetError>;
+
+    fn get_treesync(
```
franziskuskiefer, line 45: Treesync is not a great name for the outside. We may want to call this just tree.

openmls/src/storage.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+pub trait StorageProvider:
+    openmls_traits::storage::StorageProvider<1, Types = OpenMlsTypes>
+{
+}
+
+impl<P: openmls_traits::storage::StorageProvider<1, Types = OpenMlsTypes>> StorageProvider for P {}
+
+pub trait RefinedProvider:
```
franziskuskiefer, line 54: What's refined here?

franziskuskiefer (commented):
Just a few comments.

franziskuskiefer:
Closing in favour of #1565.

PR #1555, opened 2024-04-09
Disable MLS++ interop on CI

kkohbrok:
Due to a recent [`vcpkg`](https://github.com/microsoft/vcpkg/pull/38017) failure (in turn due to a bug in [CMake](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9390)), this PR disables interop testing with MLS++ on CI.

PR #1563, opened 2024-04-16
Revert "Disable MLS++ interop on CI (#1555)"

kkohbrok:
This reverts commit 7e111858904812f42f6fef922803c3a64891328a.

### DIFF
--- a/.github/workflows/interop.yml
+++ b/.github/workflows/interop.yml
@@ -26,39 +26,39 @@ jobs:
 
       # ---------------------------------------------------------------------------------------
 
-      #- name: MLS++ | Checkout
-      #  run: |
-      #    git clone https://github.com/cisco/mlspp.git
-      #    cd mlspp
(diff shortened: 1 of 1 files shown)


.github/workflows/interop.yml
```diff
@@ -89,32 +89,10 @@ jobs:
...
-#              echo "Failed";
-#              exit 1;
-#            fi
-#          done
-
-      - name: Test interoperability (OpenMLS only)
+      - name: Test interoperability
         run: |
           ./target/debug/interop_client&
+          ./mlspp/cmd/interop/mlspp_client -live 12345&
```
franziskuskiefer, line 95: Are these changes down here on purpose?

franziskuskiefer (approved):
yay

.github/workflows/interop.yml
kkohbrok: This PR really just `git revert`s the PR that disabled MLS++ interop, so there shouldn't be any changes.

Issue #1564, opened 2024-04-16
[Meta] Stable Storage Tracking Issue


keks:
- [x] First Implementation (#1562 )
- [ ] Add comments and documentation
- [x] Use constants instead of literals as versions
- [ ] Support storing more parts of the state
  - [ ] Items from the Key Store (deleting key package should also delete secrets, see note on hierarchical deletion)
  - [ ] Items that live in `CoreGroup`
    - [ ] GroupEpochSecrets
    - [ ] own leaf index
    - [ ] use ratchet tree extension
    - [ ] MessageSecretsStore (see note below)
    - [ ] ResumptionPskStore
  - [ ] Items that live in `MlsGroup`
    - [ ] JoinConfig
    - [ ] ProposalStore
      - [x] append
      - [x] query proposal
      - [x] query proposal list
      - [x] ~~is_empty~~
        - not needed, since answering that query requires fetching and parsing the list anyway. Just query the list and call is_empty on that
      - [ ] clear proposal queue
    - [ ] own leaf nodes
    - [ ] aad
    - [ ] MlsGroupState, incl. PendingCommitState
    - [x] ~~InnerState~~
      - can be removed, because we just persist everything immediately
- [ ] load PublicGroup from storage
- [ ] load CoreGroup from storage
- [ ] load MlsGroup from storage
- [ ] Review group storage granularity
- [ ] #1575

**Note on hierarchical deletion**: There are cases where removing one entry implies removing other ones, and which ones are to be removed follows from the contents of the first entry. For example, when removing a `KeyPackage`, it would be nice if we also removed the respective secrets.
However, the storage layer mostly sees the serialized data. One way this could be done is to add methods on the deleted entity to retrieve the storage keys to the transitive deletes that have to be deleted.

franziskuskiefer:
First PR to get in #1565

franziskuskiefer:
@keks can you update this and clearly list all open issues and file follow ups where necessary?

keks:
I initially started working on #1562, but it turned out that the approach to duplicate all the types of a foreign crate made it impossible to keep the approach we use for using types to indicate validated values.

In #1565 we merged an implementation of a more pragmatic approach, which kept the serialization inside the openmls crate. This is theoretically a little more fragile, but at least it is possible, and the fragility can be caught with tests.

In #1579 @franziskuskiefer added some fixes for a method that @kkohbrok eventuallly noticed was duplicate, so we removed it in #1591.

PR #1565, opened 2024-04-21
New storage APIs

### LINKED ISSUE #672: Group state serialization: Robustness: Define versions and upgrade paths

franziskuskiefer:
See #1564 for more detail.
This PR introduces a new storage API that is
* more explicit about what it is storing such that applications can be more intelligent in handling persistence
* versioned such that OpenMLS can update internal representations of persisted values without breaking persisted groups

There is still cleanup, general improvements, and more testing to be done. But all of that should be done behind the scenes such that the storage API trait stays untouched and the stored values do not change.

Fixes #672


traits/src/traits.rs
```diff
@@ -16,14 +16,14 @@ pub mod types;
 pub trait OpenMlsProvider {
     type CryptoProvider: crypto::OpenMlsCrypto;
     type RandProvider: random::OpenMlsRand;
-    type KeyStoreProvider: key_store::OpenMlsKeyStore;
+    type StorageProvider: storage::StorageProvider<{ storage::CURRENT_VERSION }>;
```
franziskuskiefer, line 19: I think I made this fixed to the current version. But I'm not sure it's what we should do. Opinions?

Cargo.toml
```diff
@@ -12,6 +12,7 @@ members = [
   "delivery-service/ds-lib",
   "basic_credential",
   "openmls-wasm",
+  # "storage-kv",
```
franziskuskiefer, line 15:
```suggestion
```

traits/src/storage.rs
```diff
@@ -0,0 +1,617 @@
...
+    ) -> Result<(), Self::Error>;
+
+    /// Delete a PSK based on an identifier.
+    fn delete_psk<PskKey: traits::PskId<VERSION>>(
+        &self,
+        psk_id: &PskKey,
+    ) -> Result<(), Self::Error>;
+}
+
+// base traits for keys and values
```
franziskuskiefer, line 574: Let's add doc comments to all these.

franziskuskiefer (commented):
* There's a `traits.rs` that is empty? Can this be deleted?
* Let's file an issue to fill in the `storage.rs` tests (and maybe remove the file until it actually contains something).

openmls/src/test_utils/test_framework/errors.rs
```diff
@@ -24,7 +23,7 @@ pub enum SetupError {
     ClientError(#[from] ClientError),
     /// See [`ExportSecretError`] for more details.
     #[error(transparent)]
-    ExportSecretError(#[from] ExportSecretError),
+    ExportSecretError(#[from] ExportSecretError<MemoryKeyStoreError>),
```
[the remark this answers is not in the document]
keks: I don't know what you want to see changed, can you elaborate?

traits/src/traits.rs
keks: I think we can leave it this way for now. I don't think it really hurts, and we were planning to take it out of that trait anyway.

traits/src/storage.rs
keks: Do you mean the `Key` and `Entity` traits or all the individual ones? For the latter I'd suggest just adding a doc comment to the `traits` module that explains the idea behind the traits. What the reader needs to know is the concept, which is difficult to explain for individual items, and when you understand the concept, there is little value in a comment on an individual item.

franziskuskiefer: Yeah that works. We just have to have enough comments here so that people know what to do with it.

openmls/src/group/public_group/staged_commit.rs
```diff
@@ -280,10 +280,12 @@ impl PublicGroup {
     ///
     /// This function should not fail and only returns a [`Result`], because it
     /// might throw a `LibraryError`.
+    /// TODO: fix comment above. Also, should we write to storage here? in that
```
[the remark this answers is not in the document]
keks: The TODO assumed that we return an error, which we didn't, because we didn't actually write to storage. Now fixed that we write to storage, and I just got rid of the section of the comment that described what kind of error we return.

openmls/src/test_utils/test_framework/errors.rs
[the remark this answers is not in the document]
franziskuskiefer: There are a `MemoryKeyStoreError`s in here. They should be somethin like `StorageError`s.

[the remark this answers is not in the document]
keks: These are concrete types, not generic types, since we are in the test framework. They are now called `MemoryStorageError`.

openmls/src/treesync/node/encryption_keys.rs
```diff
@@ -155,45 +146,49 @@ pub(crate) struct EncryptionKeyPair {
...
-    /// Write the [`EncryptionKeyPair`] to the key store of the `provider`. This
-    /// function is meant to store standalone keypairs, not ones that are
-    /// already in use with an MLS group.
+    /// Write the [`EncryptionKeyPair`] to the store of the `provider`.
     ///
-    /// Returns a key store error if access to the key store fails.
-    pub(crate) fn write_to_key_store<KeyStore: OpenMlsKeyStore>(
+    /// This must only be used for encryption key pairs that are generated for
+    /// update leaf nodes. All other encryption key pairs are stored as part
+    /// of the key package or the epoch encryption key pairs.
```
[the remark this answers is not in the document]
keks: https://github.com/openmls/openmls/issues/1567

keks:
> Let's file an issue to fill in the storage.rs tests (and maybe remove the file until it actually contains something).

https://github.com/openmls/openmls/issues/1566

storage-kv/Cargo.toml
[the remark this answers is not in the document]
keks: https://github.com/openmls/openmls/issues/1568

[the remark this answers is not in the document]
keks: I removed the directory, moved it to a [separate branch] and created [an issue]

[separate branch]: https://github.com/openmls/openmls/tree/archive/storage-with-kv-stub
[an issue]: https://github.com/openmls/openmls/issues/1568

openmls/src/storage.rs
```diff
@@ -0,0 +1,251 @@
+//! OpenMLS Storage
+//!
+//! This module serves two purposes:
+//!
+//! - It implements the Key, Entity and typ traits from `openls_traits::storage::traits`.
```
kkohbrok, line 5:
```suggestion
//! - It implements the Key, Entity and type traits from `openls_traits::storage::traits`.
```

openmls/src/treesync/mod.rs
```diff
@@ -347,21 +345,21 @@ impl fmt::Display for RatchetTree {
...
 /// the state that is synced across the group, as well as the [`LeafNodeIndex`]
 /// pointing to the leaf of this group member and the current hash of the tree.
 ///
 /// It follows the same pattern of tree and diff as the underlying
-/// [`MlsBinaryTree`], where the [`TreeSync`] instance is immutable safe for
-/// merging a [`TreeSyncDiff`], which can be created, staged and merged (see
-/// [`TreeSyncDiff`]).
+/// `MlsBinaryTree`, where the [`TreeSync`] instance is immutable safe for
+/// merging a `TreeSyncDiff`, which can be created, staged and merged (see
+/// `TreeSyncDiff`).
```
kkohbrok, line 355: If `TreeSync` is becoming `pub`, we should update the docs to something that makes sense to the application developer.

kkohbrok (changes requested):
This is quite the PR. Thanks for tackling this and digging through all this complexity!

A few comments:
- There are still a number of `todo!()`s left (mostly in the `cli`, I think)
- The `Storage` trait implements read/write/delete functions for all the individual bits, which are likely going to be implemented following a very similar pattern of more generic read/write/delete functions (as visible, I think, in the memory key store implementation). Wouldn't it make sense for the trait to declare these generic functions and then provide default implementations for each of the individual storage items?

kkohbrok:
One more general comment. At least for the proposal store it seems like we're keeping and updating state both in the provider and in the actual struct. Is that really necessary? We can probably get rid of one of the struct field at that point can't we?

memory_storage/src/lib.rs
```diff
@@ -104,10 +98,11 @@ impl MemoryStorage {
...
     ) -> Result<Option<V>, <Self as StorageProvider<CURRENT_VERSION>>::Error> {
         let values = self.values.read().unwrap();
+        let storage_key = build_key_from_vec::<VERSION>(label, key.to_vec());
 
-        let mut storage_key = label.to_vec();
-        storage_key.extend_from_slice(key);
-        storage_key.extend_from_slice(&u16::to_be_bytes(VERSION));
+        // let mut storage_key = label.to_vec();
+        // storage_key.extend_from_slice(key);
+        // storage_key.extend_from_slice(&u16::to_be_bytes(VERSION));
```
kkohbrok, line 105: Snip!

memory_storage/src/persistence.rs
```diff
@@ -0,0 +1,76 @@
...
+    fn load_from_file(&mut self, input_file: &File) -> Result<(), String> {
+        // Prepare file reader.
+        let reader = BufReader::new(input_file);
+
+        // Read the JSON contents of the file as an instance of `SerializableKeyStore`.
+        match serde_json::from_reader::<BufReader<&File>, SerializableKeyStore>(reader) {
+            Ok(ser_ks) => {
+                let mut ks_map = self.values.write().unwrap();
+                for (key, value) in ser_ks.values {
+                    ks_map.insert(base64::decode(key).unwrap(), base64::decode(value).unwrap());
```
kkohbrok, line 60: Any reason not to just use `serde_json` here to encode these? Might be a bit more inefficient, but would save us the dependency on `base64`.

kkohbrok (changes requested):
Thanks for making those changes! Just two small nits left, then we can get this merged.

memory_storage/src/persistence.rs
franziskuskiefer: No idea. This is only copied from the cli, where this was before.

kkohbrok: Okay, it's only a PoC, so it's not hugely important.

kkohbrok (approved):

PR #1570, opened 2024-04-29
Doc updates for storage

franziskuskiefer:
This PR fixes up some docs in the book after #1565, working towards #1564.


book/src/traits/traits.md
```diff
@@ -52,21 +52,80 @@ This trait defines all cryptographic functions required by OpenMLS. In particula
...
+```rust,no_run,noplayground
+{{#include ../../../traits/src/storage.rs:traits}}
+```
+
+An implementation of the storage trait should ensure that it can address and
+efficiently handle values.
+
+#### Example: Key packages
+
+This is only an example, but illustrate that the application may need to do more
```
kkohbrok, line 93:
```suggestion
This is only an example, but it illustrates that the application may need to do more
```

book/src/traits/traits.md
```diff
@@ -52,21 +52,80 @@ This trait defines all cryptographic functions required by OpenMLS. In particula
...
+when it comes to implementing storage.
+
+Key packages are only deleted by OpenMLS when they are used and _not_ last resort
+key packages (which are supposed to be used multiple times).
```
kkohbrok, line 97:
```suggestion
key packages (which may be used multiple times).
```

book/src/traits/traits.md
```diff
@@ -52,21 +52,80 @@ This trait defines all cryptographic functions required by OpenMLS. In particula
...
+The application needs to implement some logic to manage last resort key packages.
+
+```rust,no_run,noplayground
+{{#include ../../../traits/src/storage.rs:write_key_package}}
+```
+
+The application may store the has references in a separate list with a validity
```
kkohbrok, line 104:
```suggestion
The application may store the hash references in a separate list with a validity
```

kkohbrok (changes requested):
Thanks for the PR! Just some nits, otherwise this looks good!

kkohbrok (approved):

Issue #1543, opened 2024-04-04
Improve test procedures


kkohbrok:
The way we run tests is currently somewhat fragmented. Some tests only run on the RustCrypto provider and ciphersuites are hardcoded in some places. Also, the current `rstest`-based tests have hardcoded ciphersuites (as opposed to running all ciphersuites provided by a given provider).

Tests could be better in a few ways:

- Tests only need to iterate over crypto providers, not the associated storage providers 
- All test should iterate over all providers (not just RustCrypto)
- All test should iterate over all ciphersuites supported by the provider (instead of hard-coding them)
- It should be possible to create several storage provider instances inside of a test (e.g. to instantiate multiple clients)

This is related to the issue that OpenMLS internally maintains an enum of Ciphersuites instead of treating them as an integer and letting the crypto provider determine if they support a given ciphersuite.

franziskuskiefer:
I made a proc-macro for some of that a while ago, but never merged it because it was part of the libcrux work. https://github.com/openmls/openmls/blob/a4fb8358a87effd8a48ac0c114e6de386e699aa4/openmls_test/Readme.md
We could re-use that code to build something like this.

kkohbrok:
Yes, that looks like it might do the trick. It remains the issue on how to instantiate multiple clients with distinct storage providers in a given test. I don't think that's strictly necessary right now, but it might be in the future, e.g. if we store group state indexed by a group id.

keks:
On first look this seems to be resolved by the PR, only leaving the followup issue #1574. Is that right and can we close this one @franziskuskiefer?

franziskuskiefer:
> It should be possible to create several storage provider instances inside of a test (e.g. to instantiate multiple clients)

This is still open. But please update the description to reflect that most things are done and what's still open.

PR #1573, opened 2024-05-04
Test framework and enabling libcrux tests

franziskuskiefer:
This PR introduces an `opemls_test` macro to replace the `rstest` approach. This allows us checking for features in the provider at runtime and thus enabling more tests, esp. for the libcrux backend (#1250).

This works towards #1543.
And even thou it looks big, it's mostly replacing the test attributes.


openmls/src/group/mls_group/mod.rs
```diff
@@ -412,6 +412,8 @@ impl MlsGroup {
                 plaintext.into()
             }
             OutgoingWireFormatPolicy::AlwaysCiphertext => {
+                eprintln!("content_to_mls_message ...");
+                eprintln!("    ciphersuite: {}", self.ciphersuite());
```
kkohbrok, line 416: Snip?

openmls/src/messages/external_proposals.rs
```diff
@@ -41,12 +41,12 @@ impl JoinProposal {
     /// * `epoch` - group's epoch
     /// * `signer` - of the sender to sign the message
     #[allow(clippy::new_ret_no_self)]
-    pub fn new<Storage: StorageProvider>(
+    pub fn new<Provider: OpenMlsProvider>(
```
kkohbrok, line 44: Why does this have to be a full OpenMLSProvider now?

kkohbrok (changes requested):
Very nice, thanks! This should increase test coverage and make the introduction of new providers more seamless. Just two nits (mostly some errant eprintlns).

openmls/src/messages/external_proposals.rs
franziskuskiefer: It makes calling it easier. But you're right, it's not needed. I reverted the change. You see in the tests what this may require to call it.

franziskuskiefer:
There's a weird issue where coverage build fails with simd instructions in libcrux. I disabled that for now in https://github.com/openmls/openmls/pull/1573/commits/28b5d448995245675d20aeb72f68dfb501b29785.
I also kept unsupported ciphersuites in the build for cross compiled tests. This is not great, but the only way for now https://github.com/openmls/openmls/pull/1573/commits/34c48a0e6045b79de02fba7d6de3da925b51f319.

kkohbrok (approved):
Sounds good and thanks for making the changes. 

The issue with "wrong" ciphersuites when cross-compiling is annoying, but I suppose most of the time we'll be building on the same arch as we're testing. Should we make a note of this somewhere, though?

franziskuskiefer:
I filed #1574

PR #1576, opened 2024-05-09
Make StorageId Public

nplasterer:
In order to call `StorageProvider` functions that require a `SignaturePublicKey` we need the `StorageId` to be accessible.
Also adds a from function so you can get a storage id from a public key vec

For example

```rust
let signature_keys = SignatureKeyPair::new(CIPHERSUITE.signature_algorithm());
let public_key = StorageId::from(signature_keys.to_public_vec());
storage_provider.write_signature_key_pair::<StorageId, SignatureKeyPair>(&public_key, &signature_keys);
```

### DIFF
--- a/basic_credential/src/lib.rs
+++ b/basic_credential/src/lib.rs
@@ -109,7 +109,7 @@ impl SignatureKeyPair {
         }
     }
 
-    fn id(&self) -> StorageId {
+    pub fn id(&self) -> StorageId {
         StorageId {
             value: id(&self.public, self.signature_scheme),
         }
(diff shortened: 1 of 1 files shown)


basic_credential/src/lib.rs
```diff
@@ -109,7 +109,7 @@ impl SignatureKeyPair {
         }
     }
 
-    fn id(&self) -> StorageId {
+    pub fn id(&self) -> StorageId {
```
nplasterer, line 112: I didn't actually need this function to be public but figured if the struct was it should be. Totally cool with making this private atleast for my needs.

franziskuskiefer (approved):
Thanks

PR #1579, opened 2024-05-10
Proposal storage update

franziskuskiefer:
The memory storage clear proposal function is working now. This required a small change to the storage API such that we know the ProposalRef type.
There's also a test for the proposals in the storage crate now.


kkohbrok (approved):
Thanks!

nplasterer (approved):

PR #1586, opened 2024-06-04
API for processed welcome

### LINKED ISSUE #1585: API to parse Welcome | `ParsedWelcome`
The `new_from_welcome` flow to join a group does not allow to only parse the Welcome message right now. This may be seen as enforcing the "the client MUST download the tree before processing the Welcome.".

However, this makes it impossible to tell a joiner information like where to retrieve the ratchet tree from, without a DS that helps with this.

An additional API should be provided that only parses the Welcome and returns a `ParsedWelcome`, which can be inspected, and then converted into a `StagedWelcome` together with a `RatchetTree`.

franziskuskiefer:
Fixes #1585

### DIFF
--- a/openmls/src/group/mls_group/creation.rs
+++ b/openmls/src/group/mls_group/creation.rs
@@ -11,7 +11,7 @@ use crate::{
         group_info::{GroupInfo, VerifiableGroupInfo},
         Welcome,
     },
-    schedule::psk::store::ResumptionPskStore,
+    schedule::psk::{store::ResumptionPskStore, PreSharedKeyId},
     storage::OpenMlsProvider,
     treesync::RatchetTreeIn,
 };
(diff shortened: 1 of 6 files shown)


openmls/src/group/mls_group/creation.rs
```diff
@@ -147,6 +147,81 @@ fn transpose_err_opt<T, E>(v: Result<Option<T>, E>) -> Option<Result<T, E>> {
     }
 }
 
+impl ProcessedWelcome {
+    /// Creates a new processed [`Welcome`] message that can be used to parse
+    /// it before creating a [`StagedWelcome`].
```
raphaelrobert, line 152:
```suggestion
    /// Creates a new processed [`Welcome`] message , which can be
    /// inspected before creating a [`StagedWelcome`].
```

openmls/src/group/mls_group/mod.rs
```diff
@@ -494,3 +494,23 @@ pub struct StagedWelcome {
     // information.
     group: StagedCoreWelcome,
 }
+
+/// A parsed, but not fully processed `Welcome` message.
+///
+/// This may be used in order to retrieve information from the `Welcome` about
+/// the ratchet tree.
+///
+/// Use `into_staged_welcome` to get the [`StagedWelcome`] on this.
```
raphaelrobert, line 503:
```suggestion
/// A `Welcome` message that has been processed but not staged yet.
///
/// This may be used in order to retrieve information from the `Welcome` about
/// the ratchet tree and PSKs.
///
/// Use `into_staged_welcome` to stage it into a [`StagedWelcome`].
```

raphaelrobert (approved):
Thanks, looks good in general. I assume the code is essentially just C&Ced and the logic hasn't changed. I made a few tweaks to the comments.

franziskuskiefer:
> I assume the code is essentially just C&Ced and the logic hasn't changed.

Yes, there's no new code beyond the APIs.

Can you r+ again after applying the suggestions?

PR #1590, opened 2024-06-13
Add Test checking GroupContextExtensionProposal Validation: Commit contains up to one GCE Proposal

keks:
This PR adds test code, testing that the validation rule that a commit may only contain up to one GroupContextExtensionProposal. It also adds tooling needed:
- More conversion functions for frankenstein types
- An authentication function for most of a commit - it doesn't rebuild a valid confirmation tag. This would require too much work for now.
- Adding a flag for temporarily disabling the validation of confirmation tags, and a structure for adding similar exceptions in the future.

### DIFF
--- a/openmls/src/group/mls_group/test_mls_group.rs
+++ b/openmls/src/group/mls_group/test_mls_group.rs
@@ -1678,6 +1678,137 @@ fn update_group_context_with_unknown_extension<Provider: OpenMlsProvider + Defau
     );
 }
 
+#[openmls_test]
+fn test_update_group_context_with_unknown_extension_using_update_function<
+    Provider: OpenMlsProvider + Default,
+>() {
+    let alice_provider = Provider::default();
+    let (alice_credential_with_key, _alice_kpb, alice_signer, _alice_pk) =
+        setup_client("Alice", ciphersuite, &alice_provider);
+
+    // === Define the unknown group context extension and initial data ===
+    const UNKNOWN_EXTENSION_TYPE: u16 = 0xff11;
+    let unknown_extension_data = vec![1, 2];
+    let unknown_gc_extension = Extension::Unknown(
+        UNKNOWN_EXTENSION_TYPE,
+        UnknownExtension(unknown_extension_data),
+    );
+    let required_extension_types = &[ExtensionType::Unknown(UNKNOWN_EXTENSION_TYPE)];
+    let required_capabilities = Extension::RequiredCapabilities(
+        RequiredCapabilitiesExtension::new(required_extension_types, &[], &[]),
+    );
+    let capabilities = Capabilities::new(None, None, Some(required_extension_types), None, None);
+    let test_gc_extensions = Extensions::from_vec(vec![
+        unknown_gc_extension.clone(),
+        required_capabilities.clone(),
+    ])
+    .expect("error creating test group context extensions");
+    let mls_group_create_config = MlsGroupCreateConfig::builder()
+        .with_group_context_extensions(test_gc_extensions.clone())
+        .expect("error adding unknown extension to config")
+        .capabilities(capabilities.clone())
+        .ciphersuite(ciphersuite)
+        .build();
+
+    // === Alice creates a group ===
+    let mut alice_group = MlsGroup::new(
+        &alice_provider,
+        &alice_signer,
+        &mls_group_create_config,
+        alice_credential_with_key,
+    )
+    .expect("error creating group");
+
+    // === Verify the initial group context extension data is correct ===
+    let group_context_extensions = alice_group.group().context().extensions();
+    let mut extracted_data = None;
+    for extension in group_context_extensions.iter() {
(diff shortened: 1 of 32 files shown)


openmls/src/group/mls_group/test_mls_group.rs
```diff
@@ -1142,16 +1145,60 @@ fn remove_prosposal_by_ref(
 
 // Test that the builder pattern accurately configures the new group.
 #[openmls_test]
-fn group_context_extensions_proposal() {
+fn group_context_extensions_proposal<Provider: OpenMlsProvider + Default>() {
```
kkohbrok, line 1148: Doesn't the `openmls_test` macro already give you the `Provider` type to create a new provider?

openmls/src/group/mls_group/test_mls_group.rs
```diff
@@ -1142,16 +1145,60 @@ fn remove_prosposal_by_ref(
...
         .ciphersuite(ciphersuite)
-        .build(provider, &alice_signer, alice_credential_with_key)
+        .with_wire_format_policy(WireFormatPolicy::new(
+            OutgoingWireFormatPolicy::AlwaysPlaintext,
+            IncomingWireFormatPolicy::Mixed,
+        ))
+        .build(alice_provider, &alice_signer, alice_credential_with_key)
         .expect("error creating group using builder");
 
+    // === Alice addes Bob ===
```
kkohbrok, line 1168:
```suggestion
    // === Alice adds Bob ===
```

openmls/src/group/mls_group/test_mls_group.rs
```diff
@@ -1168,20 +1215,104 @@ fn group_context_extensions_proposal() {
...
+    println!("setting disable verification flag...");
+    validation_skip_handle.disable_validation();
+
+    println!("reading flag right after setting...",);
+    crate::skip_validation::is_disabled::confirmation_tag();
+
+    let err = bob_group
+        .process_message(bob_provider, fake_commit.into_protocol_message().unwrap())
+        .expect_err("expected an error");
+    assert!(matches!(
```
kkohbrok, line 1306: If (for some reason) this assert fires and the test panics the mutex should be releaseed because the guard is dropped, correct? Just making sure other tests don't starve at this point.

openmls/src/group/mls_group/test_mls_group.rs
```diff
@@ -1168,20 +1215,104 @@ fn group_context_extensions_proposal() {
...
+        err,
+        ProcessMessageError::InvalidCommit(
+            StageCommitError::GroupContextExtensionsProposalValidationError(
+                GroupContextExtensionsProposalValidationError::TooManyGCEProposals
+            )
+        )
+    ));
+    validation_skip_handle.enable_validation();
```
kkohbrok, line 1314: Instead of this disable - run - enable flow, would it make sense for `validation_skip_handle` to expose a function that does that for you? I.e. pass in a function pointer that is run with the check disabled?

kkohbrok (changes requested):
Thanks for picking this up! The general approach seems solid to me. Just a few comments and proposals.

openmls/src/lib.rs
```diff
@@ -189,6 +189,7 @@ pub mod storage;
 
 // Private
 mod binary_tree;
+mod skip_validation;
```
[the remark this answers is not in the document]
keks: I suppose we could do that, but in all the places where we optionally disable the check, instead of

```rust
if crate::skip_validation::is_disabled::confirmation_tag() {
  return Err(StageCommitError::ConfirmationTagMismatch)
}
```
we have to do

```rust
#[cfg(feature = "test-utils")]
if crate::skip_validation::is_disabled::confirmation_tag() {
  return Err(StageCommitError::ConfirmationTagMismatch)
}

#[not(cfg(feature = "test-utils"))]
return Err(StageCommitError::ConfirmationTagMismatch)
```

because we don't have access to the `is_disabled` module. That would make the code even more cluttered, which I wanted to avoid.

openmls/src/group/mls_group/test_mls_group.rs
keks: I added that, but left the old API public in case it is needed in some edge cases. Now we need to clone the `fake_commit` because we can't move it into the `FnMut` closure, but I think that is fine. The alternative would be to make it `FnOnce`, but that seems too limited.

openmls/src/lib.rs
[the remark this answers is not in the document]
kkohbrok: I think I'd be okay with that clutter. Another advantage of hiding this behind a feature flag would be that we could make that `once_cell` dependency optional and for the test-utils feature only.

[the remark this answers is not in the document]
keks: Since once_cell could simply be made an optional dependency I suggest we keep it this way, since that is equivalent and the current version seems simpler.

kkohbrok (approved):
Thanks for addressing my concerns! Looks good to me now.

PR #1593, opened 2024-06-14
Use newer go version

kkohbrok:


.github/workflows/interop.yml
```diff
@@ -80,12 +80,17 @@ jobs:
           # TODO(#1366)
           cp interop_client/docker/test-runner/main.go.patch mls-implementations/interop/test-runner
           cd mls-implementations/interop
+          go mod edit -go=1.18
```
franziskuskiefer, line 83: 1.18 is pretty old (2022). Maybe updating to something recent solves the issue?

franziskuskiefer (commented):
I'd also be ok disabling this job for now. Or at least not requiring it.
Something needs to get fixed properly here.

.github/workflows/interop.yml
raphaelrobert: 1.22 is the most recent and worked for me locally

kkohbrok:
#1594 did the trick.

PR #1600, opened 2024-06-27
Some more benchmarks

### LINKED ISSUE #1596: OpenMLS Benchmarks
Adding some more benchmarking for OpenMLS.
This especially adds some infrastructure to benchmark larger groups.

https://github.com/openmls/openmls/compare/main...franziskus/large-group-benchmarks

franziskuskiefer:
This PR adds facilities to run benchmarks on larger groups.

Design
- not criterion based
- ability read/write group state into files for testing larger groups
- different group setups
- cli interface to quickly test different scenarios

This required a few small changes to OpenMLS library code to derive some more `Clone`s for testing (benchmarking). Also, the memory key store can be written to files now.

Closes #1596

### DIFF
--- a/.github/workflows/benches.yml
+++ b/.github/workflows/benches.yml
@@ -1,8 +1,8 @@
 name: Benchmarks
 
-concurrency: 
+concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true 
+  cancel-in-progress: true
 
 on:
   push:
@@ -32,3 +32,8 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - name: Benchmarks
         run: cargo bench -p openmls --verbose
+
+      - name: Large groups
+        run: |
+          cargo run -p openmls --example large-groups --release -- --write -g 2 3
+          cargo run -p openmls --example large-groups --release -- -g 3
--- a/memory_storage/src/lib.rs
+++ b/memory_storage/src/lib.rs
@@ -1,5 +1,5 @@
 use openmls_traits::storage::*;
-use serde::{Deserialize, Serialize};
+use serde::{ser::SerializeStruct, Serialize};
 use std::{collections::HashMap, sync::RwLock};
 
 /// A storage for the V_TEST version.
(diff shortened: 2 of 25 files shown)


raphaelrobert (changes requested):
Thanks, we really need more benchmarks. We probably don't need an in-depth code review here but I do have some high-level questions and remarks:

 - Why not use criterion? Measuring is hard and using something like criterion would at least make us avoid basic mistakes.
 - Since it's a benchmark, why not have it in the `benches` folder? I think we should reserve the `examples` folder for actual examples. Benchmarks can also read command line arguments, if that's your worry.
  - This needs some high-level documentation. It's hard (if not impossible) to understand what the command line arguments are for. It's also not really obvious what exactly is being measured. I'm worried it's hard for others to maintain this code.
  - This should run on the CI to ensure it doesn't become stale when the library is refactored. I suspect this could be run with small values so that the CI doesn't take forever but still covers all relevant code paths.

franziskuskiefer:
> * Why not use criterion? Measuring is hard and using something like criterion would at least make us avoid basic mistakes.

The memory overhead from criterion is way too high unfortunately. We'd only measure memcpy.

> * Since it's a benchmark, why not have it in the `benches` folder? I think we should reserve the `examples` folder for actual examples. Benchmarks can also read command line arguments, if that's your worry.

They can, but you can't run them in the same way through cargo, so you'd need to build them separately. If there's a way around that, I'm happy to move them to benches.

> * This needs some high-level documentation. It's hard (if not impossible) to understand what the command line arguments are for. It's also not really obvious what exactly is being measured. I'm worried it's hard for others to maintain this code.

I can add some comments. The help text on the cli should have all information for using it though.

> * This should run on the CI to ensure it doesn't become stale when the library is refactored. I suspect this could be run with small values so that the CI doesn't take forever but still covers all relevant code paths.

Right, I'll put on CI.

raphaelrobert:
> > * Since it's a benchmark, why not have it in the `benches` folder? I think we should reserve the `examples` folder for actual examples. Benchmarks can also read command line arguments, if that's your worry.
> 
> They can, but you can't run them in the same way through cargo, so you'd need to build them separately. If there's a way around that, I'm happy to move them to benches.

I looked into this. Passing command line arguments also works with `cargo bench`, but it's somewhat limited because they can conflict with `libtest`. You would have to resort to using env variables instead or something like that.

> > * This needs some high-level documentation. It's hard (if not impossible) to understand what the command line arguments are for. It's also not really obvious what exactly is being measured. I'm worried it's hard for others to maintain this code.
> 
> I can add some comments. The help text on the cli should have all information for using it though.

I didn't find any explanation for what [these parameters](https://github.com/openmls/openmls/pull/1600/files#diff-e861fc216d637d56ee3f00494a1a01a925ff12dd1d67c497a8067f61ef8c458dR390-R402) do exactly.

franziskuskiefer:
> You would have to resort to using env variables instead or something like that.

Right, that's why I moved it over. I think it's nicer to have a proper tool.

I added usage infos and a CI job.

raphaelrobert (approved):
Thanks, let's get this in!

Issue #1566, opened 2024-04-24
Test Persistence/Storage Stability


keks:
Add tests that

- create a group
- perform operations that persist to storage
- load the group from storage
- check that the group state is the same
- `memory_storage::clear_proposal_queue` doesn't delete proposal refs right now

This is a follow-up issue to PR #1565, see comment https://github.com/openmls/openmls/pull/1565#pullrequestreview-2018960837

keks:
On the meeting yesterday we agreed that we should have KATs for this, i.e. we should create the test data now and then later try to consume it and verify that we have the correct group data.

We might still want to run the test with freshly generated test vectors as well, but for stability it is more important that we correctly consume old data.

PR #1601, opened 2024-06-30
De-duplicate ProposalStore

### LINKED ISSUE #1518: De-duplicate the ProposalStore
There are currently two ProposalStores in an MlsGroup. One in the PublicGroup and one in the CoreGroup. There should really only be one and that's probably the one in the PublicGroup.

raphaelrobert:
Fixes #1518.

This PR gets rid of the `ProposalStore` in the `MlsGroup` and exclusively uses the one in the `PublicGroup` instead.
There shouldn't be any functional changes, but there's quite a bit of churn because tests had to be refactored.


keks:
This is currently blocked: We need to make sure that this does not break storage stability.

The way we want to do this is to add a KAT that is generated with v0.5, and that the KAT still passes using the new proposal store mechanism.

See #1566

openmls/src/framing/test_framing.rs
```diff
@@ -448,7 +448,7 @@ fn unknown_sender<Provider: OpenMlsProvider>(ciphersuite: Ciphersuite, provider:
 
     let params = CreateCommitParams::builder()
         .framing_parameters(framing_parameters)
-        .proposal_store(&proposal_store)
+        .proposal_store(group_alice.proposal_store())
```
kkohbrok, line 451: At the risk of creating more churn: Do we still need to pass the proposal store in the CreateCommitParams? Can't create_commit() just use the internal proposal store directly?

Doesn't have to be fixed with this PR, though.

kkohbrok (commented):
Thanks for cleaning this up! Just two comments.

openmls/src/framing/test_framing.rs
keks: The current tests should already require that all parties have separate providers, so I agree, there shouldn't be a need to pass it in.

raphaelrobert: I did the proposed change and it worked as expected. I had to do all the test refactoring that I tried to avoid in the first place though. This made me also discover a bug in the public group: merging a commit didn't empty the proposal store, this is now fixed.

kkohbrok (approved):

openmls/src/extensions/test_extensions.rs
```diff
@@ -80,18 +80,18 @@ fn ratchet_tree_extension() {
...
             bob_add_proposal,
         )
-        .expect("Could not create QueuedProposal."),
+        .unwrap(),
     );
 
     let params = CreateCommitParams::builder()
         .framing_parameters(framing_parameters)
-        .proposal_store(&proposal_store)
+        //.proposal_store(&proposal_store)
```
keks, line 94: Das kann raus, oder?

openmls/src/group/core_group/staged_commit.rs
```diff
@@ -138,9 +137,11 @@ impl CoreGroup {
 
         let ciphersuite = self.ciphersuite();
 
-        let (commit, proposal_queue, sender_index) =
-            self.public_group
-                .validate_commit(mls_content, proposal_store, provider.crypto())?;
+        let (commit, proposal_queue, sender_index) = self.public_group.validate_commit(
+            mls_content,
+            self.proposal_store(),
```
keks, line 142: Wouldn't it make more sense if `PublicGroup::validation_commit` used `self.proposal_store` here? Because right now the CoreGroup first fetches the value from the PublicGroup and then passes it back into the method of PublicGroup.

keks (changes requested):
The new API looks a lot nicer! I just have one question

keks (approved):
Looks good!

kkohbrok (approved):
Thanks for cleaning this up!

Issue #1588, opened 2024-06-10
OpenMLS Book Update


franziskuskiefer:
Update the OpenMLS book to make sure that it is consistent with the current crate (APIs).

https://openmls.tech/book/

keks:
Most notably, add info on the Storage trait.

Issue #1504, opened 2024-02-19
Document application validation steps


franziskuskiefer:
Document the validation steps required by an application using OpenMLS in order to implement MLS safely.

Issue #1458, opened 2023-12-12
[Doc] `join_by_external_commit` is missing in the OpenMLS Book


Batxent:
As an important way to join a group, `join_by_external_commit` is not mentioned in the OpenMLS Book.

keks:
This one is just lacking the file in the SUMMARY.md, the fix is on PR #1603.

PR #1603, opened 2024-07-02
Assorted Documentation Fixes for v0.6

keks:
This PR covers the Book/Documentation-related issues that are targeted to land in v0.6:

- [x] #1588 - rewrite some sections 
- [x] #1458 - easy, just add link to existing file to SUMMARY.md 
- [x] #1504 - we have some notes on this, need to get this in shape and in here 

Fixes 1588
Fixes 1458
Fixes 1504


keks:
I think this now contains all _validation_ steps that the RFC asks of the application, but there are quite a few other things it wants the application to do, that are more about sane policies than validation. For example, defining a bound for how many old epochs should be kept around, or evicting group members that have been inactive for too long.

I suppose we should document them, but should we do it in this PR? @franziskuskiefer

franziskuskiefer:
> I suppose we should document them, but should we do it in this PR? @franziskuskiefer

Feel free to put then into a follow up issue.

kkohbrok (changes requested):
Thanks for adding all this to the documentation. Just a few comments. Some of them might be somewhat off-topic and more focused on the way we actually do storage.

book/src/user_manual/persistence.md
```diff
@@ -1,13 +1,16 @@
...
-
-## Group Lockout Upon State Loss
-
-MLS provides strong Post-Compromise Security properties, which means that key material is regularly refreshed and old key material becomes stale very quickly. Consequently, regularly persisting state is important, especially after the client has created a commit or issued an Update proposal, thus introducing new key material into the group. A loss of state in such a situation is only recoverable in specific cases where the commit was rejected by the Delivery Service or if the proposed Update was not committed. A re-join is required in most cases to continue participating in a group after a loss of group state. To avoid a loss of state and the associated re-join, persisting `MlsGroup` state after each state-changing group operation is mandatory.
+The state of a given `MlsGroup` instance is continuously written to the configured
+`StorageProvider`. Later, the `MlsGroup` can be loaded from the provider using
+the `load` constructor, which can be called with the respective storage provider
+as well as the `GroupId` of the group to be loaded. For this to work, the group
```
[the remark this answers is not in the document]
keks: What I meant was that the group data must have been written to the provider in the past. I'll see if I can reword it to make it clearer.

book/src/user_manual/persistence.md
```diff
@@ -1,13 +1,16 @@
...
 
 ## Forward-Secrecy Considerations
 
-The `MlsGroup` state that is persisted using the `.save()` function contains private key material. As a consequence, the application needs to delete old group states to achieve Forward-Secrecy w.r.t. that key material. Since, as detailed above, an old group state is stale immediately after most group operations, we recommend deleting old group states as soon as a new one has been written.
+The persisted `MlsGroup` state  contains
+private key material. As a consequence, the application needs to delete old group
+states to achieve Forward-Secrecy w.r.t. that key material. Since, as detailed
+above, an old group state is stale immediately after most group operations,
+we recommend that the `StorageProvider` deletes old group state when values are
+overwritten.
```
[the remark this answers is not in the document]
keks: We removed that ability by removing the `store` method. However, when the `MlsGroup` overwrites data in the provider, it's the provider's responsibility to swiftly remove the old value as to not endanger forward secrecy.

book/src/traits/traits.md
```diff
@@ -140,7 +139,9 @@ Some OpenMLS APIs require only one of the sub-traits, though.
 ## Implementation Notes
 
 It is not necessary to implement all sub-traits if one functionality is missing.
-Suppose you want to use a persisting key store. In that case, it is sufficient to do a new implementation of the key store trait and combine it with one of the provided crypto and randomness trait implementations.
+Suppose you want to use a persisting key store. In that case, it is sufficient
+to do a new implementation of the key store trait and combine it with one of the
+provided crypto and randomness trait implementations.
```
kkohbrok, line 144: key store -> storage?

book/src/user_manual/persistence.md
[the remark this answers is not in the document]
kkohbrok: What do you mean by "The persisted `MlsGroup` state"? Is that the collection of individual values in the storage provider? If so, shouldn't we write "the storage provider implementation" rather than "the application"?

Regarding deletion through the storage provider: Doesn't OpenMLS explicitly delete values using the deletion functions of the storage provider?

book/src/user_manual/persistence.md
```diff
@@ -4,13 +4,12 @@ The state of a given `MlsGroup` instance is continuously written to the configur
...
 ## Forward-Secrecy Considerations
 
-The persisted `MlsGroup` state contains
-private key material. As a consequence, the application needs to delete old group
-states to achieve Forward-Secrecy w.r.t. that key material. Since, as detailed
-above, an old group state is stale immediately after most group operations,
-we recommend that the `StorageProvider` deletes old group state when values are
-overwritten.
+The persisted `MlsGroup` state contains private key material. As a consequence,
+delete old group states need to be deleted in order to achieve Forward-Secrecy
```
kkohbrok, line 12: The `delete` at the beginning of this line is redundant, I think.

Beyond that, I still think this is somewhat confusing. What persisted `MlsGroup` state are we talking about? That persisted (in its individual parts) by the provider? Or is this here in case the application has serialized and (independent of the storage provider) stored the `MlsGroup` state?

book/src/user_manual/persistence.md
```diff
@@ -4,13 +4,12 @@ The state of a given `MlsGroup` instance is continuously written to the configur
...
+w.r.t. that key material. Since, as detailed above, an old group state is stale
+immediately after most group operations, we recommend that the `StorageProvider`
+deletes old group state when values are overwritten.
```
kkohbrok, line 15: I'm still confused as to why we have to recommend this. Doesn't OpenMLS delete the values that are no longer needed using the trait?

kkohbrok (changes requested):
Sorry to keep this in the queue, but there's one small nit and also I'm still a bit confused. I think I'm missing something. Maybe we could briefly talk about this in our call tomorrow?

book/src/user_manual/persistence.md
keks: >The delete at the beginning of this line is redundant, I think.

oops, fixed!

> Beyond that, I still think this is somewhat confusing. What persisted MlsGroup state are we talking about? That persisted (in its individual parts) by the provider? Or is this here in case the application has serialized and (independent of the storage provider) stored the MlsGroup state?

We are talking about the individual pieces of state that are stored in the provider and make up the state of the MlsGroup.

book/src/user_manual/persistence.md
keks: > I'm still confused as to why we have to recommend this. Doesn't OpenMLS delete the values that are no longer needed using the trait?

OpenMLS tells the provider to delete or overwrite the data, yes. This section tells implementors of providers that these write-requests should be applied quickly and irreversibly. For example, it doesn't make much sense to implement the storage provider on top of git or some append-only copy-on-write data structure, because then we'd keep around old keys.

book/src/user_manual/persistence.md
```diff
@@ -1,13 +1,15 @@
...
+must have been written to the provider previously.
 
 ## Forward-Secrecy Considerations
 
-The `MlsGroup` state that is persisted using the `.save()` function contains private key material. As a consequence, the application needs to delete old group states to achieve Forward-Secrecy w.r.t. that key material. Since, as detailed above, an old group state is stale immediately after most group operations, we recommend deleting old group states as soon as a new one has been written.
+The persisted `MlsGroup` state contains private key material. As a consequence,
+old group states need to be deleted in order to achieve Forward-Secrecy
+w.r.t. that key material. Since, as detailed above, an old group state is stale
+immediately after most group operations, we recommend that the `StorageProvider`
+deletes old group state when values are overwritten.
```
kkohbrok, line 15:
```suggestion
OpenMLS uses the `StorageProvider` to store sensitive key material. To achieve forward-secrecy (i.e. to prevent an adversary from decrypting messages sent in the past if a client is compromised), OpenMLS frequently deletes previously used key material through calls to the `StorageProvider`. `StorageProvider` implementations must thus take care to ensure that values deleted through any of the `delete_` functions of the trait are irrevocably deleted and that no copies are kept. Care needs to be taken especially if the medium underlying the storage provider is append-only or follows the copy-on-write paradigm.
```

kkohbrok (approved):
Thanks!

PR #1606, opened 2024-07-05
LeafNode parameters

### LINKED ISSUE #1208: Allow provisioning of new LeafNode from MlsGroup's `self_update`
`self_update` allows passing in an `encryption_key`, which is pretty useless from a consumer perspective and which will be removed with #1203. It would be useful, however, to pass in a full LeafNode, s.t. the consumer can change extensions, the credential, etc.

### LINKED ISSUE #1283: [Bug] Create commit signs the leaf twice: Once after re-keying and once after updating the parent hash
**Describe the bug**
In `compute_path`, the leaf is signed twice: Once when calling `rekey` and once when calling `apply_own_update_path`. Since we call the latter in every case, the first one can be avoided by handling TBS payloads only and passing them into `apply_own_update_path`.

### LINKED ISSUE #1605: [Bug] Updates Don't Work
The existing test suite only tests self-updates, but not the happy case alice-proposes-and-bob-commits flow for Update proposals, and it seems that it doesn't work. There are failing tests on the branch for validation testing: https://github.com/openmls/openmls/blob/e5d2a3920d386ce33fb0b6ac08f68e7a527096fa/openmls/src/group/mls_group/test_mls_group.rs#L1566-L1675

raphaelrobert:
Fixes #1208.
Fixes #1283.
Fixes #1605.

This PR does a few things:
 - cleans up how a `LeafNode` is updated internally
 - Exposes a consistent API for `self_update` and `propose_self_update()` to set `LeafNode` parameters via `LeafNodeParameters` (which includes a builder pattern)
 - Also exposes setting capabilities and extension when joining with an External Commit. Since the credential with key is mandatory in that case, the API is slightly different
 - Speeds up Commit creation by getting rid of the double signature (see #1283)


openmls/src/group/mls_group/creation.rs
```diff
@@ -77,12 +80,15 @@ impl MlsGroup {
...
     /// us, this will create a remove proposal.
+    #[allow(clippy::too_many_arguments)]
     pub fn join_by_external_commit<Provider: OpenMlsProvider>(
         provider: &Provider,
         signer: &impl Signer,
         ratchet_tree: Option<RatchetTreeIn>,
         verifiable_group_info: VerifiableGroupInfo,
         mls_group_config: &MlsGroupJoinConfig,
+        capabilities: Option<Capabilities>,
+        extensions: Option<Extensions>,
```
kkohbrok, line 91: A slightly nicer solution might be to have a new config for external joins (`MlsGroupJoinExternalConfig`?) and make these two configurable there.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -42,6 +39,101 @@ pub(crate) struct NewLeafNodeParams {
...
+pub struct LeafNodeParameters {
+    credential_with_key: Option<CredentialWithKey>,
+    capabilities: Option<Capabilities>,
+    extensions: Option<Extensions>,
+}
+
+impl LeafNodeParameters {
+    /// Create a new [`LeafNodeParametersBuilder`].
+    #[allow(clippy::new_ret_no_self)]
+    pub fn new() -> LeafNodeParametersBuilder {
```
kkohbrok, line 74: Shouldn't this be called `builder`?

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -114,6 +206,26 @@ impl LeafNode {
         Ok((leaf_node, encryption_key_pair))
     }
 
+    /// Create a new placeholder [`LeafNode`].
+    ///
+    /// Note: This is not a valid leaf node and it must be rekeyed and signed
+    /// before it can be used.
+    pub(crate) fn new_placeholder(credential_with_key: CredentialWithKey) -> Self {
```
kkohbrok, line 213: This seems a bit hacky. Is there now way to get around the need for a placeholder?

kkohbrok (commented):
Thanks for tackling this! I've left a few initial comments and will take a closer look on Monday.

openmls/src/group/mls_group/creation.rs
raphaelrobert: I agree, I created https://github.com/openmls/openmls/issues/1607 for that purpose.

openmls/src/treesync/node/leaf_node.rs
raphaelrobert: As discussed on Zulip, there's no easy way to get around a placeholder leaf. You'd have to change of the tree logic to account for a leaf that's empty but that should appear populated.

I refactored the code however so that the temporary leaf gets replaced by the real one a few lines of code later.

openmls/src/treesync/node/leaf_node.rs
raphaelrobert: Fixed.

openmls/src/treesync/node/leaf_node.rs
kkohbrok: Okay, sounds like it's the only reasonable way to do this without a major refactoring.

kkohbrok (changes requested):
One small nit, otherwise this looks good.

openmls/src/treesync/diff.rs
```diff
@@ -296,21 +297,47 @@ impl<'a> TreeSyncDiff<'a> {
...
+        // don't do that, calculating the direct path might fail. It's important
+        // to not do anything with the value of that leaf until it has been
+        // replaced.
+        let leaf_node = LeafNode::new_placeholder();
+
+        // In case of an External Commit, where the leaf is not replaced and the
+        // tree is full, we have to grow the tree first. In all other cases we
+        // just replace an existing leaf.
+        if leaf_index.u32() >= self.leaf_count() {
+            self.add_leaf(leaf_node)?;
```
keks, line 313: Should we have a `debug_assert!` here that we are processing an external commit (and one in the else branch  that we are processing a regular commit, respectively)?

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -42,6 +39,100 @@ pub(crate) struct NewLeafNodeParams {
     pub(crate) tree_info_tbs: TreeInfoTbs,
 }
 
+#[derive(Debug, PartialEq, Clone)]
+pub(crate) struct SimpleLeafNodeParams {
+    pub(crate) credential_with_key: CredentialWithKey,
+    pub(crate) capabilities: Capabilities,
+    pub(crate) extensions: Extensions,
+}
```
keks, line 47: What is the point of this type, compared to the regular `LeafNodeParams`? Is it to have something that has all fields set? Can this have a comment?

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -114,6 +205,27 @@ impl LeafNode {
...
+            leaf_node_source: LeafNodeSource::Update,
+            extensions: Extensions::default(),
+        };
+
+        Self {
+            payload,
+            signature: Vec::new().into(),
+        }
+    }
+
```
keks, line 228: I suppose we can't easily prevent using this in the wrong places using the type system, right?

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -234,86 +323,60 @@ impl LeafNode {
...
+    ///
+    /// This function generates a new encryption key pair that is stored in the
+    /// key store and also returned.
     ///
-    /// Optionally, a new leaf node can be provided to update more values such as
-    /// the credential.
-    pub(crate) fn update_and_re_sign(
+    /// This function can be used when generating an update. In most other cases
+    /// a leaf node should be generated as part of a new [`KeyPackage`].
+    #[allow(clippy::too_many_arguments)]
```
keks, line 333: I believe this isn't needed, the default threshold is 7.

https://rust-lang.github.io/rust-clippy/master/index.html#/too_many_arguments

keks (changes requested):
This generally looks good, I just have a few smaller questions and a few nits. I hope we can get this merged today or tomorrow. If we do, we should fix #1607 before v0.6.

franziskuskiefer:
Drive-by: This needs to go into changelog as a breaking change.

openmls/src/treesync/node/leaf_node.rs
raphaelrobert: The only way I saw was to refactor large parts of TreeSync to accommodate that short-lived temporary state. I'm a big fan of type safety, but that would not be reasonable at this point.

openmls/src/treesync/diff.rs
raphaelrobert: I refactored this anyway now, so I thing the comment no longer valid

openmls/src/group/tests/test_group.rs
```diff
@@ -408,14 +410,15 @@ fn group_operations() {
...
-        .unwrap()
-        .updated(
+    let mut bob_new_leaf_node = group_bob.own_leaf_node().unwrap().clone();
+    bob_new_leaf_node
+        .update(
             ciphersuite,
-            TreeInfoTbs::Update(group_bob.own_tree_position()),
             provider,
-            &alice_credential_with_keys.signer,
+            &bob_credential_with_keys.signer,
```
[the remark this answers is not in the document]
raphaelrobert: I don't think this produced an invalid signature, it rather converted the identity from Bob to Alice. Since we put the proposal in the ProposalStore directly, we bypass all validation checks we have. Feel free to convert this into an issue, however.

kkohbrok (dismissed):
Thanks again for taking this on! Looks good to merge!

keks (approved):
Looks great, let's get that in!

kkohbrok (approved):

kkohbrok (approved):

PR #1608, opened 2024-07-08
Move some external commit tests to MlsGroup

raphaelrobert:
This instead moves some of the external commit tests from the `CoreGroup` level to the `MlsGroup` level.


kkohbrok (approved):
Looks good! 👍🏻

PR #1610, opened 2024-07-09
Add a KAT for testing storage stability

### LINKED ISSUE #1566: Test Persistence/Storage Stability

keks:
We have make sure that we can seamlessly read the storage written by an older version. This PR adds 
- a test generating KAT test data: the storage for a bit of group state
- a test reading the KAT test data and making sure it contains the expected data

Currently it only has a single test that still has to be split up into the tests described above.

Fixes #1566


keks:
- This now serializes several snapshots of the storage and do some checks in between.
- Still need to split the generation+serialization and deserialization+checking into separate functions. Also, the code has grown quite a bit
- Needs to be factored into a separate module and file.
- I am also considering to move it into the context of the openmls crate, because I made a bunch of structs `pub` instead of `pub(crate)` so I have access.

keks:
Ah, also currently the tests all write to and read from the testdata file, so they can't run concurrently

keks:
Huh, not sure what the test errors are, here. `read_exact` fails with `UnexpectedEof` when reading test data from a byte slice. The byte slice has been decoded from a base64 string stored in json.

it only fails on 32 bit platforms, so it could be related to u64 <-> usize conversions. Isn't `as usize` safe for `u64` less than `usize::MAX`? I'll have to take another look tomorrow :/

memory_storage/src/lib.rs
```diff
@@ -25,6 +25,62 @@ impl Clone for MemoryStorage {
     }
 }
 
+// For testing (KATs in particular) we want to serialize and deserialize the storage
+#[cfg(feature = "test-utils")]
+impl MemoryStorage {
+    pub fn serialize(&self, w: &mut Vec<u8>) -> std::io::Result<usize> {
```
franziskuskiefer, line 31: Now we have two ways of (de)serializing this. Please file an issue to remove one.

openmls/src/storage/mod.rs
franziskuskiefer, line 1: Why did you move this? As far as I can tell there's no reason in here that makes this necessary. Having `mod.rs` files is always annoying. We should maybe enable clippy's `mod_module_files` as well. But not in this PR.

openmls/src/storage/kat_storage_stability.rs
```diff
@@ -0,0 +1,722 @@
+use base64::Engine;
```
franziskuskiefer, line 1: Can you add some description up here on what's happening in this file and how to use it?

openmls/src/storage/kat_storage_stability.rs
```diff
@@ -0,0 +1,722 @@
...
+#[cfg(not(all(
+    feature = "libcrux-provider",
+    not(any(
+        target_arch = "wasm32",
+        all(target_arch = "x86", target_os = "windows")
+    ))
+)))]
+fn write_kats() {
+    // setup
+    let rustcrypto_provider = openmls_rust_crypto::OpenMlsRustCrypto::default();
```
franziskuskiefer, line 332: Except for the provider the `write_kats` functions look very similar. Can we factor out the core to avoid duplication?

openmls/src/storage/kat_storage_stability.rs
```diff
@@ -0,0 +1,722 @@
...
+            )
+        }
+        other => panic!("expected add proposal, got {:?}", other),
+    }
+
+    // there is no pending commit
+    assert!(alice_group_pending_proposal.pending_commit().is_none());
+}
+
+//
```
franziskuskiefer, line 708: Remove?

franziskuskiefer (changes requested):
Thanks! lgtm with a few nits.

openmls/src/storage/mod.rs
keks: Huh, interesting. To me it feels weird to have storage.rs and storage/foo.rs over storage/mod.rs and storage/foo.rs. Happy to go with the other if that is what you prefer.

memory_storage/src/lib.rs
keks: Done: https://github.com/openmls/openmls/issues/1616

franziskuskiefer (approved):

PR #1599, opened 2024-06-27
Test all Validation Checks for GroupContextExtensions

keks:
The goal of this WIP PR is to test that all the checks mandated in the RFC are performed. In order to make this more verifiable, the tests are restructured to live in separate functions, and some helper functions are used to make life easier.

Status:
- 103
  -  [x] for Add proposals
  -  [x] for Update proposals
    - This fails due to the frankenstein conversion failing, need to look into it later.
- [x] ~~201~~ doesn't belong here, is not related to GCE
- [x] 308 -- this has worked even before this PR
- [x] 1001
- [x] ~~1002~~ found to be the same as 1001

Fixes #1589

Issue #1567, opened 2024-04-24
Check whether to move encryption key pair into staged commit


keks:
This is a follow-up to PR #1565. In that PR, we store `EncryptionKeyPair` as an own type, indexed by the public key.

There was the idea of moving it into the respective staged commit: https://github.com/openmls/openmls/pull/1565/files/140ef1936df2625e8268bb9b9056199ebd4e4a93#r1577373682

keks:
I think I have good new and bad news.

First the good news: The tests in #1599 pass now that we merged #1606, without changing the way we read from the store.

For the bad news; maybe I am still misunderstanding something, but in the setting

- Bob sends an Update proposal with a new EncryptionKey
- Alice processes the proposal, generates a commit, merges the commit and sends it
- Bob processes the commit

only Alice ever sees the staged commit, but the interesting key pair is only known by Bob. Maybe moving the encryption keys for commits with an update-path would work, but since it wouldn't cover the keys to update proposals, we can't get rid of the weird keys in the storage.

I suppose what we could do is put the keypair for self-updates commits (without proposals, just update_path) into the StagedCommit, and move the key pairs that belong to update proposals to a field in the store that is keyed by (GroupId, ProposalRef).

While that would certainly be possible, I am not sure that this is high-priority, because the problem with Update proposals has been fixed by #1606.

Maybe the news are not so bad after all.

@franziskuskiefer   What do you think, should we keep this in the v0.6 milestone?

franziskuskiefer:
I wrote another test in #1612 just to convince myself that this is indeed good now.
I think cleaning this up would be nice. But I don't see a reason to block 0.6 on this. I'm dropping it from the milestone.

PR #1612, opened 2024-07-12
more update proposal tests

franziskuskiefer:
Just two tests to explicitly test what was broken before.

cc #1567 #1599 #1606


kkohbrok (approved):
Looks good!

Issue #1521, opened 2024-02-29
Improve how the AAD is set


raphaelrobert:
Currently, the AAD is set permanently at the `MlsGroup` level. It is persisted and applied to every outgoing message until its value is changed. The intuition is that apps would most likely like to set it per message and don't want it to be persisted, not even in RAM.

Rust doesn't offer an elegant way to handle optional arguments, we could however do the following syntax:

```rust
mls_group.with_aad(b"This is the AAD.").add_members(...)?;
```

Internally, `with_aad()` would set a temporary AAD field and consume `mut self`. Any function that sends out messages would honor the AAD field and clear it before returning.

PR #1614, opened 2024-07-15
Add config type for joining from an external commit

### LINKED ISSUE #1607: Better group creation flow for External Commits
Creating a new group with `join_by_external_commit()` requires many parameters. It would be good to align that with how normal group creation works and introduce something like `MlsGroupJoinExternalConfig`.

keks:
This PR adds a new sort of join config to make joining using external commits easier. I believe it's mostly pretty obvious, so just a few things:

- The types have a generic lifetime parameter, because `aad` has been a reference in the past and I wanted to avoid needless copying.
- I am not quite sure about which of the arguments belong into the config and which should be arguments. The ones I included are more on the "small data" side, whereas things like group info and ratchet tree are more on the "large protocol objects" side. I wasn't sure what to do with the credential_with_key. Feedback on this is very welcome.

Fixes #1607

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -17,4 +17,4 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
-      - run: cargo doc -p openmls --message-format json
+      - run: cargo doc -p openmls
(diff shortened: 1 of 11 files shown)


raphaelrobert:
Thanks for tackling it! I think AAD really is the outlier here and I should get working on #1521 instead. I thought about it some more, and the current way the AAD is handled only came into existence because I didn't have any better ideas at the time, not because it's actually how folks are using it. All discussions I've had so far point to wanting an ad-hoc solution for the AAD. And since it's a breaking change for the storage provider, it would be good to get this in for v0.6. I'll get started with it.

keks:
Just a note: this PR also removes the `--message-format json` from the `cargo rustdoc` check in CI so it's more readable.

keks:
Do you want to do your aad changes inside this PR or do you want to do it separately @raphaelrobert? Not sure how to interpret your self-assign :)

raphaelrobert:
> Do you want to do your aad changes inside this PR or do you want to do it separately @raphaelrobert? Not sure how to interpret your self-assign :)

Self-assign is always for reviews. The PR is already is here already: https://github.com/openmls/openmls/pull/1615, but I'm still adding tests. I guess it would make sense to get in #1615 first because then you can drop the lifetime.

franziskuskiefer:
What's the plan here? @keks did you want to close this?

keks:
On a call we discussed restructuring this such that the external commit join config contains the regular join config, instead of mirroring all the fields. With that, and the changes to the AAD, I think it would make most sense to just start over on top of master, if we want to get this in.

Should I have another stab at this or should we close this unfinished?

keks:
This PR is really stale and the entire context the original issue lives in has changed since the introduction of the commit builder API, so I'll close this. The issue is still relevant, but this PR isn't, and maybe hasn't ever been a good solution.

PR #1615, opened 2024-07-15
Change how the AAD is set

### LINKED ISSUE #1119: Test authenticated data
Currently, authenticated data is not tested well.

We should have tests for setting/getting AAD for all relevant message types.

### LINKED ISSUE #1521: Improve how the AAD is set

raphaelrobert:
Fixes #1521.
Fixes #1119.

This PR changes how the AAD is handled: It is no longer persisted as part of the group state and instead needs to be set before every outgoing message. The reason for the change is that the original assumption was never motivated or validated, but instead the contrary seems to be true: the AAD is a dynamic payload that – if set at all – changes with every message. The upcoming MLS extension for structured AAD also indicates that.


CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#1559](https://github.com/openmls/openmls/pull/1559): Remove the `PartialEq` type constraint on the error type of both the `OpenMlsRand` and `OpenMlsKeyStore` traits. Additionally, remove the `Clone` type constraint on the error type of the `OpenMlsRand` trait.
 - [#1565](https://github.com/openmls/openmls/pull/1565): Removed `OpenMlsKeyStore` and replace it with a new `StorageProvider` trait in the `openmls_traits` crate.
 - [#1606](https://github.com/openmls/openmls/pull/1606): Added additional `LeafNodeParameters` argument to `MlsGroup.self_update()` and `MlsGroup.propose_self_update()` to allow for updating the leaf node with custom parameters. `MlsGroup::join_by_external_commit()` now also takes optional parameters to set the capabilities and the extensions of the LeafNode.
+- [#1615](https://github.com/openmls/openmls/pull/1615): Changes the AAD handlling. The AAD is no longer persisted and needs to be set before every API call that generates an `MlsMessageOut`.
```
franziskuskiefer, line 33: The API name changed from `authenticated_data` to `aad`. This needs to be mentioned as well, it's a breaking change.

openmls/src/group/mls_group/membership.rs
```diff
@@ -197,6 +199,7 @@ impl MlsGroup {
                 remove_proposal.clone(),
             )?);
 
+        self.reset_aad();
```
franziskuskiefer, line 202: This looks like a lot of `reset_aad` calls. How do you ensure that it's always reset when needed?

openmls/src/group/mls_group/mod.rs
```diff
@@ -447,6 +445,12 @@ impl MlsGroup {
     pub(crate) fn proposal_store_mut(&mut self) -> &mut ProposalStore {
         self.group.proposal_store_mut()
     }
+
+    /// Resets the AAD.
+    #[inline]
+    pub(crate) fn reset_aad(&mut self) {
+        self.aad = Vec::new();
```
franziskuskiefer, line 452: You could also clear here.
```suggestion
        self.aad.clear();
```

openmls/src/group/mls_group/mod.rs
```diff
@@ -471,20 +475,6 @@ impl MlsGroup {
     pub(crate) fn group(&self) -> &CoreGroup {
         &self.group
     }
-
-    /// Removes a specific proposal from the store.
-    pub fn remove_pending_proposal<Storage: StorageProvider>(
```
franziskuskiefer, line 476: Moving this function looks unrelated? Is this intentional?

franziskuskiefer (changes requested):
Thanks, lgtm with a few nits and one question.

openmls/src/group/mls_group/membership.rs
raphaelrobert: The test covers some of them, but not all of them are. If you think that's necessary I can file an issue for a follow-up.

openmls/src/group/mls_group/mod.rs
raphaelrobert: It is now unrelated (I touched it during the implementation). Moving it was however intentional because it became public at some point and was in the wrong `impl` block.

openmls/src/group/mls_group/membership.rs
franziskuskiefer: I think 100% test coverage here is difficult. I'm wondering more if we could put the reset somewhere where we can be certain that it's always called. I'm fine getting this in as is, but maybe file an issue to ensure that we always reset (if that's with tests or a different style).

franziskuskiefer (approved):
Thanks, lets get this in.

openmls/src/group/mls_group/membership.rs
raphaelrobert: https://github.com/openmls/openmls/issues/1622

PR #1619, opened 2024-07-16
Migrate tests to `MlsGroup`

kkohbrok:
This PR
- migrates a number of tests to use the `MlsGroup` instead of the `CoreGroup` API
  - where tests should have retained roughly the same coverage
- fixes a small issue uncovered in the process, where a member that processes a commit with its own removal didn't apply the path properly during commit processing
- adds helper functions for test purposes to the `MlsGroup` API where necessary
- removes functions from the `CoreGroup` API that are no longer needed


openmls/src/group/mls_group/tests_and_kats/tests/core_group.rs
```diff
@@ -563,86 +439,274 @@ fn test_proposal_application_after_self_was_removed(
...
+        &join_group_config,
+        welcome.into_welcome().unwrap(),
+        Some(ratchet_tree.into()),
     )
-    .unwrap();
+    .expect("Error staging welcome.")
+    .into_group(provider)
+    .expect("Error creating group from welcome.");
+
+    // We can now check that Bob correctly processed his and applied the changes
```
raphaelrobert, line 509:
```suggestion
    // We can now check that Bob correctly processed his commit and applied the changes
```

openmls/src/group/mls_group/tests_and_kats/tests/core_group.rs
```diff
@@ -563,86 +439,274 @@ fn test_proposal_application_after_self_was_removed(
...
-    let staged_commit = bob_group
-        .read_keys_and_stage_commit(&remove_add_commit_result.commit, &[], provider)
-        .expect("error staging commit");
     bob_group
-        .merge_commit(provider, staged_commit)
-        .expect("Error merging commit.");
+        .store_pending_proposal(provider.storage(), staged_proposal)
+        .expect("Error storing proposal");
+
+    // Alice then commit to the proposal and at the same time adds Charlie
```
raphaelrobert, line 655:
```suggestion
    // Alice then commits to the proposal and at the same time adds Charlie
```

openmls/src/group/mls_group/tests_and_kats/tests/proposals.rs
```diff
@@ -438,147 +395,62 @@ fn test_group_context_extension_proposal_fails(
...
+    //    .expect_err("Bob was able to propose an extension not supported by all other parties.");
+    //
+    //assert_eq!(
+    //    e,
+    //    ProposalError::CreateGroupContextExtProposalError(
+    //        CreateGroupContextExtProposalError::KeyPackageExtensionSupport(
+    //            KeyPackageExtensionSupportError::UnsupportedExtension
+    //        )
+    //    )
+    //);
```
raphaelrobert, line 443: Curious what is blocking this, as #1130 is not very explicit. Is it unsupported extensions? It's fine if that's out-of-scope if it's too complicated.

openmls/src/group/mls_group/tests_and_kats/utils.rs
```diff
@@ -98,3 +99,67 @@ pub(crate) fn setup_client(
...
+    ciphersuite: Ciphersuite,
+    provider: &Provider,
+) -> (
+    MlsGroup,
+    SignatureKeyPair,
+    MlsGroup,
+    SignatureKeyPair,
+    CredentialWithKey,
+) {
+    let group_aad = b"Alice's test group";
```
raphaelrobert, line 113: Does this serve a purpose?

raphaelrobert (commented):
Thanks for doing this long overdue work. Just a few nits and questions.

openmls/src/group/mls_group/tests_and_kats/tests/proposals.rs
kkohbrok: I'm not sure, I just updated the code to work with the `MlsGroup` and ran it to see if the issue had resolved itself since the issue was filed (it hasn't).

I think this is about OpenMLS ensuring that everyone in a group actually has the capability to support an extension that the application wants to introduce via this proposal. I think we already have a check that would prevent such a proposal from being committed, but again, I haven't really checked.

I think fixing and properly testing this is better done in a separate PR.

openmls/src/group/mls_group/tests_and_kats/utils.rs
kkohbrok: Not really, I think. This was already here when this function was still using the CoreGroup, so I just kept it as it was. I can remove it if you'd prefer that.

kkohbrok (commented):
Thanks for the review!

openmls/src/group/mls_group/tests_and_kats/utils.rs
raphaelrobert: I think we shouldn't carry over orphaned/unused parts of the tests, so it's best to remove it.

raphaelrobert (changes requested):
Just one more nit.

raphaelrobert (approved):
Thanks!

PR #1620, opened 2024-07-17
Move tests and kats into their own dir under mls_group

kkohbrok:
This PR does the following:
- Move tests and kats from the `core_group` dir to the `mls_group` dir in preparation to migrate all tests to use the `MlsGroup` API
- Move utility functions for tests in to separate `utils` module
- Move the `find_key_package_from_welcome_secret` function from `CoreGroup` to `Welcome` and rename it to `find_encrypted_group_secret`


franziskuskiefer (approved):
Looks much nicer, thanks!

PR #1626, opened 2024-07-23
Make the versions of all the intermediate packages pre-release as well

keks:


openmls_test/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "openmls_test"
-version = "0.1.0"
```
franziskuskiefer, line 3: This is the first release of this crate. Let's stay with `0.1.0-pre.1`

franziskuskiefer (commented):
one nit

openmls_test/Cargo.toml
keks: done

franziskuskiefer (approved):

PR #1629, opened 2024-07-24
Add `MlsGroup` endpoint to create Add commit without path

kkohbrok:

PR #1631, opened 2024-07-24
Migrate remaining `CoreGroup` tests to `MlsGroup`

kkohbrok:
Migrates all remining tests to use `MlsGroup` rather than `CoreGroup`, except for one test which is already migrated as part of #1629.

### DIFF
--- a/openmls/src/group/tests_and_kats/tests/encoding.rs
+++ b/openmls/src/group/tests_and_kats/tests/encoding.rs
@@ -2,10 +2,8 @@ use openmls_traits::crypto::OpenMlsCrypto;
 use tls_codec::{Deserialize, Serialize};
 
 use crate::group::tests_and_kats::utils::*;
-use crate::{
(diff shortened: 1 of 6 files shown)


openmls/src/group/tests_and_kats/tests/encoding.rs
```diff
@@ -2,10 +2,8 @@ use openmls_traits::crypto::OpenMlsCrypto;
 use tls_codec::{Deserialize, Serialize};
 
 use crate::group::tests_and_kats::utils::*;
-use crate::{
-    binary_tree::LeafNodeIndex, framing::*, group::*, key_packages::*, messages::*,
-    schedule::psk::store::ResumptionPskStore, test_utils::*,
-};
+use crate::prelude::LeafNodeParameters;
```
raphaelrobert, line 5: Not a strong requirement for tests, but it's nice to avoid `prelude`.

PR #1633, opened 2024-07-25
Remove `aad` getter from storage provider trait

kkohbrok:


raphaelrobert (approved):
Looks like I missed this the other day, thanks for picking it up.

PR #1639, opened 2024-08-08
Introduce PublicStorageProvider

raphaelrobert:
This PR introduces a `PublicStorageProvider` that can be used by applications that only want to use the `PublicGroup`.


kkohbrok (approved):
Looks good, thanks!

PR #1642, opened 2024-08-08
Remove OpenMlsProvider dependency of PublicGroup.

raphaelrobert:
It was impossible to instantiate a `PublicGroup` without implementing a full `StorageProvider` (via `OpenMlsProvider`), this PR removes that requirement.


franziskuskiefer (approved):
Thanks.

We should probably have some more integration tests for the public group that makes sure that it can be used as intended.

franziskuskiefer:
Looks like some tests fail on matching the correct error.

franziskuskiefer (approved):

PR #1644, opened 2024-08-13
Fix `MlsGroup::delete()` function to delete encryption keypairs

kkohbrok:


franziskuskiefer (approved):
Thanks!
We still have the other encryption keys. From what I can tell they're deleted correctly. But we should really get rid of them.

PR #1646, opened 2024-08-19
Fix Missing Persisting of Message Secrets

keks:
In `CoreGroup::merge_staged_commit` we forgot persisting the new message secrets. This PR addresses this. As part of testing this, I also updated the test to use separate providers for different group members.


franziskuskiefer (approved):
Thanks! lgtm

neekolas (approved):

PR #1647, opened 2024-08-20
Remove `CoreGroup`

kkohbrok:
This PR does the following
- inline the fields of `CoreGroup` into `MlsGroup`
- move `CoreGroup` functions into `MlsGroup`
  - if the `CoreGroup` function was called by only one `MlsGroup` function, inline it if that doesn't make the `MlsGroup` function too big
- get rid of the `core_group` module, i.e. re-structure modules s.t. all `MlsGroup` functions are in the `mls_group` module or sub-modules thereof
- simplify and refactor small bits where it was opportune

REVIEWERS: I tried to preserve the public API, but might have missed something here or there, so please look out for changes that I have overlooked.


openmls/src/group/mls_group/mod.rs
```diff
@@ -39,15 +60,57 @@ use config::*;
...
+    pub credential: Credential,
+    /// The member's public HPHKE encryption key.
+    pub encryption_key: Vec<u8>,
+    /// The member's public signature key.
+    pub signature_key: Vec<u8>,
+}
+
+impl Member {
+    /// Create new member.
+    pub fn new(
```
raphaelrobert, line 99: Does this need to be public?

openmls/src/group/mls_group/mod.rs
```diff
@@ -229,31 +304,491 @@ impl MlsGroup {
...
-        self.group
-            .public_group()
+        self.public_group
             .leaf(self.own_leaf_index())
             .map(|node| node.credential())
             .ok_or_else(|| LibraryError::custom("Own leaf node missing").into())
     }
 
+    /// Get the required capabilities extension of this group.
+    pub(crate) fn required_capabilities(&self) -> Option<&RequiredCapabilitiesExtension> {
```
raphaelrobert, line 314: `pub(crate)` methods should not be among the public methods. This also applies to quite a few other methods below.

openmls/src/group/mls_group/mod.rs
```diff
@@ -229,31 +304,491 @@ impl MlsGroup {
...
+        store: &Storage,
+    ) -> Result<(), Storage::Error> {
+        store.delete_encryption_epoch_key_pairs(
+            self.group_id(),
+            &GroupEpoch::from(self.context().epoch().as_u64() - 1),
+            self.own_leaf_index().u32(),
+        )
+    }
+
+    pub(crate) fn create_commit<Provider: OpenMlsProvider>(
```
raphaelrobert, line 454: This should not be in `mod.rs`, it's distinct enough to have its own file.

openmls/src/group/mls_group/mod.rs
```diff
@@ -418,28 +977,46 @@ impl MlsGroup {
...
                     // We can be sure the encryption will work because the plaintext was created by us
                     .map_err(|_| LibraryError::custom("Malformed plaintext"))?;
-                MlsMessageOut::from_private_message(ciphertext, self.group.version())
+                MlsMessageOut::from_private_message(ciphertext, self.version())
             }
         };
         Ok(msg)
     }
 
+    // Encrypt an PublicMessage into an PrivateMessage
```
raphaelrobert, line 997: The input is not a piblic message, the wording is misleading.

openmls/src/group/mls_group/mod.rs
```diff
@@ -549,9 +1121,29 @@ impl MlsGroup {
...
+    /// A [`MessageSecretsStore`] that stores message secrets.
+    /// By default this store has the length of 1, i.e. only the [`MessageSecrets`]
+    /// of the current epoch is kept.
+    /// If more secrets from past epochs should be kept in order to be
+    /// able to decrypt application messages from previous epochs, the size of
+    /// the store must be increased through [`max_past_epochs()`].
+    message_secrets_store: MessageSecretsStore,
+
+    /// Resumption psk store. This is where the resumption psks are kept in a rollover list.
+    pub(crate) resumption_psk_store: ResumptionPskStore,
```
raphaelrobert, line 1137: Not necessarily a big issue, but I don't think we mix fields with different visibility in the library. I think we usually have getters instead, unless all fields are `pub`.

openmls/src/group/mls_group/proposal.rs
```diff
@@ -458,4 +453,123 @@ impl MlsGroup {
             .remove(proposal_ref)
             .ok_or(RemoveProposalError::ProposalNotFound)
     }
+
+    // === Create handshake messages ===
+    // TODO: share functionality between these.
```
raphaelrobert, line 458: Maybe this should be an issue instead?

openmls/src/group/mls_group/proposal.rs
```diff
@@ -458,4 +453,123 @@ impl MlsGroup {
...
+
+    // 11.1.1. Add
```
raphaelrobert, line 460:
```suggestion
    // 12.1.1. Add
```

openmls/src/group/mls_group/proposal.rs
```diff
@@ -458,4 +453,123 @@ impl MlsGroup {
...
+            framing_parameters,
+            self.own_leaf_index(),
+            proposal,
+            self.context(),
+            signer,
+        )
+        .map_err(|e| e.into())
+    }
+
+    // 11.1.2. Update
```
raphaelrobert, line 490:
```suggestion
    // 12.1.2. Update
```

openmls/src/group/mls_group/proposal.rs
```diff
@@ -458,4 +453,123 @@ impl MlsGroup {
...
+    // struct {
+    //     KeyPackage key_package;
```
raphaelrobert, line 492:
```suggestion
    //     LeafNode leaf_node;
```

openmls/src/group/mls_group/proposal.rs
```diff
@@ -458,4 +453,123 @@ impl MlsGroup {
...
+        AuthenticatedContent::member_proposal(
+            framing_parameters,
+            self.own_leaf_index(),
+            proposal,
+            self.context(),
+            signer,
+        )
+    }
+
+    // 11.1.3. Remove
```
raphaelrobert, line 513:
```suggestion
    // 12.1.3. Remove
```

openmls/src/group/mls_group/proposal.rs
```diff
@@ -458,4 +453,123 @@ impl MlsGroup {
...
+    // struct {
+    //     KeyPackageRef removed;
```
raphaelrobert, line 515:
```suggestion
    //     uint32 removed;
```

openmls/src/group/mls_group/proposal.rs
```diff
@@ -458,4 +453,123 @@ impl MlsGroup {
...
+            framing_parameters,
+            self.own_leaf_index(),
+            proposal,
+            self.context(),
+            signer,
+        )
+        .map_err(ValidationError::LibraryError)
+    }
+
+    // 11.1.4. PreSharedKey
```
raphaelrobert, line 538:
```suggestion
    // 12.1.4. PreSharedKey
```

raphaelrobert (changes requested):
Thanks for tackling this! I made a first pass with a few suggestions and change requests.

openmls/src/group/mls_group/mod.rs
kkohbrok: Good point. I've organized the individual functions into its own impl block.

openmls/src/group/mls_group/mod.rs
kkohbrok: Fair enough. I've merged this function and the content of `create_commit_params.rs` into `create_commit.rs`.

openmls/src/group/mls_group/mod.rs
kkohbrok: As far as I remember this is just copied from elsewhere. But I'll fix it.

openmls/src/group/mls_group/mod.rs
kkohbrok: All I did was inline this from `StagedCoreWelcome`. It was public there, too. I'd rather pull this straight in a follow-up.

openmls/src/group/mls_group/proposal.rs
kkohbrok: I just copied this from `core_group` and didn't investigate further. But feel free to file an issue.

kkohbrok (commented):
Thanks for the pass. I've tried to address all the issues and comment on the ones that I didn't. I did a lot of copying stuff from `core_group` and I'm happy to fix small stuff, but larger things I'd prefer to do in a follow-up.

openmls/src/group/mls_group/proposal.rs
raphaelrobert: Rather remove the comment completely, because I don't think it makes sense.

franziskuskiefer (commented):
Thanks for getting this started. I'm wondering if the storage is stable across this change. From the changes it looks like the test doesn't complain. Just given the size of the change I was wondering if it may break things. If not, that's great! We should just make sure that it's really the case.

openmls/src/group/mls_group/mod.rs
raphaelrobert: This is the only outstanding issue, why not fix it in this PR, like the rest of it?

raphaelrobert (approved):

traits/src/storage.rs
```diff
@@ -600,7 +600,6 @@ pub mod traits {
     pub trait GroupState<const VERSION: u16>: Entity<VERSION> {}
     pub trait GroupEpochSecrets<const VERSION: u16>: Entity<VERSION> {}
     pub trait LeafNodeIndex<const VERSION: u16>: Entity<VERSION> {}
-    pub trait GroupUseRatchetTreeExtension<const VERSION: u16>: Entity<VERSION> {}
```
franziskuskiefer, line 603: This is a breaking change when it doesn't get into v0.6.

But it also seems to changes the processing. What is the change here and should it be part of this PR?

openmls/src/group/errors.rs
```diff
@@ -107,15 +107,9 @@ pub enum ExternalCommitError<StorageError> {
     /// We don't support the ciphersuite of the group we are trying to join.
     #[error("We don't support the ciphersuite of the group we are trying to join.")]
     UnsupportedCiphersuite,
-    /// Sender not found in tree.
```
franziskuskiefer, line 110: Enum changes are breaking changes. So these need to go into the changelog.

franziskuskiefer (changes requested):
Thanks! Generally looks good. I didn't go too deep into the actual changes and looked for more for breaking things.

- Enums (esp. errors) need to be listed as changes
- There's a change in the storage trait that I'm not sure I understand

traits/src/storage.rs
kkohbrok: Good question. I noticed that `use_ratchet_tree_extension` exists in the `CoreGroupConfig` (or something like that) and in the `MlsGroupJoinConfig`. So as part of the "inlining" of `CoreGroup` into `MlsGroup`, I removed that superfluous config flag. By removing it, the storage provider functions became unused, so I removed those as well. So I'm not entirely sure it was strictly in-scope for this PR, but it felt weird doing the extra work of inlining a superfluous config flag when removing it was quicker 🤷🏻.

openmls/src/group/errors.rs
kkohbrok: Good point. I reverted all changes to error enums.

franziskuskiefer (dismissed):
There are a few open comments from @raphaelrobert.
But I think this is good to go otherwise.

traits/src/storage.rs
kkohbrok: I assume as much. At least now everything is reading and writing the same thing.

raphaelrobert (approved):

PR #1648, opened 2024-08-22
Bump versions and fix build error

keks:
Bump versions for the next round of prereleases, and bump `time` (a transitive dependency) in the DS to fix [this problem](https://github.com/time-rs/time/issues/681).


franziskuskiefer (approved):
thanks

PR #1649, opened 2024-08-29
Slightly refactor public group storage interface

kkohbrok:


openmls/src/group/public_group/mod.rs
```diff
@@ -396,14 +396,14 @@ impl PublicGroup {
     }
 
     /// Deletes the [`PublicGroup`] from storage.
-    pub(crate) fn delete<Storage: PublicStorageProvider>(
-        &self,
+    pub fn delete<Storage: PublicStorageProvider>(
```
franziskuskiefer, line 399: We really need integration tests for the public group. Do we have an issue for that?

franziskuskiefer (approved):
lgtm

openmls/src/group/public_group/mod.rs
kkohbrok: Good point. I just filed #1650.

PR #1654, opened 2024-09-06
Draft of Sans-IO style message processing API

kkohbrok:


openmls/src/group/mls_group/processing_state_machine.rs
```diff
@@ -0,0 +1,110 @@
...
+        group.process_authenticated_content(
+            crypto,
+            authenticated_content,
+            credential,
+            psks,
+            old_epoch_keypairs,
+            leaf_node_keypairs,
+        )
+    }
+}
```
raphaelrobert, line 110: Can we use some more expressive terms than `perform_io` and `PerformedIo`? E.g. `io_step` and `IoResult`? Or something more specific to the data being processed.

PR #1655, opened 2024-09-19
Add check and (basic) test for check 701: remove can't contain inde to blank leaf

keks:
This PR adds check 701 from [the dashboard](https://validation.openmls.tech/) and a basic test. It doesn't properly test validation for incoming messages, because I wanted more checks done instead of futzing with Frankenstein messages for now.

### DIFF
--- a/openmls/src/group/public_group/validation.rs
+++ b/openmls/src/group/public_group/validation.rs
@@ -424,7 +424,7 @@ impl PublicGroup {
                 return Err(ProposalValidationError::UnknownMemberRemoval);
             }
 
-            // removed node can not be blank (check 701)
+            // valn0701: removed node can not be blank
             if self.treesync().leaf(removed).is_none() {
                 return Err(ProposalValidationError::UnknownMemberRemoval);
             }
--- a/openmls/src/group/public_group/validation.rs
+++ b/openmls/src/group/public_group/validation.rs
@@ -423,6 +423,11 @@ impl PublicGroup {
             if !self.treesync().is_leaf_in_tree(removed) {
                 return Err(ProposalValidationError::UnknownMemberRemoval);
             }
+
+            // removed node can not be blank (check 701)
+            if self.treesync().leaf(removed).is_none() {
+                return Err(ProposalValidationError::UnknownMemberRemoval);
+            }
         }
 
         Ok(())
--- a/openmls/src/group/tests_and_kats/tests/remove_operation.rs
+++ b/openmls/src/group/tests_and_kats/tests/remove_operation.rs
@@ -4,6 +4,132 @@ use crate::group::tests_and_kats::utils::{generate_credential_with_key, generate
 use crate::{framing::*, group::*};
 use openmls_traits::prelude::*;
 
+#[openmls_test::openmls_test]
+fn remove_blank() {
+    let group_id = GroupId::from_slice(b"Test Group");
+
+    let alice_provider = &Provider::default();
+    let bob_provider = &Provider::default();
+    let charlie_provider = &Provider::default();
+
+    // Generate credentials with keys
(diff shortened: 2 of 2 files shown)


openmls/src/group/public_group/validation.rs
```diff
@@ -423,6 +423,11 @@ impl PublicGroup {
             if !self.treesync().is_leaf_in_tree(removed) {
                 return Err(ProposalValidationError::UnknownMemberRemoval);
             }
+
+            // removed node can not be blank (check 701)
```
franziskuskiefer, line 427: It would be nice if we'd had anchors in the dashboard to link back here. Do you think that's easily doable?

openmls/src/group/public_group/validation.rs
```diff
@@ -423,6 +423,11 @@ impl PublicGroup {
             if !self.treesync().is_leaf_in_tree(removed) {
```
franziskuskiefer, line 423: How do we handle transition from ValSem annotations to the dashboard? Shall we add both where applicable?

openmls/src/group/tests_and_kats/tests/remove_operation.rs
```diff
@@ -4,6 +4,132 @@ use crate::group::tests_and_kats::utils::{generate_credential_with_key, generate
...
+    // try removing bob a second time
+    alice_group
+        .remove_members(
+            alice_provider,
+            &alice_credential_with_key_and_signer.signer,
+            &[bob_index],
+        )
+        .expect_err("using API to re-remove someone should fail");
+
+    // TODO: craft another remove commit that removes bob's (blank) index and check that merging
```
franziskuskiefer, line 129: How do we track these tests? Do you want to add it on the dashboard or an issue?

franziskuskiefer (dismissed):
Thanks lgtm.
The comments are really more procedural.

openmls/src/group/public_group/validation.rs
keks: Yeah that would be nice. Options I see:

- make each check a public function and link to the function's rustdoc
  - (+) relatively easy
  - (-) not quite the same as linking to the code
- decide on an anchor string pattern and use text fragments to link to the specific place
  - (+) relatively easy 
  - (-) [doesn't work in firefox, but might come soon](https://bugzilla.mozilla.org/show_bug.cgi?id=1753933)
  - (-) doesn't tolerate if the check moves to a different file
- decide on an anchor string and link to the github search for that string
  - (+) relatively easy 
  - (+) works everywhere
  - (-) might not work for everyone: I think you can only search when you are logged into github
    - not sure that is a problem, this is primarily an internal tool anway

Maybe we should link to github search for now and possibly pivot to text fragment links once firefox supports them.

keks: https://github.com/openmls/validation.dhall/pull/13

openmls/src/group/tests_and_kats/tests/remove_operation.rs
keks: I would track it on the dashboard. I think tracking it in individual issues would be too messy, and creating a single tracking github issue just means we re-do the work of the dashboard. Unless we automate it :nerd_face:

openmls/src/group/public_group/validation.rs
keks: https://github.com/openmls/validation.dhall/pull/14

openmls/src/group/tests_and_kats/tests/remove_operation.rs
keks: https://github.com/openmls/validation.dhall/pull/14

franziskuskiefer (approved):
Thanks, let's do this.

PR #1657, opened 2024-09-27
Fix checks for leaf node validation (valn01xx) and valn1207

keks:
This PR adds/fixes a check, adds a few tests around the check, and adds a bunch of new validation IDs into code comments, so we can better find them. The tests are relatively shallow, but a bit more tight than before.

One thing I changed is that when you specify a non-default LeafNode extension when creating a group, you also have to specify that you have the capability to deal with it. I suppose one could just always set that, but I'm not sure that is a good idea.

I removed some test cases from the valsem105 test, because they seemed somewhat nonsensical to me. Let me know if I was wrong there.

I disabled the checking of valn0106 in tests, because that contains old lifetimes and this fails with the kats. Is there a better way?

I also changed how `Capabilities` deals with defaults: before we would, at construction time, copy all the defaults in there, and then look in there if we have them. We also had checks that checked if an item is in the list of default extensions/proposals. Instead, I added a method `is_default` on `ExtensionType` and `ProposalType` for checking this, and keep the defaults out of the `Capabilities` struct. If we check whether an extension type is supported, we instead as `type.is_default() || supported.contains(type)` (paraphrased). That seems both cleaner and more efficient to me.

### DIFF
--- a/openmls/Cargo.toml
+++ b/openmls/Cargo.toml
@@ -53,6 +53,8 @@ test-utils = [
   "dep:once_cell",
   "backtrace",
 ]
+test-skip-lifetime-validity-check = [
+] # Skip the check whether a lifetime is valid. Needed for KATs with since-expired key packages.
 backtrace = ["dep:backtrace"]
 libcrux-provider = [
   "dep:openmls_libcrux_crypto",
--- a/openmls/Cargo.toml
+++ b/openmls/Cargo.toml
@@ -53,8 +53,6 @@ test-utils = [
   "dep:once_cell",
   "backtrace",
 ]
-test-skip-lifetime-validity-check = [
-] # Skip the check whether a lifetime is valid. Needed for KATs with since-expired key packages.
 backtrace = ["dep:backtrace"]
 libcrux-provider = [
   "dep:openmls_libcrux_crypto",
--- a/openmls/src/group/public_group/validation.rs
+++ b/openmls/src/group/public_group/validation.rs
@@ -168,8 +168,11 @@ impl PublicGroup {
     ///  - ValSem110: Update Proposal: Encryption key must be unique among proposals & members
     ///  - ValSem206: Commit: Path leaf node encryption key must be unique among proposals & members
     ///  - ValSem207: Commit: Path encryption keys must be unique among proposals & members
-    ///  - valn0111: Verify that the following fields are unique among the members of the group: `signature_key`
-    ///  - valn0112: Verify that the following fields are unique among the members of the group: `encryption_key`
+    ///  - [valn0111]: Verify that the following fields are unique among the members of the group: `signature_key`
+    ///  - [valn0112]: Verify that the following fields are unique among the members of the group: `encryption_key`
+    ///
+    /// [valn0111]: https://validation.openmls.tech/#valn0111
+    /// [valn0112]: https://validation.openmls.tech/#valn0112
     pub(crate) fn validate_key_uniqueness(
         &self,
         proposal_queue: &ProposalQueue,
@@ -265,7 +268,7 @@ impl PublicGroup {
 
         // Validate uniqueness of signature keys
         //  - ValSem101
-        //  - valn0111
+        //  - https://validation.openmls.tech/#valn0111
         for signature_key in signature_keys {
             if !signature_key_set.insert(signature_key) {
                 return Err(ProposalValidationError::DuplicateSignatureKey);
@@ -278,7 +281,7 @@ impl PublicGroup {
         //  - ValSem110
         //  - ValSem206
         //  - ValSem207
-        //  - valn0112
+        //  - https://validation.openmls.tech/#valn0112
         for encryption_key in encryption_keys {
             if init_key_set.contains(&encryption_key) {
                 return Err(ProposalValidationError::InitEncryptionKeyCollision);
@@ -382,7 +385,7 @@ impl PublicGroup {
                 return Err(ProposalValidationError::UnknownMemberRemoval);
             }
 
-            // valn0701: removed node can not be blank
+            // https://validation.openmls.tech/#valn0701: removed node can not be blank
             if self.treesync().leaf(removed).is_none() {
                 return Err(ProposalValidationError::UnknownMemberRemoval);
             }
@@ -581,7 +584,7 @@ impl PublicGroup {
         }
 
         // If there is a required capabilities extension, check if that one
-        // is supported (valn0103).
+        // is supported (https://validation.openmls.tech/#valn0103).
         if let Some(required_capabilities) =
             self.group_context().extensions().required_capabilities()
         {
@@ -589,7 +592,7 @@ impl PublicGroup {
             capabilities.supports_required_capabilities(required_capabilities)?;
         }
 
-        // Check that the credential type is supported by all members of the group (valn0104).
+        // Check that the credential type is supported by all members of the group (https://validation.openmls.tech/#valn0104).
         if !self.treesync().full_leaves().all(|node| {
             node.capabilities()
                 .contains_credential(leaf_node.credential().credential_type())
@@ -599,7 +602,7 @@ impl PublicGroup {
 
         // Check that the capabilities field of this LeafNode indicates
         // support for all the credential types currently in use by other
-        // members (valn0104).
+        // members (https://validation.openmls.tech/#valn0104).
         if !self
             .treesync()
             .full_leaves()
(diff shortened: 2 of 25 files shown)


openmls/src/group/tests_and_kats/tests/group.rs
```diff
@@ -104,6 +104,44 @@ fn basic_group_setup() {
         };
 }
 
+/// This test checks that we can't create a group that is invalid.
+/// Specifically, we test that the extensions in the leaf node must be supported by the lead node's
+/// own capabilties (valn0107).
```
franziskuskiefer, line 109:
```suggestion
/// own capabilities (valn0107).
```

openmls/src/group/tests_and_kats/tests/proposal_validation.rs
```diff
@@ -1042,10 +1038,10 @@ fn test_valsem105() {
     for key_package_version in [
         KeyPackageTestVersion::WrongCiphersuite,
         KeyPackageTestVersion::WrongVersion,
-        KeyPackageTestVersion::UnsupportedVersion,
-        KeyPackageTestVersion::UnsupportedCiphersuite,
         KeyPackageTestVersion::ValidTestCase,
     ] {
+        println!("running test {key_package_version:?}");
```
franziskuskiefer, line 1043: Drop debug print.

openmls/src/group/tests_and_kats/tests/proposal_validation.rs
```diff
@@ -1077,13 +1073,6 @@ fn test_valsem105() {
             KeyPackageTestVersion::WrongVersion => {
                 franken_key_package.protocol_version = 999;
             }
```
franziskuskiefer, line 1075: I'm not sure I understand why these tests need to go.

openmls/src/treesync/node/leaf_node/capabilities.rs
```diff
@@ -142,40 +142,66 @@ impl Capabilities {
...
+    pub(crate) fn contains_proposal(&self, proposal_type: ProposalType) -> bool {
+        proposal_type.is_default() || self.proposals().contains(&proposal_type)
+    }
+
+    /// Check if these [`Capabilities`] contain all the versions.
+    pub(crate) fn contains_version(&self, version: ProtocolVersion) -> bool {
+        default_versions().contains(&version) || self.versions().contains(&version)
+    }
+
+    /// Check if these [`Capabilities`] contain all the ciphersuites.
```
franziskuskiefer, line 197: It's only one ciphersuite

openmls/src/treesync/node/leaf_node/capabilities.rs
```diff
@@ -142,40 +142,66 @@ impl Capabilities {
...
+    pub(crate) fn contains_extension(&self, extension_type: ExtensionType) -> bool {
+        extension_type.is_default() || self.extensions().contains(&extension_type)
+    }
+
+    /// Check if these [`Capabilities`] contain the proposal.
+    pub(crate) fn contains_proposal(&self, proposal_type: ProposalType) -> bool {
+        proposal_type.is_default() || self.proposals().contains(&proposal_type)
+    }
+
+    /// Check if these [`Capabilities`] contain all the versions.
```
franziskuskiefer, line 192: It'd only one version.

openmls/tests/book_code.rs
```diff
@@ -121,10 +121,10 @@ fn book_operations() {
         .expect("error adding external senders extension to group context extensions")
         .ciphersuite(ciphersuite)
         .capabilities(Capabilities::new(
-            None, // Defaults to the group's protocol version
-            None, // Defaults to the group's ciphersuite
-            None, // Defaults to all basic extension types
-            None, // Defaults to all basic proposal types
+            None,                                    // Defaults to the group's protocol version
+            None,                                    // Defaults to the group's ciphersuite
+            Some(&[ExtensionType::Unknown(0xff00)]), // Defaults to all basic extension types
```
franziskuskiefer, line 126: The comment doesn't match that you pass in something.

openmls/tests/book_code.rs
```diff
@@ -137,6 +137,8 @@ fn book_operations() {
         .build();
     // ANCHOR_END: mls_group_create_config_example
 
+    println!("### {mls_group_create_config:?}");
```
franziskuskiefer, line 140: Drop debug print.

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -455,6 +456,36 @@ impl LeafNode {
         }
         Ok(())
     }
+
+    /// Perform all checks that can be done without further context:
+    /// - the used extensions are not known to be invalid in leaf leaf nodes
```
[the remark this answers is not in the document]
keks: I think that was just a typo. This was just about the entries labeled LN in this table: https://www.rfc-editor.org/rfc/rfc9420.html#name-mls-extension-types-registry

openmls/src/group/tests_and_kats/tests/proposal_validation.rs
keks: Are you sure? This is in a test (so it doesn't print to stdout in operation) and we are basically running sub-tests, and it's often difficult to understand which sub-test is failing. IMHO when running sub-tests it makes sense to print which one we are running now.

openmls/src/group/tests_and_kats/tests/proposal_validation.rs
keks: Ah, I think I misunderstood what they were testing. Let me take another look.

openmls/src/group/public_group/validation.rs
```diff
@@ -598,6 +560,81 @@ impl PublicGroup {
...
+    pub(crate) fn validate_leaf_node(
+        &self,
+        leaf_node: &crate::treesync::LeafNode,
+    ) -> Result<(), LeafNodeValidationError> {
+        // valn0103, valn0104, valn0107
+        self.validate_leaf_node_capabilities(leaf_node)?;
+
+        // valn0105 is done when sending
+
+        // valn0106
```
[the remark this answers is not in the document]
keks: I will add the CodeRefs to the dashboard. Or do you mean something else?

openmls/src/group/public_group/validation.rs
```diff
@@ -598,6 +560,81 @@ impl PublicGroup {
...
+        // don't enable in tests, because we are testing with kats that contain
+        // expired key packages
+        #[cfg(not(test))]
+        if let Some(lifetime) = leaf_node.life_time() {
```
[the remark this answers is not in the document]
keks: I added a comment.

openmls/src/group/tests_and_kats/tests/proposal_validation.rs
keks: Ah, no I think I removed these because the RFC doesn't actually require that the version and ciphersuite from the leaf node match those of the group, or the key package containing it. I suppose that is a bug in the RFC?

franziskuskiefer: The RFC has this verification of key packages
"Verify that the cipher suite and protocol version of the KeyPackage match those in the GroupContext."

https://www.rfc-editor.org/rfc/rfc9420.html#section-10.1-4.1

Some checks here look like they are for the capabilities though?
And the ciphersuite needs to be supported by everyone.

So I think these tests should still be valid. Or do I misunderstand what they are doing?

openmls/Cargo.toml
```diff
@@ -53,6 +53,8 @@ test-utils = [
   "dep:once_cell",
   "backtrace",
 ]
+test-skip-lifetime-validity-check = [
```
franziskuskiefer, line 56: I was thinking of using the `skip_validation.rs` Otherwise we'll get a lot of features ... 
Is there a reason not to use that skip validation approach?

franziskuskiefer (changes requested):

openmls/src/group/tests_and_kats/tests/proposal_validation.rs
keks: The checks you reference are for `key_package.cipher_suite` and `key_package.protocol_version`. The tests I remove check what happens for `key_package.leaf_node.capabilities.cipher_suites` and `key_package.leaf_node.capabilities.protocol_versions`.

These are two different things, and the RFC fails to actually bring them into any kind of relation.

openmls/Cargo.toml
keks: Oh, yeah good point. I forgot we had that :sweat_smile:

franziskuskiefer (approved):
Thanks, let's get this in.

PR #1660, opened 2024-10-09
Add test for incorrect max_past_epochs value

neekolas:
We noticed that there was an issue with the `max_past_epochs` flag being respected in the `MlsGroupCreateConfig`. Members of the group joining from a Welcome would see the specified value, but the group creator would always have a `max_past_epochs` set to 0.

This is because when the `MlsGroupBuilder::build` method is run, it uses the `max_past_epochs` flag from the builder instead of the `MlsGroupCreateConfig`

[This is how we fixed it](https://github.com/xmtp/openmls/pull/38/commits/043b347cb18d528647df36f500725ab57c41c7db) in our fork, but you may want to handle differently. Is the `max_past_epochs` flag on the `MlsGroupBuilder` even necessary with the group create config. From my cursory look, it seems to be always `None`.

### DIFF
--- a/openmls/src/group/mls_group/past_secrets.rs
+++ b/openmls/src/group/mls_group/past_secrets.rs
@@ -21,7 +21,7 @@ struct EpochTree {
 #[cfg_attr(feature = "crypto-debug", derive(Debug))]
 pub(crate) struct MessageSecretsStore {
     // Maximum size of the `past_epoch_trees` list.
-    max_epochs: usize,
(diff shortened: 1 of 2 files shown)


kkohbrok (commented):
Thanks for the PR! Looks like we can just remove the `max_past_epochs` field from the builder alltogether, as it appears that it's never actually set. I've filed #1662 to that effect.

keks:
Merged in #1664

PR #1662, opened 2024-10-11
Remove max_past_epochs field from MlsGroupBuilder

kkohbrok:
Fixes the issue raised in #1660.

This PR removes the `max_past_epochs` field from the builder, which is erroneously read instead of the field from the group create config. There was already no way to set it, so removing it only restores the expected behaviour of applications being able to set the `max_past_epochs` via the group create config.

### DIFF
--- a/openmls/src/ciphersuite/hpke.rs
+++ b/openmls/src/ciphersuite/hpke.rs
@@ -68,6 +68,7 @@ impl From<CryptoError> for Error {
     }
 }
 
+/// Context for HPKE encryption
 #[derive(Debug, Clone, TlsSerialize, TlsDeserialize, TlsDeserializeBytes, TlsSize)]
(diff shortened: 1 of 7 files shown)


franziskuskiefer:
This should be covered in #1661. Do you not like the solution there?

kkohbrok:
Sorry, I should really have commented in #1661.

The solution in #1661 is to set the `max_past_epochs` value in the builder to the value in the config, which is then used during group creation. However, we can just remove that middle-variable and use the variable in the config during group creation. The middle variable (i.e. the one in the builder) isn't really used anywhere else and (as pointed out in #1660) is never even set.

keks:
I think this can be closed, since I included in in #1661.

kkohbrok:
Closing, because this was included in #1661.

PR #1664, opened 2024-10-14
Add test for incorrect max_past_epochs value

keks:
I rebased @neekolas' commit from #1660  on top of our new main that already fixes the issue.


kkohbrok (commented):
Thanks for the PR! So far, we've used `#[cfg(test)]` when making stuff available selectively for tests. So for consistency reasons I have a slight preference for that. Otherwise this looks good!

franziskuskiefer (approved):
lgtm
Not sure about @kkohbrok's `cfg` comment. Is it about making `max_epochs` `pub(crate)` visible?

kkohbrok:
Yes, I was referring to the visibility of the `max_epochs` field. But as I said, I don't have a strong opinion here, so feel free to merge.

keks:
Alright, I'll get it in as-is.

PR #1672, opened 2024-10-31
`VerifiableGroupInfo::epoch()` added to public API

### LINKED ISSUE #1671: `VerifiableGroupInfo::epoch()` in public API
I have a small issue & PR and would like to contribute. Thanks!

Not exactly -- I've encountered a situation where I want to inspect the epoch in a `VerifiableGroupInfo` object (created from a received `GroupInfo` message) in addition to the group_id. As best as I can tell, the public API does not allow for reading that property.

**Describe the solution you'd like.**
A method in the `VerifiableGroupInfo` public API for reading the epoch.

josephlukefahr:
Closes #1671


franziskuskiefer (approved):
Thanks for your contribution!
I agree that this may be useful information in order to validate an incoming group info on the application level.

PR #1675, opened 2024-11-04
Add Commit Builder

### LINKED ISSUE #1674: Add Commit Builder
A common use case is to have a commit that contains multiple proposals, of different types. It would be nice if we provided an API that offered a nice way of generating these commits. Ideally, the API should also be Sans-IO-friendly, i.e. have distinct "phases" (intermediate builder types) for doing IO and mutably borrow the `MlsGroup` struct in the mean time (so it is effectively locked and can't be used for other things).

keks:
This PR is for adding a builder for commits. I have started a draft design doc [here].

[here]: https://md.cryspen.com/s/TqZXcU-gA

Closes #1674


keks:
One note: I was considering to add more tests, but I wasn't quite sure what to test - most of the builder is about the actual building of commits, not about collecting the parameters to be used, and we already tests this by way of testing how we build commits.

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -0,0 +1,795 @@
+//! This module contains the types for building commits.
+//!
+//! A living design doc can be found here: <https://md.cryspen.com/s/TqZXcU-gA>
```
kkohbrok, line 3: The docs here seem a bit draft-y. Since this is public, it would be nice to have proper documentation here.

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -0,0 +1,795 @@
...
+    pub fn propose_add(mut self, key_package: KeyPackage) -> Self {
+        self.stage
+            .own_proposals
+            .push(Proposal::Add(AddProposal { key_package }));
+        self
+    }
+
+    /// Adds an Add proposal to the provided [`KeyPackage`] to the list of proposals to be
+    /// committed.
+    pub fn propose_adds(mut self, key_packages: impl IntoIterator<Item = KeyPackage>) -> Self {
```
kkohbrok, line 201: Do we need two functions for adding and removing? I'd say it's enough to have the one that takes an IntoIterator as input.

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -0,0 +1,795 @@
...
+                    other_extensions,
+                    confirmation_tag,
+                    builder.group.own_leaf_index(),
+                )
+            };
+            // Sign to-be-signed group info.
+            Some(group_info_tbs.sign(signer)?)
+        } else {
+            None
+        };
```
kkohbrok, line 535: Just a nit (and I'm not sure if it was like this before), but I'd slightly prefer to invert the check and return `None` in the `if` and do the elaborate computation in the `else`. That way when reading the code I immediately see the obvious `None` return and don't have to parse the `else` without seeing the original condition. The same goes for the `welcome_option` below.

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -0,0 +1,795 @@
...
+
+    /// Gets a the Welcome messsage. Only [`Some`] if new clients have been added in the commit.
+    /// Performs a copy of the Welcome. For owned version, see [`Self::into_welcome_msg`].
+    pub fn to_welcome_msg(&self) -> Option<MlsMessageOut> {
+        self.welcome
+            .as_ref()
+            .map(|welcome| MlsMessageOut::from_welcome(welcome.clone(), self.version))
+    }
+
+    /// Gets a the GroupInfo message. Onlt [`Some`] if... when? TODO: finish writing this comment
```
kkohbrok, line 708: This TODO should probably be addressed before merging.

kkohbrok, line 708:
```suggestion
    /// Gets a the GroupInfo message. Only [`Some`] if... when? TODO: finish writing this comment
```

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -0,0 +1,795 @@
...
+    }
+
+    /// Gets a the Welcome messsage. Only [`Some`] if new clients have been added in the commit.
+    /// For a borrowed version, see [`Self::to_welcome_msg`].
+    pub fn into_welcome_msg(self) -> Option<MlsMessageOut> {
+        self.welcome
+            .map(|welcome| MlsMessageOut::from_welcome(welcome, self.version))
+    }
+
+    /// Gets a the GroupInfo messsage. Only [`Some`] if TODO: when???.
```
kkohbrok, line 745: Same as above.

openmls/src/group/mls_group/processing.rs
```diff
@@ -119,32 +119,24 @@ impl MlsGroup {
...
-        // storing the current [`StagedCommit`] from the commit results
-        self.group_state = MlsGroupState::PendingCommit(Box::new(PendingCommitState::Member(
-            create_commit_result.staged_commit,
-        )));
-        provider
-            .storage()
-            .write_group_state(self.group_id(), &self.group_state)
-            .map_err(CommitToPendingProposalsError::StorageError)?;
+        //let params = CreateCommitParams::builder().regular_commit().build();
+        //let create_commit_result = self.create_commit(params, provider, signer)?;
```
kkohbrok, line 123: Remove?

openmls/src/group/mls_group/proposal_store.rs
```diff
@@ -532,6 +533,139 @@ impl ProposalQueue {
...
+    ///
+    /// - Extract Adds and filter for duplicates
+    /// - Build member list with chains: Updates & Removes
+    /// - Check for invalid indexes and drop proposal
+    /// - Check for presence of Removes and delete Updates
+    /// - Only keep the last Update
+    ///
+    /// Return a [`ProposalQueue`] and a bool that indicates whether Updates for
+    /// the own node were included
+    pub(crate) fn filter_proposals_without_inline(
```
kkohbrok, line 558: Is there a reason why this wouldn't consider inline proposals? I'd think it does.

kkohbrok (changes requested):
Thanks for picking this up! Outside of a few nits, this looks very solid to me.

openmls/src/group/mls_group/proposal_store.rs
keks: That function name is a bit wonky, yes. I copied the body from `filter_proposals`, which also takes an argument `inline_proposals`. The API of that function didn't quite work with the new commit builder, so I made a copy where I changed it. Part of that was removing the `inline_proposals` argument.  I didn't want to change the existing code because (a) we'll probably get rid of `create_commit` soon anyway and we can just remove the old version then and (b) it might just break things. If you think it's important, I can look into making `create_commit` use `filter_proposals_without_inline`. We could then remove the current `filter_proposals` and rename `filter_propoals_without_inline` to `filter_proposals`. It might work, but I am not sure (I forgot some details because I was away for a week)

openmls/src/group/mls_group/commit_builder.rs
keks: I guess that's because they _are_ a draft :sweat_smile:. For now I removed the drafty part and just refer to the CommitBuilder struct docs. I guess the right thing to do would be to add proper examples to the book code, but I am not sure if I should replace all the current examples with a commit builder. Happy to hear opinions.

openmls/src/group/mls_group/commit_builder.rs
keks: It was like this before, but I think I agree with the general criticism. What I did now is that I assigned the if-conditions to variables and then do 

```rust
let group_info = if !needs_group_info {
  None
} else {
   ... ;
  Some(...)
};`

keks:
I marked all but the two things I thought require a bit more conversation as resolved, feel free to undo that.

keks:
Looks like rustdoc wants the code in doc comment code blocks to compile, so I had to remove the example from the struct comments and just link to the book.

kkohbrok (approved):
Thanks, looks good to go!

PR #1682, opened 2024-11-22
Sqlite storage provider crate

kkohbrok:
Add a sqlite-backed storage provider to the OpenMLS crate ecosystem based on the `rusqlite` crate.


strong84:
Have some questions about the [CURRENT_VERSION](https://github.com/openmls/openmls/blob/f3e165b69c04c05c9b8e2c452fd8c041da6b1334/traits/src/storage.rs#L8):

In [MemoryStorage](https://github.com/openmls/openmls/blob/ffff3f7d9897080e61018db1c22911f80dd5810b/memory_storage/src/lib.rs#L983) CURRENT_VERSION is appended after the key. Looks like CURRENT_VERSION is not used In this [SqliteStorageProvider](https://github.com/openmls/openmls/tree/konrad/sqlite_storage_provider/sqlite_storage) branch.

In the [OpenMls Book](https://book.openmls.tech/traits/traits.html) it says:
`The traits are generic over a VERSION, which is used to ensure that the values that are persisted can be upgraded when OpenMLS changes the stored structs.`

How does the CURRENT_VERSION work?

kkohbrok:
> Have some questions about the [CURRENT_VERSION](https://github.com/openmls/openmls/blob/f3e165b69c04c05c9b8e2c452fd8c041da6b1334/traits/src/storage.rs#L8):
> 
> In [MemoryStorage](https://github.com/openmls/openmls/blob/konrad/sqlite_storage_provider/memory_storage/src/lib.rs) CURRENT_VERSION is appended after the key ([here](https://github.com/openmls/openmls/blob/ffff3f7d9897080e61018db1c22911f80dd5810b/memory_storage/src/lib.rs#L983)). Looks like CURRENT_VERSION is not used in that way in this [SqliteStorageProvider](https://github.com/openmls/openmls/tree/konrad/sqlite_storage_provider/sqlite_storage) branch.
> 
> In the [OpenMls Book](https://book.openmls.tech/traits/traits.html) it says: `The traits are generic over a VERSION, which is used to ensure that the values that are persisted can be upgraded when OpenMLS changes the stored structs.`
> 
> How does the CURRENT_VERSION work?

Hi @strong84, since it might be of interest outside of that PR, I [answered your original question in the discussions section](https://github.com/openmls/openmls/discussions/1677#discussioncomment-11395082).

To answer the part of your question relevant to this PR: I had not accounted for the VERSIONING yet. I just added it in the last few commits. Now, the version is stored in a dedicated column in each table and explicitly compared against the current `STORAGE_PROVIDER_VERSION`. It should ensure that code only touches data it's meant to touch (i.e. running an updated provider version against an older (unmigrated) DB doesn't touch the data on that DB. I'm sure there are other ways of doing this, though.

keks (commented):
Thanks for tackling this! I have a few questions :)

I would say that the majority of the schema uses Sqlite as something like a denormalized relational database. It doesn't use or need JOINs and lists are encoded to fit within a single field. The group state however is modeled in a kv-store style. Is there a reason for handling these differently?

Another question I have, and I don't yet know what I think would be the right answer, is when to use INSERT and when to use INSERT OR UPDATE. Currently the latter is only used for the group state fields, and I am not 100% sure that this is correct (I also am not sure it's wrong, just wanted to make sure we understand this). The reason I am concerned here is that not overwriting sensitive data or returning stale data would both be pretty bad. Happy to talk this through on chat or a call, but we probably should document the outcome of that.

Another idea: AFAIK Sqlite supports in-memory databases. Maybe we can run the OpenMLS test suite against it on CI?

strong84:
in `sqlite_storage/src/group_data.rs` the  `UseRatchetTreeExtension` is not used anywhere. And the related fn does not exists in `StorageProvider` trait? How it is used?

strong84:
> Thanks for tackling this! I have a few questions :)
> 
> I would say that the majority of the schema uses Sqlite as something like a denormalized relational database. It doesn't use or need JOINs and lists are encoded to fit within a single field. The group state however is modeled in a kv-store style. Is there a reason for handling these differently?
> 
> Another question I have, and I don't yet know what I think would be the right answer, is when to use INSERT and when to use INSERT OR UPDATE. Currently the latter is only used for the group state fields, and I am not 100% sure that this is correct (I also am not sure it's wrong, just wanted to make sure we understand this). The reason I am concerned here is that not overwriting sensitive data or returning stale data would both be pretty bad. Happy to talk this through on chat or a call, but we probably should document the outcome of that.
> 
> Another idea: AFAIK Sqlite supports in-memory databases. Maybe we can run the OpenMLS test suite against it on CI?

For `GroupData` to be implemented as a KV store. I think it is a right call here. Otherwise there would be more tables, or one table with all `GroupDataType` as column, and you will have multiple UPDATE/DELETE/SELECT sql calls matching each column. The KV store is more convenient to my understanding. I probably will do the same here.

Because the group is evolving, group data is changing, so it should be `INSERT OR REPLACE`.

keks:
> For GroupData to be implemented as a KV store. I think it is a right call here. Otherwise there would be more tables, or one table with all GroupDataType as column, and you will have multiple UPDATE/DELETE/SELECT sql calls matching each column. The KV store is more convenient to my understanding. I probably will do the same here.

I am not sure what you mean. What I would have done is having a table where each field of the GroupState is a column, and each group has a single row. I would still denormalize lists into individual cells, so a single table is enough. That row can be selected, updated and deleted using a single SQL statement. On the other hand, updating the group state in KV style requires several statements (which is fine as well, because we don't batch updates anyway).

strong84:
> > For GroupData to be implemented as a KV store. I think it is a right call here. Otherwise there would be more tables, or one table with all GroupDataType as column, and you will have multiple UPDATE/DELETE/SELECT sql calls matching each column. The KV store is more convenient to my understanding. I probably will do the same here.
> 
> I am not sure what you mean. What I would have done is having a table where each field of the GroupState is a column, and each group has a single row. I would still denormalize lists into individual cells, so a single table is enough. That row can be selected, updated and deleted using a single SQL statement. On the other hand, updating the group state in KV style requires several statements (which is fine as well, because we don't batch updates anyway).

Yeah you are right. There are other benefits, it saves some storage space. There is no need to save `group_id` for each type of GroupData comparing to KV Style. And It is easier to view all group info, the table is more human readable if you open the db file with some viewer application. For this, I totally agree with you.

But still, I think we need multiple statement to select/update/delete each type of GroupData. Because funcs in StorageProvider trait only need one type of GroupData at a time. And the KV style here has some benefit. It just needs one generic select/update/delete fn for all type of GroupData.

kkohbrok:
Thanks for the reviews!

- KV vs. single row per GroupId: @strong84 has already listed the pros and cons. I find the KV style a little more natural given the structure of the trait and the fact that we don't batch-update. Not having to initialize with empty fields is a bonus, too.
- INSERT vs. INSERT OR REPLACE: As @strong84 has already explained, GroupData is the only place where we don't store on a per-epoch basis (and explicitly delete where necessary) or where data generally isn't update, just deleted (e.g. Proposals, KeyPackages, KeyPairs, etc.). I could change everything to `INSERT OR REPLACE` in case I'm wrong somewhere, but I'd rather be deliberate and only use it where necessary.
- Not using UseRatchetTreeExtension: UseRatchetTreeExtension has been removed from the storage provider trait in #1643 as it's is already present in the group config, which is stored elsewhere in the storage provider.
- `store` for `openmls_leaf_nodes`: Good catch! Going with AUTOINCREMENT as you suggest seems more natural to me, so I've gone with that.

keks:
Sorry for only now taking the time to review this again. Did you see my comment on running the OpenMLS tests using an in-memory sqlite db?

keks (approved):
I think this is in a good state now, let's merge it!

PR #1683, opened 2024-11-23
ci: Fix coverage

### LINKED ISSUE #1680: Fix Coverage CI
The coverage CI job is failing with 

```
 error: unknown unstable option: `profile`
```

The coverage hasn't been used well for a while. When fixing this we should make sure that the run is actually useful again.

raphaelrobert:
Fixes #1680.


franziskuskiefer (approved):
Thanks for fixing this.
We should make sure that the output is actually useful. But at least it's running again.

PR #1688, opened 2024-12-03
Fix Typo and add GroupContextExtension example for Unknown Extensions

### LINKED ISSUE #1686: Add example showing how an application can insert and access GroupContextExternsions

keks:
This PR fixes typos in the name of a commit builder function, adds an getter for unknown extensions in `Extensions` and adds a test case that serves as an example for how to access custom group context extensions, both when joining from a welcome or when processing a commit.

Fixes #1686

### DIFF
--- a/openmls/src/extensions/mod.rs
+++ b/openmls/src/extensions/mod.rs
@@ -420,6 +420,19 @@ impl Extensions {
                 _ => None,
             })
     }
+
+    /// Get a reference to the [`UnknownExtension`] with the given type id, if there is any.
+    pub fn unknown(&self, extension_type_id: u16) -> Option<&UnknownExtension> {
+        let extension_type: ExtensionType = extension_type_id.into();
+
+        match extension_type {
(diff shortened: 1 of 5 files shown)


openmls/src/extensions/mod.rs
```diff
@@ -420,6 +420,19 @@ impl Extensions {
                 _ => None,
             })
     }
+
+    /// Get a reference to the [`UnknownExtension`] with the given type id, if there is any.
+    pub fn unknown(&self, extension_type_id: u16) -> Option<&UnknownExtension> {
```
franziskuskiefer, line 425: This is a new public function that should go into the changelog.

franziskuskiefer (changes requested):
lgtm with one nit.

Looks like we also didn't add the commit builder to the changelog. Let's do that in a follow-up.

franziskuskiefer (approved):
Thanks

PR #1690, opened 2024-12-16
Fix issue uncovered by new test

keks:
I wanted to write a reproducer for an issue I think I identified, but unfortunately the test fails even before I get to that place. Now I need to debug that first.

Opening this as a Draft PR for sharing and running CI.


keks:
This is just a bug in the test; I use the wrong provider in a few places.

PR #1691, opened 2024-12-17
Correctly Look Up Credentials and Public Keys for Members for Previous Epochs

keks:
When reading through the code I found a sketchy looking bit (the removed block in framing/validation.rs). The code looks like it tries to fetch the keys and credentials for messages from old epochs, but a closer read suggest it just tries the same thing again in a convoluted way. Also, it doesn't check for the epoch, it just retries if the lookup failed. This means that it is happy with finding any leaf node, regardless of whether the message epoch matches the group epoch.

So I added a test that provokes a problem: When removing a member and adding a new member, the key material in the leaf node is replaced. If we then receive a message from the old epoch, it should load the wrong key material from the current tree, and signature verification should fail. And that's exactly what's happening.

Then I fixed the code that does the looking up. I am not sure the closure really is the best approach, but passing in the many arguments needed to handle all cases seemed bad as well. Maybe the behavior shouldn't even be attached to the `DecryptedMessage` in the first place. Anyway, this works and we can use this as a base for discussion.

### DIFF
--- a/openmls/src/group/public_group/process.rs
+++ b/openmls/src/group/public_group/process.rs
@@ -17,7 +17,6 @@ use crate::{
         past_secrets::MessageSecretsStore, proposal_store::QueuedProposal,
     },
     messages::proposals::Proposal,
-    prelude::SignaturePublicKey,
 };
 
 use super::PublicGroup;
@@ -60,18 +59,12 @@ impl PublicGroup {
             if message_epoch == self.group_context().epoch() {
                 self.treesync()
                     .leaf(leaf_node_index)
-                    .map(|leaf| CredentialWithKey {
-                        credential: leaf.credential().clone(),
-                        signature_key: leaf.signature_key().clone(),
-                    })
+                    .map(CredentialWithKey::from)
             } else if let Some(store) = message_secrets_store_option {
                 store
                     .leaves_for_epoch(message_epoch)
                     .get(leaf_node_index.u32() as usize)
-                    .map(|member| CredentialWithKey {
-                        credential: member.credential.clone(),
-                        signature_key: SignaturePublicKey::from(member.signature_key.clone()),
-                    })
+                    .map(CredentialWithKey::from)
             } else {
                 None
             }
--- a/openmls/src/group/public_group/process.rs
+++ b/openmls/src/group/public_group/process.rs
@@ -17,6 +17,7 @@ use crate::{
         past_secrets::MessageSecretsStore, proposal_store::QueuedProposal,
     },
     messages::proposals::Proposal,
+    prelude::SignaturePublicKey,
 };
 
 use super::PublicGroup;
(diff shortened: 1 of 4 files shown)


franziskuskiefer:
> So I added a test that provokes a problem: When removing a member and adding a new member, the key material in the leaf node is replaced. If we then receive a message from the old epoch, it should load the wrong key material from the current tree, and signature verification should fail. And that's exactly what's happening.

I'm not sure I fully understand the issue. User B is removed and user C is added, into the same index in the tree. Now we get a message from a previous epoch by user B, that was valid in that epoch. My understanding of what you are saying is that this message should be considered invalid. But I don't think it should, because it is a valid message. It was only received a little late.
Or do you say that it is loading the new key material and therefore fails to decrypt and validate the old message?

keks:
> My understanding of what you are saying is that this message should be considered invalid. 

Ah, that is not what I meant!

> But I don't think it should, because it is a valid message. It was only received a little late.

I agree.

> Or do you say that it is loading the new key material and therefore fails to decrypt and validate the old message?

Yes, the existing code looks up the credential and signing key in the current epoch, regardless which epoch the message originated in. This is only for looking up the signing key though, not decryption keys.

The strange thing is that the code _looks like_ it handles the case, but does something very strange: instead of taking key and credential from the Member in the old_members list, it only takes the index of that member and then looks that up in the (current epoch's) tree again.

The idea of the test case I provided is to provoke the scenario and ensure that the message decrypt correctly. It fails with the old code and passes with the new code.

raphaelrobert:
It sounds like the old code was indeed wrong. Not only does it fail in the described scenario but it would also prevent signature key rotations.

openmls/src/group/public_group/process.rs
```diff
@@ -43,14 +44,38 @@ impl PublicGroup {
...
+                        credential: leaf.credential().clone(),
+                        signature_key: leaf.signature_key().clone(),
+                    })
+            } else if let Some(store) = message_secrets_store_option {
+                store
+                    .leaves_for_epoch(message_epoch)
+                    .get(leaf_node_index.u32() as usize)
+                    .map(|member| CredentialWithKey {
+                        credential: member.credential.clone(),
+                        signature_key: SignaturePublicKey::from(member.signature_key.clone()),
```
raphaelrobert, line 73: Maybe this is the right time to introduce a conversation function between a `LeafNode` and `CredentialWithKey`? But it probably broadens the scope too much.

openmls/src/group/public_group/process.rs
```diff
@@ -17,6 +17,7 @@ use crate::{
         past_secrets::MessageSecretsStore, proposal_store::QueuedProposal,
     },
     messages::proposals::Proposal,
+    prelude::SignaturePublicKey,
```
raphaelrobert, line 20: We shall not use `prelude` in the library (I know rust-analyzer does this)

openmls/src/group/tests_and_kats/tests/group.rs
```diff
@@ -535,3 +536,164 @@ fn group_operations() {
...
+        .unwrap();
+
+    match bob_incoming_commit.into_content() {
+        ProcessedMessageContent::StagedCommitMessage(staged_commit) => group_bob
+            .merge_staged_commit(&bob_provider, *staged_commit)
+            .unwrap(),
+        _ => unreachable!(),
+    };
+
+    println!("{}", group_bob.epoch());
```
raphaelrobert, line 692: Snip?

raphaelrobert (changes requested):
Thanks for the PR! Just nits except for the closure.

openmls/src/group/public_group/process.rs
keks: Good point. I moved the conversion to `From` impls and it makes the closure look a lot less daunting.

keks:
The interop test fails because it fails to build the dependencies. Can we still merge this? It seems unrelated. I'd also like to add annotations to this on my validation branch so I can PR that later today.

raphaelrobert (approved):
Thanks for the changes, let's get this in.

PR #1692, opened 2024-12-17
update dependencies

franziskuskiefer:
Updating outdated dependencies.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -6,10 +6,10 @@ edition = "2021"
 
 [dependencies]
 url = "2.2"
-reqwest = { version = "0.11", features = ["blocking", "json"] }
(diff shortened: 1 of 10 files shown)


raphaelrobert:
It looks like building the interop tests failed, but I'm not sure if it is related to this PR or not. We should figure that out before we go ahead.

franziskuskiefer:
Yeah, I'm not sure what's going on there. I need to investigate. I didn't update the tonic dependency there because that's breaking change with the external interop crate. But maybe there's more. The last interop job earlier, not on this branch, succeeded. So I think it's in here.

raphaelrobert (approved):
Since the CI failure now appears everywhere, we should go ahead with this.

PR #1693, opened 2024-12-19
Improve `log2` function

Gusted:
- Instead of having a for-loop to compute the log2 of a number, use bit trick to compute the log2 of a number.
- This gets compiled more efficiently by rust to a few instructions: https://godbolt.org/z/ab1zPE341.
- The only usage of this function is by the `root` function to calculate the root index of the tree, I don't expect any performance improvements from this change.
- The existing treemath KAT tests verify that this still returns the correct result.


keks (changes requested):
Thank you for the contribution! I agree using that function would be nicer. Unfortunately we can't drop the zero-check in the front, because `(31-32)` will panic for unsigned values. Can you re-add that?

Gusted:
Huh weird, seems like log2(0) isn't called by any of the treemath KAT.

keks (approved):
Yeah, it might not be used now, but let's make sure it works for all inputs :)

Thanks for the quick fix!

PR #1694, opened 2024-12-19
Add Validation Comments and Fix Application Message Epoch

keks:
This PR mostly adds validation comments.

It also makes sure that the epoch on a processed message is that of the actual incoming message, not the current group epoch. This can only be different for application messages, where it doesn't matter all that much anyway, but still it was wrong.

### DIFF
--- a/openmls/src/ciphersuite/signable.rs
+++ b/openmls/src/ciphersuite/signable.rs
@@ -153,6 +153,7 @@ pub trait Verifiable: Sized {
                 return Err(SignatureError::VerificationError);
             }
         };
+        // https://validation.openmls.tech/#valn1301
         crypto
             .verify_signature(
                 pk.signature_scheme(),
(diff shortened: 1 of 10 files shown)


openmls/src/ciphersuite/signable.rs
```diff
@@ -153,6 +153,7 @@ pub trait Verifiable: Sized {
                 return Err(SignatureError::VerificationError);
             }
         };
+        // https://validation.openmls.tech/#valn1301
```
franziskuskiefer, line 156: Are these in https://github.com/openmls/validation.dhall/pull/20?

franziskuskiefer (approved):
lgtm

openmls/src/ciphersuite/signable.rs
keks: Exactly

PR #1699, opened 2025-01-09
Fixed `openmls_rust_crypto` build with `test-utils` feature

rozbb:
Previously, `cargo check --features "test-utils"` in `open_rust_crypto` would produce

```text
    Checking openmls_rust_crypto v0.3.0 (/private/tmp/openmls/openmls_rust_crypto)
error[E0277]: the trait bound `MemoryStorage: Clone` is not satisfied
  --> openmls_rust_crypto/src/lib.rs:16:5
   |
13 | #[cfg_attr(feature = "test-utils", derive(Clone))]
   |                                           ----- in this derive macro expansion
...
16 |     key_store: MemoryStorage,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `MemoryStorage`
   |
   = note: this error originates in the derive macro `Clone` (in Nightly builds, run with -Z macro-backtrace for more info)
```
This is because `MemoryStorage` only impls `Clone` when `openmls_memory_storage` crate is built with `test-utils`.

This PR simply propagates the `test-utils` feature down.

It'd also be nice to fix the CI to catch this in the future. But I don't actually know why it hasn't already.  The `build.yml` workflow seems to check the powerset of features, so why didn't this fail?

### DIFF
--- a/openmls_rust_crypto/Cargo.toml
+++ b/openmls_rust_crypto/Cargo.toml
@@ -33,4 +33,4 @@ thiserror = "2.0"
 serde = { version = "^1.0", features = ["derive"] }
 
 [features]
-test-utils = []
+test-utils = ["openmls_memory_storage/test-utils"]


franziskuskiefer (approved):
Thanks!

PR #1700, opened 2025-01-15
Change error in MemoryStorage and result handling in OpenMLS

kkohbrok:
This PR fixes two related problems:

- MemoryStorage returned an error if no epoch key pairs could be found. This is a little inconsistent with the other implementations of reader-type functions. Other storage provider implementations are not likely to implement the same semantics, which means it's probably better placed within OpenMLS. This PR thus changes MemoryStorage's return to be an `Ok(vec![])` if no values can be found.
- OpenMLS didn't actually care about any error returned by the storage provider when looking up epoch key pairs, using `unwrap_or_default` instead, which means fix 1 doesn't really have any impact. That's not great, though, so this PR also adds some error handling behaviour s.t. OpenMLS actually detects if something went wrong at the provider.

### DIFF
--- a/memory_storage/src/lib.rs
+++ b/memory_storage/src/lib.rs
@@ -246,8 +246,6 @@ pub enum MemoryStorageError {
     UnsupportedMethod,
     #[error("Error serializing value.")]
     SerializationError,
-    #[error("Value does not exist.")]
-    None,
 }
 
 const KEY_PACKAGE_LABEL: &[u8] = b"KeyPackage";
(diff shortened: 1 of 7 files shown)


memory_storage/src/lib.rs
```diff
@@ -246,8 +246,6 @@ pub enum MemoryStorageError {
     UnsupportedMethod,
     #[error("Error serializing value.")]
     SerializationError,
-    #[error("Value does not exist.")]
```
franziskuskiefer, line 249: This is a public enum, so a breaking change as well that we should add in the changelog description.

franziskuskiefer (approved):
So this is a breaking change for the API but not for the storage. lgtm with the nit.

We'll need to explain how to adapt to this change in the release notes. Do you want to add a little more to the changelog or somewhere else so that we have something for that already when releasing?

franziskuskiefer (approved):

PR #1702, opened 2025-01-15
Annotate and fix missing validation checks, including ones we hadn't catalogued before

### LINKED ISSUE #800: Public tree validation: validate unmerged leaves
Unmerged leaves should be checked according to (at least) the following criteria:

 - [ ] the should be in the subtree of the node
 - [ ] they should not be blank
 - [ ] #1120 (they should be unique & ordered)

### LINKED ISSUE #803: External Commit validation: unused variants
The variants `NoPath`, `UnknownMemberRemoval`, and `MultipleRemoveProposals` in `ExternalCommitValidationError` seem useful but are not used.

### LINKED ISSUE #807: [Meta] Tree Validation
This ticket tracks other tickets related to the validation of trees that the client receives. 

- [ ] create list of tree validation steps
- [ ] #800 
- [ ] parent hash validation (documentation and testing)
- [ ] node validation (only keypackages in leaves, only parent nodes in non-leaves)

### LINKED ISSUE #1536: Protocol version validation

keks:
This adds validation for external commits and welcomes, and protocol versions.

This is mostly annotation, but the welcome message validation checks were actually missing.

Fixes #800
Fixes #803
Fixes #807
Fixes #1536


openmls/src/binary_tree/array_representation/tree.rs
```diff
@@ -215,7 +215,6 @@ impl<L: Clone + Debug + Default, P: Clone + Debug + Default> ABinaryTree<L, P> {
     }
 }
 
-#[cfg(test)]
```
kkohbrok, line 218: Do you want to merge this into the impl block above this?

openmls/src/group/public_group/mod.rs
```diff
@@ -141,6 +147,49 @@ impl PublicGroup {
             .full_leaves()
             .try_for_each(|leaf_node| leaf_node.validate_locally())?;
 
+        // For each non-empty parent node and each entry in the node's unmerged_leaves field:
+        treesync
```
kkohbrok, line 151: Shouldn't this whole block be part of `into_verified` that's called above? In any case it might be good to make it its own function.

openmls/src/treesync/mod.rs
```diff
@@ -482,6 +488,31 @@ impl TreeSync {
...
+                    .parents()
+                    .flat_map(|(_, node)| node.node().as_ref().map(ParentNode::encryption_key)),
+            )
+            .try_fold(HashSet::new(), |mut leaves, key| {
+                if !leaves.insert(key) {
+                    return Err(RatchetTreeError::DuplicateEncryptionKey);
+                }
+
+                Ok(leaves)
+            })?;
```
kkohbrok, line 514: This probably isn't that much of a performance concern (and we can do it in a separate PR), but between the different validation steps we iterate over the whole tree multiple times. I wonder if we can't streamline that somehow.

kkohbrok (commented):
Thanks, this fills a few important gaps! Just a few comments.

openmls/src/group/public_group/mod.rs
kkohbrok: Fair enough. Let's leave it like this for now.

openmls/src/treesync/mod.rs
keks: I pasted the body of the loop that iterated over both leaves and parents into each of the two loops in that I do in the other file. There might be more to gain somewhere else, but there was nothing that I saw that would be obvious and wouldn't clutter the code.

kkohbrok: Thanks!

kkohbrok (approved):
Thanks!

PR #1707, opened 2025-01-17
fix: verify epoch equality on proposals

sandr01d:
RFC9420 states the following in the [proposals section](https://datatracker.ietf.org/doc/html/rfc9420#section-12.1):

> On receiving a FramedContent containing a Proposal, a client MUST verify the signature inside FramedContentAuthData and that the epoch field of the enclosing FramedContent is equal to the epoch field of the current GroupContext object.

From what I can see, the signature is already verified correctly, but the epoch field is not checked for equality with the current GroupContext, so I added a corresponding check.

### DIFF
--- a/openmls/src/group/mls_group/processing.rs
+++ b/openmls/src/group/mls_group/processing.rs
@@ -276,6 +276,13 @@ impl MlsGroup {
                         ))
                     }
                     FramedContentBody::Proposal(_) => {
+                        // On receiving a FramedContent containing a Proposal, a client MUST verify that the
+                        // epoch field of the enclosing FramedContent is equal to the epoch field of the current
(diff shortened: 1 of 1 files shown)


franziskuskiefer:
I didn't look whether it's correct. But according to the [validation dashboard](https://validation.openmls.tech/#valn1307), this [should be covered](https://github.com/search?type=code&q=repo%3Aopenmls%2Fopenmls%20valn1307). @keks can you have a look at this and double check?

keks:
(and please, if you think I am wrong, please let me know!)

sandr01d:
I see, that was an oversight on my side. Thanks for looking into it @keks!

PR #1708, opened 2025-01-17
fix(wasm): store pending proposals

sandr01d:
The wasm bindings currently do not store the `QueuedProposal`, which results in a `CreateCommitError::MissingProposal` being returned when trying to commit a processed proposal. I've extended the `basic()` test to cover this case.

### DIFF
--- a/openmls-wasm/src/lib.rs
+++ b/openmls-wasm/src/lib.rs
@@ -234,8 +234,10 @@ impl Group {
             openmls::framing::ProcessedMessageContent::ApplicationMessage(app_msg) => {
                 Ok(app_msg.into_bytes())
             }
-            openmls::framing::ProcessedMessageContent::ProposalMessage(_)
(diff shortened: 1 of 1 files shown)


keks:
Thank you for the PR!

In the normal API we usually ask the application to explicitly store the proposal. The idea here is that they might want to do some application-level checks that the proposal satisfies some properties (e.g. the member is allowed to perform the given change, or the member doesn't spam the group with changes which then take up storage).

I think we would _probably_ want to follow that pattern in wasm-land as well, but indeed the method `(MlsGroup::store_pending_proposal)[https://docs.rs/openmls/latest/openmls/group/struct.MlsGroup.html#method.store_pending_proposal]` is not exposed here, so the user can't.

If that method would be added to the `openmls_wasm::Group`, do you think your use-case would be served? Would you be interested in doing that?

sandr01d:
Sure, I've created #1710 to do so

PR #1715, opened 2025-02-02
Fix x25519-dalek crate link in README

shishkin:
It's not `-ng` anymore. Also mention hpke-rs.


kkohbrok (approved):
Thanks for pointing those out!

PR #1717, opened 2025-02-04
Test duplicate signature key detection is not bypassed

### LINKED ISSUE #1243: test: Add test that checks if it is possible to bypass duplicate signature key detection in proposals.
It appears to me that a set of proposals, [`Remove(a)`, `Add(a)`, `Add(a)`], may escape duplicate detection. We could extend `test_valsem101` to make sure.

wysiwys:
This PR addresses issue #1243 by adding tests for scenarios in which we create a commit with the proposals [Remove(a), Add(a), Add(a)].
* The test `mls_duplicate_signature_key_detection_different_key_package`: Ensure that duplicate signature key detection is not bypassed when a different key package is used for each Add proposal. This test ensures that the correct error is returned by `MlsGroup::commit_to_pending_proposals()`.
* The test `mls_duplicate_signature_key_detection_same_key_package`: Ensure that duplicate signature key detection is not bypassed when the same key package is used for each Add proposal. This test currently ensures that only one of the Add proposals is kept when creating the commit, and the other one is filtered out.

Fixes #1243

### DIFF
--- a/openmls/tests/mls_group.rs
+++ b/openmls/tests/mls_group.rs
@@ -22,6 +22,338 @@ fn generate_key_package<Provider: OpenMlsProvider>(
         .clone()
 }
 
+/// This test checks if it is possible to bypass duplicate signature key detection in add proposals,
+/// when the same key package is used for each of the two add proposals.
+/// - Alice creates a group
+/// - Alice adds Bob
+/// - Alice commits
+/// - Alice creates three proposals: Remove Bob, Add Bob, and Add Bob
+/// - Alice commits
+/// - Ensure that Bob is not added twice
+#[openmls_test]
+fn mls_duplicate_signature_key_detection_same_key_package() {
+    for wire_format_policy in WIRE_FORMAT_POLICIES.iter() {
+        let group_id = GroupId::from_slice(b"Test Group");
+
+        let alice_provider = &Provider::default();
+        let bob_provider = &Provider::default();
+
+        // Generate credentials with keys
+        let (alice_credential, alice_signer) =
+            new_credential(alice_provider, b"Alice", ciphersuite.signature_algorithm());
+
+        let (bob_credential, bob_signer) =
+            new_credential(bob_provider, b"Bob", ciphersuite.signature_algorithm());
+
+        // Generate KeyPackage for Bob
+        let bob_key_package = generate_key_package(
+            ciphersuite,
+            Extensions::empty(),
+            bob_provider,
+            bob_credential.clone(),
+            &bob_signer,
+        );
+
+        // Define the MlsGroup configuration
+
+        let mls_group_create_config = MlsGroupCreateConfig::builder()
+            .wire_format_policy(*wire_format_policy)
+            .ciphersuite(ciphersuite)
+            .build();
+
+        // === Alice creates a group ===
+        let mut alice_group = MlsGroup::new_with_group_id(
+            alice_provider,
+            &alice_signer,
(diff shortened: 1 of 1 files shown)


openmls/tests/mls_group.rs
```diff
@@ -22,6 +22,338 @@ fn generate_key_package<Provider: OpenMlsProvider>(
...
+
+        // Ensure that a commit is pending
+        let pending_commit = match alice_group.pending_commit() {
+            Some(pending_commit) => pending_commit,
+            None => panic!("No pending commit was created"),
+        };
+
+        // Ensure that the pending commit contains the correct number of proposals
+        // TODO: Determine whether the duplicate proposals should have been filtered out at this
+        // stage. See issue
```
keks, line 176: This sounds like there should be a reference to an issue here?

openmls/tests/mls_group.rs
```diff
@@ -22,6 +22,338 @@ fn generate_key_package<Provider: OpenMlsProvider>(
...
+            alice_credential.clone(),
+        )
+        .expect("An unexpected error occurred.");
+
+        // === Alice adds Bob ===
+        let welcome =
+            match alice_group.add_members(alice_provider, &alice_signer, &[bob_key_package]) {
+                Ok((_, welcome, _)) => welcome,
+                Err(e) => panic!("Could not add member to group: {e:?}"),
+            };
```
keks, line 79: Not sure if this is helpful feedback, because this goes into code style nitpicking territory, so ignore at will :)

I think something like

```
let (_, welcome, _) = alice_group.add_members(...).expect("could not add members to the group");
```

would read a little bit simpler to me personally.

openmls/tests/mls_group.rs
```diff
@@ -22,6 +22,338 @@ fn generate_key_package<Provider: OpenMlsProvider>(
...
+        // Ensure that there are the correct number of pending proposals
+        assert_eq!(alice_group.pending_proposals().count(), 0);
+
+        // === Alice creates three proposals: Remove Bob, Add Bob, and Add Bob ===
+        let bob_leaf_node_index = bob_group.own_leaf_index();
+        if let Err(e) =
+            alice_group.propose_remove_member(alice_provider, &alice_signer, bob_leaf_node_index)
+        {
+            panic!("Could not add member from group: {e:?}");
+        };
```
keks, line 315: similar style disclaimer as above, just calling `alice_group.propose_remove_member(...).expect(..)` would be simpler. It will print both the string and the error that was found, so you don't have to do that manually.

keks (changes requested):
I think the tests look good! There was a comment that looked like it was missing a link, and I had a little bit of stylistic feedback that I don't think is super important to agree on, but I thought might be helpful.

openmls/tests/mls_group.rs
wysiwys: Definitely agree it's simpler - I fixed that in a commit just now.

openmls/tests/mls_group.rs
wysiwys: For this one, I am considering whether it should be kept as it is for now, because it matches the syntax that is used in some of the other tests in the file (like in lines 1365-1372). Would this be a concern that should be taken into account?

keks: Like I said, ignore at will :) I also wasn't aware that this was an existing code style, and consistency is good :+1:

keks (approved):
Looks good, thanks!

PR #1721, opened 2025-02-27
Add book example for discarded welcome message

### LINKED ISSUE #1733: Add book tests and examples for discarded Welcome
- Add tests that ensure that the application is able to clear any changes to the local state that were made when processing or staging the Welcome, and
- In the OpenMLS book, provide instructions from the application's perspective that show which local state needs to be cleaned up, and how to accomplish this.

wysiwys:
This pull request adds an example to the OpenMLS book that provides guidance for the following scenario, for #1723 :

* A client is invited to a group with a Welcome message. After investigating the message, the application decides not to join the group. The Welcome message as well as any already created state must be removed. 

The book example is also accompanied by a test, which sets up the scenario, demonstrates what needs to be done (which is pulled into to the book example in the form of code snippets), and tests that all state is cleaned up correctly.

This pull request also provides a view of the current way to clean up the state, so that any new methods that are needed can be identified.

Also, in order to check that the state in the `StorageProvider` was cleaned up correctly, this pull request adds a way to compare the state associated only with a given `GroupId` to the `test_utils` module. This is done in order to be able to compare the `StorageProvider` state for any type of storage provider, which is useful for this and other testing purposes.

Resolves #1733


book/src/SUMMARY.md
```diff
@@ -21,6 +21,7 @@
   - [Persistence of group state](user_manual/persistence.md)
   - [Credential validation](user_manual/credential_validation.md)
   - [WebAssembly](user_manual/wasm.md)
+  - [Examining a welcome message](user_manual/examining_welcome_message.md)
```
franziskuskiefer, line 24: There is the `Join a group from a Welcome message` chapter. I think this could fit in there. Because it's the negative case of processing the welcome.

book/src/user_manual/examining_welcome_message.md
```diff
@@ -0,0 +1,49 @@
+# Examining a welcome message
+
+When a client is invited to join a group, the application can allow the client to decide whether or not to join the group. In order to determine whether to join the group, the application can inspect information provided in the welcome message, such as who invited them, who else is in the group, what extensions are available, and more. If the application decides not to join the group, data must be cleaned up in the storage provider.
```
franziskuskiefer, line 3: "data must be cleaned up in the storage provider." I think the application shouldn't have to worry about that level of detail. Instead, we should say something like "the welcome must be discarded to ensure that the local state is cleaned up".

openmls/tests/book_code.rs
```diff
@@ -1663,3 +1663,176 @@ fn commit_builder() {
     // ANCHOR_END: alice_adds_bob_with_commit_builder
     _ = (mls_message_out, welcome, group_info)
 }
+
+#[openmls_test]
```
franziskuskiefer, line 1667: This file gets really big. Let's move this to a separate file.

openmls/tests/book_code.rs
```diff
@@ -1663,3 +1663,176 @@ fn commit_builder() {
...
+    let state_before = GroupStorageState::from_storage(bob_provider.storage(), &group_id);
+    //ANCHOR: not_join_group_welcome
+    let welcome = match welcome.extract() {
+        MlsMessageBodyIn::Welcome(welcome) => welcome,
+        _ => unimplemented!("Handle other message types"),
+    };
+    //ANCHOR_END: not_join_group_welcome
+
+    //ANCHOR: not_join_group_welcome_inspect
+    for _secret in welcome.secrets() {
```
franziskuskiefer, line 1739: I can't think of a good reason why the secrets should be investigated. I actually wonder why this is exposed.

openmls/tests/book_code.rs
```diff
@@ -1663,3 +1663,176 @@ fn commit_builder() {
...
+
+    //ANCHOR: not_join_group_processed_welcome
+    let join_config = MlsGroupJoinConfig::default();
+    let processed_welcome = ProcessedWelcome::new_from_welcome(bob_provider, &join_config, welcome)
+        .expect("Error constructing processed welcome");
+    //ANCHOR_END: not_join_group_processed_welcome
+
+    //ANCHOR: not_join_group_processed_welcome_inspect
+    let _unverified_psks = processed_welcome.psks();
+    let _unverified_group_info = processed_welcome.unverified_group_info();
```
franziskuskiefer, line 1752: The group info is an interesting part that the application will probably want to inspect.
It provides the ciphersuite, group_id, and epoch. This may be used to get the ratchet tree when it's not in the Welcome message.

book/src/user_manual/examining_welcome_message.md
```diff
@@ -0,0 +1,49 @@
...
+```rust,no_run,noplayground
+{{#include ../../../openmls/tests/book_code.rs:not_join_group_processed_welcome}}
+```
+
+After this, some (unverified) information contained in the `ProcessedWelcome` can be investigated.
+```rust,no_run,noplayground
+{{#include ../../../openmls/tests/book_code.rs:not_join_group_processed_welcome_inspect}}
+```
+
+The next step is to stage the `ProcessedWelcome`. This causes some state to be updated in the `StorageProvider`.
```
franziskuskiefer, line 25: The relevant information here for the application developer is that up to this point, nothing is persisted. But when the message is processed, it must be discarded afterwards.

book/src/user_manual/examining_welcome_message.md
```diff
@@ -0,0 +1,49 @@
...
+```rust,no_run,noplayground
+{{#include ../../../openmls/tests/book_code.rs:not_join_group_processed_welcome_inspect}}
+```
+
+The next step is to stage the `ProcessedWelcome`. This causes some state to be updated in the `StorageProvider`. 
+```rust,no_run,noplayground
+{{#include ../../../openmls/tests/book_code.rs:not_join_group_staged_welcome}}
+```
+
+Then, more information about the welcome messages sender, such as the credential, signature public key, and encryption public key can also be individually inspected.
```
franziskuskiefer, line 30: For the application developer this part is interesting because the application may want to check or display the inviter.

openmls/tests/book_code.rs
```diff
@@ -1663,3 +1663,176 @@ fn commit_builder() {
...
+
+    //ANCHOR: not_join_group_staged_welcome
+    let staged_welcome: StagedWelcome = processed_welcome
+        .into_staged_welcome(bob_provider, None)
+        .expect("Error constructing staged welcome");
+    //ANCHOR_END: not_join_group_staged_welcome
+
+    // check storage state after staging welcome
+    let own_leaf_nodes: Vec<LeafNode> = bob_provider.storage().own_leaf_nodes(&group_id).unwrap();
+    // TODO: is this correct?
```
franziskuskiefer, line 1763: Yes, at this point there really shouldn't be anything/much persisted.

openmls/tests/book_code.rs
```diff
@@ -1663,3 +1663,176 @@ fn commit_builder() {
...
+    let welcome_sender: &LeafNode = staged_welcome
+        .welcome_sender()
+        .expect("Welcome sender could not be retrieved");
+
+    // Inspect sender's credential...
+    let _credential = welcome_sender.credential();
+    // Inspect sender's signature public key...
+    let _signature_key = welcome_sender.signature_key();
+    // Inspect sender's encryption public key...
+    let _encryption_key = welcome_sender.encryption_key();
```
franziskuskiefer, line 1782: The encryption key is not really interesting for the application. At least I can't think of a case where it should be.

book/src/user_manual/examining_welcome_message.md
```diff
@@ -0,0 +1,49 @@
...
+```rust,no_run,noplayground
+{{#include ../../../openmls/tests/book_code.rs:not_join_group_welcome_inspect}}
+```
+
+The next step is to process the `Welcome`.
+```rust,no_run,noplayground
+{{#include ../../../openmls/tests/book_code.rs:not_join_group_processed_welcome}}
+```
+
+After this, some (unverified) information contained in the `ProcessedWelcome` can be investigated.
```
franziskuskiefer, line 20: Let's write the entire part more from the perspective of application that is using this, i.e. say not only what information is available, but why that may be interesting to the application, or what it can use it for.

openmls/tests/book_code.rs
```diff
@@ -1663,3 +1663,176 @@ fn commit_builder() {
...
+        // encryption public key
+        let _encryption_key = member.encryption_key;
+        // signature public key
+        let _signature_key = member.signature_key;
+    }
+    //ANCHOR_END: not_join_group_members
+
+    //ANCHOR: not_join_group_cleanup
+    // clean up storage provider
+    PublicGroup::delete(bob_provider.storage(), &group_id).unwrap();
```
franziskuskiefer, line 1832: I'm sure most consumers won't actually do this. And it's not really clear why the user should know this detail. As discussed offline, how about we add a `discard` function to the staged welcome that can be called and does this?
A step further would be: should this even be necessary? Let's look at that together.

openmls/tests/book_code.rs
```diff
@@ -1663,3 +1663,176 @@ fn commit_builder() {
...
+
+    //ANCHOR: not_join_group_welcome_inspect
+    for _secret in welcome.secrets() {
+        // TODO: check secret
+    }
+    //ANCHOR_END: not_join_group_welcome_inspect
+
+    //ANCHOR: not_join_group_processed_welcome
+    let join_config = MlsGroupJoinConfig::default();
+    let processed_welcome = ProcessedWelcome::new_from_welcome(bob_provider, &join_config, welcome)
```
franziskuskiefer, line 1746: Here it's worthwhile mentioning that this deletes the keys used to decrypt the welcome, except if it is a last resort key package.

openmls/tests/book_code.rs
franziskuskiefer: I tried dropping the store of the public group, and it seems fine. We should double check that it's really ok. But I think the solution here is not to store the public group in `PublicGroup::from_external`. Then we don't have to delete it here.

book/src/user_manual/examining_welcome_message.md
wysiwys: I added some more info here about what particular pieces of information are useful to or needed by the application at this stage.

openmls/tests/book_code.rs
franziskuskiefer: @wysiwys do you want to do this change here? Then we can get this ready for review.

wysiwys: Sure, I can go ahead and make this change :+1: Is this something that should be a separate pull request or should it be part of this one?

wysiwys: Made a PR here #1726

franziskuskiefer:
@keks can you review?

book/src/user_manual/join_from_welcome.md
```diff
@@ -17,3 +17,41 @@ to inspect the message, e.g. to determine the identity of the sender.
...
+```rust,no_run,noplayground
+{{#include ../../../openmls/tests/book_code_discard_welcome.rs:not_join_group_processed_welcome}}
+```
+
+At this stage, there are some more pieces of information in the `ProcessedWelcome` that could be useful to the application. For example, it can be useful to check which extensions are available. Much of this information can also be checked using the `StagedWelcome` that is produced in the next step.
+```rust,no_run,noplayground
+{{#include ../../../openmls/tests/book_code_discard_welcome.rs:not_join_group_processed_welcome_inspect}}
+```
+
+The next step is to stage the `ProcessedWelcome`. This causes some state to be updated in the `StorageProvider`.
```
keks, line 40: I think this is incorrect. Staging the welcome does not touch the storage, it looks to me like it only uses the crypto part of the provider. That said, constructing a ProcessedWelcome will remove the consumed KeyPackage from the provider, so it's more the other way around.

keks: Actually I think this is something that we also should document on the function itself.

book/src/user_manual/join_from_welcome.md
```diff
@@ -17,3 +17,41 @@ to inspect the message, e.g. to determine the identity of the sender.
...
+```rust,no_run,noplayground
+{{#include ../../../openmls/tests/book_code_discard_welcome.rs:not_join_group_welcome}}
+```
+
+The next step is to process the `Welcome`.
+```rust,no_run,noplayground
+{{#include ../../../openmls/tests/book_code_discard_welcome.rs:not_join_group_processed_welcome}}
+```
+
+At this stage, there are some more pieces of information in the `ProcessedWelcome` that could be useful to the application. For example, it can be useful to check which extensions are available. Much of this information can also be checked using the `StagedWelcome` that is produced in the next step.
```
keks, line 35: I think this maybe should mention that a lot of the values here are not validated yet, and are only validated when turning it into a StagedWelcome.

keks (dismissed):
The examples and general stucture look good!

There are a few mistakes in the text, which is probably due to the distinction between ProcessedWelcome and StagedWelcome being not super clear and under-documented. Let's address these  and also improve the documentation

book/src/user_manual/join_from_welcome.md
wysiwys: As of #1726

wysiwys:
Thanks for the review  @keks ! I addressed the feedback you provided, including ensuring all text is consistent with the recent changes made in #1726. That PR updated the staging process so that the constructed `PublicGroup` is no longer stored in the `StorageProvider`.

wysiwys:
This test is now updated to use the new single-group test suite. It eliminates some boilerplate from the setup, although most of the test is unaffected for now (because it deals with more detailed inspection of the Welcome data)

franziskuskiefer (approved):
lgtm

Let's squash commits before merging to make this a little cleaner.

PR #1725, opened 2025-03-03
Add fs exporter

kkohbrok:
This PR adds the application export secret as defined in the extensions draft: https://messaginglayersecurity.rocks/mls-extensions/draft-ietf-mls-extensions.html#name-exported-secrets

More concretely, this PR adds a PPRF implementation specifically for the exporter tree. Since the exporter is stateful and the application_exporter (the root secret of the PPRF) is stored in the `StagedCommit`, the serialization format changes slightly in that the PPRF is added as an `Option` that defaults to `None`. Groups serialized prior to this PR will derive the exporter when processing/creating the next commit.

The PPRF used for the exporter tree is different than the one in SecretTree, because the exporter tree has 2^16 leaves by default and thus needs to be optimized for size. The new PPRF is generic w.r.t. the number of leaves and can in the future be used for the SecretTree, or other applications such as fork-resilient MLS (DMLS).

### DIFF
--- a/openmls/src/schedule/pprf/prefix.rs
+++ b/openmls/src/schedule/pprf/prefix.rs
@@ -1,6 +1,6 @@
 //! This module defines a trait for prefixes to instantiate a PPRF depending on
-//! the PPRF's index (i.e. input) size. It also provides implementations for
-//! different tree sizes.
+//! the PPRF's index (i.e. input) size. It also provides an implementation for
+//! trees with u16 as the leaf index type.
 //!
 //! Each prefix encodes a node in the binary tree of the PPRF. The root node has
 //! an empty prefix, which then grows in size with each step down the tree.


franziskuskiefer (commented):
The changes generally lgtm.

My main points are the following two.

As mentioned offline, I think we should introduce versioned features for draft features to i) allow us changing how it works as the draft changes, and ii) don't expose unfinished features. Let's use this to figure out how that process should work.

This is a breaking change because it changes the storage (looking at the failing tests). This means we should do a release before merging this, to get all the other changes released for this breaking change. And then it will need some docs around how to upgrade, esp. since this is the first breaking change for the storage. Hopefully this can again be a blueprint for subsequent changes like this.

kkohbrok:
Thanks for the review. I agree w.r.t. feature flags. I'll take a stab at that later in the week. I was hoping we could get away without modifying storage, although this PR currently is doing just that. I'll look into it.

openmls/src/group/mls_group/tests_and_kats/kats/welcome.rs
```diff
@@ -252,7 +255,11 @@ pub fn run_test_vector(test_vector: WelcomeTestVector) -> Result<(), &'static st
         .unwrap();
 
     let (_group_epoch_secrets, message_secrets) = {
-        let epoch_secrets = key_schedule
+        let EpochSecretsResult {
+            epoch_secrets,
+            #[cfg(feature = "extensions-draft")]
```
franziskuskiefer, line 260: Maybe `, ..` as well?

openmls/src/schedule/mod.rs
```diff
@@ -745,6 +757,57 @@ impl ExporterSecret {
...
+        Ok(ApplicationExportSecret { secret })
+    }
+
+    /// Derive a `Secret` from the exporter secret. We return `Vec<u8>` here, so
+    /// it can be used outside of OpenMLS.
+    pub fn derive_exported_secret(
+        &self,
+        ciphersuite: Ciphersuite,
+        crypto: &impl OpenMlsCrypto,
+        label: &str,
```
franziskuskiefer, line 791: Maybe add something to the doc comment that the `label` should be unique in the application context.

openmls/src/group/mls_group/tests_and_kats/tests/mls_group.rs
```diff
@@ -266,6 +266,100 @@ fn export_secret() {
...
+    let bob_derived_secret = bob_application_secret
+        .derive_exported_secret(
+            ciphersuite,
+            bob_group_state.party.core_state.provider.crypto(),
+            "test",
+            &[],
+            32,
+        )
+        .unwrap();
+    assert_eq!(alice_derived_secret, bob_derived_secret);
```
franziskuskiefer, line 360: Maybe also add a test where they are different? Because of the label for example.

openmls/src/schedule/mod.rs
```diff
@@ -745,6 +757,57 @@ impl ExporterSecret {
     }
 }
 
+/// A secret that we can derive secrets from, that are used outside of OpenMLS.
+/// In contrast to `[ExporterSecret]`, the `[ApplicationExportSecret]` is not
+/// persisted. It can be deleted after use to achieve forward secrecy.
+#[cfg(feature = "extensions-draft")]
+#[derive(Debug, Serialize, Deserialize, Clone)]
```
franziskuskiefer, line 764: Don't derive `Debug`. We shouldn't allow printing secrets. Or only with the crypto debug feature.

openmls/src/schedule/tests_and_kats/kats/key_schedule.rs
```diff
@@ -135,7 +135,11 @@ fn generate(
     key_schedule
         .add_context(crypto.crypto(), &serialized_group_context)
         .expect("An unexpected error occurred.");
-    let epoch_secrets = key_schedule
+    let EpochSecretsResult {
+        epoch_secrets,
+        #[cfg(feature = "extensions-draft")]
```
franziskuskiefer, line 140: `, ..` here and below as well?

openmls/src/schedule/mod.rs
```diff
@@ -745,6 +757,57 @@ impl ExporterSecret {
...
+#[cfg_attr(any(test, feature = "test-utils"), derive(PartialEq))]
+pub struct ApplicationExportSecret {
+    secret: Secret,
+}
+
+#[cfg(feature = "extensions-draft")]
+impl ApplicationExportSecret {
```
franziskuskiefer, line 771: The draft appears to have changed. Shall we implement what's in the draft now? This also goes back to numbering the feature with the draft number.
Adding a link here with the comments to where this is defined would also be good. Now that not everything is from rfc 9420 anymore.

franziskuskiefer (changes requested):
Generally lgtm with a couple nits and questions.
Only one bigger one.

openmls/src/schedule/mod.rs
kkohbrok: Since all it contains is `Secret`, which already redacts its content, we should be fine here.

openmls/src/framing/validation.rs
```diff
@@ -332,6 +332,24 @@ impl ProcessedMessage {
...
         &self.credential
     }
+
+    /// Safely export a value if the content of the processed message is a
+    /// [`StagedCommit`].
+    #[cfg(feature = "extensions-draft-08")]
+    pub fn safe_export_secret<Crypto: OpenMlsCrypto>(
+        &mut self,
+        crypto: &Crypto,
+        component_id: u16,
```
franziskuskiefer, line 342: @wysiwys is working on [#1821](https://github.com/openmls/openmls/issues/1821) in parallel. We should probably introduce a newtype for the `ComponentId` to use everywhere in either of the PRs.

openmls/src/group/mls_group/staged_commit/v1_storage.rs
```diff
@@ -0,0 +1,143 @@
+//! This module contains legacy structs and implementations for storage provider
```
franziskuskiefer: Is this file needed if we don't change storage versions?

openmls/src/group/mls_group/commit_builder/external_commits.rs
```diff
@@ -287,6 +287,10 @@ impl ExternalCommitBuilder {
             own_leaf_index,
             message_secrets_store,
             resumption_psk_store: ResumptionPskStore::new(32),
+            // This is set to `None` for now. It will be set once the external
+            // commit is merged.
+            #[cfg(feature = "extensions-draft-08")]
+            application_export_tree: None,
```
franziskuskiefer, line 294: Isn't this the PR we merged already?

openmls/src/group/mls_group/exporting.rs
```diff
@@ -41,6 +43,57 @@ impl MlsGroup {
...
+        crypto: &Crypto,
+        storage: &Storage,
+        component_id: u16,
+    ) -> Result<Vec<u8>, SafeExportSecretError<Storage::Error>> {
+        if !self.is_active() {
+            return Err(SafeExportSecretError::GroupState(
+                MlsGroupStateError::UseAfterEviction,
+            ));
+        }
+        let group_id = self.group_id().clone();
```
franziskuskiefer: You could scope the mutable borrow below to avoid cloning here. That probably requires a second `else` case. So I'm not sure if it's worth it.

openmls/src/group/mls_group/exporting.rs
```diff
@@ -41,6 +43,57 @@ impl MlsGroup {
...
+            return Err(PendingSafeExportSecretError::NoPendingCommit);
+        };
+        let PendingCommitState::Member(ref mut staged_commit) = **group_state else {
+            return Err(PendingSafeExportSecretError::NotGroupMember);
+        };
+        let secret = staged_commit.safe_export_secret(crypto, component_id)?;
+        storage
+            .write_group_state(&group_id, &self.group_state)
+            .map_err(PendingSafeExportSecretError::Storage)?;
+        Ok(secret.as_slice().to_vec())
```
franziskuskiefer, line 94: Why not just returning `secret`? 🤔

```suggestion
        Ok(secret)
```

openmls/src/group/mls_group/mod.rs
```diff
@@ -247,6 +250,11 @@ pub struct MlsGroup {
     // A variable that indicates the state of the group. See [`MlsGroupState`]
     // for more information.
     group_state: MlsGroupState,
+    /// The state of the Application Exporter. See the MLS Extensions Draft 07
```
franziskuskiefer:
```suggestion
    /// The state of the Application Exporter. See the MLS Extensions Draft 08
```

openmls/src/group/mls_group/staged_commit.rs
```diff
@@ -14,17 +16,22 @@ use super::{
...
     group::public_group::{
         diff::{apply_proposals::ApplyProposalsValues, StagedPublicGroupDiff},
         staged_commit::PublicStagedCommitState,
     },
-    schedule::{CommitSecret, EpochAuthenticator, EpochSecrets, InitSecret, PreSharedKeyId},
+    schedule::{CommitSecret, EpochAuthenticator, EpochSecretsResult, InitSecret, PreSharedKeyId},
     treesync::node::encryption_keys::EncryptionKeyPair,
 };
 
+pub mod v1_storage;
```
franziskuskiefer: drop?

openmls/src/schedule/pprf/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+    #[error("Evaluating on punctured input")]
+    PuncturedInput,
+    #[error("Error deriving child node: {0}")]
+    ChildDerivationError(#[from] CryptoError),
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, ZeroizeOnDrop)]
+#[cfg_attr(any(test, feature = "test-utils"), derive(PartialEq))]
+#[serde(transparent)]
+struct PprfNode(#[serde(with = "serde_bytes")] Vec<u8>);
```
franziskuskiefer, line 45: A named struct would probably be nicer here.

franziskuskiefer: Also, should this maybe be `Secret`? You convert between the two all the time and it looks to me like keeping it in a `Secret` would avoid most of the conversions.

openmls/src/schedule/pprf/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+            crypto,
+            ciphersuite,
+            "tree",
+            context,
+            ciphersuite.hash_length(),
+        )?;
+        Ok(secret.into())
+    }
+
+    fn derive_children(
```
franziskuskiefer, line 61: Could we re-use the code from the secret tree here?

openmls/src/schedule/pprf/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+        // Step 2: Derive and walk the rest of the path
+        for d in depth..P::MAX_DEPTH {
+            let (left, right) = current_node.derive_children(crypto, ciphersuite).unwrap();
+            let bit = get_bit(&leaf_index, d);
+
+            let (next_node, copath_node) = if bit { (right, left) } else { (left, right) };
+
+            let mut copath_prefix = prefix.clone();
+            copath_prefix.push_bit(!bit);
+            self.nodes.insert(copath_prefix.clone(), copath_node);
```
franziskuskiefer: This should always return `None`, right? A debug assert for that would be nice.

openmls/src/schedule/pprf/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+        // We interpret the input as a bit string indexing the leaf in our tree.
+        let leaf_index = input;
+
+        let mut prefix = P::new();
+        let mut current_node;
+        let mut depth = 0;
+
+        // Step 1: Find the deepest existing node in the cache
+        loop {
+            let key = prefix.clone();
```
franziskuskiefer: Do we really need to clone this here?

openmls/src/schedule/pprf/prefix.rs
```diff
@@ -0,0 +1,148 @@
...
+    const MAX_DEPTH: usize = 256;
+
+    fn new() -> Self {
+        Self {
+            bits: vec![],
+            len: 0,
+        }
+    }
+
+    fn push_bit(&mut self, bit: bool) {
```
franziskuskiefer: Some docs in this file would make it easier to read and allow figuring out if it is doing what you want it to do.

openmls/src/schedule/application_export_tree.rs
```diff
@@ -0,0 +1,29 @@
...
+        pprf::{Pprf, PprfError, Prefix16},
+        ApplicationExportSecret,
+    },
+};
+
+pub(crate) type ApplicationExportTree = Pprf<Prefix16>;
+pub(crate) type ApplicationExportTreeError = PprfError;
+
+impl ApplicationExportTree {
+    pub fn new(application_exporter: ApplicationExportSecret) -> Self {
```
franziskuskiefer: The type isn't `pub`. So making these `pub` doesn't make much sense. Is this supposed to be used on the outside?

openmls/src/schedule/pprf/mod.rs
```diff
@@ -0,0 +1,316 @@
+use std::collections::HashMap;
```
franziskuskiefer, line 8: I think this module needs a bit of design documentation. Since this is not part of the draft, could you describe a little what you're implementing here?

openmls/src/schedule/pprf/prefix.rs
```diff
@@ -0,0 +1,148 @@
+//! This module defines a trait for prefixes to instantiate a PPRF depending on
```
franziskuskiefer, line 1: The draft only defines `LeafIndex(ComponentID)` to index the tree. Can you describe a little what the other types are for?

openmls/src/framing/validation.rs
kkohbrok: Agreed. Since I only take it as input for the export functions in a handful of places, I'll leave the definition of the type to @wysiwys if that's okay. That type should probably live closer to the rest of the component-related machinery.

openmls/src/group/mls_group/commit_builder/external_commits.rs
kkohbrok: I'm not sure what you mean. The comment references the actual external commit that will be merged into the group right after creation.

openmls/src/group/mls_group/exporting.rs
kkohbrok: I could do that, but the current `Secret` struct is locked down such that outside of this crate you can't get hold of the internal value. So if we want this to be `Secret`, we'd have to create a way to access the actual bytes from the outside. Note that the regular `export_secret` also just returns a vec.

We could improve the situation a little by creating an `ExportedSecret` struct that gives access to its internal bytes, but also implements zeroize. What do you think?

openmls/src/group/mls_group/staged_commit/v1_storage.rs
kkohbrok: Good point, forgot to clean this up.

openmls/src/schedule/pprf/mod.rs
kkohbrok: Using `Vec<u8>` here allows me to use `serde_bytes`, which makes the serialized footprint considerably smaller given the size of the tree. We can't easily optimize `Secret` in the same way unfortunately.

openmls/src/schedule/pprf/mod.rs
kkohbrok: Yeah, good point. Not sure what I was thinking here.

openmls/src/schedule/pprf/mod.rs
kkohbrok: Good idea!

openmls/src/schedule/pprf/prefix.rs
kkohbrok: Ah, right. I originally made this implementation for DMLS, where the inputs are 32 bytes long. I also initially thought I'd just replace the SecretTree with this implementation, so I also included a Prefix for LeafIndex, which is 32 bytes. However, replacing SecretTree would require a storage provider migration, so I leave that for future work. I can remove the additional Prefix implementations if you prefer. Currently they're dead code.

openmls/src/group/mls_group/commit_builder/external_commits.rs
franziskuskiefer: 🤣 I'm not sure. But what you say makes sense.

openmls/src/group/mls_group/exporting.rs
franziskuskiefer: Ah I see. Yeah makes sense. An `ExportedSecret` sounds reasonable to have.
But we shouldn't have to copy this even without a separate type? If we do I think we are missing an API somewhere.

openmls/src/schedule/pprf/mod.rs
franziskuskiefer: That's unfortunate. But I guess nothing we can do right now.

openmls/src/schedule/pprf/prefix.rs
franziskuskiefer: Yeah let's maybe drop the unused code to keep this clean. We have it in the history if we need it again.

openmls/src/group/mls_group/exporting.rs
kkohbrok: I believe the API that's missing is missing in `tls_codec`, which doesn't allow turning a `SecretVLBytes` into `Vec<u8>`. Instead, you have to go via `as_slice` and copy it. I could do a PR in tls_codec, but then we'd have to bump the version in OpenMLS, etc. Maybe relegate this to a follow up?

PR #1726, opened 2025-03-04
Don't store `PublicGroup` in storage in `PublicGroup::from_external()`

wysiwys:
This pull request removes the call to `PublicGroup::store()` in `PublicGroup::from_external()`. This change avoids  the `PublicGroup` state being stored in the storage provider when a welcome is staged, which would then later require additional cleanup if the application were to reject the group invitation.

These changes to `PublicGroup::from_external()` also have an effect on both `ProcessedWelcome::into_staged_welcome()` and `StagedWelcome::new_from_welcome()`, as well as on `MlsGroup::join_by_external_commit()`, which all call `PublicGroup::from_external()`.

Additionally, because the `StorageProvider` no longer is required as an argument to `PublicGroup::from_external()`, the `CreationFromExternalError` that it returns is also updated, to reflect that `PublicGroup::from_external()` no longer returns errors encapsulating `StorageError`s.

### DIFF
--- a/openmls/src/binary_tree/tests.rs
+++ b/openmls/src/binary_tree/tests.rs
@@ -153,6 +153,8 @@ fn test_diff_merging() {
     assert_eq!(tree, original_tree);
 }
 
+/*
+// Disabled due to #1727
 #[test]
 fn test_new_tree_error() {
     // Let's test what happens when the tree is getting too large.
@@ -169,6 +171,7 @@ fn test_new_tree_error() {
         )
     }
 }
+*/
 
 #[test]
 fn test_diff_iter() {


kkohbrok (commented):
Thanks for the PR! It sounds reasonable to avoid storing the PublicGroup in the context of the various `MlsGroup` staging processes. However, when using `PublicGroup::from_external()` from the outside to create an actual `PublicGroup` it's a good idea to store it. 

How about we have two functions: `PublicGroup::from_external()` calls into `PublicGroup::from_external_internal()` (which doesn't store the PublicGroup) and then subsequently stores the group. `MlsGroup` functions could then just call `PublicGroup::from_external_internal()`. What do you think?

wysiwys:
Thanks! I have gone ahead and implemented those changes, so that both of those use cases are accounted for. Also, rewriting the functions in this way allows the public API to be kept the same as before, including the returned `CreationFromExternalError<StorageError>` type.

kkohbrok (approved):
Sounds good, thanks for making the changes!

Issue #1727, opened 2025-03-04
Segfault in test on 32-bit platforms with latest Rust toolchain


wysiwys:
With the latest Rust toolchain release, the test `binary_tree::tests::test_new_tree_error()` now fails on 32-bit Ubuntu and Windows platforms. In particular, this occurs when the Tests CI workflow (`tests.yml`) is run.

Description:
- The test `binary_tree::tests::test_new_tree_error()` sets the length of a `Vec` unsafely using `Vec::set_len()`, in order to test what happens when too large of an `MlsBinaryTree` is created.
- After passing in this `Vec` as an input to `MlsBinaryTree::new()`, a segfault occurs at `openmls/src/binary_tree/array_representation/tree.rs:62`.

Although the exact cause of this issue is unknown, this test did pass with previous iterations of the Rust toolchain, so further investigation would be needed to determine the cause of this issue and how to potentially redesign this test to work around it.

PR #1728, opened 2025-03-04
Disable `test_new_tree_error()` due to segfault

wysiwys:
This PR disables the test `test_new_tree_error()` due to a segfault on 32-bit platforms (see #1727)


franziskuskiefer (approved):
Thanks

PR #1730, opened 2025-03-12
Single group test framework

wysiwys:
This pull request adds a lightweight test suite that aims to avoid boilerplate in many of the tests that involve a single group with multiple members. 

It provides functionality for:
- Setting up testing scenarios (creating groups with multiple, arbitrary members)
- Delivering and applying a message for all members of a group where a provided condition is met
- Individually accessing the underlying `MlsGroup`s and performing operations on them in a more granular way
- Asserting correctness of group membership

### DIFF
--- a/openmls/src/test_utils/single_group_test_framework/errors.rs
+++ b/openmls/src/test_utils/single_group_test_framework/errors.rs
@@ -2,14 +2,14 @@ use crate::prelude::{commit_builder::*, *};
 pub use crate::utils::*;
 
 pub use openmls_traits::{
-    storage::StorageProvider as StorageProviderTrait,
+    storage::{StorageProvider as StorageProviderTrait, CURRENT_VERSION},
     types::{Ciphersuite, HpkeKeyPair},
     OpenMlsProvider,
 };
 
 // type alias for `TestError`
 pub type GroupError<Provider> =
-    TestError<<<Provider as OpenMlsProvider>::StorageProvider as StorageProviderTrait<1>>::Error>;
+    TestError<<<Provider as OpenMlsProvider>::StorageProvider as StorageProviderTrait<CURRENT_VERSION>>::Error>;
 #[derive(Debug)]
 pub enum TestError<StorageError> {
     CreateCommit(CreateCommitError),
--- a/openmls/src/test_utils/single_group_test_framework/errors.rs
+++ b/openmls/src/test_utils/single_group_test_framework/errors.rs
@@ -19,6 +19,8 @@ pub enum TestError<StorageError> {
     Welcome(WelcomeError<StorageError>),
     ProtocolMessage(ProtocolMessageError),
     MergeCommit(MergeCommitError<StorageError>),
+    CommitToPendingProposals(CommitToPendingProposalsError<StorageError>),
+    NoSuchMember,
 }
 
 impl<StorageError> From<CreateCommitError> for TestError<StorageError> {
@@ -29,6 +31,12 @@ impl<StorageError> From<CreateCommitError> for TestError<StorageError> {
 impl<StorageError> From<CommitBuilderStageError<StorageError>> for TestError<StorageError> {
     fn from(e: CommitBuilderStageError<StorageError>) -> Self {
         TestError::CommitBuilderStage(e)
+    }
+}
+
+impl<StorageError> From<CommitToPendingProposalsError<StorageError>> for TestError<StorageError> {
+    fn from(e: CommitToPendingProposalsError<StorageError>) -> Self {
+        TestError::CommitToPendingProposals(e)
     }
 }
 
--- a/openmls/src/test_utils/single_group_test_framework/errors.rs
+++ b/openmls/src/test_utils/single_group_test_framework/errors.rs
@@ -1,3 +1,5 @@
+use thiserror::Error;
+
 use crate::prelude::{commit_builder::*, *};
 pub use crate::utils::*;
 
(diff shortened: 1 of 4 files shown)


openmls/src/test_utils/single_group_test_framework/errors.rs
```diff
@@ -0,0 +1,70 @@
...
+
+pub use openmls_traits::{
+    storage::StorageProvider as StorageProviderTrait,
+    types::{Ciphersuite, HpkeKeyPair},
+    OpenMlsProvider,
+};
+
+// type alias for `TestError`
+pub type GroupError<Provider> =
+    TestError<<<Provider as OpenMlsProvider>::StorageProvider as StorageProviderTrait<1>>::Error>;
```
keks, line 12: Should that `1` be `CURRENT_VERSION`?

openmls/src/test_utils/single_group_test_framework/mod.rs
```diff
@@ -0,0 +1,478 @@
...
+                    .add_proposal(add_proposal)
+            })
+            .expect("Could not stage commit");
+
+        // Deliver_and_apply to all members but Alice
+        group_state
+            .deliver_and_apply_if(bundle.commit().clone().into(), |member| {
+                member.party.core_state.name != "alice"
+            })
+            .expect("Error deliver_and_applying commit");
```
keks, line 470: I think we should still get alice to apply the locally staged commit

keks: Actually at this point the group only contains alice, so this doesn't do anything, or am I missing something?

openmls/src/test_utils/single_group_test_framework/mod.rs
```diff
@@ -0,0 +1,478 @@
...
+
+        // Deliver_and_apply welcome to Bob
+        let welcome = bundle.welcome().unwrap().clone();
+        group_state
+            .deliver_and_apply_welcome(bob_pre_group, mls_group_join_config, welcome, None)
+            .expect("Error deliver_and_applying welcome");
+    }
```
keks, line 477: Can we call assert_membership after this?

openmls/src/test_utils/single_group_test_framework/errors.rs
```diff
@@ -0,0 +1,70 @@
...
+    fn from(e: ProtocolMessageError) -> Self {
+        TestError::ProtocolMessage(e)
+    }
+}
+
+impl<StorageError> From<MergeCommitError<StorageError>> for TestError<StorageError> {
+    fn from(e: MergeCommitError<StorageError>) -> Self {
+        TestError::MergeCommit(e)
+    }
+}
```
keks, line 70: With thiserror::Error you can both make the Error type implement std::error::Error and add all the From impls with little boilerplate

keks: I think since this is a test helper, we could also just panic in the test libary instead of returning an error, but I don't have strong feeling here.

keks (changes requested):
I think this looks good in general! There are few things that would be good to address before we merge this.

openmls/src/test_utils/single_group_test_framework/mod.rs
```diff
@@ -0,0 +1,478 @@
...
+}
+
+/// All of the state for a group and its members
+pub struct GroupState<'a, Provider> {
+    group_id: GroupId,
+    members: HashMap<Name, MemberState<'a, Provider>>,
+}
+
+impl<'b, 'a: 'b, Provider: OpenMlsProvider> GroupState<'a, Provider> {
```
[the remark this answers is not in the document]
wysiwys: This `'b` lifetime is basically introduced in order to avoid lifetime errors during compilation. The issue is related to the `FnOnce` closure that is passed in as an argument to `build_commit_and_stage()`. The mutable borrow of `self` in `build_commit_and_stage()` needs to outlive the lifetime of the `CommitBuilder` that the `FnOnce` takes ownership of. I don't know the exact cause of this, though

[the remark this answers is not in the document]
wysiwys: Hmm, actually I was looking at the wrong line here. In fact, the `'b` lifetime is not needed here. I have gone ahead and removed it.

openmls/src/test_utils/single_group_test_framework/mod.rs
```diff
@@ -227,10 +227,34 @@ impl<'b, 'a: 'b, Provider: OpenMlsProvider> GroupState<'a, Provider> {
         Ok(Self { group_id, members })
     }
 
-    /// Get mutable references to all `MemberState`s as a fixed-size array
-    pub fn groups_mut<const N: usize>(&mut self) -> [&mut MemberState<'a, Provider>; N] {
+    /// Get mutable references to all `MemberState`s as a fixed-size array,
+    /// in the order of the names provided in `names`.
+    pub fn all_members_mut<const N: usize>(
```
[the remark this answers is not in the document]
keks: Also, and this is a really tiny nitpick, I think this function could be simplified a bit using `array::map`: https://doc.rust-lang.org/std/primitive.array.html#method.map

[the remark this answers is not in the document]
wysiwys: Indeed - at this point, we may as well allow any subset of members to be requested :) I'll take a look at reimplementing this function to do that instead

PR #1731, opened 2025-03-12
Add Helpers for resolving forks

keks:
This PR adds helpers for resolving forks in MLS groups. A state is forked if the group does not agree on the sequence of commits that were merged. This can happen due to state handling bugs in clients or a buggy DS.

The helpers are implemented on top of MlsGroup. Existing clients can already do this themselves, but the helpers take some of the burden of implementing this manually. They also are behind the `fork-resolution-helpers` feature flag, in order to make it opt-in.

This PR adds helpers based on two different strategies:

1. Remove and Re-Add. This requires the client initiating the resolution to know which members are in a different state. Given that information a commit that first removes all these clients and re-adds them.

2. Reboot the Group. Creates a new group and adds the same clients again. This does not require to know which clients are forked, but is more expensive.

Both of these are written using a builder style, so the caller can provide new key packages. For the reboot helper we had to work around the fact that we couldn't return both the new group and the commit builder, so instead any changes to the commit builder have to be done through a closure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#1666](https://github.com/openmls/openmls/pull/1666): Add `members()` and `group_context()` getter methods to `StagedWelcome`.
 - [#1672](https://github.com/openmls/openmls/pull/1672): Add `epoch()` getter method to `VerifiableGroupInfo`.
 - [#1673](https://github.com/openmls/openmls/pull/1673): Return more specific error when attempting to decrypt own messages: `ProcessMessageError::ValidationError(ValidationError::CannotDecryptOwnMessage)`.
-- [#1731](https://github.com/openmls/openmls/pull/1731): Add helpers to recover from group state forks
+- [#1731](https://github.com/openmls/openmls/pull/1731): Add helpers to recover from group state forks, hidden behind the new `fork-resolution` feature flag.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#1688](https://github.com/openmls/openmls/pull/1688): Add `unknown()` getter method to `Extensions`.
 - [#1666](https://github.com/openmls/openmls/pull/1666): Add `members()` and `group_context()` getter methods to `StagedWelcome`.
 - [#1672](https://github.com/openmls/openmls/pull/1672): Add `epoch()` getter method to `VerifiableGroupInfo`.
-- [#1673](https://github.com/openmls/openmls/pull/1673): Return more specific error when attemtping to decrypt own messages: `ProcessMessageError::ValidationError(ValidationError::CannotDecryptOwnMessage)`.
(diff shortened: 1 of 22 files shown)


openmls/src/group/fork_resolution/readd.rs
```diff
@@ -0,0 +1,256 @@
...
+impl MlsGroup {
+    /// Create a [`CommitBuilder`] that is preparing to remove and re-add members from other fork
+    /// partitions. `own_partition` is the list of [`LeafNodeIndex`] that are members in the
+    /// partition that the initiating client is in. This should include the [`LeafNodeIndex`] of
+    /// the initiating client.
+    pub fn recover_fork_by_readding(
+        &mut self,
+        own_partition: &[LeafNodeIndex],
+    ) -> Result<CommitBuilder<ReAddExpectKeyPackages>, ReAddError> {
+        // Load member info. This is None of at least one of the indexes is not a valid member
```
kkohbrok, line 30:
```suggestion
        // Load member info. This is None if at least one of the indexes is not a valid member
```

kkohbrok (changes requested):
This looks like it might be useful when dealing with forks. I found it a bit hard to understand just from the function and module documentation, though. Do you want to add a section in the book? Or include an example in the module docs? I think that would be helpful. Other than that I only found a few typos.

keks:
Thanks! Yes, I'll add something to the book.

openmls/tests/book_code_fork_resolution.rs
```diff
@@ -0,0 +1,487 @@
...
+
+    // Charlie joins using Alice's invite
+    let welcome = add_charlie_messages.into_welcome().unwrap();
+    let mut charlie_group =
+        StagedWelcome::new_from_welcome(charlie_provider, mls_group_config, welcome, None)
+            .unwrap()
+            .into_group(charlie_provider)
+            .unwrap();
+
+    // We shoulkd be forked now, double-check
```
franziskuskiefer, line 144:
```suggestion
    // We should be forked now, double-check
```

franziskuskiefer (commented):
You may want to run clippy on this code.

kkohbrok (dismissed):
Looks good modulo clippy and the issues pointed out by @franziskuskiefer. Thanks for adding that book section!

franziskuskiefer:
This should also get a mention in the changelog.

CHANGELOG.md
```diff
@@ -12,7 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#1688](https://github.com/openmls/openmls/pull/1688): Add `unknown()` getter method to `Extensions`.
 - [#1666](https://github.com/openmls/openmls/pull/1666): Add `members()` and `group_context()` getter methods to `StagedWelcome`.
 - [#1672](https://github.com/openmls/openmls/pull/1672): Add `epoch()` getter method to `VerifiableGroupInfo`.
-- [#1673](https://github.com/openmls/openmls/pull/1673): Return more specific error when attemtping to decrypt own messages: `ProcessMessageError::ValidationError(ValidationError::CannotDecryptOwnMessage)`.
+- [#1673](https://github.com/openmls/openmls/pull/1673): Return more specific error when attempting to decrypt own messages: `ProcessMessageError::ValidationError(ValidationError::CannotDecryptOwnMessage)`.
+- [#1731](https://github.com/openmls/openmls/pull/1731): Add helpers to recover from group state forks
```
franziskuskiefer, line 16: It's also introducing a new feature. Let's mention that this is behind that new feature.

franziskuskiefer (approved):

franziskuskiefer (approved):

PR #1738, opened 2025-03-26
Path Required: Fix wrong negation in code comment

timokoesters:
The RFC says

> The path field MAY be omitted if (a) it covers at least one proposal and (b) none of the proposals covered by the Commit are of "path required" types.

https://www.rfc-editor.org/rfc/rfc9420.html#section-12.4-9

With a negation, this means that a path is required if there is no proposal or at least one proposal requires a path.

---

I also saw these lines were written, probably because of "External Commits MUST contain a path field" https://www.rfc-editor.org/rfc/rfc9420.html#section-12.4.3.2-11.1

            // * (or) it is an external commit
            || external_init_proposal_option.is_some()
            
But this is implied by `proposals_require_path` as far as I can tell:

External commits require "exactly one ExternalInit" proposal (https://www.rfc-editor.org/rfc/rfc9420.html#section-12.2-5), which is a proposal type that requires a path (https://www.rfc-editor.org/rfc/rfc9420.html#section-12.4-9).

I can remove this line if you agree.

### DIFF
--- a/openmls/src/group/public_group/diff/apply_proposals.rs
+++ b/openmls/src/group/public_group/diff/apply_proposals.rs
@@ -173,7 +173,7 @@ impl PublicGroupDiff<'_> {
             .any(|p| p.proposal().is_path_required());
 
         // This flag determines if the commit requires a path. A path is required if:
-        // * none of the proposals require a path
+        // * at least one proposal requires a path
         // * (or) it is an external commit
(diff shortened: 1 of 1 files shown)


timokoesters:
You've asked me to include the change I suggested above. I pushed it and it is ready to merge now.

kkohbrok (dismissed):
Thanks!

kkohbrok (approved):

PR #1739, opened 2025-03-31
Add validation test for valn0108

wysiwys:
Add validation test for https://validation.openmls.tech/#valn0108

- `KeyPackageIn::validate()` should return the error `KeyPackageVerifyError::InvalidLeafNodeSourceType` if  `leaf_node_source` is not of the correct enum variant `LeafNodeSource::KeyPackage(_)`


franziskuskiefer (approved):
Thanks, lgtm

The main question for me here is how much we should test. This tests now that the validation function is doing the right thing. But we don't know yet if the validation function is always called.
We don't need to test all cases for tha latter. But we should make sure have tests that make sure we always call into the `validate` function.

Issue #1740, opened 2025-03-31
Add validation test: valn0104


wysiwys:
https://validation.openmls.tech/#valn0104
- Verify that the credential type is supported by all members of the group, as specified by the capabilities field of each member's LeafNode, 
- and that the capabilities field of this LeafNode indicates support for all the credential types currently in use by other members.

PR #1744, opened 2025-03-31
Add validation tests for valn0104

wysiwys:
This PR adds several validation tests corresponding to https://validation.openmls.tech/#valn0104:
- `test_valn_0104_new_member_unsupported_credential_type()`: Verify that the credential type is supported by all members of the group, as specified by the capabilities field of each member's LeafNode
- `fn test_valn_0104_new_member_capabilities_not_support_all_credential_types()`: Verify that the capabilities field of the [added] LeafNode indicates support for all the credential types currently in use by other members

In addition to these tests, this PR also contains a small update to the `single_group_test_framework` :

- `pub use` the `GroupError` error type (in order to be able to also the error types it wraps) 
-  add error handling for the call to `MlsGroup::add_members()` in `GroupState::add_member()`

Resolves issue #1740

### DIFF
--- a/openmls/src/group/tests_and_kats/tests/capabilities_check.rs
+++ b/openmls/src/group/tests_and_kats/tests/capabilities_check.rs
@@ -3,7 +3,6 @@ use crate::test_utils::single_group_test_framework::*;
 use crate::treesync::errors::LeafNodeValidationError;
 
 // Helper macro for checking error matches a provided pattern
-#[cfg(test)]
 macro_rules! assert_err_matches {
     ($err:expr, $pattern:pat) => {
         assert!(matches!($err.expect_err("Expected an error"), $pattern));
@@ -11,7 +10,6 @@ macro_rules! assert_err_matches {
 }
 
 // Function to check that the correct error type was returned
-#[cfg(test)]
 fn test_valn0104_error<Provider: OpenMlsProvider>(error: Result<(), GroupError<Provider>>) {
     assert_err_matches!(
         error,
@@ -25,7 +23,6 @@ fn test_valn0104_error<Provider: OpenMlsProvider>(error: Result<(), GroupError<P
     );
 }
 
-#[cfg(test)]
 impl<'a, 'b: 'a, Provider: OpenMlsProvider + Default> GroupState<'b, Provider> {
     // add a member to the GroupState with the specified credential capabilities
     fn add_member_with_credential_capabilities(
@@ -84,7 +81,6 @@ impl<'a, 'b: 'a, Provider: OpenMlsProvider + Default> GroupState<'b, Provider> {
     }
 }
 
-#[cfg(test)]
 impl<'a, 'b: 'a, Provider: OpenMlsProvider> PreGroupPartyState<'b, Provider> {
     // Helper function to update the PreGroupPartyState to support the specified CredentialTypes in its Capabilities
     fn update_credential_capabilities(
--- a/openmls/src/group/tests_and_kats/tests/capabilities_check.rs
+++ b/openmls/src/group/tests_and_kats/tests/capabilities_check.rs
@@ -10,7 +10,7 @@ macro_rules! assert_err_matches {
 }
 
 // Function to check that the correct error type was returned
-fn test_valn0104_error<Provider: OpenMlsProvider>(error: Result<(), GroupError<Provider>>) {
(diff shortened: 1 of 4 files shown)


openmls/src/group/tests_and_kats/tests/capabilities_check.rs
```diff
@@ -0,0 +1,352 @@
...
+#[cfg(test)]
+macro_rules! assert_err_matches {
+    ($err:expr, $pattern:pat) => {
+        assert!(matches!($err.expect_err("Expected an error"), $pattern));
+    };
+}
+
+// Function to check that the correct error type was returned
+#[cfg(test)]
+fn test_valn0104_error<Provider: OpenMlsProvider>(error: Result<(), GroupError<Provider>>) {
```
keks, line 15: This function name sounds a bit like this is a test. Maybe call it `expect_valn0104_error`?

openmls/src/group/tests_and_kats/tests/capabilities_check.rs
```diff
@@ -0,0 +1,352 @@
...
+        .expect("Could not add member");
+
+    // Case with no credential capabilities; should fail
+    // Alice adds Dave
+    test_valn0104_error::<Provider>(group_state.add_member_with_credential_capabilities(
+        &dave_party,
+        "alice",
+        ciphersuite,
+        Vec::new(),
+    ));
```
keks, line 316: So this returns an error because Dave doesn't support his own CredentialType (which should be Basic)?

openmls/src/group/tests_and_kats/tests/capabilities_check.rs
```diff
@@ -0,0 +1,352 @@
...
+
+    // Case with wrong capabilities; should fail
+    // Alice adds Dave
+    test_valn0104_error::<Provider>(group_state.add_member_with_credential_capabilities(
+        &dave_party,
+        "alice",
+        ciphersuite,
+        vec![CredentialType::Basic, CredentialType::Other(2)],
+    ));
```
keks, line 325: I am not sure why this returns an error. Is there someone how has an `Other(2)` credential? If not, shouldn't this succeed?

keks (commented):
I think in general this looks pretty solid. I have two questions on what some parts of the test are actually testing, and some nits.

openmls/src/group/tests_and_kats/tests/capabilities_check.rs
wysiwys: The thing here is that Dave needs to support all of the credential types currently in use by other members. However, he doesn't support `Other(3)` (Alice's credential), even though he does support  `Basic` (Bob and Charlie's credentials)

openmls/src/group/tests_and_kats/tests/capabilities_check.rs
wysiwys: This test case should fail because Dave needs to support all of the credential types that Alice, Bob and Charlie use, but he doesn't support any at all. However, I don't know if Dave is required to support his own credential type. If Dave doesn't support any credential types, adding him to an existing group should always fail, because he doesn't support the other members' credentials, but should Dave supporting his own credential type be checked on `MlsGroup` creation?

openmls/src/group/tests_and_kats/tests/capabilities_check.rs
keks: Ah yes, I missed that Alice as an `Other(3)` credential

openmls/src/group/tests_and_kats/tests/capabilities_check.rs
keks: Ah yes, that makes sense.

Regarding the unclearity: I don't think the RFC has explicit wording for this, but I think it is implied by the way [valn0202](https://validation.openmls.tech/#valn0202) is phrased.

keks (approved):
I think there still are superfluous turbofishes (in the definition of and calls to `expect_valn0104_err`). Otherwise LGTM, Thanks!

wysiwys:
@keks Thanks for the review! For some reason, I get a compiler error (`cannot infer type`) when I remove the turbofishes for `expect_valn0104_err`. However, I'm not sure why this is, since the type of `Provider` should already be determined when the output of `add_member_with_credential_type()` is passed to this function. Leaving as is for now

PR #1748, opened 2025-04-03
Add references to validation check valn0113

wysiwys:
This PR adds references to [valn0113](https://validation.openmls.tech/#valn0113)  to the comments in the codebase, to document where this `LeafNode` validation check takes place. This check was recently added to the dashboard, but is already implemented in the proposal validation by the `CommitBuilder`.

These annotations will also allow these locations to be searchable from the `search code` link that is provided in the validation dashboard.

### DIFF
--- a/openmls/src/group/public_group/validation.rs
+++ b/openmls/src/group/public_group/validation.rs
@@ -313,9 +313,10 @@ impl PublicGroup {
         Ok(())
     }
 
-    /// Validate capablities. This function implements the following checks:
+    /// Validate capabilities. This function implements the following checks:
     /// - ValSem106: Add Proposal: required capabilities
     /// - ValSem109: Update Proposal: required capabilities
+    /// Implements check [valn0113](https://validation.openmls.tech/#valn0113).
     pub(crate) fn validate_capabilities(
         &self,
         proposal_queue: &ProposalQueue,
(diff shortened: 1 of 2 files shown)


franziskuskiefer (dismissed):
lgtm

franziskuskiefer:
clippy doesn't seem happy

wysiwys:
> clippy doesn't seem happy

@franziskuskiefer It turns out that the reason that clippy was failing was that some new lints were added in the latest version of clippy, two of which failed on existing code in OpenMLS that was not modified by this PR. 

One of the lints is [doc_overindented_list_items](https://rust-lang.github.io/rust-clippy/master/index.html#doc_overindented_list_items), which only applies to comments. 

The other one, [manual_div_ceil](https://rust-lang.github.io/rust-clippy/master/index.html#manual_div_ceil), affects one line in `TreeSize::dec()` in `treemath.rs`, and suggests replacing `(self.0 + 1)/2` with `self.0.div_ceil(2)`.  Although I would not expect this change to affect the behavior of `TreeSize::dec()`, given that this is still a change to the OpenMLS library code, is this something that should be evaluated separately or be a separate PR?

wysiwys:
> clippy doesn't seem happy

These lints were fixed in #1753 , and this PR now only contains valn0113 reference updates

keks (approved):
Looks good, thanks!

PR #1750, opened 2025-04-05
Support add proposals from external senders

### LINKED ISSUE #106: Support handshake messages from externals and new members

Snazzah:
This PR adds support processing add proposals from external senders.

Not sure if I missed anything other than adding this, please let me know.

Fixes #106.


raphaelrobert:
I'm trying to understand what the PR tries to implement. Are you referring to [External Proposals](https://www.rfc-editor.org/rfc/rfc9420.html#section-12.1.8)? In that case I think we need the External Sender Extension first. I think the library should be able to handle that automatically and enforce the adequate verification of the external senders as well.

Snazzah:
@raphaelrobert Yes I am referring to External Proposals, it seemed to only support `remove` but not `add`. I know there are others but I just added this in since I needed this for a project.

raphaelrobert (changes requested):
Thanks for making the changes, this looks solid in general. Just a nit and the CI is complaining about missing code formatting.

Do I understand correctly that you renamed `openmls/src/group/tests_and_kats/tests/external_join_add_proposal.rs` in `external_join_add_proposal.rs`? If so, that was exactly the right thing to do.

Snazzah:
Yes, I renamed the test so it would make sense name-wise.

raphaelrobert (approved):
Thanks, let's get this in.

franziskuskiefer:
Should this be added to the changelog?

raphaelrobert:
> Should this be added to the changelog?

Good point, I'll file a PR.

Issue #1751, opened 2025-04-07
Update libcrux provider


franziskuskiefer:
Update the libcrux provider to use the new pure Rust crates. This will completely drop AES support.

After the transition we should run benchmarks to see if we need to do something about performance.

- [x] use the reseedable chacha20 rng for now for rand in this provider

PR #1760, opened 2025-04-09
Update libcrux provider to use pure Rust crates

wysiwys:
This PR updates the libcrux provider to use pure Rust crates where possible (see #1751):
- `libcrux-chacha20poly1305`
- `libcrux-ed25519`
- `libcrux-hkdf`
- `libcrux-sha2`
- `hpke-rs`

For now, in the crypto and rand providers, the `libcrux::drbg::Drbg` is also replaced with a `rand::rngs::ReseedingRng<rand_chacha::ChaCha20Core,rand::rngs::OsRng>`.

This PR also drops AES support from the libcrux provider.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -68,7 +68,7 @@ jobs:
           sudo apt update && sudo apt install nodejs
           cargo install wasm-bindgen-cli
           export CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=$HOME/.cargo/bin/wasm-bindgen-test-runner
-          cargo test $TEST_MODE -p openmls -vv --target wasm32-unknown-unknown -F js libcrux-provider
+          cargo test $TEST_MODE -p openmls -vv --target wasm32-unknown-unknown -F js,libcrux-provider
 
       - name: Tests
         if: matrix.os != 'windows-latest'
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,6 +35,8 @@ jobs:
         with:
           targets: i686-pc-windows-msvc, i686-unknown-linux-gnu, wasm32-unknown-unknown
       - uses: baptiste0928/cargo-install@v3
+        # we only test wasm bindings using ubuntu-latest
+        if: matrix.os == 'ubuntu-latest'
         with:
           crate: wasm-bindgen-cli
       - uses: Swatinem/rust-cache@v2
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Tests Wasm32 on linux with libcrux provider
         if: matrix.os == 'ubuntu-latest'
         env:
-          RUSTFLAGS: --cfg getrandom_backend="wasm_js"
+          RUSTFLAGS: "--cfg getrandom_backend=\"wasm_js\""
         run: |
           sudo apt update && sudo apt install nodejs
           cargo install wasm-bindgen-cli
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -47,11 +47,10 @@ jobs:
 
       - name: Tests Wasm32 on linux with libcrux provider
         if: matrix.os == 'ubuntu-latest'
-        env:
-          RUSTFLAGS: "--cfg getrandom_backend=\"wasm_js\""
         run: |
           sudo apt update && sudo apt install nodejs
           cargo install wasm-bindgen-cli
+          export RUSTFLAGS="--cfg getrandom_backend=\"wasm_js\""
           export CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=$HOME/.cargo/bin/wasm-bindgen-test-runner
           cargo test $TEST_MODE -p openmls -vv --target wasm32-unknown-unknown -F js,libcrux-provider
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -52,7 +52,7 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         run: |
           sudo apt update && sudo apt install nodejs
-          export CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=$HOME/.cargo-install/bin/wasm-bindgen-test-runner
+          export CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=$HOME/.cargo-install/wasm-bindgen-cli/bin/wasm-bindgen-test-runner
           cargo test $TEST_MODE -p openmls -vv --target wasm32-unknown-unknown -F js
 
       - name: Tests Wasm32 on linux with libcrux provider
@@ -60,7 +60,7 @@ jobs:
         run: |
           sudo apt update && sudo apt install nodejs
           export RUSTFLAGS="--cfg getrandom_backend=\"wasm_js\""
-          export CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=$HOME/.cargo-install/bin/wasm-bindgen-test-runner
+          export CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=$HOME/.cargo-install/wasm-bindgen-cli/bin/wasm-bindgen-test-runner
           cargo test $TEST_MODE -p openmls -vv --target wasm32-unknown-unknown -F js,libcrux-provider,libcrux-provider-js
 
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,6 +36,8 @@ jobs:
           targets: i686-pc-windows-msvc, i686-unknown-linux-gnu, wasm32-unknown-unknown
       - uses: Swatinem/rust-cache@v2
       - uses: baptiste0928/cargo-install@v3
+        # we only test wasm bindings on ubuntu-latest
+        if: matrix.os == 'ubuntu-latest'
         with:
           crate: wasm-bindgen-cli
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -62,6 +62,14 @@ jobs:
           export CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=$HOME/.cargo/bin/wasm-bindgen-test-runner
           cargo test $TEST_MODE -p openmls -vv --target wasm32-unknown-unknown -F js
 
+      - name: Tests Wasm32 on linux with libcrux provder
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          sudo apt update && sudo apt install nodejs
+          cargo install wasm-bindgen-cli
+          export CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=$HOME/.cargo/bin/wasm-bindgen-test-runner
+          cargo test $TEST_MODE -p openmls -vv --target wasm32-unknown-unknown -F js libcrux-provider
+
       - name: Tests
         if: matrix.os != 'windows-latest'
         run: cargo test $TEST_MODE -p openmls --verbose
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -53,7 +53,6 @@ jobs:
       - name: Tests Wasm32 on linux
         if: matrix.os == 'ubuntu-latest'
         run: |
-          sudo apt update && sudo apt install nodejs
           export CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=$HOME/.cargo-install/wasm-bindgen-cli/bin/wasm-bindgen-test-runner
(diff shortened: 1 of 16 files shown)


libcrux_crypto/src/crypto.rs
```diff
@@ -328,247 +324,66 @@ impl OpenMlsCrypto for CryptoProvider {
...
-fn sig_alg(alg: SignatureScheme) -> Result<libcrux::signature::Algorithm, CryptoError> {
-    match alg {
-        SignatureScheme::ECDSA_SECP256R1_SHA256 => Ok(libcrux::signature::Algorithm::EcDsaP256(
-            libcrux::signature::DigestAlgorithm::Sha256,
-        )),
-        SignatureScheme::ED25519 => Ok(libcrux::signature::Algorithm::Ed25519),
-        _ => Err(CryptoError::UnsupportedSignatureScheme),
-    }
-}
+    // TODO: should we return an error on unsupported AES algorithms here?
```
keks, line 346: I don't think so. My reading is that this is just about mapping the algorithm enums, not about support.

keks (changes requested):
I think this looks pretty good, we should be able to merge that soon. I have tried responding to all the open questions, let me know if I missed anything.

libcrux_crypto/src/crypto.rs
```diff
@@ -152,38 +134,55 @@ impl OpenMlsCrypto for CryptoProvider {
...
+        // The only supported AeadType (as of openmls_traits v0.3.0) is ChaCha20Poly1305
+        if !matches!(alg, AeadType::ChaCha20Poly1305) {
+            return Err(CryptoError::UnsupportedAeadAlgorithm);
+        }
+
+        if ct_tag.len() < 16 {
             return Err(CryptoError::InvalidLength);
         }
 
         let boundary = ct_tag.len() - 16;
```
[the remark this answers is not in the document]
wysiwys: Definitely agree - I went ahead and made these replacements where needed.

libcrux_crypto/src/crypto.rs
```diff
@@ -259,21 +254,25 @@ impl OpenMlsCrypto for CryptoProvider {
...
+                input.kem_output.as_ref(),
+                &sk_r,
+                info,
+                aad,
+                input.ciphertext.as_ref(),
+                None,
+                None,
+                None,
+            )
+            // TODO: should more error types be handled?
```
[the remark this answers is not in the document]
wysiwys: I agree - it would make sense to restructure the hpke-rs errors to be separate types for open/seal/setup, since the `HpkeError` type currently does not convey which of the relevant steps the error is returned from. 

Because of this, I think it would make sense to keep the error handling here simple for now, since it's not always clear how to map the errors, and not all of the error variants we had in `libcrux::hpke` before are available, specifically `DecapError` and `ValidationError`.

libcrux_crypto/src/crypto.rs
```diff
@@ -1,52 +1,41 @@
...
-            | (AeadType::ChaCha20Poly1305, true)
-            | (AeadType::ChaCha20Poly1305, false) => Ok(()),
+        match ciphersuite.aead_algorithm() {
+            AeadType::ChaCha20Poly1305 => Ok(()),
             _ => Err(CryptoError::UnsupportedCiphersuite),
         }?;
 
         match ciphersuite.signature_algorithm() {
-            SignatureScheme::ECDSA_SECP256R1_SHA256 | SignatureScheme::ED25519 => Ok(()),
+            SignatureScheme::ED25519 => Ok(()),
```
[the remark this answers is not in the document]
wysiwys: I think it would be possible to keep ECDSA-P256 here (likely using the `libcrux-ecdsa` crate), although in this pull request, the methods `verify_signature()`, `signature_key_gen()`, and `sign()`  only implement logic for ED25519. Is that something we want to support in this PR, or possibly as part of a separate pull request?

wysiwys:
For wasm, the `libcrux-provider` feature now requires the `libcrux-provider-js` feature to be enabled as well, which enables the `wasm_js` feature for `getrandom`. On CI, this also requires the `RUSTFLAGS` to be configured according to the method described in the [documentation](https://docs.rs/getrandom/latest/getrandom/#usage).

traits/src/types.rs
```diff
@@ -181,7 +181,7 @@ pub enum HpkeKemType {
     DhKem448 = 0x0021,
 
     /// XWing combiner for ML-KEM and X25519
-    XWingKemDraft2 = 0x004D,
+    XWingKemDraft6 = 0x004D,
```
franziskuskiefer, line 184: This is a breaking change. We need to bump the version for this crate and add it to the changelog.

.github/workflows/tests.yml
```diff
@@ -45,22 +45,32 @@ jobs:
...
+          cargo test $TEST_MODE -p openmls -vv --target wasm32-unknown-unknown -F js,libcrux-provider,libcrux-provider-js
+
+
+      - name: Test with Fork Resolution helpers
+        run: cargo test $TEST_MODE -p openmls -F fork-resolution
+
+      - name: Test with libcrux provider
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          cargo install wasm-bindgen-cli
```
franziskuskiefer, line 72: We should make sure this is cached in some way (not sure if we have the cache enabled, or if we should use the prebuilt binaries that we use for coverage as well)

libcrux_crypto/Cargo.toml
```diff
@@ -10,9 +10,18 @@ repository = "https://github.com/openmls/openmls/tree/main/openmls_libcrux_crypt
...
+libcrux-hkdf = { version = "0.0.2" }
+libcrux-sha2 = { version = "0.0.2" }
 openmls_traits = { version = "0.3.0", path = "../traits" }
 openmls_memory_storage = { version = "0.3.0", path = "../memory_storage" }
-rand = "0.8.5"
+rand = "0.9.0"
 tls_codec.workspace = true
+rand_chacha = "0.9.0"
+
+hpke_rs = { git = "https://github.com/cryspen/hpke-rs", branch = "main", package = "hpke-rs" , features=["hazmat", "serialization"] }
```
franziskuskiefer, line 23: Let's make sure all of these dependencies have a version as well so that we can release this later.

.github/workflows/tests.yml
wysiwys: :+1: I used the `baptiste0928/cargo-install` action to install this on ubuntu-latest only (since we always run the wasm tests on that platform).

wysiwys:
For wasm, this PR also adds a new feature flag for OpenMLS (`libcrux-provider-js`) that will need to be enabled alongside `libcrux-provider` and `js` on the `wasm32-unknown-unknown` target. The purpose of this is to configure the correct feature flags for the `getrandom` dependency that are required by this specific combination of features.

.github/workflows/tests.yml
```diff
@@ -45,23 +50,30 @@ jobs:
...
+          export CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=$HOME/.cargo-install/wasm-bindgen-cli/bin/wasm-bindgen-test-runner
+          cargo test $TEST_MODE -p openmls -vv --target wasm32-unknown-unknown -F js
+
+      - name: Tests Wasm32 on linux with libcrux provider
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          sudo apt update && sudo apt install nodejs
+          export RUSTFLAGS="--cfg getrandom_backend=\"wasm_js\""
+          export CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=$HOME/.cargo-install/wasm-bindgen-cli/bin/wasm-bindgen-test-runner
+          cargo test $TEST_MODE -p openmls -vv --target wasm32-unknown-unknown -F js,libcrux-provider,libcrux-provider-js
```
keks, line 66: I think it would be marginally cleaner by using an [`env` declaration](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#defining-environment-variables-for-a-single-workflow) instead of exporting in the script, but I don't think that should block anything.

keks (approved):
I think this looks good, with a minor nitpick

.github/workflows/tests.yml
wysiwys: Thanks! Updated :+1:

libcrux_crypto/Cargo.toml
```diff
@@ -10,9 +10,18 @@ repository = "https://github.com/openmls/openmls/tree/main/openmls_libcrux_crypt
...
+libcrux-sha2 = { version = "0.0.2" }
+openmls_traits = { version = "0.4.0", path = "../traits" }
 openmls_memory_storage = { version = "0.3.0", path = "../memory_storage" }
-rand = "0.8.5"
+rand = "0.9.0"
 tls_codec.workspace = true
+rand_chacha = "0.9.0"
+
+hpke_rs = { version = "0.2.1-alpha.1", git = "https://github.com/cryspen/hpke-rs", branch = "main", package = "hpke-rs" , features=["hazmat", "serialization"] }
+
```
franziskuskiefer, line 24: Let's format this (and not have empty lines in between).

libcrux_crypto/src/crypto.rs
```diff
@@ -152,38 +136,56 @@ impl OpenMlsCrypto for CryptoProvider {
...
-        libcrux::aead::decrypt(&key, &mut c, iv, aad, &tag).map_err(|e| match e {
-            libcrux::aead::DecryptError::InvalidArgument(
-                libcrux::aead::InvalidArgumentError::UnsupportedAlgorithm,
-            ) => CryptoError::UnsupportedAeadAlgorithm,
-            libcrux::aead::DecryptError::DecryptionFailed => CryptoError::AeadDecryptionError,
-            _ => CryptoError::CryptoLibraryError,
-        })?;
+        let iv = <&[u8; 12]>::try_from(nonce).map_err(|_| CryptoError::InvalidLength)?;
+
+        // TODO: instead, use key conversion from the libcrux-chacha20poly1305 crate, when available
```
franziskuskiefer, line 155: Do you want to address this TODO before merging?

libcrux_crypto/src/rand.rs
```diff
@@ -1,11 +1,13 @@
...
 
-use libcrux::drbg::Drbg;
 use openmls_traits::random::OpenMlsRand;
 
-/// The libcrux-backed randomness provider for OpenMLS
+use rand::{rngs::OsRng, rngs::ReseedingRng, TryRngCore};
+use rand_chacha::ChaCha20Core;
+
+/// The randomness provider for the libcrux-backed OpenMLS provider
 pub struct RandProvider {
```
franziskuskiefer, line 9: Shall we drop this one? We could just implement `OpenMlsRand` for the crypto provider and would avoid duplicating code here.
I know that this was the style before. But I think that was more due to historic reasons, not because it made sense.

libcrux_crypto/src/crypto.rs
```diff
@@ -1,52 +1,41 @@
...
 pub struct CryptoProvider {
-    drbg: Mutex<Drbg>,
+    rng: Mutex<ReseedingRng<ChaCha20Core, OsRng>>,
 }
 
 impl Default for CryptoProvider {
     fn default() -> Self {
-        let mut seed = [0u8; 16];
-        getrandom::getrandom(&mut seed).unwrap();
+        let reseeding_rng = ReseedingRng::<ChaCha20Core, _>::new(0x100000000, OsRng).unwrap();
```
franziskuskiefer, line 23: This `unwrap` here is unfortunate. It won't happen during protocol execution. But it's still not great.
I think we really only need to implement `Default` because of the tests (`openmls_test`). We could instead call a different function there that can return an error.
But I'd be fine keeping this as is and adding a `new` function that consumers should use instead, which returns an error.

franziskuskiefer (changes requested):
Thanks! Generally looks good.  I left a couple comments inline.

libcrux_crypto/src/crypto.rs
wysiwys: Should I take these out? There is no key conversion provided by that crate currently

.github/workflows/tests.yml
```diff
@@ -45,23 +50,31 @@ jobs:
             echo "TEST_MODE=--release" >> $GITHUB_ENV
           fi
 
+      - name: Tests Wasm32 on linux
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          sudo apt update && sudo apt install nodejs
```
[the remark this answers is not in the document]
wysiwys: :+1:  As it turns out, installing it here is not necessary. I had left it in from the previous workflow configuration, but have now taken it out.

libcrux_crypto/src/crypto.rs
wysiwys: :+1:  I added a `new()` method here that returns an error, as well as to the `Provider` itself

libcrux_crypto/src/crypto.rs
wysiwys: I have taken out these TODOs since there is nothing currently that is actionable. If key conversion functionality becomes available later in the libcrux-chacha20poly1305 crate, then it can be replaced here.

keks (approved):
Still looks good to me :)

franziskuskiefer (approved):
Thanks! Let's get this in :shipit:

PR #1762, opened 2025-04-14
chore: expose LeafNodeSource

kkohbrok:
There's a public `leaf_node_source` function that allows getting the `LeafNodeSource` of a leaf. However, the `LeafNodeSource` type is not actually public. This PR fixes that.


raphaelrobert (changes requested):
Looks good, but I think we need an entry in the changelog since there's a change in the public API

PR #1766, opened 2025-04-30
Use SignerBundle instead of Signer for update

kkohbrok:
This PR introduces a `NewSignerBundle` struct that allows passing a signer together with a CredentialWithKey when using `self_update_with_new_signer`. `self_update_with_new_signer` now also throws an error if a (non-matching) `CredentialWithKey` is passed in via `LeafNodeParameters`. All this is to avoid passing in mismatching signer and credentials/keys.

### DIFF
--- a/openmls/src/group/mls_group/updates.rs
+++ b/openmls/src/group/mls_group/updates.rs
@@ -2,9 +2,15 @@ use commit_builder::CommitMessageBundle;
 use errors::{ProposeSelfUpdateError, SelfUpdateError};
 use openmls_traits::{signatures::Signer, storage::StorageProvider as _};
 
-use crate::{storage::OpenMlsProvider, treesync::LeafNodeParameters};
+use crate::{prelude::CredentialWithKey, storage::OpenMlsProvider, treesync::LeafNodeParameters};
 
 use super::*;
+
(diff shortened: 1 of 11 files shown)


openmls/src/group/mls_group/updates.rs
```diff
@@ -2,10 +2,21 @@ use commit_builder::CommitMessageBundle;
...
-use crate::{storage::OpenMlsProvider, treesync::LeafNodeParameters};
+use crate::{prelude::CredentialWithKey, storage::OpenMlsProvider, treesync::LeafNodeParameters};
 
 use super::*;
 
+/// Bundle consisting of a [`Signer`] and a [`CredentialWithKey`] to be used to
+/// update the signature key in an [`MlsGroup`]. The public key and credential
+/// in `credential_with_key` MUST match the signature key exposed by `signer`.
+#[derive(Debug, Clone)]
+pub struct NewSignerBundle<'a, S: Signer> {
```
franziskuskiefer, line 13: Should this maybe go next to `CredentialWithKey` in `credentials/mod.rs`?

franziskuskiefer (approved):
Thanks, I think this reduces the potential for misusing the API.

PR #1773, opened 2025-05-20
Add validation test for valn1207

### LINKED ISSUE #1742: Add validation test: valn1207
https://validation.openmls.tech/#valn1207

wysiwys:
Closes #1742

### DIFF
--- a/openmls/src/group/tests_and_kats/tests/leaf_node_validation.rs
+++ b/openmls/src/group/tests_and_kats/tests/leaf_node_validation.rs
@@ -132,6 +132,13 @@ impl<P: OpenMlsProvider> GroupState<'_, P> {
 ///   is set to `Update`.
 #[openmls_test]
 fn valn1207() {
+    // Set up parties
+    let alice_party = CorePartyState::<Provider>::new("alice");
+    let bob_party = CorePartyState::<Provider>::new("bob");
+    let charlie_party = CorePartyState::<Provider>::new("charlie");
+    let charlie_pre_group = charlie_party.generate_pre_group(ciphersuite);
+    let charlie_key_package = charlie_pre_group.key_package_bundle.key_package().clone();
+
     // Set up a new group with Alice
     let group_id = GroupId::from_slice(b"Test Group");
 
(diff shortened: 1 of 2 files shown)


openmls/src/group/tests_and_kats/tests/leaf_node_validation.rs
```diff
@@ -0,0 +1,196 @@
...
+        assert_eq!(
+            err,
+            ProcessMessageError::ValidationError(ValidationError::UpdatePathError(
+                UpdatePathError::InvalidType
+            ))
+        );
+    }
+}
+
+/// This test makes sure that validation check 1207 (valn1207) is performed:
```
keks, line 125: This comment should link the check on the dashboard

openmls/src/group/tests_and_kats/tests/leaf_node_validation.rs
```diff
@@ -0,0 +1,196 @@
...
+        &mls_message_out,
+        frankenstein::FrankenLeafNodeSource::Update,
+    );
+    group_state.test_valn_1207(
+        &mls_message_out,
+        frankenstein::FrankenLeafNodeSource::KeyPackage(frankenstein::FrankenLifetime {
+            not_before: 0,
+            not_after: 1,
+        }),
+    );
```
keks, line 195: The sentence from the RFC is

> If the path value is populated, validate it and apply it to the tree: Validate the LeafNode as specified in Section 7.3. The leaf_node_source field MUST be set to commit. 

I see that we check that the source is correct. Do we also check that commits that have a path with an invalid key package are rejected?

keks (changes requested):
Generally looks good, just one thing I am unclear about

openmls/src/group/tests_and_kats/tests/leaf_node_validation.rs
wysiwys: Indeed, that will also need to be tested, in addition to the rest of the leaf node validation from [Section 7.3](https://www.rfc-editor.org/rfc/rfc9420.html#section-7.3). Since that leaf node validation is part of other validation checks, and therefore also needs to be tested in other validation tests, I think it would make sense to find a reusable way to test these Section 7.3 validation checks, such as a test utility.

franziskuskiefer:
@wysiwys @keks what's the state here? We should either revive this or close it.

franziskuskiefer:
Closing for now

PR #1774, opened 2025-05-20
feat: group info creation flag in commit builder

kkohbrok:
This PR adds a way to control whether the `CommitBuilder` outputs a GroupInfo. It's backwards-compatible in that the newly added `create_group_info` flag defaults to the `use_ratchet_tree_extension` in the `MlsGroup`s `MlsGroupJoinConfig`.


franziskuskiefer (approved):
Thanks, lgtm.

This seems worth mentioning in the changelog?

franziskuskiefer (approved):

PR #1784, opened 2025-06-10
Support GroupContextExtensions proposals from external senders

wysiwys:
This pull request adds support for external GroupContextExtensions proposals.
- Implements a function on `ExternalProposal` to build a new external GroupContextExtensions proposal, which can then be committed by an existing group member.
- Adds handling logic for the processing stage in `MlsGroup::process_unverified_message()` and `PublicGroup::process_unverified_message()`
- Adds tests for this new functionality
- Consistently use `QueuedProposal::from_authenticated_content_by_ref()` for external proposals

### DIFF
--- a/cli/src/networking.rs
+++ b/cli/src/networking.rs
@@ -8,8 +8,8 @@ use openmls::prelude::tls_codec::Serialize;
 
 pub fn post(url: &Url, msg: &impl Serialize) -> Result<Vec<u8>, String> {
     let serialized_msg = msg.tls_serialize_detached().unwrap();
-    log::debug!("Post {:?}", url);
-    log::trace!("Payload: {:?}", serialized_msg);
+    log::debug!("Post {url:?}");
+    log::trace!("Payload: {serialized_msg:?}");
     let client = Client::new();
     let response = client.post(url.to_string()).body(serialized_msg).send();
     if let Ok(r) = response {
@@ -35,11 +35,11 @@ pub fn get_with_body(url: &Url, body: &impl Serialize) -> Result<Vec<u8>, String
 }
 
 fn get_internal(url: &Url, msg: Option<&impl Serialize>) -> Result<Vec<u8>, String> {
-    log::debug!("Get {:?}", url);
+    log::debug!("Get {url:?}");
     let client = Client::new().get(url.to_string());
     let client = if let Some(msg) = msg {
         let serialized_msg = msg.tls_serialize_detached().unwrap();
-        log::trace!("Payload: {:?}", serialized_msg);
+        log::trace!("Payload: {serialized_msg:?}");
         client.body(serialized_msg)
     } else {
         client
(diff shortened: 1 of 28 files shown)


franziskuskiefer (commented):
If I recall correctly, there was a question about whether using `from_authenticated_content_by_ref` or `from_authenticated_content` is the right thing, and why different ones are used. They both do exactly the same thing, except that the one with `_ref` is using the proposal reference later when committing this proposal, and the other one uses the inline proposal. That's just a bad way of configuring this in the proposal queue. I think we should be consistent here for the external proposals and always do the same. I'd suggest using `_ref` by default.

I didn't look too deep into the code yet. But what I've seen looked good to me.

Can you unify the way we queue external proposals and make this ready for review?

openmls/src/group/tests_and_kats/tests/external_group_context_extensions_proposal.rs
```diff
@@ -0,0 +1,475 @@
...
+
+    let bob_credential_with_key =
+        generate_credential_with_key("Bob".into(), ciphersuite.signature_algorithm(), provider);
+
+    let bob_key_package = KeyPackage::builder()
+        .key_package_extensions(Extensions::empty())
+        .leaf_node_capabilities(
+            Capabilities::builder()
+                .extensions(vec![ExtensionType::Unknown(0xf001)])
+                .build(),
```
franziskuskiefer, line 91: Is this c&p or do we actually care about the capabilities? The test below checks it, but that seems orthogonal to the actual thing that's being tested.

openmls/src/group/tests_and_kats/tests/external_group_context_extensions_proposal.rs
```diff
@@ -0,0 +1,475 @@
...
+        .extensions()
+        .iter()
+        .any(|e| matches!(e, Extension::ExternalSenders(senders) if senders.iter().any(|s| s.credential() == &ds_credential_with_key.credential_with_key.credential) )));
+
+    assert!(alice_group
+        .own_leaf_node()
+        .unwrap()
+        .capabilities()
+        .extensions()
+        .contains(&ExtensionType::Unknown(0xf001)));
```
franziskuskiefer: So here you want to test this? But in the other one this seems unnecessary?

openmls/src/group/tests_and_kats/tests/external_group_context_extensions_proposal.rs
wysiwys: The purpose of this assertion is to check that the setup in `new_test_group()` worked correctly (as a prerequisite to the new test `external_group_context_ext_proposal_should_succeed()`). Would it be better to move this assertion into `new_test_group()` and take out the rendundant assertions below?

wysiwys: Or remove these entirely

franziskuskiefer: If you want to keep it to make sure the setup is correct, move it to the creation, i.e. `new_test_group`.

franziskuskiefer (approved):
Thanks. Can you squash your commits before mergin?

wysiwys:
Squashed and merged

Issue #1769, opened 2025-05-15
Release 0.7


franziskuskiefer:
Release the next version of OpenMLS (double check if it should be 0.6.2 or 0.7).

raphaelrobert:
I think there are enough breaking changes to move to 0.7.

keks:
This is chugging along, see
- https://github.com/openmls/openmls/pull/1796
- https://github.com/openmls/openmls/pull/1798

franziskuskiefer:
We decided to include #1793 but not #1801, which will go into 0.7.1.

keks:
I published the -pre.1 of all crates today. We need to bump futures-util because the version we use has been yanked; I'll PR that tomorrow (or so).

keks:
The futures-util issue is a bit funny. The Cargo.lock I have in my local clone had the outdated version, but it's in .gitignore so it's not in the repo and I can't PR it.

What I can do is publish again on the same commit, but with the updated dependency. This does raise the question though what our general strategy is for updating our lockfiles before publishing.

In general what I learned is that the lockfile will be included in the crate uploaded the [crates.io](http://crates.io/), even though it maybe is only used when running tests. So maybe all we need to do is make sure we update yanked dependencies before publishing?

PR #1793, opened 2025-07-02
fix: SelfRemove proposal type and free index calculation

### LINKED ISSUE #1789: Update SelfRemove Proposal Type
#1704 introduced the SelfRemove proposal. But the proposal type changed to 0xa by now.

kkohbrok:
- Takes SelfRemove proposals into account when computing leftmost free index. This is not particularly relevant at the moment, since SelfRemoves can't be given as input when creating an external commit. However, this is in preparation of a follow up.
- Aligns the proposal types of SelfRemove and AppAck with the extensions draft 06.

Fixes #1789.

### DIFF
--- a/openmls/src/messages/proposals.rs
+++ b/openmls/src/messages/proposals.rs
@@ -162,8 +162,8 @@ impl From<u16> for ProposalType {
             5 => ProposalType::Reinit,
             6 => ProposalType::ExternalInit,
             7 => ProposalType::GroupContextExtensions,
-            8 => ProposalType::AppAck,
-            0x000c => ProposalType::SelfRemove,
+            0x000a => ProposalType::SelfRemove,
+            0x000b => ProposalType::AppAck,
             other => ProposalType::Custom(other),
         }
     }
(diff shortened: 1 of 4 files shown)


wysiwys:
@kkohbrok I wanted to add a comment here since I was just looking at this issue myself :)

Testing-wise, after updating the `ProposalType` values (`SelfRemove` -> 0x000a and `AppAck` -> 0x000b), it looks like the Franken testing utils also need to be updated to match, in particular the `FrankenProposalType` (which also maps proposal types to `u16` values). 

However, one other thing that is relevant there is that the Franken utils don't implement SelfRemove proposals yet. I'm guessing it might also be good to open an issue for that instead and implement it in another PR

openmls/src/messages/proposals.rs
```diff
@@ -162,8 +162,8 @@ impl From<u16> for ProposalType {
             5 => ProposalType::Reinit,
             6 => ProposalType::ExternalInit,
             7 => ProposalType::GroupContextExtensions,
-            8 => ProposalType::AppAck,
-            0x000c => ProposalType::SelfRemove,
+            0x000a => ProposalType::SelfRemove,
+            0x000b => ProposalType::AppAck,
```
wysiwys, line 167: In this file, in the Extensions table in the comments on lines 68-70, these `ProposalType` values would need to be updated, too

kkohbrok: Thanks for pointing that out! I'll update those as well.

kkohbrok:
Testing doesn't work just yet, because we can't add (SelfRemove) proposals when doing an external commit. I'm working on that separately. If you prefer, I can hold off with the change to the index computation and merge that together with the follow-up.

franziskuskiefer (dismissed):
The fix itself looks good to me. It would just be nice to have a test with it that we know that it's good now.

Let's sync on the order also with #1796 and the release #1769.

PR #1797, opened 2025-07-03
fix: trim diff tree after processing all proposals

### LINKED ISSUE #1794: [Bug] Incorrect state after committing remove proposals and add proposals in a single commit
**Describe the bug**

When creating a commit, during proposal processing,
the tree shrink function is called eagerly after each
remove proposal. If the shrink threshold was reached, the blanked out
nodes are removed from the diff tree.

If there are additional add proposals in a commit that trigger the
tree to grow again, this results in incorrect state if there are FEWER
add proposals than remove proposals. That's because the blanked out nodes
removed during shrinking are not re-added during growing.

**To Reproduce**
1. Start with a group size of 6. 
2. Create a commit with 2 remove proposals and 1 add proposal

**Expected behavior**

2 remove proposals and 1 add proposal are committed -> the group size should have changed from 6 members to 5 members

**Actual behavior**

The group has 6 members, one of the remove proposals is ignored.

**Suggested Fix**

I'd like submit a PR including a test demonstrating the behavior and a fix with the following approach:
Don't shrink/trim the tree eagerly after each remove proposal, but after remove and add proposals have been processed. In my experiments this fixes the issue, and as far as I can tell, this should be compatible with the protocol. What do you think?

SimonThormeyer:
Fixes #1794 

Checkout 2bef38c396bc8b6ef704677e4f568bf504f374c0 to see that the test is failing on current main.

### DIFF
--- a/openmls/src/group/public_group/diff/apply_proposals.rs
+++ b/openmls/src/group/public_group/diff/apply_proposals.rs
@@ -145,6 +145,8 @@ impl PublicGroupDiff<'_> {
             invitation_list.push((leaf_index, add_proposal.clone()))
         }
 
+        self.diff.trim_tree();
+
         // Process PSK proposals
         let presharedkeys: Vec<PreSharedKeyId> = proposal_queue
             .filtered_by_type(ProposalType::PreSharedKey)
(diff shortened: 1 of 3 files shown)


SimonThormeyer:
Thanks, looking forward to seeing it merged!

PR #1801, opened 2025-07-08
External commit builder

kkohbrok:
- Adds an `ExternalCommitBuilder` struct that can be created using `MlsGroup::external_commit_builder()`
- The ExternalCommitBuilder allows the creation of external commits using the builder pattern, in particular enabling the inclusion of by-reference SelfRemove proposals and by-value PSK proposals
- Adds a test that makes sure the index of the external joiner is computed correctly in the presence of SelfRemoves and that PSK proposals are included correctly

### DIFF
--- a/openmls/src/group/mls_group/commit_builder/external_commits.rs
+++ b/openmls/src/group/mls_group/commit_builder/external_commits.rs
@@ -70,7 +70,7 @@ pub struct ExternalCommitBuilder {
     proposals: Vec<PublicMessageIn>,
     ratchet_tree: Option<RatchetTreeIn>,
     config: MlsGroupJoinConfig,
-    aad: Option<Vec<u8>>,
+    aad: Vec<u8>,
 }
 
 impl MlsGroup {
@@ -114,7 +114,7 @@ impl ExternalCommitBuilder {
     /// Specifies additional authenticated data (AAD) to be included in the
     /// external commit.
     pub fn with_aad(mut self, aad: Vec<u8>) -> Self {
-        self.aad = Some(aad);
+        self.aad = aad;
         self
     }
 
@@ -282,8 +282,6 @@ impl ExternalCommitBuilder {
         }
 
         let mut commit_builder = CommitBuilder::<'_, Initial, MlsGroup>::new(mls_group);
-
-        let aad = aad.unwrap_or_default();
 
         commit_builder.stage.force_self_update = true;
         commit_builder.stage.external_commit_info = Some(ExternalCommitInfo {


openmls/src/framing/validation.rs
```diff
@@ -206,7 +206,11 @@ impl DecryptedMessage {
 #[derive(Debug, Clone)]
 pub(crate) enum SenderContext {
     Member((GroupId, LeafNodeIndex)),
-    ExternalCommit((GroupId, LeafNodeIndex)),
+    ExternalCommit {
```
franziskuskiefer, line 209: Would it make sense to these given that one got bigger? No strong opinion though.

openmls/src/group/mls_group/commit_builder/external_commits.rs
```diff
@@ -0,0 +1,371 @@
...
+    #[error("No external_pub extension available to join group by external commit.")]
+    MissingExternalPub,
+    /// We don't support the ciphersuite of the group we are trying to join.
+    #[error("We don't support the ciphersuite of the group we are trying to join.")]
+    UnsupportedCiphersuite,
+    /// This error indicates the public tree is invalid. See
+    /// [`CreationFromExternalError`] for more details.
+    #[error(transparent)]
+    PublicGroupError(#[from] CreationFromExternalError<StorageError>),
+    /// An erorr occurred when writing group to storage
```
franziskuskiefer:
```suggestion
    /// An error occurred when writing group to storage
```

openmls/src/group/mls_group/commit_builder/external_commits.rs
```diff
@@ -0,0 +1,371 @@
...
+/// This is the builder for external commits. It allows you to build an external
+/// commit that can be used to join a group externally. Parameters such as
+/// optional SelfRemove proposals from other members, the ratchet tree, and the
+/// group join configuration can be set in the first builder stage.
+///
+/// The second stage of this builder is a [`CommitBuilder`] that can be used to
+/// add one or more [`PreSharedKeyProposal`]s to the external commit and specify
+/// [`LeafNodeParameters`].
+#[derive(Default)]
+pub struct ExternalCommitBuilder {
```
franziskuskiefer, line 69: It looks like this is not exposed publicly, hence missing in the docs. Can you make sure that everything public is exposed in some way?

openmls/src/group/mls_group/creation.rs
```diff
@@ -100,127 +98,33 @@ impl MlsGroup {
         credential_with_key: CredentialWithKey,
     ) -> Result<(Self, MlsMessageOut, Option<GroupInfo>), ExternalCommitError<Provider::StorageError>>
```
franziskuskiefer, line 103: I deprecated this function in #1810. I'd suggest we do the same here? Or we do it in #1810.

openmls/src/group/mls_group/commit_builder/external_commits.rs
```diff
@@ -0,0 +1,371 @@
...
+    proposals: Vec<PublicMessageIn>,
```
franziskuskiefer, line 70: Do we need to own all these values? We probably only need to own the proposals and the ratchet tree.

openmls/src/group/tests_and_kats/tests/external_commit_builder.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    // Make sure we support SelfRemoves
+    let capabilities = Capabilities::builder()
+        .proposals(vec![ProposalType::SelfRemove])
+        .build();
+
+    // Since SelfRemoves and PSK proposals need to be sent as public
+    // messages if we want to use them with an external commit, we need to
+    // set the wire format policy to PURE_PLAINTEXT_WIRE_FORMAT
+    const POLICY: WireFormatPolicy = PURE_PLAINTEXT_WIRE_FORMAT_POLICY;
```
franziskuskiefer, line 47: So we can't send self removes if the policy is different? That wouldn't be good, we should allow sending self removes and psks and don't apply the policy. This is similar to other messages, like application messages, which have a mandatory wire format.

openmls/src/messages/mod.rs
```diff
@@ -231,28 +231,39 @@ impl CommitIn {
...
+                                    .then_some(self_remove.sender)
+                            })
+                        });
+
+                    let new_leaf_index = [leftmost_blank_index]
+                        .into_iter()
+                        .chain(former_sender_index)
+                        .chain(self_removed_indices)
+                        .min()
+                        .expect("We know the iterator has at least one element.");
```
franziskuskiefer: We return a result anyway. I'd rather return an error than panic here.

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -252,17 +250,60 @@ impl<'a> CommitBuilder<'a, Initial> {
         self
     }
 
+    /// Adds a proposal to the proposals to be committed.
+    pub fn add_proposal(mut self, proposal: Proposal) -> Self {
```
franziskuskiefer, line 257: Maybe add that there's `add_proposals` as well to add multiple in one go? Taking an iterator there looks a little odd though. I guess that's how we'd use it internally. But is this what users should use? Wouldn't they rather have a slice `&[Proposals]`?

openmls/src/framing/validation.rs
kkohbrok: I think there's a verb missing in the first sentence. Did you mean box?

openmls/src/group/mls_group/commit_builder.rs
kkohbrok: I didn't come up with this API. The same function (and the iterator-based variant) existed previously at line 210.

openmls/src/group/mls_group/commit_builder/external_commits.rs
kkohbrok: Thanks for pointing this out, I've explicitly exported it as part of the `prelude`.

openmls/src/group/mls_group/commit_builder/external_commits.rs
kkohbrok: Good point, but other than the AAD we'd have to clone everyting eventually anyway, so we can at least give the caller a chance to prevent the clone.

openmls/src/group/mls_group/creation.rs
kkohbrok: No need. Added a deprecation note.

openmls/src/group/tests_and_kats/tests/external_commit_builder.rs
kkohbrok: This isn't something I added in this PR, so I'd rather file an issue. 

It's a bit different than application messages, though, because this is mandatory plaintext, not mandatory encryption. If an application configures a group with `PURE_CIPHERTEXT_WIRE_FORMAT_POLICY` the assumption is that every output is safely encrypted. I don't have a particularly strong opinion here, though, so we could also just add some strong language in the docs.

openmls/src/group/mls_group/commit_builder/external_commits.rs
```diff
@@ -70,7 +70,7 @@ pub struct ExternalCommitBuilder {
     proposals: Vec<PublicMessageIn>,
     ratchet_tree: Option<RatchetTreeIn>,
     config: MlsGroupJoinConfig,
-    aad: Vec<u8>,
+    aad: Option<Vec<u8>>,
```
franziskuskiefer: Or just an empty vector instead of an option?

openmls/src/framing/validation.rs
franziskuskiefer: Yes, I meant box 🙈

openmls/src/group/mls_group/commit_builder/external_commits.rs
franziskuskiefer: Looks like not owning is breaking some patterns and needs a more comprehensive change to the commit builder. Let's not do it for now.

openmls/src/group/tests_and_kats/tests/external_commit_builder.rs
franziskuskiefer: Yeah it needs to be clearly documented. But then we should allow it. I opened #1818

franziskuskiefer (approved):
Thanks

openmls/src/framing/validation.rs
kkohbrok: I don't think it makes that much of a difference. I'll leave it as is.

franziskuskiefer (approved):

PR #1805, opened 2025-07-16
Bump openmls and openmls_test to -pre.2

keks:
These two had the issue with the yanked version in the lockfile. Bump to -pre.2 for another attempt.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,10 +21,10 @@ resolver = "2"
 [workspace.dependencies]
 tls_codec = { version = "0.4.2", features = ["derive", "serde", "mls"] }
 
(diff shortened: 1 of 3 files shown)


keks:
Huh, this seems fishy but unrelated. Does this ring a bell, @kkohbrok?

https://github.com/openmls/openmls/actions/runs/16320187796/job/46095672055?pr=1805#step:6:2182

I think it's related to this code: https://github.com/openmls/openmls/blob/69729d572160038b656d2b391a6921fb9e7fd5af/openmls/src/binary_tree/tests.rs#L38

Anyway, I am going to merge this as is, because it's unrelated.

PR #1807, opened 2025-07-17
feat: set migration table name for refinery migrations

kkohbrok:
The sqlite storage provider uses refinery to manage migrations, i.e. creation of tables for storage provider data. So far, the migrations in the provider use the general migrations table used by refinery, which means that it doesn't compose well with external apps that want to use the same DB for other data (and have their own migrations). This PR mitigates that by storing OpenMLS-specific migrations in a separate, named table.


franziskuskiefer (approved):
lgtm

PR #1814, opened 2025-07-30
Allow configuration of leaf node lifetime checks on join

### LINKED ISSUE #1810: Lifetime verification in leaf nodes is stricter than the RFC
When creating a group from a Welcome, OpenMLS is [strict on the lifetimes of leaf nodes](https://github.com/openmls/openmls/blob/main/openmls/src/group/public_group/validation.rs#L663) in the ratchet tree.

The RFC is more relaxed about this issue:

> it is RECOMMENDED that the client verifies that the current time is within the range of the lifetime field. (This check is not mandatory because the LeafNode might have expired in the time between when the message was sent and when it was received.)

This can lead to issues where group members who have not updated their keys since joining will eventually put the group into a "bricked" state for new members until the offending leafs are removed. That removal and validation should be the choice of the application and not a hard requirement from OpenMLS

franziskuskiefer:
Allow configuration of leaf node validation when joining a group.
To this end this changes the `join_by_external_commit` into a builder,
and introduces new functions `MlsGroup::from_external_commit` and
`StagedWelcome::build_from_welcome` to make joining of a group more
ergonomic with the new option. As part of this,
`MlsGroup::join_by_external_commit` is deprecated.

This also renames `PublicGroup::from_external_internal` to
`from_ratchet_tree` and introduces `Lifetime::init` to set lifetime
values explicitly.

Fixes #1810

### DIFF
--- a/openmls/src/group/mls_group/creation.rs
+++ b/openmls/src/group/mls_group/creation.rs
@@ -336,7 +336,7 @@ impl ProcessedWelcome {
         &self.group_secrets.psks
     }
 
-    /// Consume the `ProcessedWelcome` and combine it witht he ratchet tree into
+    /// Consume the `ProcessedWelcome` and combine it with the ratchet tree into
     /// a `StagedWelcome`.
     pub fn into_staged_welcome<Provider: OpenMlsProvider>(
         mut self,
--- a/openmls/src/group/mls_group/creation.rs
+++ b/openmls/src/group/mls_group/creation.rs
@@ -1,7 +1,9 @@
 use std::iter;
 
 use errors::NewGroupError;
-use openmls_traits::{signatures::Signer, storage::StorageProvider as StorageProviderTrait};
+use openmls_traits::{
+    signatures::Signer as SignerTrait, storage::StorageProvider as StorageProviderTrait,
+};
 
 use super::{builder::MlsGroupBuilder, *};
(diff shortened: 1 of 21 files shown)


openmls/src/group/mls_group/creation.rs
```diff
@@ -507,3 +536,69 @@ fn keys_for_welcome<Provider: OpenMlsProvider>(
...
+        self
+    }
+
+    /// Skip the validation of lifetimes in leaf nodes in the ratchet tree.
+    ///
+    /// Note that only the leaf nodes are checked that were never updated.
+    pub fn skip_lifetime_validation(mut self) -> Self {
+        self.validate_lifetimes = LeafNodeLifetimePolicy::Skip;
+        self
+    }
```
kkohbrok, line 588: The ExternalCommitBuilder and the JoinBuilder have different interfaces w.r.t. lifetime validation policy. Maybe it would be worth aligning the two?

kkohbrok (approved):
Looks good! Just one nit.

openmls/src/group/mls_group/creation.rs
franziskuskiefer: Yes, good point. I think I changed my mind on how this should look like while doing it. I'll align them to this one.

PR #1825, opened 2025-08-15
adding the hmac function to OpenMlsCrypto

### LINKED ISSUE #1375: [Bug] MACs are computed using HKDF extract
MACs are currently computed using HKDF extract. That works, but only because all HPKE ciphersuites use HKDF as their key derivation function and MLS hard-codes HMAC as the MAC to use. The crypto provider should expose a separate HMAC functionality s.t. we can use it to compute MACs.

tbehner:
Coses #1375.

Since `OpenMlsCrypto` is also used by `Mac` and `Secret` we also need to add the respective methods here.

Both the rust_crypto and libcrux crypto providers have been updated to implement the new trait. Although `hkdf_expand` could have been used, we opted for the explicit `hmac` implementations provided by both crates for clarity and consistency.

A separate module was created for the HMAC implementation in rust_crypto to resolve import conflicts (multiple applicable methods in the same scope).

The test_utils have also been updated to use the new hmac function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `StagedWelcome::build_from_welcome`: Alternative to `new_from_welcome` in a builder style that allows disabling lifetime validation of the incoming ratchet tree.
   - `Lifetime::init`: Set explicit lifetimes for a key package.
 - [#1801](https://github.com/openmls/openmls/pull/1801): Added `MlsGroup::external_commit_builder`.
-- [#1825](https://github.com/openmls/openmls/pull/1825): Add the `hmac` method for hashing to the `OpenMlsCrypto` trait.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `StagedWelcome::build_from_welcome`: Alternative to `new_from_welcome` in a builder style that allows disabling lifetime validation of the incoming ratchet tree.
   - `Lifetime::init`: Set explicit lifetimes for a key package.
 - [#1801](https://github.com/openmls/openmls/pull/1801): Added `MlsGroup::external_commit_builder`.
+- [#1825](https://github.com/openmls/openmls/pull/1825): Add the `hmac` method for hashing to the `OpenMlsCrypto` trait.
 
 ### Fixed
 
--- a/libcrux_crypto/src/crypto.rs
+++ b/libcrux_crypto/src/crypto.rs
@@ -395,6 +395,7 @@ fn hkdf_alg(hash_type: HashType) -> libcrux_hkdf::Algorithm {
         HashType::Sha2_512 => libcrux_hkdf::Algorithm::Sha512,
     }
 }
+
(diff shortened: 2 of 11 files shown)


libcrux_crypto/src/crypto.rs
```diff
@@ -384,6 +395,13 @@ fn hkdf_alg(hash_type: HashType) -> libcrux_hkdf::Algorithm {
         HashType::Sha2_512 => libcrux_hkdf::Algorithm::Sha512,
     }
 }
```
franziskuskiefer, line 397: Let's add some new lines above and below to make this a little more readable.

traits/src/crypto.rs
```diff
@@ -28,6 +28,13 @@ pub trait OpenMlsCrypto: Send + Sync {
         ikm: &[u8],
     ) -> Result<SecretVLBytes, CryptoError>;
 
+    fn hmac(
```
franziskuskiefer, line 31: This is a breaking change and should get and entry in the changelog.

openmls_rust_crypto/src/hmac.rs
```diff
@@ -0,0 +1,30 @@
...
+
+type HmacSha256 = hmac::Hmac<Sha256>;
+type HmacSha384 = hmac::Hmac<Sha384>;
+type HmacSha512 = hmac::Hmac<Sha512>;
+
+macro_rules! hmac_digest {
+    ($hash_t:ty, $key:ident, $message:ident) => {{
+        let mut hmac = <$hash_t>::new_from_slice($key).map_err(|_e| CryptoError::InvalidLength)?;
+        hmac.update($message);
+        hmac.finalize().into_bytes().into_iter().collect()
```
franziskuskiefer: Collecting here seems a little inefficient (unless the compiler is clever here). Why not just `.as_slice().to_vec()`?

tbehner: Yes, absolutely!

traits/src/crypto.rs
tbehner: I added a respective line to the changelog.

openmls/src/ciphersuite/secret.rs
```diff
@@ -114,7 +135,7 @@ impl Secret {
                 ciphersuite.hash_algorithm(),
                 self.value.as_slice(),
                 info,
-                okm_len,
+        okm_len,
```
franziskuskiefer: Looks like rustfmt is needed.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `StagedWelcome::build_from_welcome`: Alternative to `new_from_welcome` in a builder style that allows disabling lifetime validation of the incoming ratchet tree.
   - `Lifetime::init`: Set explicit lifetimes for a key package.
 - [#1801](https://github.com/openmls/openmls/pull/1801): Added `MlsGroup::external_commit_builder`.
+- [#1825](https://github.com/openmls/openmls/pull/1825): Add the `hmac` method for hashing to the `OpenMlsCrypto` trait.
```
franziskuskiefer: The line lgtm, but I should've been more specific. The traits crate has it's own changelog. Can you move this line over there?

tbehner: My mistake, I could have looked in the package before adding the line.

PR #1826, opened 2025-08-19
Attaching Application Data to MLS Messages (Section 4.6)

### LINKED ISSUE #1821: Attaching Application Data to MLS Messages (Section 4.6)
Approach for implementing the basic functionality of [Section 4.6 of the extensions draft](https://datatracker.ietf.org/doc/html/draft-ietf-mls-extensions#section-4.6):

- Add a variant to the [Extension](https://latest.openmls.tech/doc/openmls/extensions/enum.Extension.html) enum representing the `AppDataDictionary` extension
- An `AppDataDictionary` struct containing the component_data entries (e.g. as `Vec<ComponentData>`)
- A `ComponentData` struct containing the component_id and the specific data as `Vec<u8>`
- Include the `AppDataDictionary` type in the [ExtensionType](https://docs.rs/openmls/latest/openmls/extensions/enum.ExtensionType.html) enum

Entries in the `AppDataDictionary` must be sorted by component_id. There must also be at most one entry for each component_id.

`KeyPackage`s and `LeafNode`s can be created with this extension using existing functionality. To allow creating a `GroupInfo` with the extension, we'll also need to provide functionality to add extensions to the `GroupInfo`, e.g. on the `CommitBuilder`.

wysiwys:
Resolves #1821

### DIFF
--- a/openmls/src/extensions/application_data_dict_extension.rs
+++ b/openmls/src/extensions/application_data_dict_extension.rs
@@ -2,6 +2,7 @@ use super::{Deserialize, Serialize};
 use tls_codec::{TlsDeserialize, TlsDeserializeBytes, TlsSerialize, TlsSize, VLBytes};
 
 // TODO: use newtype or type alias defined elsewhere?
+/// The unique ComponentId.
 pub type ComponentId = u32;
 
 /// An entry in the [`AppDataDictionary`].
@@ -50,6 +51,10 @@ impl<Data: Into<VLBytes>> From<BTreeMap<ComponentId, Data>> for AppDataDictionar
 }
 
 /// App data dictionary in the [`AppDataDictionaryExtension`].
+///
+/// This struct contains a list of [`ComponentData`] entries.
+/// Entries are in order, and there is at most one entry per [`ComponentId`].
+/// These properties are checked upon creation, as well as upon deserialization.
 #[derive(
     PartialEq,
     Eq,
@@ -75,7 +80,7 @@ enum AppDataDictionaryError {
 }
 
 impl AppDataDictionary {
-    /// Returns an [`AppDataDictionaryBuilder`].
+    /// Returns a builder for this type.
     pub fn builder() -> AppDataDictionaryBuilder {
         AppDataDictionaryBuilder::new()
     }
@@ -101,10 +106,9 @@ impl AppDataDictionary {
     /// Creates an [`AppDataDictionary`] from a Vec of [`ComponentData`] entries.
     ///
     /// Ensures that the list is ordered by [`ComponentId`], and that there is at most one entry per [`ComponentId`].
-    // <https://datatracker.ietf.org/doc/html/draft-ietf-mls-extensions#section-4.6-5>
+    /// <https://datatracker.ietf.org/doc/html/draft-ietf-mls-extensions#section-4.6-5>
     fn try_from_vec(data: Vec<ComponentData>) -> Result<Self, AppDataDictionaryError> {
-        // Use an ordered map of processed ComponentIds to check conditions
-        // See
+        // Use an ordered set of processed ComponentIds to check conditions
         let mut seen = std::collections::BTreeSet::<ComponentId>::new();
 
         for ComponentData { component_id, .. } in data.iter() {
@@ -131,6 +135,10 @@ impl AppDataDictionary {
 }
 
 impl tls_codec::Deserialize for AppDataDictionary {
+    /// Deserialize from bytes.
+    ///
+    /// This function also ensures that the [`ComponentData`] entries are in order by
+    /// [`ComponentId`], and there is at most one entry per [`ComponentId`].
     fn tls_deserialize<R: std::io::Read>(bytes: &mut R) -> Result<Self, tls_codec::Error> {
         let vec = Vec::<ComponentData>::tls_deserialize(bytes)?;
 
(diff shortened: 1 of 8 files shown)


openmls/src/group/mls_group/exporting.rs
```diff
@@ -83,16 +97,21 @@ impl MlsGroup {
...
             } else {
-                Extensions::single(external_pub_extension()?)
-            }
+                vec![external_pub_extension()?]
+            };
+
+            extensions.extend(
+                additional_extensions
+                    .into_iter()
+                    .filter(|extension| extension.extension_type() != ExtensionType::RatchetTree),
```
keks: Are additional ExternalPubExtensions allowed here, or should they also be filtered out? I wonder if we should return an error if there is an illegal extension type here instead of silently dropping it.

openmls/src/extensions/application_data_dict_extension.rs
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    /// Consumes the struct and returns its component parts.
+    pub fn into_parts(self) -> (ComponentId, VLBytes) {
+        (self.component_id, self.data)
+    }
+}
+
+/// Build an [`AppDataDictionary`] from a [`BTreeMap`].
+impl<Data: Into<VLBytes>> From<BTreeMap<ComponentId, Data>> for AppDataDictionary {
```
keks, line 44: Could this be a FromIterator<Item=(ComponentId, Data)> instead? BTreeMap should implement that.

openmls/src/extensions/application_data_dict_extension.rs
```diff
@@ -0,0 +1,328 @@
...
+    /// Build an [`AppDataDictionary`] from a [`BTreeMap`].
+    pub fn from_btree_map(map: BTreeMap<ComponentId, impl Into<VLBytes>>) -> Self {
+        Self::from(map)
+    }
+
+    /// Creates an [`AppDataDictionary`] from a Vec of [`ComponentData`] entries.
+    ///
+    /// Ensures that the list is ordered by [`ComponentId`], and that there is at most one entry per [`ComponentId`].
+    /// <https://datatracker.ietf.org/doc/html/draft-ietf-mls-extensions#section-4.6-5>
+    fn try_from_vec(data: Vec<ComponentData>) -> Result<Self, BuildAppDataDictionaryError> {
```
keks: This could take an IntoIter<Item= ComponentData> instead

openmls/src/extensions/application_data_dict_extension.rs
```diff
@@ -0,0 +1,328 @@
...
+        // Use an ordered set of processed ComponentIds to check conditions
+        let mut seen = std::collections::BTreeSet::<ComponentId>::new();
```
keks: I think so far we mostly used HashMap/HashSet instead of the BTree constructions, but I am not sure using these would be a problem.

openmls/src/extensions/application_data_dict_extension.rs
```diff
@@ -0,0 +1,328 @@
...
+                return Err(BuildAppDataDictionaryError::DuplicateEntries);
+            }
+
+            // Check the ordering
+            // The component id must be greater than all previous component ids
+            if let Some(max) = seen.last() {
+                if max > component_id {
+                    return Err(BuildAppDataDictionaryError::EntriesNotInOrder);
+                }
+            }
```
keks: I am not sure we need to do a set lookup every time. Wouldn't it suffice to have a mutable `prev: Option<&ComponentId>` as a local variable, and use that to ensure monotonicity? Or am I missing something?

keks: I guess a completely different approach would be to just sort the Vec, not sure if that would break anything.

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -343,6 +352,39 @@ impl<'a, G: BorrowMut<MlsGroup>> CommitBuilder<'a, Initial, G> {
...
+    ) -> Self {
+        self.stage.group_info_config.create_group_info = true;
+        self.stage.group_info_config.other_extensions = extensions
+            .into_iter()
+            .filter(|extension| extension.extension_type() != ExtensionType::RatchetTree)
+            .collect();
+
+        self
+    }
+
```
keks: Is there a reason these setting were moved from the early phase into the late phase? Does this somehow depend on the PSKs being loaded?

openmls/src/extensions/application_data_dict_extension.rs
wysiwys: Ah, yeah, the set lookup isn't needed here - I replaced this monotonicity check with a `prev` variable :+1:  Thanks!

openmls/src/extensions/application_data_dict_extension.rs
wysiwys: The reason it's for only a BTreeMap now is that `BTreeMap::into_iter()` preserves the properties that need to hold for the `ComponentId`s (in order and no duplicate keys). The `FromIterator` wouldn't ensure those properties, but something like a `TryFrom` could work

openmls/src/group/mls_group/commit_builder.rs
wysiwys: The main reason is that this stage is the stage right before `build()` is called. This is to avoid the need to move this config between all intermediate stages, including the fork resolution stages. However, it doesn't depend on the PSKs being loaded, so I'd agree that it doesn't really fit in this stage from a naming perspective :)

Do you think it would be better to pass the config between stages? Or maybe store the config outside of the stage, as a member of the commit builder?

openmls/src/group/mls_group/exporting.rs
wysiwys: Ah, yeah - errors there would be more informative. I added error handling to express this :+1:

franziskuskiefer:
Converting this to a draft for now to get it out of the review queue. Let's add the uses and compile draft feedback and then get back to this.

PR #1831, opened 2025-08-21
Update validation for extension types of leaf node extensions

wysiwys:
For these functions where the leaf node extension types are validated, this PR updates the validation to use the same [list of allowed types](https://www.rfc-editor.org/rfc/rfc9420.html#section-17.3-4) as in the RFC:
- `MlsGroupCreateConfig::with_leaf_node_extensions()` 
- `group::public_group::TempBuilderPG1::with_leaf_node_extensions()`

The extension types are also validated in `KeyPackageBuilder::leaf_node_extensions()` and `LeafNodeParametersBuilder::with_extensions()`. 
- Error handling is added to these methods, to return an error if an extension with an invalid type is provided. 

This PR also adds tests for this functionality.

### DIFF
--- a/openmls/src/treesync/node/leaf_node.rs
+++ b/openmls/src/treesync/node/leaf_node.rs
@@ -485,10 +485,13 @@ impl LeafNode {
     /// - the type of the credential is covered by the capabilities
     pub(crate) fn validate_locally(&self) -> Result<(), LeafNodeValidationError> {
         // Check that no extension is invalid when used in leaf nodes.
-        // NOTE: This check is conducted manually, instead of using
-        // Extensions::validate_extension_types_for_leaf_node(),
-        // in order to collect the invalid extension types for a log message.
         // https://validation.openmls.tech/#valn1601
+        // NOTE: This check is conducted manually for now, instead of using the method
+        // Extensions::validate_extension_types_for_leaf_node(),
+        // in order to collect the invalid extension types for the log message below.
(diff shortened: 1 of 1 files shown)


openmls/src/treesync/node/leaf_node.rs
```diff
@@ -477,6 +484,8 @@ impl LeafNode {
     /// - the type of the credential is covered by the capabilities
     pub(crate) fn validate_locally(&self) -> Result<(), LeafNodeValidationError> {
         // Check that no extension is invalid when used in leaf nodes.
+        // NOTE: This check is conducted manually, instead of using
+        // Extensions::validate_extension_types_for_leaf_node().
         let invalid_extension_types = self
             .extensions()
             .iter()
```
keks: Can you explain why here?

wysiwys: I added a note to the comments with an explanation of the reason, which is that the list of all invalid `ExtensionType` is collected here in order to be output in a log message, but `Extensions::validate_extension_types_for_leaf_node()` does not return any information about the invalid types, so the list is built here, instead. 

One possible solution could be to return the invalid `ExtensionType`s in the `LeafNodeValidationError`, and then to include them in the log message, instead, or to collect the list separately after calling `Extensions::validate_extension_types_for_leaf_node()`.

keks: Thanks!

PR #1832, opened 2025-08-26
Improve Lifetime log message

### LINKED ISSUE #1815: Improve Lifetime error
The lifetime error should provide additional information about the leaf node that is invalid. Providing the credential in addition to the `Lifetime` value should allow the caller to identify the offending leaf node.

> Also, please return a clearer error. The current message on the receiver side is:
> `offending lifetime: Lifetime { not_before, not_after }`
> This isn’t tied to the receiver’s LeafNode, so it’s unclear why it occurs or which LeafNode caused it. 

 _Originally posted by @mchenani in [#1810](https://github.com/openmls/openmls/issues/1810#issuecomment-3114262426)_

wysiwys:
In the log message for an invalid lifetime, include the Credential, so that the relevant leaf node can be identified.

See #1815

### DIFF
--- a/openmls/src/group/public_group/validation.rs
+++ b/openmls/src/group/public_group/validation.rs
@@ -696,7 +696,10 @@ impl PublicGroup {
         {
             if let Some(lifetime) = leaf_node.life_time() {
                 if !lifetime.is_valid() {
-                    log::warn!("offending lifetime: {lifetime:?}");
+                    log::warn!(
+                        "offending lifetime: {lifetime:?} for leaf node with {credential:?}",
+                        credential = leaf_node.credential()
+                    );
                     return Err(LeafNodeValidationError::Lifetime(LifetimeError::NotCurrent));
                 }
             }


openmls/src/group/public_group/validation.rs
```diff
@@ -696,7 +696,10 @@ impl PublicGroup {
         {
             if let Some(lifetime) = leaf_node.life_time() {
                 if !lifetime.is_valid() {
-                    log::warn!("offending lifetime: {lifetime:?}");
+                    log::warn!(
```
franziskuskiefer, line 699: Let's maybe put some information into the error as well?

wysiwys: Ah, I had thought this referred to just the log message! I'll add something to the error, too :+1:

wysiwys: Should information to identify the leaf node be included more generally in the `LeafNodeValidationError`?

wysiwys: To allow identifying the offending leaf node for other validation errors, something like this could be possible:
```rust
+#[derive(Clone, Debug, Error, Eq, PartialEq)]
pub struct LeafNodeValidationError {
    /// The error that occurred during validation.
    pub content: LeafNodeValidationErrorContent,
    /// Optional credential of the offending leaf node.
    pub credential: Option<Box<Credential>>,
}

```
Replacing the current `LeafNodeValidationError` enum with this error type would allow the offending leaf node to be identified for all error variants (where a credential can be associated to the leaf node)

wysiwys: I'm not sure if this PR is the best place for this though (since it would be a more involved reworking of error types)

franziskuskiefer: Adding the credential to the error types in `LeafNodeValidationError` that are due to the credential makes sense. But that doesn't mean it should go onto the entire type.

wysiwys: Got it, then would it be best just to add the `Lifetime` to the `LifetimeError`? That wouldn't allow the node to be identified from the error type, but since the credential's in the log message, the leaf node can at least be identified from a debugging perspective

franziskuskiefer: I think the `Lifetime`, `UnsupportedCredentials`, `CredentialNotInCapabilities`, `LeafNodeCredentialNotSupportedByMember`, and `MemberCredentialNotSupportedByLeafNode` variants in there could benefit from the credential. But looking at the error type, I'm not sure that's the best way forward.
All of the leaf node errors lack information about where the error occurs. We should provide more information about the error, like the credential. But maybe just a pointer to the entire leaf node would be better, because the error is not always the credential.

But that indeed makes this much more involved. How about we merge this, just to get some logging in and design how to get more information onto that error type in general on the issue?

PR #1834, opened 2025-08-28
check extensions in GroupContextExtensionProposals for validity

tbehner:
Some tests are failing and have to be adapted.


openmls/src/extensions/mod.rs
```diff
@@ -130,6 +130,17 @@ impl ExtensionType {
...
+    pub(crate) fn is_valid_group_context_extension_proposal(self) -> Option<bool> {
+        match self {
+            ExtensionType::RequiredCapabilities | ExtensionType::ExternalSenders => Some(true),
+            ExtensionType::ApplicationId
+            | ExtensionType::RatchetTree
+            | ExtensionType::ExternalPub
+            | ExtensionType::LastResort => Some(false),
+            ExtensionType::Unknown(_) => Some(true),
+        }
+    }
```
wysiwys: One general question about these extension type validation functions is whether they should return `None` for Unknown extensions, or whether they should simply return `true`. For example, for leaf nodes, we currently always accept Unknown extensions. 

@franziskuskiefer Do you have thoughts on this?

openmls/src/extensions/mod.rs
```diff
@@ -130,6 +130,17 @@ impl ExtensionType {
             ExtensionType::Unknown(_) => None,
         }
     }
+
+    pub(crate) fn is_valid_group_context_extension_proposal(self) -> Option<bool> {
```
wysiwys: Should the names of these functions match each other, to convey that they serve similar purposes? For example, something like `is_valid_in_group_context()` could be chosen to match the function `is_valid_in_leaf_node()`.

tbehner: Good point :+1:

tbehner:
> One overarching point is that due to the added error handling, several functions now need to return `Result<T, E>`, instead of just `T`, because the `Extensions` list passed in needs to be validated inside the function. The main functionality affected here is proposal creation. It may be better to decouple the `ExtensionType` validation from the proposal creation, instead, by passing in an already-validated Extensions list first, which could look something like an `Extensions<GroupContext>`, or another type that indicates the extension types have been pre-validated.

I really like the idea! Not only does this decouple the validation from the creation, but also communicates the intend and requirements of the extensions more clearly. I will take a look into this.

openmls/src/extensions/mod.rs
franziskuskiefer: Yes, we always have to accept unknown extensions. They are usually application defined and OpenMLS shouldn't interfere. Really, that should be only in the private space. But I think we can be a little more lenient.

openmls/src/extensions/mod.rs
```diff
@@ -236,34 +236,104 @@ pub enum Extension {
...
+
+    /// An [`ExternalSendersExtension`]
+    ExternalSenders(ExternalSendersExtension),
+
+    /// A currently unknown extension.
+    Unknown(u16, UnknownExtension),
+}
+
+/// TypedExtension is implemented by all types that can be used as a Extension.
+pub trait TypedExtension:
```
tbehner: I am not happy with this traits name, but could not come up with something better.

openmls/tests/book_code_discard_welcome.rs
```diff
@@ -167,7 +167,8 @@ fn not_join_group() {
     // Inspect ciphersuite...
     let _ciphersuite = group_context.ciphersuite();
     // Inspect extensions...
-    let extensions: &Extensions = group_context.extensions();
+    // TODO TIM ich bin mir noch nicht sicher was ich hiermit mache
+    let extensions: Extensions<Extension> = group_context.extensions().clone().into();
```
franziskuskiefer: The idea is, as below, to get the extensions and look inside. There could be something in there the client doesn't like. Right now there may not be anything and we have to add something.

openmls/src/extensions/mod.rs
```diff
@@ -236,34 +236,104 @@ pub enum Extension {
...
+/// # GroupContextExtension
+///
+/// A group context extension is one of the [`GroupContextExtension`] enum values.
+/// The enum provides a set of common functionality for extensions that are valid in a group
+/// context.
+///
+/// See the individual extensions for more details on each extension.
+///
+/// ```c
+/// // draft-ietf-mls-protocol-16
```
franziskuskiefer: This is pre-RFC. We can drop these.

```suggestion
```

tbehner:
With the current approach, I followed the pattern of other parsed and validated proposals.
The parsed but not yet validated =GroupContextExtensionProposalIn= can be validated by calling =validate()=, which returns a =GroupContextExtensionProposal=.

To represent parsed and validated extensions, the =Extensions= type now takes a type parameter that implements =TypedExtension=. (See my comment about that name.)
Currently, this can be either =Extension= or =GroupContextExtension=. =GroupContextExtension= is the more specific extension type required in a =GroupContext=. An =Extension= can be cast to a =GroupContextExtension= by calling =try_into()=. Since not every type of =Extension= can be used in a =GroupContext=, this method might return an error.
As all methods expecting extensions valid for a =GroupContext= only accept =Extensions<GroupContextExtension>= as arguments, no further validation or checks are needed.

In a next step, we could implement more specific extension types for =GroupInfo=, =LeafNodes=, and =KeyPackages=.

openmls/src/extensions/mod.rs
```diff
@@ -517,6 +623,87 @@ impl Extension {
...
+        }
+    }
+}
+
+impl TryFrom<Extensions<Extension>> for Extensions<GroupContextExtension> {
+    type Error = InvalidExtensionError;
+
+    fn try_from(value: Extensions<Extension>) -> Result<Self, Self::Error> {
+        let group_extensions: Result<Vec<GroupContextExtension>, Self::Error> =
+            value.iter().map(|e| e.clone().try_into()).collect();
```
franziskuskiefer: No need to clone here.
```suggestion
            value.unique.into_iter().map(|e| e.try_into()).collect();
```

openmls/src/extensions/mod.rs
```diff
@@ -370,13 +440,49 @@ impl TryFrom<Vec<Extension>> for Extensions {
...
-    fn find_by_type(&self, extension_type: ExtensionType) -> Option<&Extension> {
+impl<T: TypedExtension> Extensions<T> {
+    fn find_by_type(&self, extension_type: ExtensionType) -> Option<&T> {
         self.unique
             .iter()
             .find(|ext| ext.extension_type() == extension_type)
     }
+}
+
+impl Extensions<GroupContextExtension> {
```
franziskuskiefer: Hm, I don't like this. As long as we have only one that may be fine, but if we have all of the extensions in separate types, this will blow up. Any ideas for something better, without macros?

openmls/src/extensions/mod.rs
```diff
@@ -517,6 +623,87 @@ impl Extension {
     }
 }
 
+impl GroupContextExtension {
```
franziskuskiefer: Same here, this is duplicating a lot of code for any extension that can show up in multiple places.

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -243,13 +243,16 @@ impl<'a> CommitBuilder<'a, Initial, &mut MlsGroup> {
 
     /// Adds a GroupContextExtensions proposal for the provided [`Extensions`] to the list of
     /// proposals to be committed.
-    pub fn propose_group_context_extensions(mut self, extensions: Extensions) -> Self {
+    pub fn propose_group_context_extensions(
+        mut self,
+        extensions: Extensions<Extension>,
```
franziskuskiefer: The question here for the API is whether the caller should use the correct extension type already, or if we do this in here and then return an error.

openmls/src/group/mls_group/staged_commit.rs
```diff
@@ -512,29 +512,29 @@ impl StagedCommit {
...
+                        //         .into_iter()
+                        //     })
+                        //     // TODO: ideally we wouldn't collect in between here, but the match arms
+                        //     //       have to all return the same type. We solve this by having them all
+                        //     //       be vec::IntoIter, but it would be nice if we just didn't have to
+                        //     //       do this.
+                        //     //       It might be possible to solve this by letting all match arms
+                        //     //       evaluate to a dyn Iterator.
+                        //     .collect::<Vec<_>>()
+                        //     .into_iter(),
```
franziskuskiefer: This should go back in?

openmls/src/group/public_group/staged_commit.rs
```diff
@@ -294,6 +294,7 @@ impl PublicGroup {
             return Err(StageCommitError::RequiredPathNotFound);
         };
 
+        // TODO we should probably not unwrap here
```
franziskuskiefer: Obsolete comment?

openmls/src/messages/proposals_in.rs
```diff
@@ -395,3 +432,30 @@ impl From<crate::messages::proposals::ProposalOrRef> for ProposalOrRefIn {
...
+)]
+pub struct GroupContextExtensionProposalIn {
+    extensions_tbv: Extensions<Extension>,
+}
+
+impl GroupContextExtensionProposalIn {
+    pub(crate) fn validate(self) -> Result<GroupContextExtensionProposal, ValidationError> {
+        let group_context_extensions = self
+            .extensions_tbv
+            .clone()
```
franziskuskiefer: No need to clone here.

openmls/src/messages/proposals_in.rs
```diff
@@ -395,3 +432,30 @@ impl From<crate::messages::proposals::ProposalOrRef> for ProposalOrRefIn {
...
+            .try_into()
+            .map_err(|_| ValidationError::InvalidExtension)?;
```
franziskuskiefer: Giving some more information here which extension was bad may be nice?

franziskuskiefer (commented):
some preliminary comments

openmls/src/extensions/mod.rs
wysiwys: To add on to this comment, one possible option for avoiding code duplication could be:
```rust
#[derive(Default, Debug, Clone, PartialEq, Eq, Serialize, Deserialize, TlsSize)]
pub struct ExtensionsForObject<T> {
    unique: Vec<Extension>,
    #[serde(skip)]
    _object: core::marker::PhantomData<T>,
}

#[derive(Clone, Copy, PartialEq, Eq, Debug, Default)]
/// Any extensions
pub struct AnyObject;
/// Type alias for AnyObject extensions
pub type Extensions = ExtensionsForObject<AnyObject>;
```
This approach (or a similar one) would mean that the plain `Extensions` can still be called `Extensions`, so that not so much code needs to be replaced immediately, and we only specify a validated version when needed.

openmls/src/extensions/mod.rs
```diff
@@ -236,34 +236,104 @@ pub enum Extension {
...
 }
 
-impl TlsSerializeTrait for Extensions {
+impl<T: TypedExtension> Default for Extensions<T> {
+    fn default() -> Self {
+        Self { unique: vec![] }
+    }
+}
+
+impl<T: TypedExtension> TlsSerializeTrait for Extensions<T> {
```
wysiwys: One thing to check here is whether these serialization implementations should produce identical results for each `T: TypedExtension`. This could be done by converting to a plain `Extension` first, for example, or by ensuring that the enum variants have the same discriminants and layouts.

openmls/src/extensions/mod.rs
```diff
@@ -236,34 +236,104 @@ pub enum Extension {
...
     fn tls_serialize<W: Write>(&self, writer: &mut W) -> Result<usize, Error> {
         self.unique.tls_serialize(writer)
     }
 }
 
-impl TlsDeserializeTrait for Extensions {
+impl<T: TypedExtension> TlsDeserializeTrait for Extensions<T> {
```
wysiwys: I haven't looked into this deeply yet, but it could be better to always deserialize plain `Extensions` lists for incoming messages, and then validate them afterwards, to avoid validation during deserialization. This would mean that the `tls_codec::Deserialize`/`DeserializeTrait` implementation would only be needed for the `Extensions<Extension>`.

openmls/src/group/public_group/builder.rs
```diff
@@ -41,21 +42,16 @@ impl TempBuilderPG1 {
 
     pub(crate) fn with_group_context_extensions(
         mut self,
-        extensions: Extensions,
+        extensions: Extensions<Extension>,
```
wysiwys: An `Extensions<GroupContextExtension>` could be passed in as an argument here instead

openmls/src/group/group_context.rs
```diff
@@ -45,7 +45,7 @@ pub struct GroupContext {
     epoch: GroupEpoch,
     tree_hash: VLBytes,
     confirmed_transcript_hash: VLBytes,
-    extensions: Extensions,
+    extensions: Extensions<GroupContextExtension>,
```
wysiwys: Since the group context is deserialized using `TlsDeserialize`, keeping this field as a plain extensions list would avoid validation during deserialization

openmls/src/extensions/mod.rs
```diff
@@ -236,34 +236,104 @@ pub enum Extension {
...
+///
+/// ```c
+/// // draft-ietf-mls-protocol-16
+/// struct {
+///     ExtensionType extension_type;
+///     opaque extension_data<V>;
+/// } Extension;
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum GroupContextExtension {
```
wysiwys: I think the main question here is whether to define a separate extension kind for each object, or whether to use the same extension kind everywhere and make the `Extensions` generic over the object types instead (`GroupContext`, `LeafNode`, etc.). 

One strong advantage of separate kinds is that the extension type validity is enforced from the ground up. However, the different extension enums could add complexity from the public API perspective. 

I think this is something it could be good to come back and revisit when using the `Extensions<GroupContextExtension>` as an input parameter in other places in the public API, such as  `MlsGroupBuilder::with_group_context_extensions()` and `CommitBuilder::propose_group_context_extensions()`.

wysiwys (commented):
Thanks! I think this goes in the right direction, although I'd agree it would be good to reduce some of the overlap in enums and method implementations. I added some specific feedback about types and serialization/deserialization.

openmls/src/extensions/mod.rs
tbehner: I tried this out, and it works well, but only if the extensions in `GroupContext` are of type `Extensions<Extension>`. Otherwise, the `TlsSerialize` and `TlsDeserialize` traits can't be derived and must be implemented manually, though that's also an option.

But isn't it counterintuitive that the extensions in `GroupContext` are **not** of type `Extensions<GroupContextExtension>`? This feels more like a workaround to me, so I'm not entirely convinced.

An alternative would be to implement `TlsSerialize`, `TlsDeserialize`, and `Size` for `GroupContextExtension` by first converting it to an `Extension`, for example:

```rust
impl Size for GroupContextExtension {
    #[inline]
    fn tls_serialized_len(&self) -> usize {
        Into::<Extension>::into(self.clone()).tls_serialized_len()
    }
}
```

openmls/src/extensions/mod.rs
tbehner: After attempting different approaches, the core problem remains that enum variants are not types on their own. This prevents us from defining generic methods across multiple enums with semantically equal variants.

This leaves us with as far as I can tell with two main options.

Option 1: Redefine the `Extension` Enum

This approach involves redefining the plain `Extension` enum to contain other extension enums as variants.

```rust
enum Extension {
    GroupContext(GroupContextExtension),
    LeafNode(LeafNodeExtension),
    // ...
}
```

This would eliminate the need to redefine the same methods for accessing internal fields multiple times. However, this only works if the extension variants don't overlap. So far, is this only not true for `Unknown`? If so, we'd still have duplicate `fn unknown` instances across all variants.

Option 2: Use Macros

A macro like `#[derive(/*...*/ GetterMethods)]` could be used to automatically generate getter methods.

```rust
#[derive(/*...*/ GetterMethods)]
enum Extension {
    ApplicationId(ApplicationIdExtension),
    RatchetTree(RatchetTreeExtension),
    // ...
}
```

could then produce the following methods:

```rust
pub fn as_application_id_extension(&self) -> Result<&ApplicationIdExtension, ExtensionError> {}
pub fn as_ratchet_tree_extension(&self) -> Result<&RatchetTreeExtension, ExtensionError> {}
// ...
```

This method would also significantly reduce code duplication within the `impl Extension` itself, as all these getter methods behave identically, with the only changes being the concrete type and arity.

openmls/src/extensions/mod.rs
tbehner: I would also prefer separate enums to enforce the validity of extensions early. From a user's perspective, this might be even better since the intent of the API method is expressed more clearly.

openmls/src/group/mls_group/commit_builder.rs
tbehner: Yes, absolutely! I am also in favor of asking for `Extension<GroupContextExtension>` to enforce validity.

openmls/src/extensions/mod.rs
tbehner: This solution works well. I believe it's a solid option, provided we don't need to add more functionality to the context-specific extensions down the line.

openmls/src/extensions/mod.rs
tbehner: The PhantomData solution is another viable option for this problem.

franziskuskiefer: Thanks for detailing a couple more options.

Option 1 is pretty disruptive to consumers and with the [extensions draft](https://messaginglayersecurity.rocks/mls-extensions/draft-ietf-mls-extensions.html#section-7.2.1) we have more overlapping extension. So I don't think that will work.

For option 2 we'd need to write and maintain the proc macro. That wouldn't need too much code, but requires a the new crate. I don't think there's a reasonable existing crate we could use here.

Between those three options I think the phantom data version sounds best to me.

Taking a step back to what we want to achieve. We want to make it easier to validate the correctness of the extensions used in a certain context. This could be done just on the plain extension by ensuring the correct validation function is called. That's the least amount of code but the most error prone. Using the type system to make this more robust is good. But then we have to think about what the cost of this added safety is. In this setting, using a marker like @wysiwys suggested sounds like a good middle ground to me.

openmls/src/messages/proposals_in.rs
```diff
@@ -317,6 +318,39 @@ impl From<UpdateProposal> for UpdateProposalIn {
...
+// The following `From` implementation breaks abstraction layers and MUST
+// NOT be made available outside of tests or "test-utils".
+#[cfg(any(feature = "test-utils", test))]
+impl From<GroupContextExtensionProposalIn> for GroupContextExtensionProposal {
+    fn from(value: GroupContextExtensionProposalIn) -> Self {
+        Self::new(value.extensions_tbv.try_into().unwrap())
+    }
+}
+
+impl From<GroupContextExtensionProposalIn> for Box<GroupContextExtensionProposal> {
```
wysiwys, line 341: Should this `From` implementation also be annotated with
`#[cfg(any(feature = "test-utils", test))]`, like the trait implementation above it?

openmls/src/extensions/mod.rs
```diff
@@ -349,13 +397,41 @@ impl Extensions {
     }
 }
 
-impl TryFrom<Vec<Extension>> for Extensions {
+/// Can be implemented by a type to validate extensions.
+pub trait ExtensionValidator {
+    /// Check if the extension is valid.
+    fn valid_extension(ext: &Extension) -> bool;
```
wysiwys: A nit on this method name - the name could indicate that only the extension type is being checked, and not other aspects of the extension. E.g. `fn validate_extension_type()`

openmls/src/extensions/mod.rs
```diff
@@ -517,6 +596,30 @@ impl Extension {
...
+        value.unique.try_into()
+    }
+}
+
+// TODO get rid of this
+impl From<ExtensionsForObject<GroupContext>> for Extensions {
+    fn from(value: ExtensionsForObject<GroupContext>) -> Self {
+        (&value).into()
+    }
+}
```
wysiwys: This method could also instead move the `value.unique` member directly into a new `Extensions` (if the method is still needed)

openmls/src/extensions/errors.rs
```diff
@@ -92,11 +95,16 @@ pub enum InvalidExtensionError {
     /// The specified extension could not be found.
     #[error("The specified extension could not be found.")]
     NotFound,
-    /// The provided extension list contains an extension that is not allowed in group contexts
+    /// The provided extension list contains an extension that is not allowed
```
wysiwys:
```suggestion
    /// The provided extension list contains an extension type that is not allowed
```

openmls/src/extensions/errors.rs
wysiwys, line 112: The `NotValid` error variant could also be renamed to something like `ExtensionTypeNotValidInObject`, to more closely indicate the case where it occurs. Also, for the `IllegalInLeafNodes` variant, since it's already covered by the new variant, it looks like it could be taken out, too

openmls/src/extensions/mod.rs
```diff
@@ -296,6 +329,21 @@ impl Extensions {
         extensions.try_into()
     }
 
+    /// Validate if the extensions are valid for this context
+    pub fn validate<'a>(
+        extensions: impl Iterator<Item = &'a Extension>,
+    ) -> Option<InvalidExtensionError> {
```
wysiwys: One stylistic comment about this function: instead of returning an `Option<_>` from this method, a `Result<(), _>` could be returned here instead, to match the style in other parts of the codebase

wysiwys (commented):
Thanks, this looks good to me, with a few comments. 

About the `MlsGroupBuilder::with_group_context_extensions()` API:

 Indeed, adapting this API to accept a list with validated extension types would currently result in a conversion to a plain `Extensions` list internally. This could be done, for example, when building the `GroupContext` using the `GroupContext::new()` method, since the `GroupContext` struct represents its extensions as a plain `Extensions` list. 

If an internal conversion is acceptable, it  could make sense to already adapt the `MlsGroupBuilder::with_group_context_extensions()` API, in order to still provide a public API that enforces extension type validity, even though the validated list type isn't used throughout the internal handling. I definitely can see pros and cons to adapting this method at this stage, though!

openmls/src/extensions/mod.rs
tbehner: That's better! (yes, the method is still needed)

PR #1836, opened 2025-08-30
fix(external-commit): persist initial group state in finalize(); add regression test

pwnsdx:
Linked issue
- Ref: #1801

Summary
- Persist the initial `MlsGroup` state when joining via External Commit by calling
  `group.store(provider.storage())` in `external_commit_builder::finalize()` before entering `PendingCommit`.
- Add a regression test with two providers to ensure the group is reloadable from storage after `finalize()`.

Motivation & context
- After #1801 refactoring, the initial `store()` call was no longer present in the External Commit path.
- Since an External Commit implies creating a group, that initial state must be persisted once so `MlsGroup::load()` can succeed.

Design / Approach
- `external_commit_builder::finalize()` now calls `group.store(provider.storage())` before setting `PendingCommit` and merging.
- No changes to wire or on-disk formats; behavior aligns with the documented contract that `store()` is used by constructors.

Alternatives considered
- Persist in `merge_commit()` (GroupMember path): rejected to follow the documentation (“store() only from constructors”) and avoid mixing construction and merge responsibilities.

Backwards compatibility
- No breaking changes; no format changes; aligns persistence with expected semantics.

Tests & verification
- New regression test: `external_commit_persistence_bug`:
  - Uses two providers (Alice/Bob) with distinct storages.
  - Fails without the fix (group not reloadable from Bob's storage).
  - Passes with the fix across variants.

Reproduction
```bash
CARGO_ENCODED_RUSTFLAGS= cargo test -p openmls \
  --target aarch64-apple-darwin \
  --no-default-features --features test-utils \
  -- external_commit_persistence_bug --nocapture
```

Potential side-effects / risks
- Minimal. One additional storage write during `finalize()` (constructor phase) for External Commit. No changes to merge-time writes.

Documentation
- Matches the intended contract: initial state persisted by constructor-like code paths.

Checklist
- [x] Small, focused changeset
- [x] Issue linked
- [x] Tests added and passing
- [x] No public API or format changes
- [x] Commit message in imperative mood; clear rationale

### DIFF
--- a/openmls/src/group/mls_group/commit_builder/external_commits.rs
+++ b/openmls/src/group/mls_group/commit_builder/external_commits.rs
@@ -369,6 +369,12 @@ impl CommitBuilder<'_, super::Complete, MlsGroup> {
             group.mls_group_config.wire_format_policy = wire_format_policy;
         }
 
+        // Persist the initial group state before entering PendingCommit.
+        // Per documentation, store() should be called by constructors only.
+        group
+            .store(provider.storage())
+            .map_err(ExternalCommitBuilderFinalizeError::StorageError)?;
+
         // Set the current group state to [`MlsGroupState::PendingCommit`],
         // storing the current [`StagedCommit`] from the commit results
         group.group_state = MlsGroupState::PendingCommit(Box::new(PendingCommitState::Member(
--- a/openmls/src/group/tests_and_kats/tests/external_commit_persistence_bug.rs
+++ b/openmls/src/group/tests_and_kats/tests/external_commit_persistence_bug.rs
@@ -1,0 +1,111 @@
+//! Test to reproduce the persistence bug in External Commits
+//! This test verifies that own_leaf_index and join_group_config are correctly persisted
+//! after an external commit.
+
+use crate::group::{
+    tests_and_kats::utils::generate_credential_with_key, MlsGroup, MlsGroupCreateConfig,
+    PURE_PLAINTEXT_WIRE_FORMAT_POLICY,
+};
+
+// Test to reproduce the external commit persistence bug
+#[openmls_test::openmls_test]
+fn test_external_commit_persistence_bug() {
+    // Separate providers for Alice and Bob (distinct storage spaces)
+    let alice_provider = &Provider::default();
+    let bob_provider = &Provider::default();
+
+    // Generate credentials
+    let alice_credential = generate_credential_with_key(
+        "Alice".into(),
+        ciphersuite.signature_algorithm(),
+        alice_provider,
+    );
+
+    let bob_credential = generate_credential_with_key(
+        "Bob".into(),
+        ciphersuite.signature_algorithm(),
+        bob_provider,
+    );
+
+    // Group configuration
+    let mls_group_create_config = MlsGroupCreateConfig::builder()
+        .wire_format_policy(PURE_PLAINTEXT_WIRE_FORMAT_POLICY)
+        .ciphersuite(ciphersuite)
+        .build();
+
+    // Alice creates a group
+    let mut alice_group = MlsGroup::new(
+        alice_provider,
+        &alice_credential.signer,
+        &mls_group_create_config,
+        alice_credential.credential_with_key.clone(),
+    )
+    .unwrap();
+
+    // Get group information for the external commit
+    let verifiable_group_info = alice_group
+        .export_group_info(alice_provider.crypto(), &alice_credential.signer, false)
+        .unwrap()
+        .into_verifiable_group_info()
+        .unwrap();
+    let tree_option = alice_group.export_ratchet_tree();
+
+    // Bob performs an external commit
+    let (bob_group, _public_message_commit) = MlsGroup::external_commit_builder()
+        .with_config(alice_group.configuration().clone())
+        .with_ratchet_tree(tree_option.into())
+        .build_group(
+            bob_provider,
+            verifiable_group_info,
+            bob_credential.credential_with_key.clone(),
+        )
+        .unwrap()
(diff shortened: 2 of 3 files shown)


pwnsdx:
### Note for Reviewers: Concurrency Considerations

While this PR fixes the immediate issue (groups not being reloadable after External Commits), I wanted to document a related challenge discovered during extensive testing:

### Race Condition in High-Concurrency Scenarios

When multiple clients join simultaneously via External Commits:

1. `finalize()` persists the group state to storage (including with this fix)
2. The commit is sent to the server
3. If the server rejects it (e.g., HTTP 409 - leaf index conflict), the client has already persisted invalid state
4. Subsequent retry attempts fail with SQLite constraint violations

### Current Impact

- **Single client joins**: ✅ Working perfectly with this fix
- **Sequential joins**: ✅ Working perfectly
- **Parallel joins (10+ simultaneous)**: ⚠️ Requires cleanup between retries

### Workaround in Production

Applications experiencing this can handle it by:
```rust
  // If server rejects the commit
  if response.status() == 409 {
      // Clean up SQLite tables manually
      delete_from_storage(group_id)?;
      // Retry with fresh GroupInfo
  }
```

### Potential Future Improvements

This could be addressed in a follow-up PR through:

1. Transaction support in the storage provider
2. Deferred persistence - only persist after server confirmation
3. Official cleanup API for failed External Commits

### Testing

The included regression test verifies the core fix. The concurrency issue only manifests under specific conditions (high parallel load) and doesn't affect the primary use case.

This PR remains focused on fixing the regression. The concurrency challenge is documented here for transparency and potential future work.

openmls/src/group/tests_and_kats/tests/mod.rs
```diff
@@ -7,6 +7,7 @@ mod encoding;
 mod external_add_proposal;
 mod external_commit;
 mod external_commit_builder;
+mod external_commit_persistence_bug;
```
kkohbrok:
```suggestion
mod external_commit_persistence;
```
So far, module names describe what is tested. If you want to reference the bug specifically, you could add an issue number to the test doc comment.

openmls/src/group/tests_and_kats/tests/external_commit_persistence_bug.rs
```diff
@@ -0,0 +1,111 @@
...
+    );
+
+    // Verify that the group configuration is properly persisted
+    assert_eq!(
+        reloaded_group.configuration(),
+        &expected_join_config,
+        "join_group_config should be persisted and match the original value"
+    );
+
+    println!("✅ Test passed: External commit persistence works correctly");
```
kkohbrok, line 110: Doesn't `cargo test` only print stdout if the test fails? If so, this will never be shown.

openmls/src/group/tests_and_kats/tests/external_commit_persistence_bug.rs
```diff
@@ -0,0 +1,111 @@
+//! Test to reproduce the persistence bug in External Commits
```
kkohbrok, line 1: If you want to reference a specific bug, it might be good to include a GH issue number here.

openmls/src/group/mls_group/commit_builder/external_commits.rs
```diff
@@ -369,6 +369,12 @@ impl CommitBuilder<'_, super::Complete, MlsGroup> {
             group.mls_group_config.wire_format_policy = wire_format_policy;
         }
 
+        // Persist the initial group state before entering PendingCommit.
+        // Per documentation, store() should be called by constructors only.
```
kkohbrok:
```suggestion
        // Persist the initial group state.
```

openmls/src/group/tests_and_kats/tests/external_commit_persistence_bug.rs
```diff
@@ -0,0 +1,111 @@
...
+            bob_provider.crypto(),
+            &bob_credential.signer,
+            |_| true,
+        )
+        .unwrap()
+        .finalize(bob_provider)
+        .unwrap();
+
+    // Verify that Bob has a valid own_leaf_index (it must be >= 0)
+    let _idx = bob_group.own_leaf_index();
```
kkohbrok, line 76: I don't think this actually tests anything.

openmls/src/group/tests_and_kats/tests/external_commit_persistence_bug.rs
```diff
@@ -0,0 +1,111 @@
...
+        expected_own_leaf_index,
+        "own_leaf_index should be persisted and match the original value"
+    );
+
+    // Verify that the group configuration is properly persisted
+    assert_eq!(
+        reloaded_group.configuration(),
+        &expected_join_config,
+        "join_group_config should be persisted and match the original value"
+    );
```
kkohbrok, line 108: Instead of checking the config and the own leaf index, you can just check if the exported authenticator matches.

openmls/src/group/tests_and_kats/tests/external_commit_persistence_bug.rs
```diff
@@ -0,0 +1,111 @@
...
+//! after an external commit.
+
+use crate::group::{
+    tests_and_kats::utils::generate_credential_with_key, MlsGroup, MlsGroupCreateConfig,
+    PURE_PLAINTEXT_WIRE_FORMAT_POLICY,
+};
+
+// Test to reproduce the external commit persistence bug
+#[openmls_test::openmls_test]
+fn test_external_commit_persistence_bug() {
```
kkohbrok:
```suggestion
fn external_commit_persistence() {
```

PR #1838, opened 2025-09-02
fix large enum variants

### LINKED ISSUE #1670: Box large enum variants in ProtocolMessage, Node and NodeIn
This is waiting for [PR 1586 to tls_codec](https://github.com/RustCrypto/formats/pull/1586).

franziskuskiefer:
While working on the partial MLS draft I ran into the large enums again and again. So I fixed them. This is a pretty mechanic changes. I only had to touch logic in a few places where matching doesn't work anymore.

Fixes #1670

### DIFF
--- a/openmls/src/binary_tree/array_representation/tree.rs
+++ b/openmls/src/binary_tree/array_representation/tree.rs
@@ -23,8 +23,25 @@ where
     L: Clone + Debug + Default,
     P: Clone + Debug + Default,
 {
-    Leaf(L),
-    Parent(P),
+    Leaf(Box<L>),
+    Parent(Box<P>),
+}
+
+#[cfg(test)]
+impl<L, P> TreeNode<L, P>
+where
+    L: Clone + Debug + Default,
+    P: Clone + Debug + Default,
+{
+    /// Create a new leaf.
+    pub(crate) fn leaf(l: L) -> Self {
+        Self::Leaf(Box::new(l))
+    }
+
+    /// Create a new parent.
+    pub(crate) fn parent(p: P) -> Self {
+        Self::Parent(Box::new(p))
+    }
 }
 
 #[cfg_attr(any(test, feature = "test-utils"), derive(PartialEq))]
@@ -63,14 +80,14 @@ impl<L: Clone + Debug + Default, P: Clone + Debug + Default> ABinaryTree<L, P> {
             match node {
                 TreeNode::Leaf(l) => {
                     if i % 2 == 0 {
-                        leaf_nodes.push(l)
+                        leaf_nodes.push(*l)
                     } else {
                         return Err(ABinaryTreeError::WrongNodeType);
                     }
                 }
                 TreeNode::Parent(p) => {
                     if i % 2 == 1 {
-                        parent_nodes.push(p)
+                        parent_nodes.push(*p)
                     } else {
                         return Err(ABinaryTreeError::WrongNodeType);
                     }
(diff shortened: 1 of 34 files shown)


openmls/src/framing/mls_content.rs
```diff
@@ -66,14 +65,22 @@ impl From<AuthenticatedContent> for FramedContent {
 #[repr(u8)]
 pub(crate) enum FramedContentBody {
     #[tls_codec(discriminant = 1)]
-    Application(VLBytes),
+    Application(Box<VLBytes>),
```
keks: This shouldn't be big. It's just a Vec, or am I missing something?

openmls/src/framing/mls_content_in.rs
```diff
@@ -112,14 +111,19 @@ impl From<AuthenticatedContentIn> for FramedContentIn {
 #[repr(u8)]
 pub(crate) enum FramedContentBodyIn {
     #[tls_codec(discriminant = 1)]
-    Application(VLBytes),
+    Application(Box<VLBytes>),
```
keks: same here

openmls/src/binary_tree/array_representation/tree.rs
```diff
@@ -23,8 +23,25 @@ where
     L: Clone + Debug + Default,
     P: Clone + Debug + Default,
 {
-    Leaf(L),
-    Parent(P),
+    Leaf(Box<L>),
+    Parent(Box<P>),
```
keks, line 27: I am not convinced that this is a good idea. It seems values are only kept temporarily, and only in one case we have a large amount of them. And in [that one instance](https://github.com/openmls/openmls/blob/19f144ef108809836312540df1e144943b6269fa/openmls/src/treesync/mod.rs#L451-L470), we create a long list, and then immediately consume it and call ABinaryTree::new on it, pulling the leafs out of the enum again. If we box these, we have a lot of small heap allocations so we keep a bit less memory for a short time.

My impression is that if we want to reduce the memory consumption over there, we shouldn't first create a Vec<LeafNode> and then immediate consuming it, but instead directly creating the ABinaryTree, maybe using a builder or a from_components function.

Do we even need the TreeNode type?

franziskuskiefer: Well, this PR is not about changing the tree implementation, but fixing warnings. And the leaf and parent nodes have significantly different sizes here, hence the warning.

If you think there's something in the design of the tree that's worth changing, file an issue with those thaughts?

openmls/src/framing/mls_content.rs
franziskuskiefer: Ah yeah, I just did all of them. This is of course unnecessary.

openmls/src/binary_tree/array_representation/tree.rs
keks: Well, so far this might have produced warnings but the code was fine. Now I am not so sure from a performance perspective, but I guess this is difficult to argue without having measurements. I can open an issue that flags the unnecessary allocations here.

franziskuskiefer: Sure. I think the type here can be dropped. But because there are a bunch of tests and other things, that's a bit of work.

keks: Alright, let's do it this way. Let's merge this and I'll open an issue.

PR #1840, opened 2025-09-03
Add function to check if there are pending proposals

timokoesters:
This is useful, to see if a commit is required or if the client is able to send messages without a commit.

Alternative: The client can also try to do an operation and look for a PendingProposal error.


openmls/src/group/mls_group/processing.rs
```diff
@@ -95,6 +95,11 @@ impl MlsGroup {
         Ok(())
     }
 
+    /// Returns true if there are pending proposals queued in the proposal store.
+    pub fn has_pending_proposal(&self) -> bool {
```
raphaelrobert:
```suggestion
    pub fn has_pending_proposals(&self) -> bool {
```

raphaelrobert (approved):

kkohbrok (approved):

PR #1844, opened 2025-09-09
Updating Application Data in the GroupContext (Section 4.7)

### LINKED ISSUE #1843: Updating Application Data in the GroupContext (Section 4.7)
https://datatracker.ietf.org/doc/html/draft-ietf-mls-extensions#section-4.7

wysiwys:
This draft pull request implements an approach for [Section 4.7](https://datatracker.ietf.org/doc/html/draft-ietf-mls-extensions#section-4.7) of draft-ietf-mls-extensions-08, for updating application data in the `GroupContext`, using the `AppDataUpdateProposal`. This PR is based on PR #1826, which implements [Section 4.6](https://datatracker.ietf.org/doc/html/draft-ietf-mls-extensions#section-4.6).

### Proposals
The `AppDataUpdateProposal` is introduced, which includes the `ComponentId` and the `AppDataUpdateOperation` to apply. For creating and adding these proposals, following the same format as existing APIs for other proposal types, new APIs are added to the `MlsGroup`. Validation is also added for the `AppDataUpdateProposal`s, as well as for the the conditions that need to hold for the `GroupContextExtensionProposal`, to the `CommitBuilder::build_internal()` and `PublicGroup::validate_commit()` methods.

### Applying application logic
This PR adds a new public API for applying application logic for `AppDataUpdateProposal`s. A new enum variant is added to `ProcessedMessage` that represents a `StagedCommit` with pending `AppDataUpdateProposal`s. This message can be processed using a method that takes a reference to a struct (`AppDataUpdateLogic`) representing the application logic registered for each component, and returns a `StagedCommit` with the changes applied. 
```rust
// set up application-level component logic
let mut registered_components = AppDataUpdateLogic::new();
// register logic for each component
registered_components.register(component_id, |_data| { /* logic for processing update */ });
```
- Ensure that the application only updates state in the dictionary when expected, in order to avoid inconsistencies in the state across clients. An alternative approach would be to iterate through the proposals and mutate the state in the `StagedCommit`. Then, the application would need to ensure that it processes all of the AppDataUpdate proposals, and makes updates in the correct order.
- Provide a convenient API that avoids a need to mutate the `StagedCommit` and introduce incomplete or invalid states to the `StagedCommit`.
- Keep the registered application logic separate from the `MlsGroup`, while not requiring the application logic to be passed in as an argument to `MlsGroup::process_message()`. 

This approach is one of several possible approaches to organizing this functionality, and I think in its current state could provide a useful basis for discussion of the pros and cons of different approaches.

```rust
if let ProcessedMessageContent::PendingAppDataUpdates(commit) = processed_message.into_content() {
    let staged_commit = commit.apply_app_logic(&registered_components)
        .expect("Applying application logic failed");

    // handle staged commit

}
```

### Additional changes
- In the test utils, adds a builder struct for the `PreGroupPartyState`
- Represent the `AppDataDictionaryExtension`'s dictionary internally using a `BTreeMap`, to ensure that ordering and uniqueness properties hold, and for simpler implementation of functions that modify the dictionary. However, the dictionary is still serialized as an ordered vector of `ComponentData` entries.

### Some open questions
- What is the best way to represent the `ComponentId`s that are known to the application, as well as the application logic registered to each one? When to provide this information in the steps for processing a commit?
- Would it be better to represent the `AppDataDictionary` internally as a vector?
- When merging pending commits on the committer's side: API for updating the `AppDataDictionary`

### TODO:
- [ ] Add more tests (including validation of incoming proposals)

Resolves #1843

### DIFF
--- a/openmls/src/group/mls_group/app_data_update.rs
+++ b/openmls/src/group/mls_group/app_data_update.rs
@@ -99,6 +99,10 @@ impl AppDataUpdateLogic {
 }
 
 impl PendingAppDataUpdates {
+    /// Retrieve a reference to the underlying [`StagedCommit`].
+    pub fn staged_commit(&self) -> &StagedCommit {
+        &self.0
+    }
     /// Apply the application logic registered in the [`AppDataUpdateLogic`] to the [`StagedCommit`].
     pub fn apply_app_logic(
         self,


openmls/src/messages/proposals.rs
```diff
@@ -81,6 +81,8 @@ pub enum ProposalType {
     GroupContextExtensions,
     AppAck,
     SelfRemove,
+    #[cfg(feature = "extensions-draft-08")]
```
franziskuskiefer, line 84: Guarding enum variants is of course a little ugly. But I think it's fine here for the extensions.

openmls/src/messages/proposals.rs
```diff
@@ -755,6 +770,113 @@ pub(crate) struct MessageRange {
     last_generation: u32,
 }
 
+// TODO: move AppDataUpdate proposal to separate file?
```
franziskuskiefer, line 773: yeah lets do that

openmls/src/messages/proposals.rs
```diff
@@ -755,6 +770,113 @@ pub(crate) struct MessageRange {
...
+        Deserialize,
+        TlsDeserialize,
+        TlsDeserializeBytes,
+        TlsSerialize,
+        TlsSize,
+    )]
+    /// The operation that is part of an [`AppDataUpdateProposal`].
+    pub enum AppDataUpdateOperation {
+        /// Update operation, containing update data
+        // TODO: is this the correct use of tls_codec(discriminant = ..)?
```
franziskuskiefer, line 805: Yes, that's for the encoding, i.e. we don't need to encode the `Type` above but just that here from the discriminant.
But why the values ` = 1` in the enum?

openmls/tests/app_data_update.rs
```diff
@@ -0,0 +1,281 @@
...
+    let processed_message = bob
+        .group
+        .process_message(
+            &bob_party.provider,
+            message_in.try_into_protocol_message().unwrap(),
+        )
+        .unwrap();
+
+    // set up component logic
+    let mut registered_components = RegisteredComponentsWithLogic::new();
```
franziskuskiefer: We should make this a little closer to a real live use case.
This is Bob getting a commit, right? So Bob would look inside and see if there's an app data update proposal inside. If yes, Bob would process it by getting the `data` from the proposal, getting the existing data for the component, and doing some processing, before then writing it back.
I'm not quite sure that's doable with the API right now?

openmls/src/group/mls_group/app_data_update.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// [`AppDataDictionaryExtension`](crate::extensions::AppDataDictionaryExtension).
+    ComponentNotAvailable,
+}
+
+// helper type for the application logic stored in the [`RegisteredComponentsWithLogic`]
+type RegisteredComponentLogic = Box<dyn Fn(&[u8]) -> Result<Vec<u8>, ApplyAppLogicError>>;
+
+/// A struct representing the [`ComponentId`]s known to the application,
+/// with the application logic registered to each component.
+pub struct RegisteredComponentsWithLogic(BTreeMap<ComponentId, RegisteredComponentLogic>);
```
franziskuskiefer: I'm not sure about the name and where this should live.

openmls/src/group/mls_group/app_data_update.rs
```diff
@@ -0,0 +1,188 @@
...
+impl RegisteredComponentsWithLogic {
+    /// Initialize a new, empty [`RegisteredComponentsWithLogic`].
+    pub fn new() -> Self {
+        Self(BTreeMap::new())
+    }
+    /// Register a [`ComponentId`] alongside its application logic.
+    pub fn register(
+        &mut self,
+        component_id: ComponentId,
+        app_logic: impl Fn(&[u8]) -> Result<Vec<u8>, ApplyAppLogicError> + 'static,
```
franziskuskiefer, line 68: This seems pretty limited for the client.
Didn't we discuss that the caller should get an iterator for the app data proposals and run its update on that? Didn't that work out? This design seems pretty different.

openmls/tests/app_data_update.rs
```diff
@@ -0,0 +1,281 @@
...
+        let mut new_data = b"new_data:".to_vec();
+        new_data.extend(data.to_vec());
+        Ok(new_data)
+    });
+
+    let commit = match processed_message.into_content() {
+        ProcessedMessageContent::StagedCommitWithPendingAppDataUpdates(commit) => commit,
+        _ => panic!("Should be a processed commit with app data updates"),
+    };
+    let staged_commit = commit.apply_app_logic(&registered_components).unwrap();
```
franziskuskiefer: At this stage we could just return an iterator with the app data dictionary proposals and have the caller iterate over it, with individual store calls to update the data.

openmls/src/messages/proposals.rs
wysiwys: It's another way to set the discriminant for the enum variant, but is this redundant/not needed?

keks: I think that makes sense especially for type-style enums like https://github.com/openmls/openmls/blob/2849f05eab2eadc9fe150fd3419e41732c88bec8/openmls/src/framing/mod.rs#L188-L196 or https://github.com/openmls/openmls/blob/2849f05eab2eadc9fe150fd3419e41732c88bec8/openmls/src/framing/mod.rs#L127-L139.

Note that neither of these carry any data, do they can easily be converted to the number they map to. I suppose we could introduce a separate enum here that has just the numbers and use that there in there.

PR #1849, opened 2025-09-24
perf: Avoid storing the whole group state after decryption

kkohbrok:
Small follow-up to #1846. Instead of storing the whole group state after decrypting a message, it only stores the message_secrets_store.

### DIFF
--- a/openmls/src/group/mls_group/processing.rs
+++ b/openmls/src/group/mls_group/processing.rs
@@ -87,7 +87,9 @@ impl MlsGroup {
 
         // Persist the group state if the secret tree was modified to ensure forward secrecy
         if will_modify_secret_tree {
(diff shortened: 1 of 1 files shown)


openmls/src/group/mls_group/processing.rs
```diff
@@ -87,7 +87,9 @@ impl MlsGroup {
 
         // Persist the group state if the secret tree was modified to ensure forward secrecy
         if will_modify_secret_tree {
-            self.store(provider.storage())
+            provider
+                .storage()
+                .write_message_secrets(self.group_id(), &self.message_secrets_store)
```
raphaelrobert, line 92: Can you change the comment accordingly as well, please?

PR #1854, opened 2025-09-30
feat: add sqlx based storage provider

kkohbrok:
This PR adds a sqlx based storage provider, which only supports Sqlite for now.

### DIFF
--- a/sqlx_storage/src/epoch_key_pairs.rs
+++ b/sqlx_storage/src/epoch_key_pairs.rs
@@ -1,0 +1,15 @@
+// SPDX-FileCopyrightText: 2023 Phoenix R&D GmbH <hello@phnx.im>
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+use std::marker::PhantomData;
+
+use openmls_traits::storage::{CURRENT_VERSION, Entity};
+
+pub(crate) struct StorableEpochKeyPairs<EpochKeyPairs: Entity<CURRENT_VERSION>>(
+    PhantomData<EpochKeyPairs>,
+);
+
+pub(super) struct StorableEpochKeyPairsRef<'a, EpochKeyPairs: Entity<CURRENT_VERSION>>(
+    pub &'a [EpochKeyPairs],
+);
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ members = [
     "interop_client",
     "memory_storage",
     "sqlite_storage",
+    "sqlx_storage",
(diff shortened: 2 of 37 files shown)


sqlx_storage/src/epoch_key_pairs.rs
```diff
@@ -0,0 +1,11 @@
+use std::marker::PhantomData;
+
+use openmls_traits::storage::{CURRENT_VERSION, Entity};
+
+pub(crate) struct StorableEpochKeyPairs<EpochKeyPairs: Entity<CURRENT_VERSION>>(
+    PhantomData<EpochKeyPairs>,
+);
```
keks: I don't think this will be instantiable, since the field is private. Is that correct?

sqlx_storage/src/migrator.rs
```diff
@@ -0,0 +1,164 @@
...
+            Ok(elapsed)
+        })
+    }
+
+    fn revert<'e: 'm, 'm>(
+        &'e mut self,
+        migration: &'m Migration,
+    ) -> BoxFuture<'m, Result<Duration, MigrateError>> {
+        Box::pin(async move {
+            // Use a single transaction for the actual migration script and the essential bookeeping so we never
```
keks:
```suggestion
            // Use a single transaction for the actual migration script and the essential bookkeeping so we never
```

sqlx_storage/src/epoch_key_pairs.rs
kkohbrok: Good point. Turns out this was just unnecessary boilerplate code. I removed it.

sqlx_storage/src/wrappers.rs
```diff
@@ -0,0 +1,58 @@
...
+    ProposalRef: Entity<CURRENT_VERSION>,
+>(pub ProposalRef, pub Proposal);
+
+pub(super) struct StorableProposalRef<
+    'a,
+    Proposal: Entity<CURRENT_VERSION>,
+    ProposalRef: Entity<CURRENT_VERSION>,
+>(pub &'a ProposalRef, pub &'a Proposal);
+
+pub(crate) struct StorablePskBundle<PskBundle: Entity<CURRENT_VERSION>>(PskBundle);
```
keks: I must have missed that earlier, but this one also looks uninstantiable.

kkohbrok: Good catch, thanks. Removed it.

keks (approved):
Nice, looks good!

PR #1855, opened 2025-10-02
re-add api

franziskuskiefer:
A simple re-add API that allows removing and adding the same members. This may help with recovering from bad group states.

Adds
- `MlsGroup::readd_members`
- `MlsGroup::member_leaf_index`
- `WelcomeCommitMessages`

### DIFF
--- a/openmls/src/group/mls_group/membership.rs
+++ b/openmls/src/group/mls_group/membership.rs
@@ -58,7 +58,7 @@ impl MlsGroup {
     ///
     /// This function can be used in scenarios where members are no
     /// longer in sync with the rest of the group and need to be re-added.
-    /// Note however that this function _does not_ enforce that the 
+    /// Note however that this function _does not_ enforce that the
     /// removed `members` and new members in the `key_packages` correspond.
     pub fn swap_members<Provider: OpenMlsProvider>(
         &mut self,


openmls/src/group/mls_group/membership.rs
```diff
@@ -46,6 +50,94 @@ impl MlsGroup {
...
+
+        let bundle = self
+            .commit_builder()
+            .propose_removals(members.iter().cloned())
+            .propose_adds(key_packages.into_iter())
+            .load_psks(provider.storage())?
+            .build(provider.rand(), provider.crypto(), signer, |_| true)?
+            .stage_commit(provider)?;
+
+        provider
```
kkohbrok: Why do we store the group state here? Shouldn't the commit builder have taken care of persistence?

openmls/src/group/mls_group/membership.rs
```diff
@@ -46,6 +50,94 @@ impl MlsGroup {
...
     }
 
+    /// Re-add members.
+    ///
+    /// If there are issues with a set of members and they are suspected to be
+    /// forked, this functions allows to remove and re-add them.
+    /// This is a convenience function that first removes, and then re-adds the
+    /// list of provided users.
+    ///
+    /// Note that this functions _does not_ enforce that the removed `members`
```
kkohbrok:
```suggestion
    /// Note that this function _does not_ enforce that the removed `members`
```

openmls/src/group/mls_group/membership.rs
keks: You're right, this happens in stage_commit: https://github.com/openmls/openmls/blob/2849f05eab2eadc9fe150fd3419e41732c88bec8/openmls/src/group/mls_group/commit_builder.rs#L794-L797

openmls/src/group/mls_group/membership.rs
```diff
@@ -46,6 +50,94 @@ impl MlsGroup {
...
+    /// and new members in the `key_packages` correspond.
+    ///
+    /// For the best performance it is important to sort the key packages in the
+    /// order they appear in the tree (left to right). If `sort` is set to
+    /// `true`, this function sorts the indices internally, based on the
+    /// credential. Note that this only works if the credential doesn't change
+    /// and the credential allows uniquely identifying a leaf node.
```
raphaelrobert: I don't understand what this is supposed to do. You can influence the order in which members are re-added, but you cannot guarantee in which leaf they will be added. The rule is always to use the left-most free leaf. If there are blanks, the members will be added in there. I also don't understand why the order matters at all.

keks: This was a misunderstanding of the RFC on our end. While of course leaves won't end up in the same position if there already are blanked leaves in the lower parts of the tree, we thought we'd at least achieve it in _some_ of the cases.

However, we also realized that the whole effort didn't work as we thought. We hoped to be able to avoid blanking the entire path up the re-added leaves by immediately re-adding them in the same position, but when we re-read the RFC it became clear that this isn't happening. Thanks for flagging that!

openmls/src/group/mls_group/errors.rs
```diff
@@ -187,6 +187,42 @@ pub enum AddMembersError<StorageError> {
     StorageError(StorageError),
 }
 
+/// Add members error
+#[derive(Error, Debug, PartialEq, Clone)]
+pub enum SwapMembersError<StorageError> {
+    /// Unable to map the key packages to the given leaf indices.
+    #[error("Unable to map the key packages to the given leaf indices.")]
```
raphaelrobert:
```suggestion
    #[error("Number of added and removed members is not the same")]
```

openmls/src/group/mls_group/membership.rs
```diff
@@ -46,6 +50,52 @@ impl MlsGroup {
...
+    /// This function swaps out the `members` of the group with new clients,
+    /// identified by the provided `key_packages`.
+    ///
+    /// One scenarion where this is useful is if there are issues with a set of
+    /// members and they are suspected to be forked, this functions allows to
+    /// remove and re-add them. This is a convenience function that first
+    /// removes, and then re-adds the list of provided users.
+    ///
+    /// Note that this function _does not_ enforce that the removed `members`
+    /// and new members in the `key_packages` correspond.
```
raphaelrobert:
```suggestion
    /// This function replaces a set of `members` of the group with new members.
    /// The members-to-be-replaced are identified by their index, and the new
    /// members are identified by the provided `key_packages`.
    ///
    /// This function can be used in scenarios where members are no
    /// longer in sync with the rest of the group and need to be re-added.
    /// Note however that this function _does not_ enforce that the 
    /// removed `members` and new members in the `key_packages` correspond.
```

Issue #1122, opened 2022-11-28
Validation for PSK proposals


raphaelrobert:
**General**

* [x] `psk_nonce` must be of length `KDF.Nh`.
  * [x] In Welcome
  * [x] In Proposal (#1331)
* [x] Application SHOULD specify an upper limit on the number of past epochs for which the resumption_psk may be stored.

**In Welcome**
* [x] PSK in welcome must be `Resumption/{ReInit,Branch}` or `External`.

**In Proposal**
* [x] PSK in proposal must be `Resumption/Application` or `External`. (#1331)
* [x] Proposal list must not contain multiple PreSharedKey proposals that reference the same PreSharedKeyID.  (#1331)

PR #1862, opened 2025-10-15
chore: psk validation when processing welcome messages

kkohbrok:
Fixes parts of #1122.

- Only reinit and branch resumption PSKs in Welcomes
- Only one reinit or branch resumption PSK in Welcomes
- Check PSK nonce length in Welcomes

### DIFF
--- a/openmls/src/schedule/errors.rs
+++ b/openmls/src/schedule/errors.rs
@@ -39,6 +39,7 @@ pub enum PskError {
         /// Got PSK type.
         got: ResumptionPskUsage,
     },
+    /// Usage conflict.
     #[error("Usage conflict. First found `{first:?}`, now additionally found `{second:?}`.")]
     UsageConflict {
         /// First found PSK type.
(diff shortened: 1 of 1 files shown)


openmls/src/schedule/psk.rs
```diff
@@ -327,6 +327,60 @@ impl PreSharedKeyId {
...
+                            allowed: vec![ResumptionPskUsage::Reinit, ResumptionPskUsage::Branch],
+                            got: resumption_psk.usage,
+                        });
+                    }
+                    ResumptionPskUsage::Reinit => {
+                        if contains_reinit_psk || contains_resumption_psk {
+                            return Err(PskError::UsageMismatch {
+                                allowed: vec![ResumptionPskUsage::Reinit],
+                                got: resumption_psk.usage,
+                            });
```
keks: Do I understand correctly that this error basically reports something liE "usage mismatch: expected Reinit, got Reinit"? That seems off

kkohbrok: Good catch. In fact, some of the error types were off. Should be fixed now.

PR #1864, opened 2025-10-19
feat: add from_components method to avoid allocations in TreeNode

### LINKED ISSUE #1839: Avoid allocations in TreeNode
In #1838 we introduced `Box`es around the children in `TreeNode`. The type is only used once, where a `Vec<TreeNode>` is created and then immediately turned into a `ABinaryTree`. This means that that change introduced a lot of small allocations, which isn't ideal.

Instead, we could create a different way to create the `ABinaryTree`, maybe using a builder or a `from_components` function.

WaterWhisperer:
- Add `ABinaryTree::from_components` method that accepts separate `Vec<L>` and `Vec<P>` to avoid intermediate Box allocations
- Refactor `TreeSync::from_ratchet_tree` to use `from_components` directly
- Add tests for the new method
- Remove unused `TreeNode` import from `treesync` module

This change eliminates unnecessary heap allocations when creating binary trees, improving performance by avoiding the creation of Vec<TreeNode> and subsequent Box unwrapping.

Fixes #1839

### DIFF
--- a/openmls/src/treesync/mod.rs
+++ b/openmls/src/treesync/mod.rs
@@ -45,7 +45,7 @@ use crate::binary_tree::array_representation::ParentNodeIndex;
 use crate::{binary_tree::array_representation::level, test_utils::bytes_to_hex};
 use crate::{
     binary_tree::{
-        array_representation::{is_node_in_tree, tree::TreeNode, LeafNodeIndex, TreeSize},
+        array_representation::{is_node_in_tree, LeafNodeIndex, TreeSize},
         MlsBinaryTree, MlsBinaryTreeError,
     },
     ciphersuite::{signable::Verifiable, Secret},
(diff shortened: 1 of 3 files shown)


openmls/src/treesync/mod.rs
```diff
@@ -449,25 +449,42 @@ impl TreeSync {
         ratchet_tree: RatchetTree,
     ) -> Result<Self, TreeSyncFromNodesError> {
         // TODO #800: Unmerged leaves should be checked
-        let mut ts_nodes: Vec<TreeNode<TreeSyncLeafNode, TreeSyncParentNode>> =
-            Vec::with_capacity(ratchet_tree.0.len());
+        let mut leaf_nodes = Vec::new();
+        let mut parent_nodes = Vec::new();
```
kkohbrok: Could you use `with_capacity` here?

WaterWhisperer: Understood, thanks!

kkohbrok (approved):

Issue #1866, opened 2025-11-04
`cargo build -F extensions-draft-08` fails in workspace root


wysiwys:
When building with `cargo build -F extensions-draft-08` in the workspace root, the following build error occurs:

```
error[E0046]: not all trait items implemented, missing: `write_application_export_tree`, `application_export_tree`, `delete_application_export_tree`
  --> sqlx_storage/src/storage_provider.rs:27:1
   |
27 | impl<C: Codec> StorageProvider<CURRENT_VERSION> for SqliteStorageProvider<'_, C> {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `write_application_export_tree`, `application_export_tree`, `delete_application_export_tree` in implementation
   |
   = help: implement the missing item: `fn write_application_export_tree<GroupId, ApplicationExportTree>(&self, _: &GroupId, _: &ApplicationExportTree) -> Result<(), <Self as StorageProvider<CURRENT_VERSION>>::Error> where GroupId: GroupId, ApplicationExportTree: ApplicationExportTree { todo!() }`
   = help: implement the missing item: `fn application_export_tree<GroupId, ApplicationExportTree>(&self, _: &GroupId) -> Result<std::option::Option<ApplicationExportTree>, <Self as StorageProvider<CURRENT_VERSION>>::Error> where GroupId: GroupId, ApplicationExportTree: ApplicationExportTree { todo!() }`
   = help: implement the missing item: `fn delete_application_export_tree<GroupId, ApplicationExportTree>(&self, _: &GroupId) -> Result<(), <Self as StorageProvider<CURRENT_VERSION>>::Error> where GroupId: GroupId, ApplicationExportTree: ApplicationExportTree { todo!() }`

For more information about this error, try `rustc --explain E0046`.
error: could not compile `openmls_sqlx_storage` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```

This is relevant to the CI workspace tests, since adding the `extensions-draft-08` feature here would cause this build error to occur.

kkohbrok:
I can take a look later today.

PR #1871, opened 2025-11-06
fix: store application export tree after group creation

kkohbrok:
Previously, a freshly created group wouldn't store the application export secret (tree) properly. This PR works towards fixing #1866.


franziskuskiefer (approved):
Thanks, lgtm.

What order shall we get things and at which one should the `extensions-draft-08` feature build again?

PR #1874, opened 2025-11-11
Add AES-GCM support to libcrux provider, and update libcrux dependencies

### LINKED ISSUE #1859: Add AES-GCM ciphersuite to libcrux provider
Add support for aes-gcm in the libcrux provider now that this landed in libcrux.

Either use the aes-gcm crate directly, or start using the aead crate instead.

wysiwys:
* Upgrade the libcrux provider's libcrux dependencies to the latest versions
* Using the `libcrux-aead` crate, add AES-GCM support to `aead_encrypt()` and `aead_decrypt()` in the libcrux provider

This depends on https://github.com/cryspen/hpke-rs/pull/98

Closes #1859

### DIFF
--- a/.github/workflows/benches.yml
+++ b/.github/workflows/benches.yml
@@ -25,7 +25,7 @@ jobs:
           - windows-latest
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           ref: ${{ github.event.pull_request.head.sha }}
       - uses: dtolnay/rust-toolchain@stable
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,7 +48,7 @@ jobs:
           - release
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       # Install the required target
       - uses: dtolnay/rust-toolchain@stable
         with:
@@ -75,7 +75,7 @@ jobs:
 
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
       - uses: taiki-e/install-action@cargo-hack
--- a/.github/workflows/build_test_workspace.yml
+++ b/.github/workflows/build_test_workspace.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Install dependencies
         if: matrix.os == 'macos-latest'
         run: brew install protobuf
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
       - name: Build workspace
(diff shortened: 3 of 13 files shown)


libcrux_crypto/src/crypto.rs
```diff
@@ -96,12 +98,16 @@ impl OpenMlsCrypto for CryptoProvider {
...
-        libcrux_hkdf::expand(alg, prk, info, okm_len)
+        let mut okm = vec![0u8; okm_len];
+
+        libcrux_hkdf::expand(alg, &mut okm, prk, info)
             .map_err(|e| match e {
-                libcrux_hkdf::Error::OkmTooLarge => CryptoError::HkdfOutputLengthInvalid,
-                libcrux_hkdf::Error::ArgumentsTooLarge => CryptoError::InvalidLength,
+                libcrux_hkdf::ExpandError::OutputTooLong => CryptoError::HkdfOutputLengthInvalid,
+                libcrux_hkdf::ExpandError::ArgumentTooLong => CryptoError::InvalidLength,
+                libcrux_hkdf::ExpandError::PrkTooShort => CryptoError::InvalidLength,
```
wysiwys, line 111: Should a new variant of `CryptoError` be added for the `PrkTooShort` case?

franziskuskiefer: This really shouldn't happen (by definition of the protocol). Let's skip for now but add a comment here to maybe add in future.

libcrux_crypto/src/crypto.rs
```diff
@@ -33,8 +33,7 @@ impl CryptoProvider {
 impl OpenMlsCrypto for CryptoProvider {
     fn supports(&self, ciphersuite: Ciphersuite) -> Result<(), CryptoError> {
         match ciphersuite.aead_algorithm() {
-            AeadType::ChaCha20Poly1305 => Ok(()),
-            _ => Err(CryptoError::UnsupportedCiphersuite),
+            AeadType::ChaCha20Poly1305 | AeadType::Aes128Gcm | AeadType::Aes256Gcm => Ok(()),
```
franziskuskiefer, line 36: We need to update `supported_ciphersuites` as well. (Ideally we'd use the same list for both.) We can add `Ciphersuite::MLS_128_DHKEMP256_AES128GCM_SHA256_P256, Ciphersuite::MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519` now.
Given CI time etc. we should make sure we don't make things too slow though.

wysiwys: The `supported_ciphersuites` is now updated to include  `MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519` :+1: 

For adding support for the ciphersuite `MLS_128_DHKEMP256_AES128GCM_SHA256_P256`: does its signature algorithm also need to be supported in the libcrux provider, in `sign()` and `verify_signature()`?

franziskuskiefer: Yes. I thought we do that already. But apparently we don't. Let's do that in a separate PR. We support this https://github.com/cryspen/libcrux/tree/main/ecdsa.

franziskuskiefer (dismissed):
Thanks, lgtm.
Let's do a follow-up for ecdsa.

franziskuskiefer (approved):

PR #1878, opened 2025-11-14
update dependabot

franziskuskiefer:
Dependabot hasn't been running for a while. It runs into issues I don't quite understand. But let's try the full workspace, we should check more than the main openmls crate anyway.


raphaelrobert (approved):
Let's try this.

PR #1898, opened 2025-11-19
Implement HPKE safe encrypt/decrypt with label (MLS Extensions Draft)

wysiwys:
This pull request implements the SafeEncryptWithLabel and SafeDecryptWithLabel functions, as defined in Section 4.2 of the MLS Extensions Draft.
- https://messaginglayersecurity.rocks/mls-extensions/draft-ietf-mls-extensions.html#section-4.2

This new functionality wraps some of the existing `encrypt_with_label()` and `decrypt_with_label()` functionality in this module. Because the label for the new safe encrypt/decrypt functionality is constructed in a slightly different way, these existing methods are split into their core functionality (`encrypt_with_label_internal()` and `decrypt_with_label_internal()`) as well as the `EncryptContext` construction. 

This pull request also adds a test for this new functionality

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -84,4 +84,4 @@ jobs:
       - name: Cargo hack
         run: |
           cargo hack check --feature-powerset --no-dev-deps --verbose -p openmls \
-          --group-features test-utils,crypto-debug,content-debug,crypto-subtle,backtrace,libcrux-provider,libcrux-provider-js,sqlite-provider
+          --group-features test-utils,crypto-debug,content-debug,backtrace,libcrux-provider,libcrux-provider-js,sqlite-provider
--- a/book/book.toml
+++ b/book/book.toml
@@ -1,6 +1,5 @@
 [book]
 authors = ["OpenMLS Authors"]
 language = "en"
-multilingual = false
 src = "src"
 title = "OpenMLS Book"
--- a/openmls/src/ciphersuite/signable.rs
+++ b/openmls/src/ciphersuite/signable.rs
@@ -50,7 +50,7 @@ pub enum SignatureError {
 /// This trait must be implemented by all structs that contain a self-signature.
 pub trait SignedStruct<T> {
     /// Build a signed struct version from the payload struct.
-    fn from_payload(payload: T, signature: Signature) -> Self;
+    fn from_payload(payload: T, signature: Signature, serialized_payload: Vec<u8>) -> Self;
 }
 
 /// The `Signable` trait is implemented by all struct that are being signed.
@@ -75,19 +75,23 @@ pub trait Signable: Sized {
         let payload = self
             .unsigned_payload()
             .map_err(|_| SignatureError::SigningError)?;
-        let payload = match SignContent::new(self.label(), payload.into()).tls_serialize_detached()
-        {
-            Ok(p) => p,
-            Err(e) => {
-                log::error!("Serializing SignContent failed, {e:?}");
-                return Err(SignatureError::SigningError);
-            }
-        };
+        let payload_sign_content =
+            match SignContent::new(self.label(), payload.clone().into()).tls_serialize_detached() {
+                Ok(p) => p,
+                Err(e) => {
+                    log::error!("Serializing SignContent failed, {e:?}");
+                    return Err(SignatureError::SigningError);
+                }
+            };
         let signature = signer
-            .sign(&payload)
+            .sign(&payload_sign_content)
             .map_err(|_| SignatureError::SigningError)?;
 
-        Ok(Self::SignedOutput::from_payload(self, signature.into()))
+        Ok(Self::SignedOutput::from_payload(
+            self,
+            signature.into(),
+            payload,
+        ))
     }
 }
 
--- a/openmls/src/ciphersuite/signature.rs
+++ b/openmls/src/ciphersuite/signature.rs
@@ -139,7 +139,7 @@ impl Signature {
 }
 
 impl<T> SignedStruct<T> for Signature {
-    fn from_payload(_payload: T, signature: Signature) -> Self {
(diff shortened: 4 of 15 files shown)


openmls/src/ciphersuite/tests_and_kats/tests.rs
```diff
@@ -78,3 +78,86 @@ fn test_hpke_seal_open() {
...
+        )
+        .expect("error deriving hpke key pair");
+
+    const COMPONENT_ID: u16 = 1;
+
+    let ciphertext = hpke::safe_encrypt_with_label(
+        &kp.public,
+        COMPONENT_ID,
+        "label",
+        &[1, 2, 3],
```
franziskuskiefer: Put these two into constants to re-use in decrypt.

openmls/src/ciphersuite/tests_and_kats/tests.rs
```diff
@@ -78,3 +78,86 @@ fn test_hpke_seal_open() {
...
+        ciphersuite,
+        provider.crypto(),
+    )
+    .expect("Unexpected error while decrypting a valid ciphertext.");
+    assert_eq!(decrypted_payload, plaintext);
+
+    let mut broken_kem_output = ciphertext.kem_output.clone();
+    broken_kem_output.pop();
+    let mut broken_ciphertext = ciphertext.ciphertext.clone();
+    broken_ciphertext.pop();
```
franziskuskiefer: This changes the size so will probably be rejected because of that, rather than an actual decryption error. You could xor a byte with `0xFF` to get a bad ciphertext.

openmls/src/ciphersuite/hpke.rs
```diff
@@ -54,6 +57,11 @@ pub(crate) enum Error {
     /// Decryption failed.
     #[error("Decryption failed.")]
     DecryptionFailed,
+    #[cfg(feature = "extensions-draft-08")]
+    // TODO: these variants could be a different error type.
+    /// Serializing label failed
+    #[error("Serializing label failed.")]
+    LabelSerializationFailed,
```
franziskuskiefer: I think this is always `MissingBoundCheck`? This should only fail if something too large was put in.

openmls/src/ciphersuite/hpke.rs
```diff
@@ -71,17 +79,34 @@ impl From<CryptoError> for Error {
...
 
 impl EncryptContext {
     /// Create a new [`EncryptContext`] from a string label and the content bytes.
-    pub fn new(label: &str, context: VLBytes) -> Self {
+    /// Ensures that the prefix LABEL_PREFIX is prepended to the label.
+    pub(crate) fn new(label: &str, context: VLBytes) -> Self {
         let label_string = LABEL_PREFIX.to_owned() + label;
         let label = label_string.as_bytes().into();
         Self { label, context }
     }
```
franziskuskiefer, line 90: Without newlines the code gets really hard to read.

```suggestion
    }

```

openmls/src/ciphersuite/hpke.rs
```diff
@@ -123,6 +160,49 @@ pub(crate) fn encrypt_with_label(
...
+            base_label: base_label.as_bytes().into(),
+            component_id,
+            label: label.as_bytes().into(),
+        }
+    }
+}
+
+#[allow(unused)]
+#[cfg(feature = "extensions-draft-08")]
+pub(crate) fn safe_encrypt_with_label(
```
franziskuskiefer: These should actually be public for applications to use.

openmls/src/ciphersuite/hpke.rs
```diff
@@ -123,6 +160,49 @@ pub(crate) fn encrypt_with_label(
     Ok(cipher)
 }
 
+/// Label for safe encryption/decryption as defined in Section 4.2 of the MLS Extensions draft
+#[derive(Debug, TlsSerialize, TlsSize)]
+#[cfg(feature = "extensions-draft-08")]
+pub(crate) struct ComponentOperationLabel {
```
franziskuskiefer: This is also used for other things, so we should move this next to the `ComponentId`. Maybe we should move both of them into a new module like `components`.

openmls/src/ciphersuite/hpke.rs
```diff
@@ -123,6 +160,49 @@ pub(crate) fn encrypt_with_label(
...
+    /// LABEL_PREFIX + "Application"
+    base_label: VLBytes,
+    component_id: ComponentId,
+    label: VLBytes,
+}
+#[cfg(feature = "extensions-draft-08")]
+impl ComponentOperationLabel {
+    /// Creates a new ComponentOperationLabel, prefixed with LABEL_PREFIX
+    pub fn new(component_id: ComponentId, label: &str) -> Self {
+        let base_label = super::LABEL_PREFIX.to_owned() + "Application";
```
franziskuskiefer: This will actually result in a wrong encoding.
Encoding this will be "MLS 1.0 Application" as bytes, i.e. a string of 19 bytes. But the expected encoding is two byte strings, i.e. "MLS 1.0 " of length 8, and the "Application" of length 11.
Also, this may be used somewhere else without that prefix. So we should keep them separate.

openmls/src/ciphersuite/hpke.rs
wysiwys: Would this module be a new module in the `openmls` crate root, or would it also make sense to add it as a submodule of `extensions`?

openmls/src/ciphersuite/hpke.rs
wysiwys: Ah, right, for this error type, there is already a`From<tls_codec:Error>` implementation :+1: Since the `Serialize` implementation for the label is derived using `TlsSerialize`, is it correct that that means that no other `tls_codec::Error` variants should be expected?

PR #1900, opened 2025-11-19
Add handling and helpers for GREASE values

erskingardner:
This PR implements GREASE (Generate Random Extensions And Sustain Extensibility) support as defined in [RFC 9420 Section 13.5](https://www.rfc-editor.org/rfc/rfc9420.html#section-13.5). GREASE injects reserved values (`0x0A0A`, `0x1A1A`, etc.) to ensure implementations handle unknown values correctly and maintain forward compatibility.

### Key Changes

- Added `Grease(u16)` variants to `ProposalType`, `ExtensionType`, and `CredentialType` enums
- Implemented `is_grease()` methods for all GREASE-capable types including `VerifiableCiphersuite`
- **GREASE values are now automatically injected into all KeyPackages and MlsGroups**, regardless of whether custom or default capabilities are used
- At least one GREASE value is added to each capability list (ciphersuites, extensions, proposals, credentials) unless already present
- GREASE values are recognized during deserialization and filtered during validation
- Members with different GREASE values can interoperate seamlessly

### Testing & Documentation

- Added integration tests covering GREASE scenarios
- Updated existing tests to filter GREASE values when comparing capabilities
- Added user manual documentation with usage examples at `book/src/user_manual/grease.md`
- Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,13 +19,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Added comprehensive unit and integration tests for GREASE handling
   - Added user manual documentation for GREASE support
 
-
 ### Fixed
 - [#1868](https://github.com/openmls/openmls/pull/1868): The implementation of [valn0311](https://validation.openmls.tech/#valn0311), was updated to check support for all non-default proposals, instead of only checking support for Custom proposals.
-
-### Fixed
-
 - [#1871](https://github.com/openmls/openmls/pull/1871): Fixed a bug where the application export tree (part of the `extensions-draft-08` feature) was not stored properly after group creation.
+
+### Changed
+- [#1874](https://github.com/openmls/openmls/pull/1874): Changed `ProposalType`, `ExtensionType`, and `CredentialType` enums to include `Grease(u16)` variant.
 
 ## 0.7.1 (2025-09-24)
 
--- a/book/src/user_manual/grease.md
+++ b/book/src/user_manual/grease.md
@@ -10,6 +10,22 @@ GREASE values are special reserved values that follow a specific pattern (`0x0A0
 2. **Prevent ossification**: Help maintain forward compatibility by exercising unknown value handling paths
 3. **Identify bugs**: Catch implementations that incorrectly assume all possible values are known
 
+### RFC-defined vs. Custom GREASE Values
+
+The [15 values defined in RFC 9420](#grease-values) are the "official" GREASE values. When OpenMLS generates GREASE values (e.g., via `with_grease()`), it uses these RFC-defined values. The `is_grease()` method only returns `true` for these specific values.
+
+However, **any** unknown value can serve a similar purpose. The difference is that for non-RFC values, we cannot distinguish whether they are intentionally injected "GREASE-like" values or genuinely unknown identifiers from future protocol extensions.
+
+### The Purpose of GREASE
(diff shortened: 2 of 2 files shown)


franziskuskiefer (changes requested):
Just leaving a request changes here to reflect the discussion on Zulip.

CHANGELOG.md
```diff
@@ -11,6 +11,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#1855](https://github.com/openmls/openmls/pull/1855): Added the `swap_members()` method to `MlsGroup` to replace members in a group, as well as the `WelcomeCommitMessages` and `SwapMembersError` structs.
 - [#1868](https://github.com/openmls/openmls/pull/1868): Implemented AppEphemeral functionality as defined in the MLS Extensions draft and replaced the existing AppAck proposal with the AppAck object, which can now be conveyed inside an AppEphemeral proposal. These features are behind the `extensions-draft-08` feature flag.
 - [#1874](https://github.com/openmls/openmls/pull/1874): In the `openmls_libcrux_crypto` provider, added AES-GCM support.
+- Implemented GREASE (Generate Random Extensions And Sustain Extensibility) support as defined in [RFC 9420 Section 13.5](https://www.rfc-editor.org/rfc/rfc9420.html#section-13.5):
+  - Added `Grease(u16)` variants to `ProposalType`, `ExtensionType`, and `CredentialType` enums
```
franziskuskiefer, line 15: These are actually changes to existing enums. So this should also get mentioned in the Changed section.

book/src/user_manual/grease.md
```diff
@@ -0,0 +1,256 @@
...
+GREASE (Generate Random Extensions And Sustain Extensibility) is a mechanism defined in [RFC 9420 Section 13.5](https://www.rfc-editor.org/rfc/rfc9420.html#section-13.5) to ensure that implementations properly handle unknown values and maintain protocol extensibility.
+
+## What is GREASE?
+
+GREASE values are special reserved values that follow a specific pattern (`0x0A0A, 0x1A1A, 0x2A2A, ..., 0xEAEA`) and are used to:
+
+1. **Test extensibility**: Ensure implementations don't reject messages containing unknown values
+2. **Prevent ossification**: Help maintain forward compatibility by exercising unknown value handling paths
+3. **Identify bugs**: Catch implementations that incorrectly assume all possible values are known
+
```
franziskuskiefer, line 12: I think we need to make clear here that there are

- the grease values defined in the RFC. And when not using a manually defined value, that's what's being used by OpenMLS. Also any check like `is_grease()` is only defined for these values.
- any other grease value. For these we can't distinguish whether they are supposed to be grease values or just unknown IDs.

But since the entire point of grease values is *NOT* to check for them. We need to make sure that it's clear that this is not how they should be used.
But instead that this allows applications to inject grease values to discourage other implementations from failing on unknown values and hard coding too many things.

There's some more text about this further down. Maybe linking there would be ok as well.

erskingardner: I added some extra text here to clarify and linked the section below.

CHANGELOG.md
erskingardner: Removed the duplicate `### Fixed` section and added a `### Changed` section.

franziskuskiefer (approved):
Thanks!

Issue #160, opened 2020-10-26
Optimize Signable


kkohbrok:
Signable currently encodes the contents of the objects before signing. In many cases, this leads to the object being encoded twice, once to sign it and once to send it over the wire. Maybe `sign` could also return the encoded payload so it can be reused?

PR #1901, opened 2025-11-20
feat: optimize Signable to reuse serialized payload

WaterWhisperer:
modifies the `Signable` and `SignedStruct` traits to allow reusing the serialized payload generated during the signing process.
This prevents double serialization of payloads (once for signing, once for wire transmission) for `GroupInfo` and `KeyPackage`.

Fixes: #160


franziskuskiefer:
Merging with CI error (see https://github.com/openmls/openmls/issues/1905)

PR #1903, opened 2025-11-25
Improve the error return types in TreeSync

WaterWhisperer:
Fixes: #804

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,7 +7,7 @@ on:
   pull_request:
   workflow_dispatch:
 
-concurrency: 
+concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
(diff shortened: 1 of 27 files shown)


WaterWhisperer:
Seems like almost all ci failed due to "failed to select a version for `libcrux-macros` which could resolve this conflict"

openmls/src/treesync/errors.rs
```diff
@@ -72,6 +72,12 @@ pub enum ApplyUpdatePathError {
     /// External Committer used the wrong index.
     #[error("External Committer used the wrong index.")]
     InconsistentSenderIndex,
+    /// The own leaf is not in the tree.
+    #[error("The own leaf is not in the tree.")]
+    MissingOwnLeaf,
```
franziskuskiefer, line 77: This requires an entry in the changelog. Can you add that?

openmls/src/treesync/diff.rs
```diff
@@ -458,7 +462,9 @@ impl TreeSyncDiff<'_> {
         // Compute the filtered copath corresponding to the given filtered
         // direct path.
         let filtered_copath = self.filtered_copath(leaf_index);
-        debug_assert_eq!(filtered_copath.len(), path.len());
+        if filtered_copath.len() != path.len() {
```
franziskuskiefer, line 466: We still want to keep the debug assert.

franziskuskiefer (changes requested):
Thanks, generally lgtm.

The error should go away when #1911 is merged.

openmls/src/treesync/errors.rs
WaterWhisperer: Sure, thanks for the review!

franziskuskiefer (approved):

PR #1913, opened 2025-12-02
sub group branching

franziskuskiefer:
Adds support for [subgroup branching](https://www.rfc-editor.org/rfc/rfc9420.html#name-subgroup-branching): seeding a new group from a subset of a parent group's members, mixing the parent's resumption PSK into the new key schedule.

API
- `CommitBuilder::branch(rand, parent)`: branches a fresh group off parent, adding a Branch-usage resumption PSK (fresh KDF.Nh nonce) to the initial commit and injecting the parent's PSK secret.
- `StagedWelcome::new_from_branch`: receiver side; injects the parent PSK and enforces the RFC join checks.
- `CommitBuilder::propose_psks`: helper for regular PSK proposals.

Validation
- Branch PSKs only accepted in the initial commit (epoch 0); duplicate PSK IDs rejected (ValSem403).
- Receiver checks: version/ciphersuite match parent, subgroup at epoch 1, every subgroup member matches a parent member (gated by check_members).
- `load_psks` resolves the branch PSK at sentinel epoch 0.

New errors
- `WelcomeError::{SubgroupParameterMismatch, SubgroupEpochInvalid, SubgroupLeafMismatch}`
- `PskError::{Duplicate, NotAllowed}`

Tests:
- `openmls/tests/sub-groups.rs` covering the branch flow and receiver-side validation.

Follow ups:
- [x] Add interop tests #2127


openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -253,6 +257,32 @@ impl<'a> CommitBuilder<'a, Initial, &mut MlsGroup> {
...
+        self = self.propose_psks([psk_id]);
+
+        // We clear the resumption PSK store first and then add the branching PSK.
+        self.group.borrow_mut().resumption_psk_store.clear();
+        self.group
+            .borrow_mut()
+            .resumption_psk_store
+            .add(0.into(), secret);
+        self
+    }
```
keks, line 401: Should these already check that the PSKs are suitable for the use case and do not contradict?

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -345,6 +376,27 @@ impl<'a, G: BorrowMut<MlsGroup>> CommitBuilder<'a, Initial, G> {
...
+
+    /// Branches the group by adding a branch PreSharedKey proposals.
+    pub fn branch(mut self, psk_id: PreSharedKeyId) -> Self {
+        self.stage.branch = true;
+        self = self.propose_psks([psk_id]);
+        self
+    }
+
```
keks: I don't think these should be here. The LoadedPsk phase is too late to add more proposals, especially PSK proposals, because then these are not loaded anymore and the later computations won't have the keys (just the IDs).

openmls/src/group/mls_group/group_builder.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// The initial state.
+    init: Init<'a>,
+
+    /// The bundle used for decrypting this welcome.
+    key_package_bundle: KeyPackageBundle,
+}
+
+/// The state after decrypting the group secrets in the welcome message.
+pub struct Decrypted<'a> {
+    /// The initial state.
```
keks:
```suggestion
    /// The previous state.
```

openmls/src/group/mls_group/group_builder.rs
```diff
@@ -0,0 +1,204 @@
...
+    key_package: KeyPackageState<'a>,
+
+    /// The decrypted group secrets.
+    group_secrets: GroupSecrets,
+}
+
+/// The state after loading and generating the PSK secret.
+pub struct Psks<'a> {
+    /// The initial state.
```
keks:
```suggestion
    /// The previous state.
```

keks (commented):
Some early feedback. I think this PR is doing a lot, and maybe we can talk offline about the bigger approach here, which builders are responsible for what, etc.

openmls/src/group/mls_group/commit_builder.rs
franziskuskiefer: Oh yeah, this bit shouldn't be here.

openmls/src/group/mls_group/commit_builder.rs
franziskuskiefer: it could ... but we need the check later again because more psks may have been added after this. So I'm not sure if checking here gives us so much.

kkohbrok:
I've mostly looked at the API so far. Two questions: 

- Is there a reason not to do group- and commit creation in one go, i.e. make `branch` part of the MlsGroup builder?
- As I understand it, when processing a Welcome to create a branch, the application has to know somehow (OOB?) that the incoming Welcome is a branch welcome. It would make more sense to me if the client would/could learn as part of Welcome processing that the Welcome is creating a branched group. I believe you don't have to pass in the full group as input here. Instead, it should be possible to load the Resumption PSK store from the storage provider during processing.

franziskuskiefer:
The internals aren't that interesting, so yes the API is probably the most interesting.

> * Is there a reason not to do group- and commit creation in one go, i.e. make `branch` part of the MlsGroup builder?

Do you mean something like `MlsGroupBuilder::new().sub_group(parent ...)` for the committer? No good reason, I think that should be doable as well. I think having the manual flow would be nice as well. But having a builder on the group for a somewhat higher level API is a good idea.

> * As I understand it, when processing a Welcome to create a branch, the application has to know somehow (OOB?) that the incoming Welcome is a branch welcome. It would make more sense to me if the client would/could learn as part of Welcome processing that the Welcome is creating a branched group. I believe you don't have to pass in the full group as input here. Instead, it should be possible to load the Resumption PSK store from the storage provider during processing.

We don't really have a way to get the resumption PSK of a different group. But you mean we allow the group to access the group secrets of the parent group? I'm not sure if we really want to do that. And the plumbing will be pretty ugly.

Instead of passing in the group, we can of course pass in only the necessary information. But we need (to check) a couple of things of the parent group. So I think the alternative would be to create a sub-group package from the parent group with all that's needed and pass that in instead. wdyt?

kkohbrok:
> We don't really have a way to get the resumption PSK of a different group. But you mean we allow the group to access the group secrets of the parent group? I'm not sure if we really want to do that. And the plumbing will be pretty ugly.

I think we do. Can't you just use the storage provider to pull the ResumptionsPskStore of another group from the storage provider and pull the specific resumption PSK you need from there?

franziskuskiefer:
> Can't you just use the storage provider to pull the ResumptionsPskStore of another group from the storage provider and pull the specific resumption PSK you need from there?

Yes we could. But that would block the other group. I'm not sure I like mixing that. I think passing things in explicitly will be nicer. But I'll try both

kkohbrok:
Fair point. These cross-group interactions are gnarly. No strong opinion here. We can also start with the API you propose and then see how it works for applications.

franziskuskiefer:
I did a version with a branch info struct.
While doing I also added a check for the epoch to avoid running into a bad state without noticing. That's another reason to not just take the PSK from parent group. The parent group may have moved on. Passing in the info explicitly avoids racing other commits on the parent group.

kkohbrok:
The parent group having moved on should be fine. Isn't that why we have a store that retains past keys? But I agree that this would be easier if we stored individual keys rather than the whole store as a blob.

book/src/user_manual/sub-groups.md
```diff
@@ -0,0 +1,95 @@
+# Sub-group branching
+
+Sometimes a subset of the members in a group want to continue in a separate, smaller
+group without going through a full fresh key exchange. MLS supports this with
```
kkohbrok: It is a full fresh key exchange, though, right? It's just that additionally, the PSK is injected.

kkohbrok (approved):
Looks good, thanks! Just one nit in the book.

PR #1914, opened 2025-12-03
ci update

### LINKED ISSUE #1896: Reduce CI time for tests
The tests take too long on CI right now.

franziskuskiefer:
CI for tests takes too long right now. I parallelised it and updated the matrix a bit. This should be testing the same things as before and a little more. Each run only takes about 5 minutes now. But because there are so many different targets, it still takes about 20 minutes total. We can further reduce this with other runners or moving some targets to nightly jobs.

I also changed the `openmls_test` macro to run only with rust crypto for now to speed up local tests. I'd like to change this to libcrux at some point, but that's a bigger change. CI still runs both for now.

Fixes #1896

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,7 +7,7 @@ on:
   pull_request:
   workflow_dispatch:
 
-concurrency: 
+concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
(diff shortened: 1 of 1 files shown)


bennycode:
Would using [larger Ubuntu runners](https://docs.github.com/en/actions/how-tos/manage-runners/larger-runners/manage-larger-runners#adding-a-larger-runner-to-an-organization) help cut down the 5 minute runtime?

raphaelrobert:
I think we should drop either `debug` or `release` mode tests, they are redundant.

franziskuskiefer:
> I think we should drop either `debug` or `release` mode tests, they are redundant.

They are not redundant. They are different code paths in some areas.
We do not always need to run everything. But that goes for all combinations.

Issue #1215, opened 2023-01-22
[Feature request] Return indices of added members when merging a commit


kkohbrok:
At the moment, applications have to explicitly look up the new members in the group to figure out which index they are in. Returning the leaf indices of new members when merging a commit makes it easier for the application to maintain a mapping between Members and their indices.

PR #1918, opened 2025-12-10
feat: Return indices of added members when merging a commit

WaterWhisperer:
Fixes: #1215


kkohbrok:
Thanks for the PR @WaterWhisperer!

We discussed the PR and the underlying issue internally and came to the conclusion that it's not quite in line with our API anymore. Instead of returning the indices at the point of merging a commit, they should instead be made accessible from the `StagedCommit` struct.

Sorry you wasted time on this stale issue. We do review open issues regularly, but apparently this snuck past us. I'll make sure to change the issue accordingly and review all issues marked as "good first issue".

PR #1923, opened 2026-01-07
feat: async

raphaelrobert:
This PR introduces maybe_async for the storage provider. A few things to note:
 - the changes to the core library are fairly minimal
 - the memory & sqlite providers remain exclusively "sync"
 - the sqlx provider is now exclusively "async"
 - since both "sync" and "async" variants cannot be simultaneously be used in a workspace, the sqlx provider was moved out of the workspace
 - there's a new crate to test the "async" API by using the sqlx provider


tylerhawkes:
@raphaelrobert We're evaluating this branch for XMTP and merged it onto our fork. We hit three bugs, then prototyped the two receiver shapes that would prevent the third. Everything below reproduces on `raphael/async` at 18a7243.

## `openmls_memory_storage` forces the whole graph to sync

`memory_storage/Cargo.toml:13` pins `openmls_traits = { workspace = true, features = ["sync"] }`, and `openmls_rust_crypto/Cargo.toml:14` depends on `memory_storage` unconditionally. `maybe-async/is_sync` is a global switch on a proc-macro crate, so any dependency graph that reaches `openmls_rust_crypto` compiles every trait in sync shape. `openmls_sqlx_storage` is async, so it fails with 53 signature mismatches, and `openmls_async_tests` does not build at all.

We annotated `MemoryStorage`'s `StorageProvider` impl with `#[maybe_async::maybe_async(AFIT)]` and gave `memory_storage` and `openmls_rust_crypto` `sync`/`async` passthrough features.

## Nothing in CI compiles the async crates

The root `Cargo.toml` carries `exclude = ["sqlx_storage"]`, and `openmls_async_tests` is not listed at all. That is why the bug above survived. We understand the exclusion, since their async traits conflict with the workspace's sync default. A separate job that builds both would catch this class of drift.

## Storage futures are not `Send`

`SqliteStorageProvider` holds `RefCell<&mut SqliteConnection>` at `sqlx_storage/src/lib.rs:41` and keeps the borrow across await points, at 56 `borrow_mut()` sites in `sqlx_storage/src/storage_provider.rs`. `RefCell` is not `Sync`, so no MLS operation backed by this provider can be handed to `tokio::spawn`. Any server running a task per session is blocked.

`openmls_async_tests` misses it because that binary drives everything from `main`'s future and never spawns.

We swapped the `RefCell` for `tokio::sync::Mutex`, which is `Send + Sync` and meant to be held across await, and added a compile-time `Send` assertion to the test crate.

## Pools do not need a cell at all

`&sqlx::Pool` implements `Executor`, so a pool-backed provider satisfies the existing `&self` methods with no interior mutability. Shipping a pool flavor next to the borrowed-connection one would give most clients a `Send` provider by default, and keep the borrowed form for callers who need the provider to join a transaction they opened.

## Receiver shape for 0.9

All of this follows from `StorageProvider` methods taking `&self` while a bare sqlx `Connection` only executes through `&mut`. Something has to close that gap, and the choice of primitive decides whether the futures are `Send`. We built both alternatives to get real numbers. Both are breaking, so 0.9 is the moment if either happens.

Measure with an isolated `CARGO_TARGET_DIR`. A shared one produces phantom `&`-vs-`&mut` errors across sibling checkouts, which cost us time twice.

### `&mut self`

Branch `tyler/exp-storage-mut-self`. `OpenMlsProvider::storage` (`traits/src/traits.rs:35`) becomes a `&mut self` accessor. No GAT needed.

`openmls` compiles green in async, sync and test-utils modes. The library conversion is 67 files and about 790 lines.

The call sites are free. All 256 `.storage()` sites in `openmls/src` are byte-identical afterwards, because implicit reborrow of `&mut Provider` absorbs them, and the whole library needed two explicit `&mut *x` reborrows. We had budgeted this as the bulk of the work.

Both provider shapes then work with no cell, pool-backed and borrowed `&mut Connection`, and their futures are `Send` and `tokio::spawn`-able. That removes 56 lock acquisitions and the `run_task` helper from `openmls_sqlx_storage`.

The cost is 1383 unique error locations in test, bench, example and bin code. It does not respond to mechanical fixing: 846 regex substitutions closed 220 of them, because each `&mut` added exposes a new `E0596` or `E0499` one level up. The cause is harnesses that hold a provider in a struct and hand out overlapping shared borrows, not `.storage()`.

### `self` by value, the serde `Serializer` pattern

Branch `tyler/exp-storage-by-value`. Implementors pick their own receiver:

```rust
impl StorageProvider for &PostgresPoolProvider      // shared, concurrent
impl StorageProvider for &mut SqliteStorageProvider // exclusive, no cell
```

This is the smaller diff at 38 files and about 830 lines, and the only shape where one trait serves both providers natively. `&P` is `Copy`, so a pool handle drives two live futures with no reborrow at all.

It needs a GAT:

```rust
type StorageError: Debug + Error;
type StorageProvider<'a>: storage::StorageProvider<{CURRENT_VERSION}, Error = Self::StorageError> where Self: 'a;
fn storage(&mut self) -> Self::StorageProvider<'_>;
```

Hoisting `StorageError` out is mandatory, since `Self::StorageProvider<'?>::Error` cannot be named without picking a lifetime.

A provider that itself carries a lifetime parameter then fails `Send` with "implementation of OpenMlsProvider is not general enough". That is a GAT well-formedness limit rather than a genuinely `!Send` future, and it clears only when the provider has no lifetime parameter. Two of our three providers carry one, so this is the reason we are not adopting it.

The friction is also not reborrows. rustc auto-reborrows `&mut` receivers even for by-value `self`. The failures land in functions holding an opaque `Storage: StorageProvider` by value, where a reborrow cannot be expressed, and they became 10 signature changes: `MlsGroup::{load,delete,store}`, `PublicGroup::{load,delete,store,from_external,merge_commit}` and `load_psks` twice. `PublicGroup`'s storage methods widen from `PublicStorageProvider` to a full `OpenMlsProvider`, which is a capability regression for public-group-only clients.

Its test tail is not comparable to the 1383 above. `openmls/src/test_utils` still has 47 errors, so the downstream crates are blocked behind them and unmeasured.

### What both variants hit

A function returning an elided-lifetime type alongside `&mut Provider` silently pins the provider exclusively, so the next call in the chain is `E0499`. Two independent conversions hit this, which points at `&mut` itself rather than either encoding. The case we found is `build_group` at `openmls/src/group/mls_group/commit_builder/external_commits.rs:136`.

`CommitBuilder::load_psks(storage: &'a Storage)` at `openmls/src/group/mls_group/commit_builder.rs:315` ties the storage borrow to the builder's lifetime even though the builder never keeps it. Under `&self` that is invisible. Under `&mut` it freezes the provider for the rest of every commit chain, and decoupling it dropped one conversion from 37 errors to 8 in a single edit.

Expressions mixing `crypto()` and `storage()` become `E0502` and need the crypto call hoisted to a `let`.

### Whether it is worth doing

Branch `tyler/exp-storage-pool-baseline` shows a pool-backed provider satisfying the current `&self` trait with no interior mutability, on an otherwise unmodified tree. So the receiver change buys a cell-free borrowed connection and nothing else. Clients that only need a pool are already fine today.

## Branches

The three fixes are on `xmtp/openmls` `tyler/maybe-async`, which also adds a PostgreSQL backend to `openmls_sqlx_storage` behind a `postgres` feature. That is what made the `Send` behavior matter to us. The three experiment branches are based on it rather than on `raphael/async`, so a diff against upstream will include that work.

Happy to open PRs for any of this, whichever pieces are useful to you. The three fixes are the obvious ones, but the same goes for the pool provider, the CI job, or either receiver shape if you decide to take one for 0.9.

raphaelrobert:
Thanks for the feedback. Note that this is still in an early stage and should not be expected to work flawlessly. It's out of sync with the main branch and will likely see substantial changes. We'll resume work on this after v0.9 is published.

PR #1925, opened 2026-01-12
[Proof of concept] Per-group locking in storage provider

wysiwys:
As a basis for discussion, this draft PR includes a minimal proof of concept of per-group locking functionality in the storage provider. Allowing the storage provider to be locked for a given `GroupId` is a step towards having async APIs in OpenMLS, as well as towards allowing safe parallel processing of multiple groups.

The main proof-of-concept is implemented in the `openmls_memory_storage` crate. The APIs in the `openmls_traits` crate have also been modified/extended, so that the lock manager is retrieved instead of the storage provider, and to provide trait APIs for retrieving a lock from the lock manager. 

In addition, so that an example can be created to show how the new APIs on the `OpenMlsProvider` look, this PR also adapts some APIs in the `openmls_libcrux_crypto` crate). 

This draft PR didn't make other updates to the crates in the workspace yet (so some other crates will not be able to build yet).

### DIFF
--- a/memory_storage/Cargo.toml
+++ b/memory_storage/Cargo.toml
@@ -18,7 +18,7 @@ serde = { version = "1.0", features = ["derive"] }
 log = { version = "0.4" }
 hex = { version = "0.4", features = ["serde"], optional = true }
 base64 = { version = "0.22", optional = true }
-# TODO: which features are needed?
+# for `tokio::sync::Mutex`
 tokio = { version = "1.49.0", features = ["full"] }
 
 [features]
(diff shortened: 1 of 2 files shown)


memory_storage/tests/provider.rs
```diff
@@ -0,0 +1,46 @@
...
+    Missing,
+    Operational,
+}
+impl traits::GroupState<CURRENT_VERSION> for TestGroupState {}
+impl Entity<CURRENT_VERSION> for TestGroupState {}
+
+/// A basic example with `openmls_libcrux_provider::Provider`, which uses the
+/// `MemoryStorageManager`.
+#[tokio::test]
+async fn basic_example() {
```
wysiwys, line 22: The main example for this draft PR is here, showing how the `OpenMlsProvider` can be used  to retrieve the lock manager, to retrieve a lock for a specific `GroupId`.

PR #1926, opened 2026-01-12
Update rand dependency for `openmls` tests and `ds-lib` crate

wysiwys:
This PR updates the `rand` dependency in the tests for the `openmls` crate, as well as in the `ds-lib` crate (the companion library for the proof-of-concept delivery service).

It also replaces calls to deprecated functions, and adapts tests and test utilities to use the updated APIs from `rand=0.9`.

See Dependabot PR #1885


franziskuskiefer:
@wysiwys looks like some of the jobs are not happy with it though.

franziskuskiefer (dismissed):
Thanks, lgtm. Can you add a line to the changelog

franziskuskiefer:
@wysiwys ready to merge with the changelog update

PR #1927, opened 2026-01-13
Enforce allowed extension types using Rust types

### LINKED ISSUE #1830: GroupContextExtensions proposal with invalid extension types can be applied
At the moment, a GroupContextExtensions proposal can be used to add `application_id`, `ratchet_tree`, and `external_pub` extensions to the GroupContext, although these extension types are not valid in the GroupContext ([MLS Extension Types](https://www.rfc-editor.org/rfc/rfc9420.html#section-17.3-4))

tbehner:
Closes #1830 

Extends PR #1834 by integrating @keks’s work on extension validation.

### DIFF
--- a/openmls/src/framing/private_message_in.rs
+++ b/openmls/src/framing/private_message_in.rs
@@ -46,6 +46,13 @@ pub struct PrivateMessageIn {
 }
 
 impl PrivateMessageIn {
+    /// Retrieve the additional authenticated data (AAD) from the [`PrivateMessageIn`].
+    ///
+    /// NOTE: This AAD is unverified.
+    pub fn aad(&self) -> &[u8] {
+        self.authenticated_data.as_slice()
+    }
+
     /// Decrypt the sender data from this [`PrivateMessageIn`].
     pub(crate) fn sender_data(
         &self,


openmls/src/extensions/mod.rs
```diff
@@ -396,11 +465,81 @@ impl TryFrom<Vec<Extension>> for Extensions {
...
+    type Error = ExtensionTypeNotValidInGroupInfoError;
+
+    fn validate_extension_type(
+        ext: &Extension,
+    ) -> Result<(), ExtensionTypeNotValidInGroupInfoError> {
+        matches!(
+            ext.extension_type(),
+            ExtensionType::RatchetTree | ExtensionType::ExternalPub | ExtensionType::Unknown(_)
+        )
+        .then_some(())
```
kkohbrok: This seems a bit contrived. Not a strong opinion, but I'd prefer an `if .. else` for readability. Especially if you add something additional like `.unwrap_or` as done below.

openmls/src/extensions/mod.rs
```diff
@@ -396,11 +465,81 @@ impl TryFrom<Vec<Extension>> for Extensions {
...
+// https://validation.openmls.tech/#valn1601
+impl ExtensionValidator for LeafNode {
+    type Error = ExtensionTypeNotValidInLeafNodeError;
+
+    fn validate_extension_type(
+        ext: &Extension,
+    ) -> Result<(), ExtensionTypeNotValidInLeafNodeError> {
+        ext.extension_type()
+            .is_valid_in_leaf_node()
+            .unwrap_or(true) // allow extension types we don't know
```
kkohbrok: I think this is what we do everywhere. Maybe it's worth changing `is_valid_in_leaf_node` to just return `true` if the extension is unknown?

kkohbrok (approved):
Very nice! Just two nits.

openmls/src/extensions/mod.rs
keks: I was sure we are using this distinction somewhere, but it looks like you are right. Let's change it to just return a bool.

tbehner: For Unknown extensions, we cannot decide if the extension is allowed or not (as they are defined by the application), right? Therefore, we would have to apply ternary logic. We are already using this here: https://github.com/tbehner/openmls/blob/86162baac5dd0486cb8c0f659daf520d9f840f39/openmls/src/treesync/node/leaf_node.rs#L492

openmls/src/extensions/mod.rs
tbehner: :+1:

openmls/src/extensions/mod.rs
kkohbrok: The RFC is pretty explicit in saying that we must ignore unknown extensions in leaf nodes. Section 13.4 says the following.

> A client processing a KeyPackage object MUST ignore all unrecognized values in the capabilities field of the LeafNode and all unknown extensions in the extensions and leaf_node.extensions fields. Otherwise, it could fail to interoperate with newer clients.

openmls/src/extensions/mod.rs
kkohbrok: There are three more such checks below this one. Could you change them as well?

PR #1928, opened 2026-01-13
fix: check for duplicate PSK IDs in commits

kkohbrok:


openmls/src/group/mls_group/proposal_store.rs
```diff
@@ -285,6 +287,16 @@ impl ProposalQueue {
                         return Err(FromCommittedProposalsError::SelfRemoval);
                     };
 
+                    // https://validation.openmls.tech/#valn0804
```
kkohbrok: I don't think there's a number for this in our dashboard so I've given it the number 0804 for now.

franziskuskiefer: @keks can you add something to the dashboard and let us know what the number is?

openmls/src/group/errors.rs
```diff
@@ -558,6 +561,9 @@ pub(crate) enum FromCommittedProposalsError {
     /// The sender of a Commit tried to remove themselves.
     #[error("The sender of a Commit tried to remove themselves.")]
     SelfRemoval,
+    /// Commit contains two PSK proposals with the same PSK ID.
+    #[error("Commit contains two PSK proposals with the same PSK ID.")]
+    DuplicatePskId,
```
franziskuskiefer: We had the `PreSharedKeyId` in the errors before. That would make these errors a little more helpful.

franziskuskiefer (approved):
lgtm with one nit about the errors

Otherwise we can get this merged and update the number when there's one on the dashboard.

openmls/src/group/mls_group/proposal_store.rs
keks: I think it's this one: https://validation.openmls.tech/#valn0307

kkohbrok: Thanks! I changed the number and deduplicated some code.

PR #1929, opened 2026-01-14
feat: don't replace groups by default

kkohbrok:
This PR changes `MlsGroup` constructors (both for fresh groups, as well as groups created from a `Welcome`) s.t. they don't replace pre-existing groups with the same `GroupId`. If the application would like to replace an existing group, it can either just delete the group manually before recreating it, or call the `replace_old_group` on the `MlsGroupBuilder` or the `JoinBuilder`.

This PR also changes the semantics on the `store` functions of the `sqlx_storage` and the `sqlite_storage` from `INSERT` to `INSERT OR REPLACE`.


keks (approved):
I think this looks good.

You are right, we should streamline these APIs somewhat. It looks like this might be harder than it needs.

Do you think we should open a followup issue to add an API to the storage provider to check for the existence of group, without loading all of it?

PR #1930, opened 2026-01-14
Add AppData mechanisms from the extensions draft

keks:
This commit adds an implementation of the AppDataDictionary (and related) structs, as well as the AppDataUpdate proposal (and handling) as defined in the extensions draft.

Since the proposals contain diffs in an appplication-defined format, processing of these proposals requires extra work from the application.

While working on this we also found a missing validation check that we added

### DIFF
--- a/openmls/src/group/public_group/validation.rs
+++ b/openmls/src/group/public_group/validation.rs
@@ -1,7 +1,7 @@
 //! This module contains validation functions for incoming messages
 //! as defined in <https://github.com/openmls/openmls/wiki/Message-validation>
 
-use std::collections::{BTreeSet, HashSet};
+use std::collections::HashSet;
 
 use openmls_traits::types::VerifiableCiphersuite;
 
@@ -27,7 +27,6 @@ use crate::{
         proposals::{Proposal, ProposalOrRefType, ProposalType},
         Commit,
     },
-    schedule::errors::PskError,
 };
 
 use crate::treesync::errors::LifetimeError;
@@ -465,30 +464,17 @@ impl PublicGroup {
     ///
     /// * ValSem401: The nonce of a PreSharedKeyID must have length KDF.Nh.
     /// * ValSem402: PSK in proposal must be of type Resumption (with usage Application) or External.
-    /// * ValSem403: Proposal list must not contain multiple PreSharedKey proposals that reference the same PreSharedKeyID.
     pub(crate) fn validate_pre_shared_key_proposals(
         &self,
         proposal_queue: &ProposalQueue,
     ) -> Result<(), ProposalValidationError> {
-        // ValSem403 (1/2)
-        // TODO(#1335): Duplicate proposals are (likely) filtered.
-        //              Let's do this check here until we haven't made sure.
-        let mut visited_psk_ids = BTreeSet::new();
-
         for proposal in proposal_queue.psk_proposals() {
             let psk_id = proposal.psk_proposal().clone().into_psk_id();
 
             // ValSem401
             // ValSem402
             // https://validation.openmls.tech/#valn0803
-            let psk_id = psk_id.validate_in_proposal(self.ciphersuite())?;
-
-            // ValSem403 (2/2)
-            if !visited_psk_ids.contains(&psk_id) {
-                visited_psk_ids.insert(psk_id);
-            } else {
-                return Err(PskError::Duplicate { first: psk_id }.into());
-            }
+            psk_id.validate_in_proposal(self.ciphersuite())?;
         }
 
         Ok(())
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - [#1874](https://github.com/openmls/openmls/pull/1874): Changed `ProposalType`, `ExtensionType`, and `CredentialType` enums to include `Grease(u16)` variant.
 - [#1924](https://github.com/openmls/openmls/pull/1924): Exposed `JoinBuilder::new` as public API.
+- [#1928](https://github.com/openmls/openmls/pull/1928): Processing a commit now fails if it contains a duplicate PSK proposal.
+
 - [#1926](https://github.com/openmls/openmls/pull/1926):
   - Updated `getrandom` dependency in `js` feature to `0.3.4`
   - Removed `libcrux-provider-js` feature (the `libcrux-provider`,`js` features are now sufficient to enable the libcrux crypto provider with support for compiling to wasm)
--- a/openmls/src/group/errors.rs
+++ b/openmls/src/group/errors.rs
@@ -17,7 +17,7 @@ use crate::{
     group::commit_builder::external_commits::ExternalCommitBuilderError,
     key_packages::errors::{KeyPackageExtensionSupportError, KeyPackageVerifyError},
     messages::{group_info::GroupInfoError, GroupSecretsError},
-    schedule::errors::PskError,
+    schedule::{errors::PskError, PreSharedKeyId},
     treesync::errors::*,
 };
 
@@ -275,6 +275,9 @@ pub enum StageCommitError {
     #[cfg(feature = "extensions-draft-08")]
     #[error(transparent)]
     ApplyAppDataUpdateError(#[from] ApplyAppDataUpdateError),
+    /// Duplicate PSK Proposal.
+    #[error("Duplicate PSK proposal with PSK ID {0:?}.")]
+    DuplicatePskId(PreSharedKeyId),
 }
 
 /// Create commit error
@@ -577,6 +580,9 @@ pub(crate) enum FromCommittedProposalsError {
     /// The sender of a Commit tried to remove themselves.
     #[error("The sender of a Commit tried to remove themselves.")]
     SelfRemoval,
+    /// Commit contains two PSK proposals with the same PSK ID.
+    #[error("Commit contains two PSK proposals the PSK ID {0:?}.")]
+    DuplicatePskId(PreSharedKeyId),
 }
 
 /// Create group context ext proposal error
--- a/openmls/src/group/mls_group/proposal_store.rs
+++ b/openmls/src/group/mls_group/proposal_store.rs
@@ -14,6 +14,7 @@ use crate::{
         AddProposal, PreSharedKeyProposal, Proposal, ProposalOrRef, ProposalOrRefType,
         ProposalType, RemoveProposal, UpdateProposal,
     },
+    schedule::PreSharedKeyId,
     utils::vector_converter,
 };
 
@@ -220,6 +221,30 @@ impl OrderedProposalRefs {
     }
 }
 
+#[derive(Default)]
+struct PskProposalDuplicateChecker {
+    psk_proposals: HashSet<PreSharedKeyId>,
+}
+
+impl PskProposalDuplicateChecker {
+    /// Checks whether the given proposal is a duplicate Psk proposal
+    /// (https://validation.openmls.tech/#valn0307). If it is, it returns an
+    /// error.
+    ///
+    ///
+    fn check(&mut self, proposal: &Proposal) -> Result<(), FromCommittedProposalsError> {
+        if let Proposal::PreSharedKey(psk_proposal) = proposal {
+            let psk_id = psk_proposal.clone().into_psk_id();
+            if self.psk_proposals.contains(&psk_id) {
+                return Err(FromCommittedProposalsError::DuplicatePskId(psk_id));
+            } else {
+                self.psk_proposals.insert(psk_id);
+            }
+        }
+        Ok(())
+    }
+}
+
 /// Proposal queue that helps filtering and sorting Proposals received during
 /// one epoch. The Proposals are stored in a `HashMap` which maps Proposal
 /// references to Proposals, such that, given a reference, a proposal can be
@@ -269,6 +294,8 @@ impl ProposalQueue {
         // Build the actual queue
         let mut proposal_queue = ProposalQueue::default();
 
+        let mut psk_proposal_duplicate_checker = PskProposalDuplicateChecker::default();
+
         // Iterate over the committed proposals and insert the proposals in the queue
         log::trace!("   committed proposals ...");
         for proposal_or_ref in committed_proposals.into_iter() {
@@ -289,6 +316,9 @@ impl ProposalQueue {
                         return Err(FromCommittedProposalsError::SelfRemoval);
                     };
 
+                    // https://validation.openmls.tech/#valn0307
+                    psk_proposal_duplicate_checker.check(&proposal)?;
+
                     QueuedProposal::from_proposal_and_sender(
                         ciphersuite,
                         crypto,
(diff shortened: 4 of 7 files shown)


openmls/src/extensions/app_data_dict_extension.rs
```diff
@@ -0,0 +1,275 @@
...
+
+/// Serializable app data dictionary in the [`AppDataDictionaryExtension`].
+///
+/// This struct contains a list of [`ComponentData`] entries.
+/// Entries are in order, and there is at most one entry per [`ComponentId`].
+/// These properties are checked upon creation, as well as upon deserialization.
+#[derive(PartialEq, Eq, Clone, Debug, Default, Serialize, Deserialize)]
+pub struct AppDataDictionary {
+    // NOTE: A BTreeMap is used here to ensure that the data is ordered by keys,
+    // and unique. Since this also goes into the actual MLS Extension message, you could argue that
```
kkohbrok, line 22: What do you mean by "this also goes into the actual MLS Extension message"? Wouldn't that break interop?

openmls/src/extensions/app_data_dict_extension.rs
```diff
@@ -0,0 +1,275 @@
...
+        Ok(Self {
+            component_data: map,
+        })
+    }
+}
+
+impl tls_codec::Size for AppDataDictionary {
+    fn tls_serialized_len(&self) -> usize {
+        // get length without copying
+        let data: Vec<&ComponentData> = self.entries().collect();
```
kkohbrok, line 124: Why not use `len()` here? That would allow you to get away without allocations, wouldn't it?

openmls/src/framing/mls_auth_content_in.rs
```diff
@@ -187,6 +187,11 @@ impl VerifiableAuthenticatedContentIn {
     pub(crate) fn content_type(&self) -> ContentType {
         self.tbs.content.body.content_type()
     }
+
+    /// Get the content.
+    pub(crate) fn content(&self) -> &FramedContentIn {
```
kkohbrok: This kind of bypasses type-based verification. As far as I can see, this is only called in `proposals()` in `UnverifiedMessage`, which in turn is only used in tests. Could we make this available only for `test` and `test-utils`? Or is there another use-case that I'm not seeing?

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -753,14 +841,55 @@ impl<'a, G: BorrowMut<MlsGroup>> CommitBuilder<'a, LoadedPsks, G> {
...
+    /// Sets the [`AppDataUpdates`] that contain the changes made by the AppDataUpdate proposals
+    #[cfg(feature = "extensions-draft-08")]
+    pub fn with_app_data_dictionary_updates(
+        &mut self,
+        app_data_dictionary_updates: Option<AppDataUpdates>,
+    ) {
+        self.stage.app_data_dictionary_updates = app_data_dictionary_updates;
+    }
+
+    /// Returns an iterator over all AppDataUpdate propoas in the proposal store of the group
```
kkohbrok:
```suggestion
    /// Returns an iterator over all AppDataUpdate proposals in the proposal store of the group
```

openmls/src/extensions/app_data_dict_extension.rs
keks: This is a list of varlen buffers, so the total length depends not just one the number of items but also the respective lengths.

openmls/src/extensions/app_data_dict_extension.rs
keks: Ah. The AppDataDictionaryExtension that goes into the group context contains this data type. One perspective should be that the memory layout of the extension struct should closely mirror the definition of the extension in the RFC (which is a sequence of `ComponentData`), which would suggest it should be a `Vec<ComponentData>`.

However, the entries should be ordered by `ComponentId`, and inserting into the middle of a Vec is not super efficient. So instead we use a BTreeMap and (to maintain interop) de/serialize it as if it were a Vec (hence the custom de/serialization impls below). 

Does that make sense?

openmls/src/framing/mls_auth_content_in.rs
keks: That's no good! I thought it was fine because we are returning another `*In` type.  You are right in that it is only called in `proposals()` - but that function must be public!

Basically the caller needs to be able to iterate over the AppDataUpdate proposals to assemble a changeset, which is then passed into the function that processes the unverified message. So we need to expose some access to something unverified here.

Maybe we can, instead of exposing all of the contents, just the proposals, and then in the `UnverifiedMessage` we filter that for AppDataUpdate proposals? That would limit the unverified stuff we expose a little. We could also change the name of these methods here to indicate that they are not verified yet.

kkohbrok: Good idea, exposing just AppDatatProposals sounds like it's what we want.

openmls/src/extensions/app_data_dict_extension.rs
kkohbrok: That does make sense, thanks.

openmls/Cargo.toml
```diff
@@ -45,6 +45,7 @@ zeroize = "1.8.1"
 serde_bytes = "0.11.17"
 
 [features]
+crypto-subtle = [] # Enable subtle crypto APIs that have to be used with care.
```
kkohbrok: I'm surprised to see this here. I don't think it's used anywhere in the code.

openmls/src/framing/mls_auth_content_in.rs
```diff
@@ -187,6 +187,13 @@ impl VerifiableAuthenticatedContentIn {
     pub(crate) fn content_type(&self) -> ContentType {
         self.tbs.content.body.content_type()
     }
+
+    /// If this message is a commit, this returns the unverified list of committed porposals.
+    /// Otherwise it returns `None`.
+    #[cfg(feature = "extensions-draft-08")]
+    pub(crate) fn committed_proposals(&self) -> Option<&[ProposalOrRefIn]> {
+        self.tbs.content.proposals()
```
kkohbrok, line 198: Shouldn't this be filtered for AppData relevant proposals?

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -55,13 +61,19 @@ struct ExternalCommitInfo {
     wire_format_policy: WireFormatPolicy,
 }
 
+#[derive(Debug, Default)]
+struct GroupInfoConfig {
```
kkohbrok, line 65: What is this for? Just to organize the configs a little more?

openmls/src/group/mls_group/exporting.rs
```diff
@@ -116,6 +117,20 @@ impl MlsGroup {
...
         with_ratchet_tree: bool,
+    ) -> Result<MlsMessageOut, ExportGroupInfoError> {
+        self.export_group_info_with_additional_extensions(crypto, signer, with_ratchet_tree, None)
+    }
+
+    /// Export a group info object for this group, with additional extensions.
+    ///
+    ///  Returns an error if a  [`RatchetTreeExtension`] or [`ExternalPubExtension`] is added
+    ///  directly here.
+    pub fn export_group_info_with_additional_extensions<CryptoProvider: OpenMlsCrypto>(
```
kkohbrok, line 128: I understand why this may be useful. But how is it related to the AppData mechanism?

openmls/src/group/mls_group/processing.rs
```diff
@@ -13,8 +13,82 @@ use crate::{
...
+            new_entries: None,
+        }
+    }
+
+    /// Looks up the old value for a component.
+    pub fn old_value(&self, component_id: ComponentId) -> Option<&[u8]> {
+        self.old_dict?.get(&component_id)
+    }
+
+    /// helper method that returns a mutable reference to the
```
kkohbrok:
```suggestion
    /// Helper method that returns a mutable reference to the
```

openmls/src/group/mls_group/processing.rs
```diff
@@ -13,8 +13,82 @@ use crate::{
...
+    /// [`AppDataUpdates`], creating the struct if it does not exist.
+    fn new_entries_mut(&mut self) -> &mut AppDataUpdates {
+        self.new_entries
+            .get_or_insert_with(|| AppDataUpdates(BTreeMap::new()))
+    }
+
+    /// sets a value in the new_entries. if we already have data for that component id, overwrite
+    /// it. else add it in the right position.
```
kkohbrok:
```suggestion
    /// Sets a value in the new_entries. if we already have data for that component id, overwrite
    /// it. Else add it in the right position.
```

openmls/src/group/mls_group/processing.rs
```diff
@@ -13,8 +13,82 @@ use crate::{
...
+    }
+
+    /// Flags an entry in the dictionary for removal
+    pub fn remove(&mut self, id: &ComponentId) {
+        self.new_entries_mut().0.insert(*id, None);
+    }
+
+    /// Consumes the updater and returns just the changes, so we can pass them into
+    /// process_unverified_message
+    /// only returns Some if we actually called set
```
kkohbrok:
```suggestion
    /// Consumes the updater and returns just the changes, so we can pass them into
    /// process_unverified_message.
    /// Only returns Some if we actually called set.
```

openmls/src/group/mls_group/processing.rs
```diff
@@ -30,6 +104,24 @@ impl MlsGroup {
         provider: &Provider,
         message: impl Into<ProtocolMessage>,
     ) -> Result<ProcessedMessage, ProcessMessageError<Provider::StorageError>> {
+        let unverified_message = self.unprotect_message(provider, message)?;
+
+        self.process_unverified_message(provider, unverified_message)
+    }
+
+    #[cfg(feature = "extensions-draft-08")]
+    /// returns a new helper struct for updating the app data
```
kkohbrok:
```suggestion
    /// Returns a new helper struct for updating the app data
```

openmls/src/group/mls_group/processing.rs
```diff
@@ -13,8 +13,82 @@ use crate::{
...
+    extensions::AppDataDictionary,
+};
+
+#[cfg(feature = "extensions-draft-08")]
+use std::collections::BTreeMap;
+
 use super::{errors::ProcessMessageError, *};
 
+#[cfg(feature = "extensions-draft-08")]
+/// keeps the old dictionary as well as the values that are being overwritten
```
kkohbrok:
```suggestion
/// Keeps the old dictionary as well as the values that are being overwritten
```

openmls/src/group/public_group/diff/apply_proposals.rs
```diff
@@ -181,12 +189,108 @@ impl PublicGroupDiff<'_> {
...
+
+        let updates = match (has_app_data_update_proposals, app_data_dict_updates) {
+            // there are AppDataUpdate proposals and we were provided the AooDataUpdates
+            (true, Some(updates)) => updates,
+            // there are no AppDataUpdate proposals in the queue, and we weren't provided any updates
+            (false, None) => return Ok(updated_group_context_extensions),
+
+            // If there are app data update proposals, there must be a list
+            (true, None) => {
+                return Err(ApplyAppDataUpdateError::MissingAppDataUpdates);
```
kkohbrok, line 259: Is this really always an error? What if two updates cancel one-another out? That may make sense in some cases, where the proposals originate from different senders.

openmls/src/group/public_group/diff/apply_proposals.rs
```diff
@@ -181,12 +189,108 @@ impl PublicGroupDiff<'_> {
...
+            extensions: updated_group_context_extensions,
         })
     }
+
+    /// Return the updated GroupContext extensions after applying AppDataUpdates.
+    #[cfg(feature = "extensions-draft-08")]
+    pub(crate) fn apply_app_data_update_proposals(
+        &self,
+        // group_context_extensions if updated by a GroupContextExtensions proposal
+        mut updated_group_context_extensions: Option<Extensions>,
```
kkohbrok: Wouldn't it make more sense to pass in a mutable reference here?

openmls/src/group/public_group/validation.rs
```diff
@@ -600,6 +605,140 @@ impl PublicGroup {
...
+                })
+                .unwrap_or(false);
+
+            // From https://datatracker.ietf.org/doc/html/draft-ietf-mls-extensions#section-4.7-6:
+            // When an MLS group contains the AppDataUpdate proposal type in the proposal_types list in
+            // the group's required_capabilities extension, a GroupContextExtensions proposal MUST NOT
+            // add, remove, or modify the app_data_dictionary GroupContext extension. In other words,
+            // when every member of the group supports the AppDataUpdate proposal, a
+            // GroupContextExtensions proposal could be sent to update some other extension(s), but the
+            // app_data_dictionary GroupContext extension, if it exists, is left as it was.I
```
kkohbrok:
```suggestion
            // app_data_dictionary GroupContext extension, if it exists, is left as it was.
```

openmls/src/messages/group_info.rs
```diff
@@ -259,13 +259,20 @@ impl GroupInfoTBS {
...
         signer: LeafNodeIndex,
-    ) -> Self {
-        Self {
+    ) -> Result<Self, InvalidExtensionError> {
+        // validate the extensions
+        for extension_type in extensions.iter().map(Extension::extension_type) {
+            if extension_type.is_valid_in_group_info() == Some(false) {
+                return Err(InvalidExtensionError::IllegalInGroupInfo);
+            }
+        }
```
kkohbrok, line 268: This looks good, but is also somewhat out of scope for the PR. If you want to include it, could you add a note in the commit message and/or PR description?

openmls/src/framing/mls_auth_content_in.rs
keks: Unfortunately we can't do that here, because we don't know the type of proposal references.

openmls/Cargo.toml
keks: Good catch! We added that because we thought we might be adding unsafe APIs and wanted to gate them, but I think we managed to avoid that.

openmls/src/group/public_group/diff/apply_proposals.rs
keks: You can still have an empty `AppDataUpdates`, but there should be one.

openmls/src/framing/mls_auth_content_in.rs
kkohbrok: Ah, of course!

openmls/src/group/mls_group/commit_builder.rs
keks: It's part of the implementation of `export_group_info_with_additional_extensions`

openmls/src/group/mls_group/exporting.rs
keks: This is needed for us to put the AppDataDictionary into the GroupInfo, I think

openmls/src/group/public_group/diff/apply_proposals.rs
keks: :+1: done

openmls/src/messages/group_info.rs
keks: We noticed this while working on the other group info stuff, I can add it to the PR description

kkohbrok (approved):

PR #1933, opened 2026-01-20
Add and document more validation checks

keks:
This PR adds and fixes documentation for existing validation checks and adds (and documents) new ones.

There is a sister PR to this one on the validation dashboard repo that adds new checks that we reference here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 0.8.0 (2026-02-04)
+
 ### Added
 
 - [#1855](https://github.com/openmls/openmls/pull/1855): Added the `swap_members()` method to `MlsGroup` to replace members in a group, as well as the `WelcomeCommitMessages` and `SwapMembersError` structs.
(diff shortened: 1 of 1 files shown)


openmls/src/group/errors.rs
```diff
@@ -69,8 +69,8 @@ pub enum WelcomeError<StorageError> {
     /// Could not decrypt the Welcome message.
     #[error("Could not decrypt the Welcome message.")]
     UnableToDecrypt,
-    /// Unsupported extensions found in the KeyPackage of another member.
-    #[error("Unsupported extensions found in the KeyPackage of another member.")]
+    /// Unsupported extensions found in the GroupContext or  KeyPackage of another member.
```
kkohbrok:
```suggestion
    /// Unsupported extensions found in the GroupContext or KeyPackage of another member.
```

openmls/src/group/tests_and_kats/tests/group_context_extensions.rs
```diff
@@ -1142,3 +1142,74 @@ fn proposal() {
         )
         .expect_err("expected an error building GCE proposal with bad required_capabilities");
 }
+
+// NOTE: Need to disable the check for valn0502 in group/public_group/validation.rs for this to
```
franziskuskiefer, line 1720: You could allow this with `skip_validation.rs`.

keks: I know, but that seemed like a bit of a hack. The proper solution would be to create the message using the Frankenstein framework, IMHO, but that would have been a lot of work.

Do you think we should do this here now?

franziskuskiefer: I think it would be a reasonable way to automate the disabling of that check. But I'm fine without.

PR #1934, opened 2026-01-22
add AAD getter to `PrivateMessageIn`

wysiwys:


openmls/src/framing/private_message.rs
```diff
@@ -302,6 +302,11 @@ impl PrivateMessage {
     pub(crate) fn ciphertext(&self) -> &[u8] {
         self.ciphertext.as_slice()
     }
+
+    /// Retrieve the additional authenticated data (AAD) from the [`PrivateMessage`].
+    pub(crate) fn aad(&self) -> &[u8] {
```
franziskuskiefer: Looks like this is dead code.

wysiwys: I'll leave this out for now, to keep the scope of this pull request focused on `PrivateMessageIn` for now

PR #1935, opened 2026-01-22
[Proof of concept] Provider-level storage transactions

wysiwys:
As a basis for discussion, this draft PR includes a proof of concept for a way to provide provider-level transactions on a storage provider, in particular in an async context where several tasks may access a storage provider.

This draft PR also includes some documentation around ensuring exclusive reads and writes, as well as atomicity, when using the storage provider to perform group operations.

### DIFF
--- a/openmls/src/group/mls_group/mod.rs
+++ b/openmls/src/group/mls_group/mod.rs
@@ -219,6 +219,12 @@ pub enum MlsGroupState {
 /// incorrect, an error event will be returned with a corresponding error
 /// message and the state of the group will remain unchanged.
 ///
+/// `MlsGroup` operations write state to the storage provider, and should be
+/// performed by the application in a context that ensures atomicity, as well as
+/// exclusive access per group. See
+/// [storage](https://github.com/openmls/openmls/tree/main/traits#storageprovider)
+/// for more details.
+///
 /// An `MlsGroup` has an internal state variable determining if it is active or
 /// inactive, as well as if it has a pending commit. See [`MlsGroupState`] for
 /// more information.


sqlx_storage/src/lib.rs
```diff
@@ -37,36 +36,97 @@ mod wrappers;
 /// It is generic over any codec `C` that implements the [`Codec`] trait.
 /// The codec is used to serialize and deserialize the data stored in the
 /// underlying database.
-pub struct SqliteStorageProvider<'a, C> {
-    connection: RefCell<&'a mut SqliteConnection>,
+pub struct SqliteStorageProvider<C> {
+    pool: sqlx::SqlitePool,
```
kkohbrok, line 40: Is it really necessary to upgrade this to a pool? A mutable reference to a `SqliteConnection` should be enough to get a transaction.

sqlx_storage/src/lib.rs
```diff
@@ -37,36 +36,97 @@ mod wrappers;
...
+    ///
+    /// This transaction is started in EXCLUSIVE mode, so it locks
+    /// the database for all reads and writes.
+    ///
+    /// If the transaction falls out of scope without being committed,
+    /// it is automatically rolled back.
+    pub async fn get_transaction<'a>(
+        &'a self,
+    ) -> Result<SqliteStorageProviderWithTransaction<'a, C>, sqlx::Error> {
+        let transaction = self.pool.begin_with("BEGIN EXCLUSIVE").await?;
```
kkohbrok, line 63: Wouldn't `BEGIN IMMEDIATE` be enough?

sqlx_storage/src/lib.rs
```diff
@@ -37,36 +36,97 @@ mod wrappers;
...
+pub struct SqliteStorageProvider<C> {
+    pool: sqlx::SqlitePool,
     codec: PhantomData<C>,
 }
 
-impl<'a, C: Codec> SqliteStorageProvider<'a, C> {
+/// A [`StorageProvider`] implementation of a transaction wrapper, for an EXCLUSIVE transaction.
+///
+/// Default behavior: rolls back on drop
+pub struct SqliteStorageProviderWithTransaction<'a, C> {
```
kkohbrok, line 47: You may have tried it already, instead of having two separate provider structs, wouldn't it be easier to have one struct that is generic over the `connection`, requiring it only to implement `sqlx::Connection` for sqlite?

wysiwys: Ah, I had not tried this yet! Yes, that allows the `StorageProvider` implementation to be simplified quite a bit

sqlx_storage/src/lib.rs
wysiwys: Indeed, it is not strictly necessary to use a `SqlitePool` in the provider - the main reason a `SqlitePool` is used in this example is to show how this approach to transactions can also be extended to an async context, where several tasks may need access to the underlying database. Since this aspect of the proof of concept goes beyond a minimal implementation, I have now added a note about it it to the PR description!

In an async context, depending on the use case, I think that it it could be preferable to use a single `SqliteConnection` as well, where access to the single connection is governed e.g. by a `Mutex`, which would already enforce exclusivity of reads and writes at the application level.

sqlx_storage/src/lib.rs
wysiwys: I think that this is only needed in the case where multiple database connections can be open (to prevent re-entrant reads), and `BEGIN IMMEDIATE` is enough when only a single connection is open.

PR #1936, opened 2026-01-22
Add APIs to access information about the next epoch from a StagedCommit

keks:
These changes allow "peeking" into the next epoch's state before merging.

This is needed e.g. if you want to encrypt the application data using an exporter secret and send it along a commit. Right now the only way to access these values is to merge the commit, but you shouldn't really merge the commit yet (because the DS can't have ack'd it before you sent it).

It did require a bit of refactoring around the key, so (when passing in the previous/current epoch's ratchet tree) you can export the new tree. This is because the Staged trees don't have the old tree anymore (which is nice becaue it gets rid of the lifetimes, but is annoying in this case. I moved the code into functions that take the values as inputs instead of operating on the struct fields. That way I could reuse the code for both cases.

### DIFF
--- a/interop_client/Cargo.toml
+++ b/interop_client/Cargo.toml
@@ -18,7 +18,7 @@ openmls_traits = { path = "../traits" }
 
 mls_interop_proto = { git = "https://github.com/mlswg/mls-implementations/" }
 
-tonic = "0.8"
+tonic = "0.14"
 tokio = { version = "1.19.2", features = ["macros", "net", "rt-multi-thread"] }
 clap = "4.1"
 clap_derive = "4.1"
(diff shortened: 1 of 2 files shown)


openmls/src/group/mls_group/staged_commit.rs
```diff
@@ -580,6 +584,29 @@ impl StagedCommit {
...
+        crypto: &impl OpenMlsCrypto,
+        original_tree: RatchetTree,
+    ) -> Option<RatchetTree> {
+        match &self.state {
+            StagedCommitState::PublicState(public_staged_commit_state) => None,
+            StagedCommitState::GroupMember(member_staged_commit_state) => Some(
+                member_staged_commit_state
+                    .staged_diff
+                    .export_ratchet_tree(crypto, self.group_context().ciphersuite(), original_tree)
+                    // XXX: figure out error handling
```
kkohbrok: Maybe at least log for now?

keks: I made the function `Result<Optiond<_>, _>`

kkohbrok (approved):

PR #1941, opened 2026-01-29
Don't use git dependencies for hpke-rs, update tonic in interop_client

keks:
I think this should fix CI.


franziskuskiefer (approved):
We merged the version bump for the interop protobuf upstream. So this may just go through. If now, let's do it separately.

PR #1943, opened 2026-02-02
fix: skip removed members in capabilities check

kkohbrok:
RFC 9420 says a commit is invalid if (among other things):

> It contains a Proposal with a non-default proposal type that is not supported by some members of the group that will process the Commit (i.e., members being added or removed by the Commit do not need to support the proposal type).

Our validation is currently stricter than necessary in that it also requires members removed in a commit to support all proposal types in that commit. This PR fixes that.

### DIFF
--- a/openmls/src/group/public_group/validation.rs
+++ b/openmls/src/group/public_group/validation.rs
@@ -148,11 +148,6 @@ impl PublicGroup {
             })
             .collect();
 
-        println!(
-            "Signature keys of removed members: {:?}",
-            signature_keys.len()
-        );
-
         // Iterate over all leaf nodes except the removed ones
         let mut leaves = self
             .treesync()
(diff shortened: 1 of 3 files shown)


openmls/src/group/public_group/validation.rs
```diff
@@ -136,10 +136,27 @@ impl PublicGroup {
         &self,
         proposal_queue: &ProposalQueue,
     ) -> Result<(), ProposalValidationError> {
-        let mut leaves = self.treesync().full_leaves();
+        // Collect signature keys of removed members s.t. we can skip them
+        // when checking capabilities.
+        let signature_keys: HashSet<_> = proposal_queue
```
franziskuskiefer, line 141: Do we have any case where we may not commit all the proposals in the queue? Maybe we should at least add a comment to the actual commit step, that we rely on all proposals being committed here.

kkohbrok: The queue is specifically created from all proposals in the commit, both during creation and during processing, so there can't be a gap. Do you maybe mean the proposal store?

PR #1944, opened 2026-02-02
fix: correctly access credentials from past epochs

kkohbrok:


franziskuskiefer (commented):
lgtm. Can you add some comments to both changed functions with a little description what's happening

franziskuskiefer (approved):
Thanks

PR #1959, opened 2026-02-10
Expose TreeSync for MlsGroup and PublicGroup

### LINKED ISSUE #1939: Tree inspection unavailable

I need to inspect the tree to determine if all members of a group support an extension before using it. We are changing to use proposals separate from commits and old code doesn't support this, so it has to be signaled through an extension / capability. 

We also need to limit commit sizes, so this at least splits them up, but having support for seeing which nodes are unmerged or which parents are unfilled to determine which leaves should do a self update is also needed to keep the commit sizes down. Right now we do a self update before sending our first message, but then never again, so removing members can leave empty parent nodes

A way to iterate though all leaf nodes in the tree, all parent nodes in the tree, all parent nodes for a leaf, and all leaf nodes for a parent.

All that I can see is available in the api right now is `export_ratchet_tree` but the internals are private, so using the api as is can be done using `transmute`, but that's not safe. `members` doesn't give the necessary information for leaves right now, and that doesn't include parent information.

**Additional context**

I'm happy to make a PR for this, but I'd like feedback on the best way to implement it. The `RatchetTree` type is public, but isn't inspectable. The `TreeSync` type has almost all the methods needed to implement this (just need to make some of them `pub`), but it isn't public. The `members` function doesn't return the information that I need. What is the preferred way of being able to inspect the tree so that we can make better use of extensions and ensure that we getting as close to log2 commit sizes as possible?

keks:
Yesterday we merged #1955, which exposed methods to get the nodes from a TreeSync, but I forgot adding a way to actually get that TreeSync from a group. This PR adds methods to do that for both MlsGroup and PublicGroup.

Closes #1939

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -78,8 +78,3 @@ jobs:
         run: |
           export CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=$HOME/.cargo-install/wasm-bindgen-cli/bin/wasm-bindgen-test-runner
           cargo test -p openmls -vv --target wasm32-unknown-unknown -F js,js-test
-
-      - name: Tests Wasm32 on linux with libcrux provider
-        run: |
-          export CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=$HOME/.cargo-install/wasm-bindgen-cli/bin/wasm-bindgen-test-runner
-          cargo test -p openmls -vv --target wasm32-unknown-unknown -F js,js-test,libcrux-provider
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 0.8.1 (2026-02-13)
+
+### Added
+
+- [#1955](https://github.com/openmls/openmls/pull/1955): Expose functions that allow access to (blank) leaves and parent nodes
+
+### Changed
+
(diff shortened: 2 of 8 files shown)


openmls/src/group/mls_group/mod.rs
```diff
@@ -347,6 +347,11 @@ impl MlsGroup {
         self.proposal_store().proposals()
     }
 
+    /// Returns the current tree state of the group, in the form of a [`TreeSync`].
+    pub fn treesync(&self) -> &TreeSync {
```
franziskuskiefer, line 351: Or should this be `tree_sync`? 🤷🏻 No strong opinion. Just wondering.

franziskuskiefer:
CI isn't happy though

franziskuskiefer:
Another thought: To make sure this is actually usable, can you add a short integration test to make sure the tree and its values are actually accessible.

openmls/src/group/mls_group/mod.rs
keks: It already was treesync in PublicGroup and didn't want more churn, but yeah making it public is a good time to ask that question.

tree_sync seems like the correct adaptation of TreeSync, but somehow it looks wrong to me :sweat_smile: So yeah, also not strong opinions either way.

openmls/tests/treesync_api.rs
```diff
@@ -0,0 +1,53 @@
...
+            adder: "alice",
+            addees: vec![bob_pre_group],
+            join_config,
+            tree: None,
+        })
+        .unwrap();
+
+    // 4. Bob inspects the tree
+    let [bob] = group_state.members_mut(&["bob"]);
+    let leaves: Vec<_> = bob.group.treesync().full_leaves().collect();
```
franziskuskiefer, line 36: The intermediate nodes are important as well. Can we get them too?

franziskuskiefer:
@keks can this get merged or is there anything else you wanted to do?

PR #1961, opened 2026-02-11
Add API to export secret from staged welcome

wysiwys:
This PR adds an API to the `StagedWelcome` that allows exporting a secret for the group that can be joined from the `StagedWelcome`. This is useful for decrypting data that was encrypted with an exporter secret, before creating the `MlsGroup` from the `StagedWelcome`.


openmls/tests/data_next_epoch.rs
```diff
@@ -229,3 +230,63 @@ fn staged_commit_self_removed_returns_none() {
...
+        .unwrap();
+
+    // 7. Bob creates an MlsGroup from the StagedWelcome
+    let bob_group = staged_welcome
+        .into_group(&bob_party.provider)
+        .expect("error creating group");
+
+    // === Verify staged values match group values ===
+    assert_eq!(
+        staged_export,
```
franziskuskiefer, line 288: Can you also add a check that this matches the export from Alice?

PR #1962, opened 2026-02-13
update libcrux and hpke-rs depdencenies

franziskuskiefer:


raphaelrobert (approved):
It looks like there is currently a regression in the rust nighly toolchain. I tried it with an older version and the tests passed. I think this is good to go.

PR #1963, opened 2026-02-13
feat: add virtual-clients feature flag and confirmable app messages

kkohbrok:
This PR is the first in a number of PRs towards implementing the [virtual clients draft](https://github.com/mlswg/mls-virtual-clients).

This PR adds the `virtual-clients-draft` feature flag. Guarded by that flag, it adds a two-stage mechanism that allows sending an application message and deleting the corresponding message secret in two separate steps.

This PR also generally enables clients to decrypt their own application messages by introducing a new `DualUseRatchet` in addition to the sender and decryption ratchets. The `DualUseRatchet` stores additional metadata to properly maintain the retention window for missed messages. In particular, ratchet secrets used for encryption shouldn't influence the retention of secrets for decryption of missed messages.


openmls/src/tree/sender_ratchet.rs
```diff
@@ -207,6 +207,37 @@ impl DecryptionRatchet {
         &mut self.ratchet_head
     }
 
+    #[cfg(feature = "virtual-clients-draft")]
+    pub(crate) fn delete_secret_for_generation(&mut self, generation: Generation) {
+        let window_index = ((self.generation() - generation) as i32) - 1;
+        if window_index >= 0 {
+            let index = window_index as usize;
+            self.past_secrets.get_mut(index).take();
```
franziskuskiefer: This doesn't actually do anything. You need to actually take the value from entry.

openmls/src/group/mls_group/errors.rs
```diff
@@ -166,13 +167,19 @@ pub enum ProcessMessageError<StorageError> {
 
 /// Create message error
 #[derive(Error, Debug, PartialEq, Clone)]
-pub enum CreateMessageError {
+pub enum CreateMessageError<StorageError> {
```
franziskuskiefer: Do we need this breaking change even without the feature?

openmls/src/framing/private_message.rs
```diff
@@ -9,8 +9,16 @@ use crate::{
...
 
+#[cfg(feature = "virtual-clients-draft")]
+use crate::tree::sender_ratchet::SenderRatchetConfiguration;
+
 use super::*;
 
+#[cfg(not(feature = "virtual-clients-draft"))]
+pub(crate) type EncryptionOutput = PrivateMessage;
+#[cfg(feature = "virtual-clients-draft")]
+pub(crate) type EncryptionOutput = (u32, PrivateMessage);
```
franziskuskiefer: We could always use this type and just ignore the unused `u32`. Would make the code a bit cleaner.

franziskuskiefer (commented):
Thanks. Generally lgtm. A couple first comments.

openmls/src/framing/private_message.rs
```diff
@@ -11,6 +11,13 @@ use crate::{
 
 use super::*;
 
+/// The result of encrypting an [`AuthenticatedContent`] into a
+/// [`PrivateMessage`]. The `u32` is the generation of the encryption secret
+/// used. With the `virtual-clients-draft` feature, callers use the generation
+/// to confirm the message and delete the corresponding encryption secret.
+/// Without the feature, the generation is unused.
+pub(crate) type EncryptionOutput = (u32, PrivateMessage);
```
boxdot: Minor: WDYT about making it a struct with two fields? Then it is possible to attach the docs to both fields more precisely, and we can access the message field just immediately after calling `encrypt`?

boxdot: Minor: Maybe it also makes sense to use the type alias `Generation` for u32 here.

openmls/src/tree/sender_ratchet.rs
```diff
@@ -278,3 +285,214 @@ impl DecryptionRatchet {
...
+/// decrypt their own ciphertexts (e.g. when receiving a message encrypted by
+/// another emulating client). Encryption secrets are kept in a past-secrets
+/// window like a [`DecryptionRatchet`] until they are explicitly dropped by
+/// confirming the corresponding generation via
+/// [`Self::delete_secret_for_generation`].
+#[cfg(feature = "virtual-clients-draft")]
+#[derive(Serialize, Deserialize)]
+#[cfg_attr(any(feature = "test-utils", test), derive(PartialEq, Clone))]
+#[cfg_attr(any(feature = "crypto-debug", test), derive(Debug))]
+pub struct DualUseRatchet {
```
boxdot: Minor: to make guarding include with the feature flag it is easier to introduce a mod for `DualUserRatchet` guarded by the feature flag. Then it is possible just to expose this type and keep everything inside contained.

openmls/src/group/mls_group/application.rs
```diff
@@ -37,15 +74,56 @@ impl MlsGroup {
...
+    }
+
+    /// Confirms that a message has been successfully sent without a generation
+    /// collision. This deletes the encryption secrets for the given generation.
+    #[cfg(feature = "virtual-clients-draft")]
+    pub fn confirm_message<Storage: StorageProvider>(
+        &mut self,
+        storage: &Storage,
+        generation: u32,
+    ) -> Result<(), CreateMessageError<Storage::Error>> {
```
keks: Should this maybe have its own error type?

openmls/src/tree/dual_use_ratchet.rs
```diff
@@ -0,0 +1,224 @@
...
+        ) {
+            Self::AwaitingConfirmation(ratchet_secret)
+            | Self::RetainedForDecryption(RetainedDecryptionSecret::Available(ratchet_secret)) => {
+                *self = Self::RetainedForDecryption(RetainedDecryptionSecret::Consumed);
+                Ok(ratchet_secret)
+            }
+            Self::RetainedForDecryption(RetainedDecryptionSecret::Consumed) => {
+                *self = Self::RetainedForDecryption(RetainedDecryptionSecret::Consumed);
+                Err(SecretTreeError::SecretReuseError)
+            }
```
keks, line 65: Do we need to assign self in these? I think this is already done by `std::replace`?

openmls/src/tree/dual_use_ratchet.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    fn error_for_missing_past_secret(&self, generation: Generation) -> SecretTreeError {
+        if self
+            .past_secrets
+            .iter()
+            .find_map(|(retained_generation, entry)| {
+                entry
+                    .is_retained_for_decryption()
+                    .then_some(*retained_generation)
+            })
```
keks: I think this could just `find` for `entry.is_retained_for_decryption()` and the closure to `is_some_and` below just takes a pair? Slightly simpler, but either are fine I think.

keks (commented):
First pass. The one about filters really is just a nitpick.

openmls/src/group/mls_group/application.rs
```diff
@@ -2,32 +2,71 @@ use openmls_traits::signatures::Signer;
...
+
+    /// Creates an application message. Returns
+    /// `CreateMessageError::MlsGroupStateError::UseAfterEviction` if the member
+    /// is no longer part of the group. Returns
+    /// `CreateMessageError::MlsGroupStateError::PendingProposal` if pending
+    /// proposals exist. In that case `.process_pending_proposals()` must be
+    /// called first and incoming messages from the DS must be processed
+    /// afterwards.
+    #[cfg(feature = "virtual-clients-draft")]
+    pub fn create_message<Provider: OpenMlsProvider>(
```
raphaelrobert, line 42: This confirms messages immediately. Is there a good use case for this? It sounds like a shortcut that applications should not take.

openmls/src/group/mls_group/errors.rs
raphaelrobert: If we got rid of `create_message()` (see comment above), we could have a clean divide that would not lead to a breaking change (I think).

raphaelrobert (commented):
This looks good in general. I have some questions regarding the API & breaking changes.

raphaelrobert (dismissed):
Thanks for making the change!

openmls/src/tree/dual_use_ratchet.rs
```diff
@@ -0,0 +1,218 @@
...
+        }
+        if matches!(
+            self.past_secrets.get(&generation),
+            Some(DualUsePastSecret::AwaitingConfirmation(_))
+        ) {
+            self.past_secrets.remove(&generation);
+        }
+    }
+
+    /// Gets a secret for encryption. The secret is also recorded in the
```
franziskuskiefer, line 105: I think it's worth commenting here that this copies the AEAD key on purpose.

openmls/src/framing/errors.rs
```diff
@@ -35,7 +35,7 @@ pub enum MessageDecryptionError {
 
 /// Message encryption error
 #[derive(Error, Debug, PartialEq, Clone)]
-pub(crate) enum MessageEncryptionError<StorageError> {
+pub enum MessageEncryptionError<StorageError> {
```
franziskuskiefer, line 38: Making this public should probably go into the changelog.

openmls/src/group/mls_group/errors.rs
```diff
@@ -191,6 +197,48 @@ pub enum CreateMessageError {
     GroupStateError(#[from] MlsGroupStateError),
 }
 
+/// Create message error
+#[cfg(feature = "virtual-clients-draft")]
```
franziskuskiefer: Generally, grouping things into a module where possible would make the `cfg`s simpler and the code a bit cleaner.

kkohbrok:
Thanks for all the feedback! I hope I got everything. If there are no more objections, I'd like to merge this soon s.t. I can get on with the follow-up PRs.

raphaelrobert (approved):

PR #1967, opened 2026-02-23
add tree maintainence helpers

keks:
This PR adds a helper crate for working towards good tree health. The code is in a separate crate so it can be added optionally.

So far it only adds a function that finds good candidates for issuing self-update commits after a Remove or Update proposal is committed (or for committing such a proposal and adding an update_path).

More helpers are planned.


franziskuskiefer:
@keks shall we get this cleaned up and merged. As a separate crate this is not so critical, and exposing more information about the tree makes sense.

PR #1972, opened 2026-03-12
Add APIs for time-based deletion of past epoch secrets

### LINKED ISSUE #1966: Remove past epoch secrets not just based on number of epochs, but also time

wysiwys:
This PR adds APIs for deletion of past epoch secrets before a provided timestamp or older than a provided duration.

Resolves #1966

### DIFF
--- a/openmls/src/group/mls_group/tests_and_kats/tests/past_secrets.rs
+++ b/openmls/src/group/mls_group/tests_and_kats/tests/past_secrets.rs
@@ -410,6 +410,42 @@ fn setup_tree_store_with_timestamps<Provider: OpenMlsProvider>(
     message_secrets_store
 }
 
+/// Test persistence of an update to the past epoch deletion policy.
+#[openmls_test::openmls_test]
+fn test_update_policy_persistence<Provider: OpenMlsProvider>(ciphersuite: Ciphersuite) {
+    // initial policy
+    let initial_policy = PastEpochDeletionPolicy::MaxEpochs(3);
+    // set up a provider, signer and group
+    let (alice_provider, _alice_signer, mut alice_group) =
+        setup::<Provider>(ciphersuite, initial_policy.clone());
+
+    // load group from storage
+    let alice_group_stored = MlsGroup::load(alice_provider.storage(), alice_group.group_id())
+        .expect("error loading group")
+        .expect("no group for id");
+    // check policy
+    assert_eq!(
+        alice_group_stored.past_epoch_deletion_policy(),
+        &initial_policy
+    );
+
+    for new_policy in [
+        PastEpochDeletionPolicy::KeepAll,
+        PastEpochDeletionPolicy::MaxEpochs(3),
+        PastEpochDeletionPolicy::MaxEpochs(2),
+    ] {
+        alice_group
+            .set_past_epoch_deletion_policy(&alice_provider, new_policy.clone())
+            .unwrap();
+        // load group from storage
+        let alice_group_stored = MlsGroup::load(alice_provider.storage(), alice_group.group_id())
+            .expect("error loading group")
+            .expect("no group for id");
+        // check policy
+        assert_eq!(alice_group_stored.past_epoch_deletion_policy(), &new_policy);
+    }
+}
+
 /// Test changing the past epoch deletion policy on a group.
 #[openmls_test::openmls_test]
 fn test_update_policy<Provider: OpenMlsProvider>(ciphersuite: Ciphersuite) {
@@ -527,7 +563,7 @@ fn test_update_policy<Provider: OpenMlsProvider>(ciphersuite: Ciphersuite) {
             .num_past_epoch_trees(),
         1
     );
-    assert_eq!(alice_group.past_epoch_deletion_policy(), &new_policy);
+    assert_eq!(alice_group_stored.past_epoch_deletion_policy(), &new_policy);
 }
 
 /// Test a secret tree store with all legacy timestamps, where a timestamp is available for the


book/src/user_manual/past_epoch_secret_deletion.md
```diff
@@ -0,0 +1,45 @@
+# Past epoch secret deletion
+
+
+## Setting a past epoch secrets deletion policy
+
+The Delivery Service cannot guarantee that application messages from one epoch are sent before the beginning of the next epoch. To address this, applications can configure their groups to keep the necessary key material around for past epochs by configuring the past epoch deletion policy on the `MlsGroupCreateConfig`.
```
franziskuskiefer: I know this is copied over. But I find it a bit harsh.

```suggestion
The Delivery Service may not be able to guarantee that application messages from one epoch are sent before the beginning of the next epoch. To address this, applications can configure their groups to keep the necessary key material around for past epochs by configuring the past epoch deletion policy on the `MlsGroupCreateConfig`.
```

book/src/forward_secrecy.md
```diff
@@ -46,7 +46,7 @@ When an encrypted message is received, the corresponding decryption key is deriv
 
 OpenMLS can address 3 scenarios:
 
-- The Delivery Service cannot guarantee that application messages from one epoch are sent before the beginning of the next epoch. To address this, applications can configure their groups to keep the necessary key material around for past epochs by setting the `max_past_epochs` field in the `MlsGroupCreateConfig` to the desired number of epochs.
+- The Delivery Service cannot guarantee that application messages from one epoch are sent before the beginning of the next epoch. To address this, applications can configure their groups to keep the necessary key material around for past epochs by configuring the past epoch deletion policy on the `MlsGroupCreateConfig`.
```
franziskuskiefer: This doesn't tell the reader much anymore now. Can you link the new page to get more information?

book/src/user_manual/past_epoch_secret_deletion.md
```diff
@@ -0,0 +1,45 @@
+# Past epoch secret deletion
+
```
franziskuskiefer, line 2: There is also a description of the group config [in the book](https://book.openmls.tech/user_manual/group_config.html?highlight=max_past_epochs#group-configuration). That needs updating as well.

openmls/src/group/mls_group/tests_and_kats/tests/past_secrets_storage_compatibility.rs
```diff
@@ -0,0 +1,160 @@
...
+use crate::prelude::*;
+use crate::prelude::{mls_group::MessageSecretsStore, past_secrets::MessageSecretsWithTimestamp};
+use crate::schedule::message_secrets::MessageSecrets;
+
+use openmls_rust_crypto::RustCrypto as CryptoProvider;
+use openmls_sqlite_storage::{Codec, Connection, SqliteStorageProvider};
+use openmls_traits::storage::StorageProvider;
+use serde::Serialize;
+
+// TODO: load from file at runtime instead
```
franziskuskiefer: Is this an open todo?

openmls/src/group/mls_group/builder.rs
```diff
@@ -223,6 +226,10 @@ impl MlsGroupBuilder {
     ///
     /// **WARNING**
     ///
+    /// This method overrides the policy set by [`Self::set_past_epoch_deletion_policy()`],
+    /// and is equivalent to setting the past epoch deletion policy to
+    /// `PastEpochDeletionPolicy::MaxEpochs(max_past_epochs)`.
```
franziskuskiefer, line 229: Add a note that this function will be deprecated in future releases.
Let's also add a deprecation warning to it for `0.9`.

openmls/src/group/mls_group/builder.rs
```diff
@@ -235,6 +242,19 @@ impl MlsGroupBuilder {
         self
     }
 
+    // TODO: Add more information about past epochs to the documentation
```
franziskuskiefer: Is this TODO still open?

openmls/src/group/mls_group/mod.rs
```diff
@@ -644,6 +644,18 @@ impl MlsGroup {
...
+    /// Delete all past epoch secrets.
+    ///
+    /// For more information on the arguments to this method, see [`PastEpochDeletion`].
+    ///
+    /// NOTE: the past epoch deletion policy determines the minimum number of past
+    /// epoch secrets that will be kept. If this value is a `Some(n)`, at least
+    /// `n` elements will be kept, regardless of whether their duration
+    /// is allowed.
+    pub fn delete_past_epoch_secrets(&mut self, policy: PastEpochDeletion) {
+        self.message_secrets_store.delete_past_epoch_secrets(policy);
```
franziskuskiefer, line 686: This will need to be persisted. Right now, this only has an effect as long as the group is in memory.

openmls/src/group/mls_group/past_secrets.rs
```diff
@@ -39,19 +77,43 @@ impl core::fmt::Debug for MessageSecretsStore {
...
 
     /// Resize the store.
-    pub(crate) fn resize(&mut self, max_past_epochs: usize) {
+    pub(crate) fn resize(&mut self, policy: &PastEpochDeletionPolicy) {
+        // max or the limit of the storage size
+        let max_past_epochs = max_epochs(policy);
+
         let old_size = self.max_epochs;
         self.max_epochs = max_past_epochs;
         if old_size > max_past_epochs {
```
franziskuskiefer, line 129: This (the rotation) could panic (now) if `old_size > self.past_epoch_trees.len()`. We need to take the minimum `num_epochs_out.min(self.past_epoch_trees.len())`

openmls/src/group/mls_group/mod.rs
```diff
@@ -644,6 +644,18 @@ impl MlsGroup {
         )
     }
 
+    /// Delete all past epoch secrets.
+    ///
+    /// For more information on the arguments to this method, see [`PastEpochDeletion`].
+    ///
+    /// NOTE: the past epoch deletion policy determines the minimum number of past
```
franziskuskiefer: This comment doesn't seem right. There's no minimum number of past epoch secrets.

openmls/src/group/mls_group/past_secrets.rs
```diff
@@ -39,19 +77,43 @@ impl core::fmt::Debug for MessageSecretsStore {
...
+const VECDEQUE_MAX_CAPACITY: usize = isize::MAX as usize;
+
+// XXX: the VecDeque capacity is not checked elsewhere in this module.
+/// Helper function to map a policy to a maximum number of past epochs
+fn max_epochs(policy: &PastEpochDeletionPolicy) -> usize {
+    // get the `max_epochs`, or the maximum capacity of a `VecDeque`
+    let max_epochs = policy.max_epochs().unwrap_or(VECDEQUE_MAX_CAPACITY);
+
+    // cap at max capacity
+    max_epochs.min(VECDEQUE_MAX_CAPACITY)
```
franziskuskiefer, line 99: This is a no op and not needed.

openmls/src/group/mls_group/config.rs
```diff
@@ -139,8 +245,24 @@ impl MlsGroupJoinConfigBuilder {
...
+    /// and is equivalent to setting the past epoch deletion policy to
+    /// `PastEpochDeletionPolicy::MaxEpochs(max_past_epochs)`.
     pub fn max_past_epochs(mut self, max_past_epochs: usize) -> Self {
-        self.join_config.max_past_epochs = max_past_epochs;
+        self.join_config.past_epoch_deletion_policy =
+            PastEpochDeletionPolicy::MaxEpochs(max_past_epochs);
+        self
+    }
+
+    // TODO: Add more information about past epochs to the documentation
```
franziskuskiefer: Open todo?

openmls/src/group/mls_group/config.rs
```diff
@@ -36,6 +36,100 @@ use crate::{
 };
 use serde::{Deserialize, Serialize};
 
+/// Configures the automatic deletion of past epoch secrets.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub enum PastEpochDeletionPolicy {
+    /// Keep at most `n` past epoch secrets.
+    MaxEpochs(usize),
+    /// Keep all past epoch secrets.
```
franziskuskiefer, line 52: This should also have the same warning sign as the functions, and maybe even more, i.e. that the user is responsible for calling delete.

openmls/src/group/mls_group/past_secrets.rs
```diff
@@ -172,11 +253,71 @@ impl MessageSecretsStore {
...
+                }
+                PastEpochDeletionTimeConfig::Duration(duration) => {
+                    self.delete_past_epoch_secrets_older_than_duration(duration)
+                }
+            }
+        } else {
+            // delete all
+            self.past_epoch_trees.clear();
+        }
+        // ensure at most `max_past_epochs` entries are included
```
franziskuskiefer: This doesn't do anything in the `else` case above. So maybe it should go into the `if` case?

openmls/src/group/mls_group/config.rs
```diff
@@ -278,7 +404,19 @@ impl MlsGroupCreateConfigBuilder {
     /// the same epoch in which they were generated. The number for `max_epochs` should be
     /// as low as possible.
     pub fn max_past_epochs(mut self, max_past_epochs: usize) -> Self {
-        self.config.join_config.max_past_epochs = max_past_epochs;
+        self.config.join_config.past_epoch_deletion_policy =
+            PastEpochDeletionPolicy::MaxEpochs(max_past_epochs);
+        self
+    }
+
+    // TODO: Add more information about past epochs to the documentation
```
franziskuskiefer: Open todo?

openmls/src/group/mls_group/past_secrets.rs
```diff
@@ -172,11 +253,71 @@ impl MessageSecretsStore {
...
+        // retain entries that are older than the duration.
+        self.past_epoch_trees.retain(|tree| {
+            let Some(added_at) = tree.message_secrets.added_at else {
+                return true;
+            };
+            let Ok(elapsed) = std::time::SystemTime::now().duration_since(added_at) else {
+                return true;
+            };
+
+            // retain entried where elapsed is less than the provided duration
```
franziskuskiefer:
```suggestion
            // retain entries where elapsed is less than the provided duration
```

openmls/src/group/mls_group/past_secrets.rs
wysiwys: It's not a no op per se, but rather would replace an assumption that the `usize` value provided does not exceed `isize::MAX`. Do we assume here that only a valid size is provided?

openmls/src/group/mls_group/past_secrets.rs
wysiwys: This is now updated, with a test added :+1:

openmls/src/group/mls_group/past_secrets.rs
```diff
@@ -5,15 +5,63 @@ use crate::schedule::message_secrets::MessageSecrets;
...
+#[derive(Serialize, Deserialize)]
+#[cfg_attr(any(test, feature = "test-utils"), derive(Clone, PartialEq))]
+#[cfg_attr(feature = "crypto-debug", derive(Debug))]
+pub(crate) struct EpochTree {
+    epoch: u64,
     leaves: Vec<Member>,
+    message_secrets: MessageSecretsWithTimestamp,
 }
 
+// TODO: check timestamp ordering of past epoch trees on deserialization?
```
franziskuskiefer: Open todo

openmls/src/group/mls_group/mod.rs
```diff
@@ -644,6 +661,24 @@ impl MlsGroup {
         )
     }
 
+    /// Delete all past epoch secrets.
+    ///
+    /// For more information on the arguments to this method, see [`PastEpochDeletion`].
```
franziskuskiefer, line 679: The `PastEpochDeletion` then doesn't say much unfortunately.

openmls/src/group/mls_group/past_secrets.rs
```diff
@@ -172,11 +265,97 @@ impl MessageSecretsStore {
...
+            match before {
+                PastEpochDeletionTimeConfig::Timestamp(timestamp) => {
+                    self.delete_past_epoch_secrets_before_timestamp(timestamp)
+                }
+                PastEpochDeletionTimeConfig::Duration(duration) => {
+                    self.delete_past_epoch_secrets_older_than_duration(duration)
+                }
+            };
+            // ensure at most `max_past_epochs` entries are included
+            if let Some(max_past_epochs) = policy.max_past_epochs {
```
franziskuskiefer, line 364: Epochs with `None` are not deleted. That's the right thing. But the API needs to tell the user. So the public function should have something like: "Epochs without timestamp will not be deleted with this function when using a time based deletion schedule. Instead, use ... in order to delete those epochs."

openmls/src/group/mls_group/mod.rs
```diff
@@ -523,12 +523,29 @@ impl MlsGroup {
...
+        self.message_secrets_store.resize(policy);
+    }
+
+    /// Set the past epoch secret deletion policy for the group.
+    pub fn set_past_epoch_deletion_policy<Provider: OpenMlsProvider>(
+        &mut self,
+        provider: &Provider,
+        policy: PastEpochDeletionPolicy,
+    ) -> Result<(), SetPastEpochDeletionPolicyError<Provider::StorageError>> {
+        //
```
franziskuskiefer: Add comment or drop?

openmls/src/group/mls_group/tests_and_kats/tests/past_secrets.rs
```diff
@@ -275,13 +356,248 @@ fn delete_all<Provider: crate::storage::OpenMlsProvider>() {
...
+        .expect("no group for id");
+    assert_eq!(
+        alice_group_stored
+            .message_secrets_store()
+            .num_past_epoch_trees(),
+        1
+    );
+}
+
+// TODO: add more test cases
```
franziskuskiefer: Open todo.

openmls/src/group/mls_group/tests_and_kats/tests/past_secrets.rs
```diff
@@ -275,13 +356,248 @@ fn delete_all<Provider: crate::storage::OpenMlsProvider>() {
...
+    // ensure that the past epoch secrets are now empty
+    // assert all past secrets deleted
+    // NOTE: the `None` tree at the end will not be deleted
+    assert_eq!(message_secrets_store.num_past_epoch_trees(), 3);
+    assert!(message_secrets_store.secrets_for_epoch(4).is_some());
+    assert!(message_secrets_store.secrets_for_epoch(5).is_some());
+    assert!(message_secrets_store.secrets_for_epoch(6).is_some());
+}
+
+// TODO: add more test cases
```
franziskuskiefer: Open todo

openmls/src/group/mls_group/past_secrets.rs
franziskuskiefer: Oh I thought that happens in the policy already. It's not so this is fine.

openmls/src/group/mls_group/past_secrets.rs
wysiwys: I think this could go along with general validation of the `MessageSecretsStore` on deserialization, or validation of other structs in storage. This could be something to open another issue for, or leave as is

openmls/src/group/mls_group/mod.rs
wysiwys: :+1: Some more documentation is now added

book/src/user_manual/past_epoch_secret_deletion.md
```diff
@@ -0,0 +1,67 @@
...
+Delete all past epoch secrets older than a provided duration, leaving at most a provided number of epochs:
+```rust,no_run,noplayground
+{{#include ../../../openmls/tests/book_code_past_epoch.rs:duration_with_max_epochs}}
+```
+
+## Migration and deleting legacy entries
+
+Epoch secrets that were created before upgrading to `openmls=0.8.2` will not yet include a timestamp.
+
+After migration, these may not always be deleted by applying a time-based `PastEpochDeletion`. Only if a new secret that does include a timestamp is added later, and it matches the time-based condition in the `PastEpochDeletion`, all earlier past epoch secrets without timestamps will be deleted, as well. However, otherwise, past epoch secrets without timestamps will not be affected by applying time-based `PastEpochDeletion`s.
```
neekolas, line 51: I think this makes sense for a migration path

openmls/src/group/mls_group/config.rs
```diff
@@ -36,6 +36,140 @@ use crate::{
...
+///
+/// An [`MlsGroup`] also applies automatic deletion of past epoch secrets by default.
+///
+/// For more information, see [`PastEpochDeletionPolicy`] and [`MlsGroup::set_past_epoch_deletion_policy()`].
+///
+/// These methods can be used by the application to set up time-based deletion schedules:
+/// - [`PastEpochDeletion::before_timestamp()`]
+/// - [`PastEpochDeletion::older_than_duration()`]
+///
+/// **NOTE**: Epoch secrets that were created before upgrading to `openmls=0.8.2` will not yet include a timestamp.
```
franziskuskiefer: Let's not use the version, it may go into another one. Let's use the versions that have the old behaviour, i.e. the current one.

book/src/user_manual/past_epoch_secret_deletion.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+Delete all past epoch secrets older than a provided duration, leaving at most a provided number of epochs:
+```rust,no_run,noplayground
+{{#include ../../../openmls/tests/book_code_past_epoch.rs:duration_with_max_epochs}}
+```
+
+## Migration and deleting legacy entries
+
+Epoch secrets that were created before upgrading to `openmls=0.8.2` will not yet include a timestamp.
```
franziskuskiefer: Let's use the current release here as well.

openmls/src/group/mls_group/mod.rs
```diff
@@ -523,12 +523,29 @@ impl MlsGroup {
...
+    }
+
+    /// Set the past epoch secret deletion policy for the group.
+    pub fn set_past_epoch_deletion_policy<Provider: OpenMlsProvider>(
+        &mut self,
+        provider: &Provider,
+        policy: PastEpochDeletionPolicy,
+    ) -> Result<(), SetPastEpochDeletionPolicyError<Provider::StorageError>> {
+        // apply the new policy to the store
+        self.set_max_past_epochs(&policy);
```
franziskuskiefer: The policy here is stored in the group, so we need to persist that.

openmls/src/group/mls_group/config.rs
```diff
@@ -188,8 +365,8 @@ impl MlsGroupCreateConfig {
     }
 
     /// Returns the [`MlsGroupCreateConfig`] max past epochs.
-    pub fn max_past_epochs(&self) -> usize {
-        self.join_config.max_past_epochs
+    pub fn max_past_epochs(&self) -> Option<usize> {
```
franziskuskiefer, line 368: This is a breaking API change and needs to be documented in the changelog as well. Alternatively, it could return the actual max epoch size.

wysiwys: :+1: The changelog now includes an entry for this API change.

PR #1974, opened 2026-03-19
feat(sqlite_storage): implement transactional sqlite storage provider

quangtruongqualgo:
## Summary

This PR adds transaction support to the existing `openmls_sqlite_storage` crate.

### What changed
- Added `TransactionalSqliteStorageProvider<'tx, C>` to run storage operations
- Extended the SQLite storage helpers to support both normal and transaction-backed execution
- Used cached prepared statements

### Test coverage
Added integration tests that verify:
- successful transactions commit persisted changes
- failed transactions roll back persisted changes

### DIFF
--- a/sqlite_storage/src/encryption_key_pairs.rs
+++ b/sqlite_storage/src/encryption_key_pairs.rs
@@ -1,12 +1,12 @@
 use std::marker::PhantomData;
 
 use openmls_traits::storage::{Entity, Key};
-use rusqlite::{Connection, OptionalExtension, params};
+use rusqlite::{params, Connection, OptionalExtension};
 
 use crate::{
-    STORAGE_PROVIDER_VERSION,
     codec::Codec,
     wrappers::{EntityRefWrapper, EntityWrapper, KeyRefWrapper},
+    STORAGE_PROVIDER_VERSION,
 };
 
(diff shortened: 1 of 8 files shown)


kkohbrok:
Hi @quangtruongqualgo, the crate already supports the use of transactions via something like this:

```rust
let tx = conn.transaction()?;
let provider = SqliteStorageProvider::<MyCodec, _>::new(&*tx);
```

As such, I think the changes you propose are a little large for very limited ergonomic benefits, but please do let me know if I'm missing something.

What I do find interesting is the added `transaction` function that allows spawning a transaction-based storage provider from one based on a connection. Something like

```rust
pub fn with_transaction<F, T, E>(&mut self, f: F) -> Result<T, E>
where
    F: FnOnce(SqliteStorageProvider<C, &Connection>) -> Result<T, E>,
    E: From<rusqlite::Error>,
{
    let tx = self.connection.borrow_mut().transaction()?;
    let provider = SqliteStorageProvider::<C, &Connection>::new(&*tx);
    let result = f(provider);
    match result {
        Ok(val) => {
            tx.commit()?;
            Ok(val)
        }
        Err(e) => Err(e), // tx rolls back on drop
    }
}
```

PR #1982, opened 2026-03-25
update the `rusqlite` version used in `openmls_sqlite_storage`

### LINKED ISSUE #1904: Update `rusqlite` version in `sqlite_storage`
Since we have `sqlite_storage` and `sqlx_storage` in the same workspace, we have to align their `libsqlite-sys` dependency. This will only be possible when `sqlx` version 9.0 is released, which allows a range of `libsqlite-sys` versions.

wysiwys:
This pull request updates the `rusqlite` version in `openmls_sqlite_storage` to `0.37`.

It also moves the `openmls_sqlx_storage` crate out of the workspace. This allows `openmls_sqlite_storage` to use a version of `libsqlite3-sys` that does not need to be compatible with that required in `openmls_sqlx_storage`

Lastly, the workspace tests on CI now test the `openmls_sqlx_storage` crate separately.

A note on the `rusqlite` version: `rusqlite` versions higher than `0.37` are not yet compatible with the latest version of `refinery` used in the `openmls_sqlite_storage` crate.

Resolves #1904

### DIFF
--- a/.github/workflows/build_test_workspace.yml
+++ b/.github/workflows/build_test_workspace.yml
@@ -33,11 +33,23 @@ jobs:
       - uses: actions/checkout@v6
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: |
+            . -> target
+            sqlx_storage -> target
       - name: Build workspace
-        run: cargo build --workspace --all-targets --exclude openmls-fuzz
+        run: |
+          cargo build --workspace --all-targets --exclude openmls-fuzz
+          cargo build --manifest-path sqlx_storage/Cargo.toml
       - name: Test workspace
-        run: cargo test --workspace --all-targets --exclude=openmls --exclude openmls-fuzz
+        run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/build_test_workspace.yml
```diff
@@ -33,11 +33,23 @@ jobs:
...
       - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: |
+            . -> target
+            sqlx_storage -> target
       - name: Build workspace
-        run: cargo build --workspace --all-targets --exclude openmls-fuzz
+        run: |
+          cargo build --workspace --all-targets --exclude openmls-fuzz
+          cargo build --manifest-path sqlx_storage/Cargo.toml
```
franziskuskiefer: I'd suggest to move this into a separate workflow to keep things a little cleaner.

franziskuskiefer (commented):
lgtm with a nit on the ci.

kkohbrok (dismissed):
Thanks! Looks good minus [Franziskus's comment](https://github.com/openmls/openmls/pull/1982/changes#r2989331666).

kkohbrok (approved):

PR #1990, opened 2026-04-01
Fix backward incompatible changes

### LINKED ISSUE #1989: [Bug] serde untagged and flatten break bincode
**Describe the bug**

The `MessageSecretsWithTimestamp` struct (openmls/src/group/mls_group/past_secrets.rs#L14) struct uses `#[serde(flatten)]` on the `message_secrets` field. 

The `Format` enum (openmls/src/group/mls_group/config.rs#L149) uses `#[serde(untagged)]`.

Both of these are bincode incompatible. Xmtp has been using bincode as it's serde implementation provider and these latest changes won't allow any db operations for openmls to succeed.

**To Reproduce**

Use bincode as the serde implementation inside of the StorageProvider trait.

**Expected behavior**

All data structures work with all serde implementations.

**Actual behavior**

Everything fails with a bincode `SerializationError`

**Screenshots or debug log**

 - OS & OS version: macOS 25.3.0
 - Architecture: aarch64
 - Version: latest main

**Additional context**

tylerhawkes:
Fixes https://github.com/openmls/openmls/issues/1989

The contract is with serde and not just serde_json, so all fields changes are essentially breaking changes.

To test this more formally, we could use multiple serde providers in the tests, especially those that have the hardest compatibility guarantees.


openmls/src/group/mls_group/config.rs
```diff
@@ -63,6 +63,34 @@ impl Default for PastEpochDeletionPolicy {
...
+        D: serde::Deserializer<'de>,
+    {
+        let epochs = u64::deserialize(deserializer)?;
+        let epochs = usize::try_from(epochs).map_err(serde::de::Error::custom)?;
+        if epochs == usize::MAX {
+            Ok(Self::KeepAll)
+        } else {
+            Ok(Self::MaxEpochs(epochs))
+        }
+    }
```
wysiwys, line 94: One minor point about the `usize` conversion - a policy set to `Self::MaxEpochs(usize::MAX)` and then loaded from storage would instead be read as a `Self::KeepAll`. Despite these policies being identical in effect, this conversion might be unexpected. 

I think a line in the documentation for `PastEpochDeletionPolicy` could be helpful here, noting that `MaxEpochs(usize::MAX)` will be converted to `KeepAll`. It could also make sense to consider ensuring that the policy is converted earlier on, e.g. when creating or updating the `MessageSecretsStore` for the group

wysiwys (dismissed):
Thanks, this looks good to me, with one small point on the `usize` conversion.

Additionally, I agree that it makes sense to add tests for other serde providers, or to adapt the existing tests focusing on backwards compatibility to test multiple serde providers

openmls/src/group/mls_group/config.rs
tylerhawkes: I've added a comment to address this in the latest push

openmls/src/schedule/message_secrets.rs
```diff
@@ -70,6 +76,29 @@ impl MessageSecrets {
     pub(crate) fn secret_tree_mut(&mut self) -> &mut SecretTree {
         &mut self.secret_tree
     }
+
+    pub(crate) fn timestamp(&self) -> Option<&std::time::SystemTime> {
```
keks, line 80: Is there a reason this returns an `Option<&_>`? `SystemTime` is `Copy`, so this could just be `self.added_at`, unless I am missing something. That would also save us some `.copied()` and `.into()` elsewhere.

keks (changes requested):
Looks pretty good to me.

One thing that I also noticed but don't have a better idea for off the top of my head is that the methods are all called `timestamp`, but the field is called `added_at`. The former is not super descriptive, but `with_added_at` also roll off the tongue super well. Not sure what to do there.

openmls/src/schedule/message_secrets.rs
tylerhawkes: There was no good reason. I've changed it in the most recent commit.

tylerhawkes:
I somehow pulled in all of xmtp's changes yesterday. That is fixed now. Let me know if there's anything else I need to change.

keks:
I think this looks good now but wanted to way for CI before I give this my green light - but looks Actions is not picking up the jobs.

keks (approved):

PR #1993, opened 2026-04-13
chore: add lockfile to repository

boxdot:
Rust official blog changed the guidance on committing Lockfiles in
<https://blog.rust-lang.org/2023/08/29/committing-lockfiles>.

There are several benefits of having a lockfile in the repository, even
for libraries:

* Each commit in repository is buildable, even after a long time because
  it tracks which dependencies were used *per* commit. In particular,
  it make bisecting possible.
* It allows for tooling to check for vulnerabilities in the
  dependencies.
* It is easy to test in CI with latest dependencies by just running
  `cargo update` (in case such job is needed).
* It gives a supply chain security: committed lockfile ensures that CI
  builds and tests exactly what is specified in the lockfile. Note that
  lockfile contains the hashes of the content of the dependencies not
  only their versions.


boxdot:
I am proposing to add the lockfile also because I hit the wall today when building openmls. My local lockfile was outdated and contained an old version of <https://github.com/mlswg/mls-implementations.git> which was built with a different version of tonic. It took me some time before I realized that an update is needed.

raphaelrobert (approved):
Let's try this out and see how it goes. Reverting it would be easy. I'm curious if we should also add a job that regularly runs `cargo update` as the blog post suggests.

kkohbrok (approved):

PR #1994, opened 2026-04-13
feat: Post-Quantum Cryptography support (ML-KEM + ML-DSA)

boxdot:
Co-authored-by: Konrad Kohbrok <konrad.kohbrok@datashrine.de>


boxdot:
Waiting for https://github.com/cryspen/hpke-rs/pull/150 before *ready for review*.

p-barabas:
Hi, 

is there any progress with PQC support? Is there any deadline for it?

Thanks in advance.

kkohbrok:
@p-barabas OpenMLS generally already supports the experimental XWING-based ciphersuite `MLS_256_XWING_CHACHA20POLY1305_SHA256_Ed25519`, for example, via the libcrux crypto provider. This PR is about adding support for the pure PQ ciphersuites. All of this support is somewhat experimental, however, as we're waiting for the [PQ ciphersuites draft](https://www.ietf.org/archive/id/draft-ietf-mls-pq-ciphersuites-04.html) to be finalized in the MLS WG. I believe that draft is again waiting for the draft introducing the necessary HPKE ciphersuites.

But all of this is somewhat off-topic for this issue. Please file a new issue, or join our [Zulip instance](https://openmls.zulipchat.com/) if you have more questions.

openmls/src/key_packages/key_package_in.rs
```diff
@@ -209,6 +209,24 @@ impl KeyPackageIn {
     pub(crate) fn version_is_supported(&self, protocol_version: ProtocolVersion) -> bool {
         self.payload.protocol_version == protocol_version
     }
+
+    /// Assume that the signature is valid and return the [`KeyPackage`].
+    ///
+    /// Use with caution, the client must guarantee that the key package is verified.
+    pub fn unwrap_verified(self) -> Result<KeyPackage, KeyPackageVerifyError> {
```
boxdot, line 216: Here and below, this is an unrelated change, however I am using it downstream already. It introduces an explicit mechanism to bypass the type checked In/Out parameters and conversion between them via verification.

Rationale: Sometimes, we can guarantee that a type is already verified however we loaded it via TLS from some other source (DB, Networks, Disk). This allows to skip the expensive signature verification. I called it `unwrap_` because it is very similar to Rust's Option or Result unwrap, where the developer must guarantee invariants.

I am open moving it out from this PR and also choosing a different name. Also I introduced it only for the types I am using it with.

boxdot:
I will split this PR in 4 parts:

* P384, ML-DSA and general PQ ciphersuite support for RustCrypto provider
* missing processing functions for PublicGroup 
* unwrap-style functions for skipping validation
* new PSK-type

openmls/src/key_packages/key_package_in.rs
boxdot: Will be moved out into a different PR.

boxdot:
Fully split into

* #2046 
* #2045 
* #2044 
* #2043 

For the main implementation see #2046. Closing this one.

PR #2010, opened 2026-04-23
Allow proposing a self update that includes a new signer

### LINKED ISSUE #2008: [Bug] Not possible to propose self update with new signer
**Describe the bug** 

Currently, it isn't possible to do a _full_ self key rotation as a proposal. While `propose_self_update` has space for a new signing key within credential_with_key in the LeafParameters, if a new public key is supplied to this function both the new leaf node and proposal will be signed by the _current_ signer. To propose a self update that also changes the member's signer, the leaf would need to be signed by the new signer and proposal signed by the old signer.

**Suggested fix**

`propose_self_update_with_new_signer` should exist (similar to the committing `self_update_with_new_signer`) which wires up the new and old signers together to construct a valid signing key rotation.

If the above makes sense, I can follow this up with a PR to implement this. I wanted to create this issue to verify I'm not missing something that would make this functionality undesirable / a reason this doesn't already exist.

aidantwoods-1p:
Resolves https://github.com/openmls/openmls/issues/2008

Adds `propose_self_update_with_new_signer`. I opted to refactor the internals of `propose_self_update` to plumb in an optional new signer. The new public function and the existing propose now defer to the same internal helper (which differs only by an optional signer).

### DIFF
--- a/openmls/src/group/mls_group/updates.rs
+++ b/openmls/src/group/mls_group/updates.rs
@@ -94,7 +94,7 @@ impl MlsGroup {
     /// Creates a proposal to update the own leaf node. Optionally, a
     /// [`LeafNode`] can be provided to update the leaf node. Note that its
     /// private key must be manually added to the key store.
-    fn _create_self_update_proposal<Provider: OpenMlsProvider, S: Signer>(
+    fn create_self_update_proposal_internal<Provider: OpenMlsProvider, S: Signer>(
         &mut self,
         provider: &Provider,
         old_signer: &impl Signer,
@@ -169,14 +169,14 @@ impl MlsGroup {
         Ok(update_proposal)
     }
 
-    fn _propose_self_update<Provider: OpenMlsProvider, S: Signer>(
+    fn propose_self_update_internal<Provider: OpenMlsProvider, S: Signer>(
         &mut self,
         provider: &Provider,
         old_signer: &impl Signer,
         new_signer: Option<NewSignerBundle<'_, S>>,
         leaf_node_parameters: LeafNodeParameters,
     ) -> Result<(MlsMessageOut, ProposalRef), ProposeSelfUpdateError<Provider::StorageError>> {
-        let update_proposal = self._create_self_update_proposal(
+        let update_proposal = self.create_self_update_proposal_internal(
             provider,
             old_signer,
             new_signer,
@@ -209,7 +209,7 @@ impl MlsGroup {
         signer: &S,
         leaf_node_parameters: LeafNodeParameters,
     ) -> Result<(MlsMessageOut, ProposalRef), ProposeSelfUpdateError<Provider::StorageError>> {
-        self._propose_self_update(
+        self.propose_self_update_internal(
             provider,
             signer,
             None::<NewSignerBundle<'_, S>>,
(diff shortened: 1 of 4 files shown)


openmls/src/group/mls_group/updates.rs
```diff
@@ -169,4 +199,60 @@ impl MlsGroup {
...
+        // shares almost all logic.
+        enum NoSigner {}
+        impl Signer for NoSigner {
+            fn sign(&self, _payload: &[u8]) -> Result<Vec<u8>, SignerError> {
+                unreachable!()
+            }
+            fn signature_scheme(&self) -> SignatureScheme {
+                unreachable!()
+            }
+        }
```
kkohbrok: It may technically change the public API, but I'd rather introduce a new type parameter `S` to mirror the approach in `CommitBuilder`'s `build`.

openmls/src/group/mls_group/updates.rs
```diff
@@ -90,11 +94,12 @@ impl MlsGroup {
     /// Creates a proposal to update the own leaf node. Optionally, a
     /// [`LeafNode`] can be provided to update the leaf node. Note that its
     /// private key must be manually added to the key store.
-    fn _propose_self_update<Provider: OpenMlsProvider>(
+    fn _create_self_update_proposal<Provider: OpenMlsProvider, S: Signer>(
```
kkohbrok: While you're here, could you add an `_internal` suffic instead of sticking with the `_` prefix? Same with `_propose_self_update` below.

openmls/src/group/mls_group/updates.rs
```diff
@@ -108,14 +113,36 @@ impl MlsGroup {
...
-            leaf_node_parmeters,
-        )?;
+        if let Some(new_signer) = new_signer {
+            // Reconcile `leaf_node_parameters.credential_with_key` with
+            // `new_signer.credential_with_key`. Mirrors the commit-path logic in
+            // `CommitBuilder::build_internal`.
+            if let Some(ln_cred) = leaf_node_parameters.credential_with_key() {
+                if ln_cred != &new_signer.credential_with_key {
+                    return Err(ProposeSelfUpdateError::InvalidLeafNodeParameters);
+                }
```
kkohbrok, line 123: Just for completeness, could you add a check here to make sure that the `credential_with_key`'s signature scheme matches that of the `new_signer`? I just noticed that it's also missing in `build_with_new_signer`, but that should probably go into its own PR.

kkohbrok (changes requested):
Thanks for the PR! Just a few nits and then this is good to merge.

openmls/src/group/mls_group/updates.rs
aidantwoods-1p: The downside of a dedicated type param for the new signer is we put it on the caller to specify a signer type for the _new_ signer that isn't being used. A minor hack we could do is type param the existing signer and then use that to satisfy the type param for the unused new signer generics.

openmls/src/group/mls_group/updates.rs
aidantwoods-1p: The signature scheme should be included in the equality check between the two `CredentialWithKey` objects right?

aidantwoods-1p: Possible the check you want is to reject if `self.ciphersuite().signature_algorithm() != new_signer.signer.signature_scheme()`?

kkohbrok: Ah, you're right of course. The Credential doesn't actually expose that.

aidantwoods-1p:
Ready for re-review!

kkohbrok (dismissed):
Thanks for making those additional changes! Let's get this merged.

aidantwoods-1p:
Apologies, I forgot to format before the last push. Pipelines should be happy now!

kkohbrok (approved):

PR #2012, opened 2026-04-29
Don't explicitly include `openmls_sqlite_storage` in `extensions-draft-08` feature

wysiwys:
This PR updates the `extensions-draft-08` feature in the `openmls` library to no longer explicitly enable `openmls_sqlite_storage?/extensions-draft-08`.

This change prevents the crate's optional dependency `openmls_sqlite_storage` from being included in the lockfile when the `extensions-draft-08` feature is enabled, which could lead to unexpected version conflicts, in particular related to `libsqlite3-sys`. See https://github.com/rust-lang/cargo/issues/10801

For the `openmls` crate tests, to allow running these tests using the `openmls_sqlite_storage` provider when the `extensions-draft-08` feature is enabled, this PR also adds a new testing feature to `openmls`, for use with the `extensions-draft-08` feature. There may be a better way to allow this, though, for example potentially by reorganizing crate tests that use the `openmls_sqlite_storage` crate.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -85,4 +85,4 @@ jobs:
         run: |
           cargo hack check --feature-powerset --no-dev-deps --verbose -p openmls \
           --group-features test-utils,crypto-debug,content-debug,backtrace,libcrux-provider,sqlite-provider \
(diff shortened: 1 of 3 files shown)


kkohbrok (approved):
👍🏻

PR #2016, opened 2026-04-30
Migrate to web-time

tylerhawkes:
This avoids panics from

/library/std/src/sys/time/unsupported.rs:35:9:
      time not implemented on this platform

and fixes https://rustsec.org/advisories/RUSTSEC-2024-0384 that `cargo deny check advisories` warns about.


franziskuskiefer (approved):
lgtm, thanks!

PR #2017, opened 2026-05-06
Virtual clients: Allow processing of own commits

kkohbrok:
This PR allows clients to send and receive commits via a virtual client.

- Allow derivation of virtual-client specific secrets from emulation group and store them in the storage provider
- Allow creation of virtual client commits from those secrets and embed information in leaf component
- When receiving a virtual client commit sent by another emulation client, read embedded information, derive the correct secrets and process commit
- Extended storage interface for virtual client secrets 

External commits are not covered yet and will be handled in a follow-up.


.github/workflows/build.yml
```diff
@@ -82,4 +82,4 @@ jobs:
           cargo hack check --feature-powerset --no-dev-deps --verbose -p openmls \
           --group-features test-utils,crypto-debug,content-debug,backtrace,libcrux-provider,sqlite-provider \
           --group-features extensions-draft-08,extensions-draft-08-test-dependencies \
-          --group-features js,js-test
+          --group-features virtual-clients-draft,virtual-clients-draft-test-dependencies
```
raphaelrobert, line 86: Why remove the js group here?

openmls/src/components/vc_derivation_info.rs
```diff
@@ -0,0 +1,567 @@
...
+    messages::PathSecret,
+    schedule::pprf::{Pprf, PprfError, Prefix256},
+    treesync::node::encryption_keys::EncryptionKeyPair,
+};
+
+/// Component ID under which the virtual-clients derivation info is carried in
+/// the leaf node's `app_data_dictionary` extension.
+///
+/// `0xFFFF` is a placeholder until the IETF draft is assigned an IANA value.
+pub const VC_COMPONENT_ID: u16 = 0xFFFF;
```
raphaelrobert: I think it would be good to request a specific value in the draft and use that here going forward, because this might collide with other components.

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -981,6 +1162,199 @@ impl CommitBuilder<'_, Complete, &mut MlsGroup> {
...
+    resolved_dictionary: AppDataDictionary,
+    rand: &impl OpenMlsRand,
+    crypto: &impl OpenMlsCrypto,
+    ciphersuite: openmls_traits::types::Ciphersuite,
+) -> Result<crate::treesync::diff::OwnUpdatePathOverride, CreateCommitError> {
+    // Build the encrypted EpochInfoTbe payload the receiver will decrypt
+    // with `epoch_encryption_key`. The same struct is hashed to produce
+    // the PPRF input so that sender and receiver derive the same
+    // per-commit secret.
+    let random = rand.random_vec(32).map_err(|e| {
```
raphaelrobert: I know this is hard-coded in the draft, but a constant would still be nice here rather than just the numeral.

openmls/src/schedule/pprf/mod.rs
```diff
@@ -132,7 +132,7 @@ impl<P: Prefix> Pprf<P> {
     }
 
     /// Evaluates the PPRF at the given input.
-    pub(super) fn evaluate<Input: AsIndexBytes>(
+    pub(crate) fn evaluate<Input: AsIndexBytes>(
```
raphaelrobert, line 138: It's not part of this PR, but there's an `unwrap()` in line 176 that we should get rid of.

openmls/src/schedule/pprf/prefix.rs
```diff
@@ -63,3 +64,49 @@ impl From<SerdePrefix16> for Prefix16 {
...
+            bits: [0; 32],
+            len: 0,
+        }
+    }
+
+    fn push_bit(&mut self, bit: bool) {
+        if bit {
+            let byte_idx = (self.len / 8) as usize;
+            let bit_idx = 7 - (self.len % 8) as u8;
+            self.bits[byte_idx] |= 1 << bit_idx;
```
raphaelrobert, line 168: This can potentially panic when len >= 255

openmls/src/components/vc_derivation_info.rs
```diff
@@ -0,0 +1,567 @@
...
+        crypto: &impl OpenMlsCrypto,
+        ciphersuite: Ciphersuite,
+    ) -> Result<EpochId, VirtualClientsError> {
+        let secret = self
+            .0
+            .derive_secret(crypto, ciphersuite, EPOCH_ID_LABEL)
+            .map_err(|e| {
+                log::error!("vc: derive epoch id failed: {e:?}");
+                VirtualClientsError::CryptoError
+            })?;
```
raphaelrobert: There are a few instances in this file where the original error is only logged but not propagated. This might be ok, if it is really intentional. I want to make sure that's the case.

openmls/src/components/vc_derivation_info.rs
```diff
@@ -0,0 +1,567 @@
...
+            .finish()
+    }
+}
+
+impl EmulatorEpochSecret {
+    /// Construct an `EmulatorEpochSecret` from raw bytes. Bytes are
+    /// expected to be the output of the emulation group's
+    /// `safe_export_secret(VC_COMPONENT_ID)`.
+    pub(crate) fn new(bytes: &[u8]) -> Self {
+        Self(Secret::from_slice(bytes))
```
raphaelrobert, line 153: We should have a debug_assert at least to make sure the length is ok, ideally, we'd have better type safety

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -272,8 +282,12 @@ impl LeafNode {
         group_id: GroupId,
         leaf_index: LeafNodeIndex,
         signer: &impl Signer,
+        encryption_key_pair_override: Option<EncryptionKeyPair>,
```
raphaelrobert: Why not use the feature flag here as well? That would be safer and avoid unrelated code changes

openmls/src/group/mls_group/processing.rs
```diff
@@ -321,6 +324,114 @@ impl MlsGroup {
...
+
+        let pprf_input_bytes = pprf_input(crypto, ciphersuite, &epoch_info)?;
+        let operation_secret: OperationSecret = pprf
+            .evaluate(crypto, ciphersuite, &pprf_input_bytes)
+            .map_err(VirtualClientsError::from)?
+            .into();
+
+        // Persist puncture immediately. Reprocessing the same commit will
+        // fail with `PprfError::PuncturedInput`, which is exactly what we
+        // want for forward secrecy.
```
raphaelrobert, line 426: Could this be a DoS vector? In case we receive a commit that is not valid for unrelated reasons, we couldn't later process a valid commit with the same input to the PPRF because we punctured the PPRF. I might also misunderstand the scope here, and it's not an issue.

openmls/src/components/vc_derivation_info.rs
```diff
@@ -0,0 +1,567 @@
...
+pub const VC_COMPONENT_ID: u16 = 0xFFFF;
+
+// Operation-secret child labels. Each child is derived from the per-commit
+// operation secret produced by evaluating the per-epoch PPRF on the per-commit
+// input. Only `Encryption Key` and `Path Generation` are wired up at the
+// moment; the remaining ones are reserved so that `key_package` / `application`
+// derivation can be added without churning the constants.
+const ENCRYPTION_KEY_LABEL: &str = "Encryption Key";
+const PATH_GENERATION_LABEL: &str = "Path Generation";
+#[allow(dead_code)] // reserved for KeyPackage/Application operation types
```
raphaelrobert: Why is this dead code? If it gets resolved in subsequent PRs, it's better to make this a comment.

openmls/src/group/mls_group/processing.rs
```diff
@@ -482,17 +593,34 @@ impl MlsGroup {
...
+                // and must fall through to the normal HPKE path.
+                #[cfg(feature = "virtual-clients-draft")]
+                let vc_material = if matches!(&sender, Sender::Member(idx) if *idx == self.own_leaf_index())
+                {
+                    self.load_vc_commit_material(provider, commit)?
+                } else {
+                    None
+                };
+                #[cfg(not(feature = "virtual-clients-draft"))]
+                let _ = commit;
```
raphaelrobert, line 614: You could also gate the binding with something like `#[cfg_attr(not(feature = "virtual-clients-draft"), allow(unused))]` instead of doing this

raphaelrobert (changes requested):
Good progress on the feature! A few remarks and questions

.github/workflows/build.yml
kkohbrok: Good question. Now sure how that happened.

openmls/src/components/vc_derivation_info.rs
kkohbrok: I've filed a PR against the spec: https://github.com/mlswg/mls-virtual-clients/pull/20

openmls/src/components/vc_derivation_info.rs
kkohbrok: It's hard to check anything here without taking the emulation group ciphersuite as additional input. And for more type safety, we'd have to change the safe_export_secret interface, which I'd rather not do in this PR.

openmls/src/components/vc_derivation_info.rs
kkohbrok: Good point.

openmls/src/group/mls_group/processing.rs
kkohbrok: Good question. I think this is not a problem, because you first have to AEAD decrypt the derivation info. That should be guard enough.

kkohbrok (commented):
Thanks for the review! All good points. I think I addressed everything.

openmls/src/components/vc_derivation_info.rs
raphaelrobert: The new value should still be used here

raphaelrobert (approved):
Thanks for making the changes

PR #2019, opened 2026-05-07
docs: clarify that merge_pending_commit is required after add_members

Thisiskarthikraj:
## Summary

Clarifies that `merge_pending_commit()` must be called after `add_members()`.

## Problem

The current documentation does not explicitly mention that `add_members()` only creates a pending commit. This can lead to confusion because the group state (epoch) is not updated unless `merge_pending_commit()` is called.

## Solution

Added a note explaining:
- the creation of a pending commit
- the requirement to call `merge_pending_commit()`
- the impact of not doing so

## Impact

Improves developer experience and prevents incorrect usage of the API.


raphaelrobert:
Good point, could you please also check where else this needs to be clarified? Pending commits are also generated for removing group members.

franziskuskiefer (changes requested):
changing state

PR #2021, opened 2026-05-11
Fix bincode serde compatibility

tylerhawkes:
## Summary                          
                                                                                                                                                                                                                                                              
  Restore openmls-0.7.x serde wire-format compatibility for non-self-describing formats (bincode, postcard, …). Builds on #1990, which fixed the same class of bug for `serde_json` only.                                                                     
                                                                                                                                                                                                                                                              
  `serde`'s contract is with the underlying format, not with `serde_json`. Several changes since 0.7.1 — new enum variants inserted mid-declaration, new `#[serde(default)]` tail fields, a `usize`-shaped sentinel — produce stable JSON output but corrupt or refuse to load data persisted by 0.7.x bincode/postcard clients.                                                                                                                                                                
                                                                                                                                                                                                                                                              
  ## What broke since 0.7.1                                                                                                                                                                                                                                   
                                      
  | Type | Change | bincode impact |                                                                                                                                                                                                                          
  |---|---|---|                                                                                                                                                                                                                      
  | `ProposalType` / `Proposal` | `AppAck` removed; `Grease`, `AppEphemeral`, `AppDataUpdate` inserted | persisted `Custom(_)` decodes as `SelfRemove`; `SelfRemove` decodes as `Grease` |                                                                    
  | `Extension` / `ExtensionType` | `Grease`, `AppDataDictionary` inserted before `Unknown` | persisted `Unknown(_)` decodes as one of the new variants |                                                                                                     
  | `CredentialType` | `Grease` inserted before `Other` | persisted `Other(_)` decodes as `Grease(_)` |                                                                                                                                                       
  | `MessageSecrets` | `added_at: Option<SystemTime>` added with `#[serde(default)]` | bincode hits EOF; `MessageSecretsStore` fails to load |                                                                                                                
  | `MemberStagedCommitState` | `application_export_tree: Option<…>` added with `#[serde(default)]` (feature-gated) | same EOF when feature is enabled |                                                                                                      
  | `PastEpochDeletionPolicy::KeepAll` | sentinel was `usize::MAX as u64` | `wasm32` (`u32::MAX`) and 64-bit (`u64::MAX`) wrote different bytes; cross-host `KeepAll` failed to deserialize on `wasm32` |                                                     
                                                                                                                                                                                                                                                              
  ## Commits                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                              
  1. fix: stabilize serde variant indices for persisted enums — restore 0.7.x declaration order; new variants moved to the end. The removed AppAck variant is replaced with a hidden _AppAck unit placeholder so subsequent indices don't shift. openmls never originated AppAck (no public API constructs one), and ProposalQueue::filter_proposals panics on any inbound Proposal::AppAck, so no openmls deployment plausibly has persisted state containing this variant. If any does survive (e.g. from a fork), the unit placeholder will misalign subsequent QueuedProposal fields and the deserialize will fail loudly in the common case.                                      
  2. **`fix: tolerantly deserialize trailing optional fields on persisted structs`** — hand-written `Deserialize` for `MessageSecrets` and `MemberStagedCommitState` that probes the trailing field via `SeqAccess::next_element` and treats EOF / `Ok(None)`
  as "field absent → `None`". Both structs are positioned as the final field of their containing record, so the EOF-tolerant fallback can't misinterpret unrelated bytes.                                                                                     
  3. **`fix: encode persisted usize values as portable u64`** — fix the `KeepAll` sentinel (`u64::MAX`, checked before narrowing). Defensively annotate the remaining persisted `usize` fields with a `usize_as_u64` adapter; this is a wire no-op against
  today's `serde` `usize` impl but pins the convention.                                                                                                                                                                                                       
  4. **`refactor: lock persisted enum variant indices via custom serde impls`** — replace the derived `Serialize`/`Deserialize` on the five enums with hand-written impls that pin each variant's `variant_index` as a numeric literal. Wire bytes
  byte-identical to the previous positional encoding. **This commit is independent — feel free to drop it if the size isn't worth the safety; the rest of the fix stands.**                                                                                   
  5. **`test: lock serde variant indices for persisted enums`** — minimal `#[cfg(test)]` `serde::Serializer` whose `Ok` is `(u32, &'static str)` and whose `serialize_*_variant` methods capture and return the call. Five unit tests pin every variant of
  each enum.                                                                                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                     
  ## What this does *not* break                                                                                                                                                                                                                               
                                                                                                                                                                                                                                     
  - **JSON / CBOR / msgpack** — every existing variant name is preserved. The `_AppAck` placeholder serializes as `"AppAck"`.                                                                                                                                 
  - **TLS wire format** — `From<u16>` mappings on the type-ID enums and the hand-written codecs in `messages/codec.rs` / `extensions/codec.rs` are untouched.
  - **Public API** — every variant name, public method, and trait impl preserved. `_AppAck` is `#[doc(hidden)]`.                                                                                                                                              
  - **Cross-architecture key derivation** — not affected; the `usize` fields touched are bookkeeping (cache caps, ring-buffer cursor, padding length, PPRF tree width). HKDF inputs are sized by ciphersuite constants and TLS-length-prefixed bytes.         
  - **Derived `Ord` on the type-ID enums** — restored to 0.7.x order for existing variants; new variants sort higher.                                                                                                                                         
                                                                                                                                                                                                                                                              
## Caveats worth flagging                                                                                                                                                                                                                                    
  - Exhaustive matches. External code that exhaustively matches on Proposal or ProposalType will need a _AppAck arm. The variant is #[doc(hidden)] but not #[non_exhaustive]. If you'd prefer marking these enums #[non_exhaustive] as a one-time API change instead, happy to do that — would let us drop the placeholder entirely and just delete the variant.
  - Inbound AppAck from non-openmls peers. If a peer ever sends a Proposal with proposal_type = 0x000b, the unit _AppAck placeholder consumes the variant tag and leaves the wrapped payload in the stream, causing the surrounding QueuedProposal deserialization to fail. This was already broken before (the receive path panics in filter_proposals), so this is a strict improvement, but worth flagging as an explicit non-fix.

### DIFF
--- a/openmls/src/credentials/mod.rs
+++ b/openmls/src/credentials/mod.rs
@@ -77,10 +77,7 @@ pub mod errors;
 /// | 0xDADA           | GREASE                   | Y | RFC XXXX |
 /// | 0xEAEA           | GREASE                   | Y | RFC XXXX |
 /// | 0xF000  - 0xFFFF | Reserved for Private Use | - | RFC XXXX |
-// Variant order is part of the serde wire format for non-self-describing
-// serializers like bincode. Do not reorder existing variants; append new
-// variants at the end of the enum.
-#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
 #[repr(u16)]
 pub enum CredentialType {
     /// A [`BasicCredential`]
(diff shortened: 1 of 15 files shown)


raphaelrobert:
Thanks for the PR. This looks like something we should have outside the library to facilitate one-time migrations.

tylerhawkes:
@raphaelrobert This PR has no impact on json or other self describing (human readable) serializations. This just fixes bincode for us to be backwards compatible back to 0.7.1 (which we have several million clients currently using)

raphaelrobert:
I'm not questioning your need for it, but this is a pretty substantial departure from the status quo:
 - So far, the standing rule was that potential data format migrations should not live inside the library. We have not had to deal with migration so far because the changes haven't broken self-describing formats. Compatibility for bincode was not something the library ever promised.
 - This PR is 100s of lines of obviously vibe-coded code. It is not on par in terms of structure with the rest of the codebase. This lowers the quality and adds to the maintenance cost.

I don't see why this application-specific backwards compatibility is something that should be part of the library. A one-time migration should not be much harder to implement, and would avoid the two issues mentioned above.

PR #2023, opened 2026-05-15
Stabilize serde variant indices for persisted enums

tylerhawkes:
bincode and postcard encode enum variant tags as positional u32s, so when
variants get inserted mid-declaration the index of every following variant
shifts and persisted data silently decodes as the wrong variant.

Since 0.7.x, `ProposalType`/`Proposal`, `Extension`/`ExtensionType`, and
`CredentialType` have all had variants added (`Grease`, `AppEphemeral`,
`AppDataUpdate`, `AppDataDictionary`) — and `AppAck` removed from
`Proposal`/`ProposalType` — in positions that shift the indices.

This PR restores the 0.7.x order, appends every new variant to the end,
and replaces the removed `AppAck` with a hidden `_AppAck` unit placeholder
at its original index 7. JSON/CBOR variant names, the TLS wire format,
and the public API are all unchanged.

The four enums with feature-gated variants also get a note in the
existing variant-order comment: new non-optional variants must not be
added after a `#[cfg(feature = ...)]` variant, because toggling the
feature would otherwise shift every following index and silently corrupt
persisted data.

This means that for non human-readable serde implementations that after a
non-optional variant is added that enabling the feature would be a breaking
change for consumers of the library.

#### On the `_AppAck` placeholder

Open to discussion. The placeholder is doing two jobs: it occupies index 7
(which no real persisted data hits — openmls never produced `AppAck`) and,
more importantly, it keeps `SelfRemove` at index 8 and `Custom` at index 9,
which 0.7.x bincode/postcard data *does* hit. Without the placeholder,
persisted `Custom(N: u16)` silently decodes as `Grease(N)` on
`ProposalType`, and persisted `SelfRemove` either fails or silently
corrupts on `Proposal`.

The cost is that exhaustive matches on `Proposal`/`ProposalType` in
external code need an `_AppAck` arm. If a tidier alternative is preferred,
marking these enums `#[non_exhaustive]` as a one-time API change would
let the placeholder be dropped and the variant deleted cleanly — happy
to take that direction.

This can go away by using #2026 for the serde impls which wouldn't break 0.8 users.

Split from #2021.

### DIFF
--- a/openmls/src/credentials/mod.rs
+++ b/openmls/src/credentials/mod.rs
@@ -77,6 +77,9 @@ pub mod errors;
 /// | 0xDADA           | GREASE                   | Y | RFC XXXX |
 /// | 0xEAEA           | GREASE                   | Y | RFC XXXX |
 /// | 0xF000  - 0xFFFF | Reserved for Private Use | - | RFC XXXX |
+// Variant order is part of the serde wire format for non-self-describing
+// serializers like bincode. Do not reorder existing variants; append new
+// variants at the end of the enum.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 #[repr(u16)]
 pub enum CredentialType {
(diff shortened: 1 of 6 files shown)


wysiwys:
Closing this PR in favor of the changes in #2034

PR #2025, opened 2026-05-15
Encode persisted usize values as portable u64

tylerhawkes:
The hand-written `PastEpochDeletionPolicy::KeepAll` serializer from #1990
writes `usize::MAX as u64`, which is `u32::MAX` on `wasm32` and `u64::MAX`
on 64-bit hosts.

So a `KeepAll` policy persisted on a 64-bit host fails to deserialize on
`wasm32` — the deserializer narrows to `usize` before checking the
sentinel, and `u64::MAX` doesn't fit in a 32-bit `usize`.

Fix: emit `u64::MAX` unconditionally and check the sentinel before
narrowing. Legacy `wasm32`-written `KeepAll` loads as `MaxEpochs(u32::MAX)`,
which is functionally "keep all" given the practical ceiling on
past-epoch counts, so existing wasm32 data is not broken.

Also adds a `usize_as_u64` adapter to the remaining persisted `usize`
fields (`padding_size`, `number_of_resumption_psks`, `max_epochs`,
`max_number_of_secrets`, `cursor`, `Pprf::width`). A wire no-op against
today's serde `usize` impl, but it pins the convention so a future format
change can't reintroduce platform-native encoding by accident.

Split from #2021.

### DIFF
--- a/openmls/src/group/mls_group/config.rs
+++ b/openmls/src/group/mls_group/config.rs
@@ -76,11 +76,16 @@ impl Serialize for PastEpochDeletionPolicy {
     where
         S: serde::Serializer,
     {
-        let usize = match self {
-            Self::MaxEpochs(epochs) => *epochs,
-            Self::KeepAll => usize::MAX,
+        // Wire encoding is always `u64` (never platform-dependent `usize`) so
+        // bytes written on a 64-bit host are still readable on `wasm32` and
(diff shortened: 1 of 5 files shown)


franziskuskiefer (approved):
Thanks! We should get this in as well.

franziskuskiefer:
@tylerhawkes can you resolve the conflict? Then we can get this merged.

PR #2026, opened 2026-05-15
Lock persisted enum variant indices via custom serde impls

tylerhawkes:
After #2023 the five enum variant indices are pinned by source-file order
alone, which is fragile: serde's derive macro assigns indices via
`iter().enumerate()`, so a future PR that inserts a variant mid-enum
silently re-breaks bincode/postcard the same way 0.7.x → 0.8.x did.

This replaces the derived `Serialize`/`Deserialize` on `CredentialType`,
`ExtensionType`, `Extension`, `ProposalType`, and `Proposal` with
hand-written impls that pin each variant's `variant_index` as a numeric
literal at the call site.

Wire bytes are byte-identical to the previous positional encoding — the
same `serialize_*_variant` calls, just with the indices spelled out — and
the compiler's exhaustiveness check on the new `Serialize` match now
forces anyone adding a variant to make a deliberate choice about its
index.

Stacked on top of #2023; the diff will include those commits until #2023
lands, head commit is the only new content. Drop this one if the size
isn't worth the safety — #2023 stands without it.

Split from #2021.

### DIFF
--- a/openmls/src/credentials/mod.rs
+++ b/openmls/src/credentials/mod.rs
@@ -77,6 +77,9 @@ pub mod errors;
 /// | 0xDADA           | GREASE                   | Y | RFC XXXX |
 /// | 0xEAEA           | GREASE                   | Y | RFC XXXX |
 /// | 0xF000  - 0xFFFF | Reserved for Private Use | - | RFC XXXX |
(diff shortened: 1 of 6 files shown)


wysiwys:
Closing this PR as addressed by #2034

PR #2027, opened 2026-05-15
Unit tests for serde variant indices on persisted enums

tylerhawkes:
Adds unit tests that pin the `(variant_index, variant_name)` pair for
every variant of `CredentialType`, `ExtensionType`, `Extension`,
`ProposalType`, and `Proposal`. The index is the bincode/postcard wire
encoding, the name is the JSON/CBOR encoding, and either drifting
silently corrupts persisted state.

The mechanism is a minimal `serde::Serializer` (`utils::variant_index_probe`)
whose `Ok` type is `(u32, &'static str)` and whose only working methods are
`serialize_*_variant`. Everything else errors, so misuse fails loudly,
and the tests don't have to construct full payloads for tagged-union
variants.

The variants that actually drifted (`_AppAck`, `SelfRemove`, `Custom`,
the `Unknown` arms) are each pinned explicitly.

Stacked on top of #2023 — works against either the derived impls from
that PR or the hand-rolled impls from #2026.

Split from #2021.

### DIFF
--- a/openmls/src/credentials/mod.rs
+++ b/openmls/src/credentials/mod.rs
@@ -77,6 +77,9 @@ pub mod errors;
 /// | 0xDADA           | GREASE                   | Y | RFC XXXX |
 /// | 0xEAEA           | GREASE                   | Y | RFC XXXX |
 /// | 0xF000  - 0xFFFF | Reserved for Private Use | - | RFC XXXX |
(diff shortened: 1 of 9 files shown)


wysiwys:
Closing this PR in favor of the tests added in #2034

PR #2028, opened 2026-05-15
feat: targeted messages

raphaelrobert:
A first implementation of the [targeted messages draft](https://datatracker.ietf.org/doc/draft-ietf-mls-targeted-messages/).

The implementation is gated with a feature flag for now (the draft is still evolving), but the overlap with RFC9420 in the code is almost non-existent. It comes with tests and KAT vectors.


openmls/src/framing/message_in.rs
```diff
@@ -116,6 +103,16 @@ impl MlsMessageIn {
         self.try_into()
     }
 
+    /// Extract a [`TargetedMessageIn`] from this message, if it contains one.
+    #[cfg(feature = "targeted-messages-draft")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "targeted-messages-draft")))]
```
kkohbrok, line 107: To keep the API consistent, we should either make this available for tests only or remove the `cfg` constraint on all other `into_x` functions. I have a slight preference for the latter.

kkohbrok (approved):
Generally looks good, but I'm a bit wary of just throwing away underlying errors using `map(|_| ...)`. Could you log an error in those places? That would make debugging easier.

EDIT: Another thing: Should groups allow decryption of targeted messages from past epochs within the configured window? Maybe for a follow-up.

openmls/src/group/mls_group/targeted_messages.rs
```diff
@@ -0,0 +1,123 @@
...
+        &mut self,
+        provider: &Provider,
+        signer: &impl Signer,
+        recipient_leaf_index: LeafNodeIndex,
+        application_data: &[u8],
+        padding_length: usize,
+    ) -> Result<MlsMessageOut, CreateTargetedMessageError> {
+        if !self.is_active() {
+            return Err(CreateTargetedMessageError::GroupNotActive);
+        }
```
kkohbrok, line 31: Probably more of a spec-level question, but should a client be allowed to send targeted messages if there is a proposal pending?

openmls/src/ciphersuite/hpke.rs
```diff
@@ -267,3 +272,65 @@ pub fn safe_decrypt_with_label(
 
     decrypt_with_label_internal(private_key, context, ciphertext, ciphersuite, crypto)
 }
+
+#[cfg(feature = "targeted-messages-draft")]
+#[allow(clippy::too_many_arguments)]
```
franziskuskiefer: Should these use some `EncryptParams` instead of disabling the linter?

openmls/src/targeted_messages/mod.rs
```diff
@@ -0,0 +1,746 @@
...
+    pub(crate) fn application_data(&self) -> &[u8] {
+        self.application_data.as_slice()
+    }
+
+    /// Serialize to bytes: VLBytes application_data followed by raw zero
+    /// padding (no length prefix on the padding).
+    fn serialize_detached(&self) -> Result<Vec<u8>, tls_codec::Error> {
+        use std::io::Write;
+        let app_data_len = self.application_data.tls_serialized_len();
+        let mut buffer = Vec::with_capacity(app_data_len + self.padding_length);
```
franziskuskiefer: We should make sure this is not too large, i.e. `<= isize::MAX`.

openmls/src/framing/message_in.rs
raphaelrobert: Fixed in the line above.

openmls/src/group/mls_group/targeted_messages.rs
raphaelrobert: Spec-level indeed, and the spec does not prohibit it.

traits/src/crypto.rs
```diff
@@ -157,6 +168,39 @@ pub trait OpenMlsCrypto: Send + Sync {
...
+    #[allow(clippy::too_many_arguments)]
+    fn hpke_seal_psk_resolved_aad<F, E>(
+        &self,
+        config: HpkeConfig,
+        pk_r: &[u8],
+        info: &[u8],
+        ptxt: &[u8],
+        psk: &[u8],
+        psk_id: &[u8],
+        aad_builder: F,
```
kkohbrok, line 201: Might be worth explicitly saying what this does. I assume F is meant to be a function that takes the KEM ciphertext bytes as input and is expected to output the AAD?

openmls/src/ciphersuite/hpke.rs
```diff
@@ -267,3 +272,68 @@ pub fn safe_decrypt_with_label(
...
+
+/// Parameters shared by [`encrypt_with_label_psk_resolved_aad`] and
+/// [`decrypt_with_label_psk_aad`]: the label and context that make up the HPKE
+/// info, and the PSK material for the HPKE PSK mode.
+#[cfg(feature = "targeted-messages-draft")]
+pub(crate) struct PskEncryptParams<'a> {
+    pub label: &'a str,
+    pub context: &'a [u8],
+    pub psk: &'a [u8],
+    pub psk_id: &'a [u8],
```
kkohbrok, line 284: You could also add the ciphersuite here and theoretically even the crypto provider.

openmls/src/group/mls_group/targeted_messages.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// `padding_length` is the number of zero bytes appended to the plaintext
+    /// before encryption to obscure the application data length. Pass `0` for
+    /// no padding.
+    pub fn create_targeted_message<Provider: OpenMlsProvider>(
+        &mut self,
+        provider: &Provider,
+        signer: &impl Signer,
+        recipient_leaf_index: LeafNodeIndex,
+        application_data: &[u8],
+        padding_length: usize,
```
kkohbrok: If we use the AAD from the group config, we should probably also use the padding size that's configured there. Alternatively, this could also be a builder that defaults to the values configured in the group.

kkohbrok (commented):
Looks good! Just a few suggestions.

kkohbrok (approved):
Sorry, that should have been a ✅ .

openmls/src/ciphersuite/hpke.rs
raphaelrobert: I added `ciphersuite`, but since the struct is used in two functions with slightly different sets of parameters, we can't add all of them.

traits/src/crypto.rs
raphaelrobert: Addressed in the comment above

openmls/src/framing/message_in.rs
kkohbrok: Thanks!

PR #2029, opened 2026-05-15
chore(ci): make the CI leaner

raphaelrobert:
An attempt to make the CI faster with the following changes:

 - Test matrix flattened: full coverage only on Linux x86_64, single debug+default run on other platforms.
  - Dropped 32-bit targets (i686 Linux/Windows) and the iOS simulator.
  - macOS tests moved to Apple silicon (also matches the runner btw).
  - Cross-build matrix: dropped debug pass, release only.
  - Benches and fuzz: Linux-only.
  - Cargo-hack powerset: tighter grouping (js+js-test), combos dropped 129 → 17.
  - paths-ignore on compile-heavy workflows so doc-only PRs skip them.
  - README's platform list resynced to the new matrix.

We need to adjust the required tests before we can merge this.

### DIFF
--- a/.github/workflows/benches.yml
+++ b/.github/workflows/benches.yml
@@ -8,7 +8,17 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
+      - '**/*.md'
+      - 'book/**'
+      - 'LICENSE*'
+      - '.gitignore'
   pull_request:
+    paths-ignore:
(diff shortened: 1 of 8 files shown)


kkohbrok (approved):

franziskuskiefer:
lgtm, thanks! You need to update the required jobs to make GH happy.

PR #2030, opened 2026-05-18
Virtual clients: Compute reuse guard instead of generating randomly

kkohbrok:
This PR implements the reuse guard computation/derivation aspect of the virtual clients draft.

When sending a PrivateMessage, clients compute the reuse guard based on the emulation group state that was last used to update the group. This binding between higher-level group (ID) and emulation group epoch is stored in the storage provider.

The reuse guard is computed using a small-space PRP, which is pulled in as a dependency.
On the spec-level, computing the reuse guard in this way avoids reuse guard collisions between emulator clients.

Emulator clients that receive a PrivateMessage from a sibling emulator client in a higher-level group can reverse the reuse guard computation to determine which emulator client has sent the message. This change, along with the failure modes associated with this reversing operation required some threading of information through various function calls. I apologize for the churn.


openmls/src/framing/private_message.rs
```diff
@@ -58,21 +73,6 @@ pub(crate) struct MlsMessageHeader {
 }
 
 impl PrivateMessage {
-    /// Returns the [`GroupId`] of the group this message was sent in
```
raphaelrobert, line 61: This looks like a regression

openmls/src/group/mls_group/mod.rs
```diff
@@ -678,13 +685,52 @@ impl MlsGroup {
...
+                            );
+                            MessageEncryptionError::VirtualClientsError(
+                                crate::components::vc_derivation_info::VirtualClientsError::MissingEmulationEpochState,
+                            )
+                        })?,
+                ),
+                None => None,
+            }
+        };
+        #[cfg(feature = "virtual-clients-draft")]
```
raphaelrobert, line 746: You could guard a block instead of individual lines, but I don't have a strong opinion.

openmls/src/group/mls_group/processing.rs
```diff
@@ -181,13 +181,65 @@ impl MlsGroup {
...
+                ),
+                None => None,
+            }
+        } else {
+            None
+        };
+        #[cfg(feature = "virtual-clients-draft")]
+        let emulator_ctx: Option<crate::framing::EmulatorReuseGuardCtx<'_>> = emulation_state
+            .as_ref()
+            .map(|state| state.reuse_guard_inputs());
```
raphaelrobert, line 214: This looks very similar to the code in `mls_group/mod.rs`, maybe this could be extracted.

openmls/src/group/mls_group/mod.rs
```diff
@@ -514,6 +514,13 @@ impl MlsGroup {
         #[cfg(feature = "extensions-draft-08")]
         storage.delete_application_export_tree::<_, ApplicationExportTree>(self.group_id())?;
 
+        // Drop this group's emulation-epoch bindings. `EmulationEpochState`
+        // and `VcPprf` are keyed on the emulation epoch and may still be
+        // referenced by other higher-level groups, so they're not
+        // deleted here.
+        #[cfg(feature = "virtual-clients-draft")]
+        storage.delete_vc_emulation_bindings(self.group_id())?;
+
```
raphaelrobert, line 533: Don't we need to also call `delete_vc_emulation_epoch_state()` here?

openmls/src/group/mls_group/processing.rs
```diff
@@ -279,6 +331,47 @@ impl MlsGroup {
...
+                bindings.insert(staged_commit.epoch(), epoch_id, max_entries);
+                provider
+                    .storage()
+                    .write_vc_emulation_bindings(self.group_id(), &bindings)
+                    .map_err(|e| {
+                        log::error!("vc: persist emulation bindings at merge failed: {e:?}");
+                        MergeCommitError::StorageError(e)
+                    })?;
+            }
+        }
```
raphaelrobert, line 373: It might be better to do all this after the merge, in case the merge fails

openmls/src/framing/validation.rs
```diff
@@ -151,7 +176,19 @@ impl DecryptedMessage {
...
+            verifiable_content,
+            #[cfg(feature = "virtual-clients-draft")]
+            emulator_sender_leaf_index,
+        })
+    }
+
+    /// Recovered sender emulation-group leaf index, if the message came
+    /// from a sibling emulator client.
+    #[cfg(feature = "virtual-clients-draft")]
+    #[allow(dead_code)]
```
raphaelrobert: Why the dead code allow?

openmls/src/components/vc_derivation_info.rs
```diff
@@ -262,32 +382,65 @@ impl std::fmt::Debug for EpochEncryptionKey {
...
+    /// registration time. Hashed into `EpochInfoTbe` and used as the
+    /// sender's `leaf_index_e` in the reuse-guard derivation.
     pub(crate) leaf_index: LeafNodeIndex,
     pub(crate) epoch_encryption_key: EpochEncryptionKey,
+    pub(crate) reuse_guard_secret: ReuseGuardSecret,
+    /// Number of leaves `N_e` in the emulation group at registration time.
+    pub(crate) emulation_group_size: TreeSize,
+    /// Ciphersuite of the emulation group at registration time. Used by
+    /// the reuse-guard derivation.
+    pub(crate) emulation_ciphersuite: Ciphersuite,
```
raphaelrobert, line 405: There are field changes here that don't use `#[serde(default)]`, this can break

openmls/src/group/mls_group/application.rs
```diff
@@ -90,6 +88,38 @@ impl MlsGroup {
...
+        let EncryptionOutput {
+            generation,
+            private_message,
+        } = self.encrypt(authenticated_content, provider)?;
+
+        let output = MlsMessageOut::from_private_message(private_message, self.version());
+        self.reset_aad();
+        Ok((generation, output))
+    }
+
```
raphaelrobert, line 101: It would be nice if the functions weren't duplicated, this is hard to maintain. Maybe there's a smaller scope for the fork?

openmls/src/framing/validation.rs
```diff
@@ -354,6 +416,13 @@ impl ProcessedMessage {
         &self.authenticated_data[self.safe_aad_prefix_len..]
     }
 
+    /// Returns the sender's leaf index in the emulation group when this
+    /// message is an application message from a sibling emulator client.
```
raphaelrobert: This is not only for application messages, but more generally for private messages. I think there are more occurances of that in the comments.

raphaelrobert (changes requested):
I did a first pass. I wonder if FF1 shouldn't go to the crypto provider instead. It can still be guarded by the feature flag.

I think it would also be good to add more tests, e.g. for collisions: two distinct leaf indices, same secret and nonce, should return distinct guards.

openmls/src/group/mls_group/mod.rs
kkohbrok: No, the `vc_emulation_epoch_state` is emulation group state that's used across higher-level groups. Managing that state is up to the application. I've added some documentation in that regard.

openmls/src/components/vc_derivation_info.rs
kkohbrok: I'd like to make a separate backwards-compatibility pass in a follow-up. This PR is already large enough.

raphaelrobert (approved):
Thanks for making the changes, let's get this in

PR #2033, opened 2026-05-20
allow external commit wuth pure ciphertext

### LINKED ISSUE #1818: Allow external commits with pure ciphertext
So we can't send self removes if the policy is different? That wouldn't be good, we should allow sending self removes and psks and don't apply the policy. This is similar to other messages, like application messages, which have a mandatory wire format.

_Originally posted by @franziskuskiefer in https://github.com/openmls/openmls/pull/1801#discussion_r2245117007_

seun-ja:
Fixes #1818 

Allows support for external commit with pure ciphertext. It also includes test


raphaelrobert:
I don't understand what this PR does. External commits have to always be public messages.

PR #2034, opened 2026-05-21
`serde` serialization helpers for non-self-describing storage representations

wysiwys:
This pull request proposes adding serialization helper macros for use with `serde` non-self-describing formats, which pin the storage tag index used for each enum variant. It uses these macros to pin storage tags for  `Proposal`/`ProposalType`, `Extension`/`ExtensionType`, and `CredentialType`, and also adds an `openmls_compat_tests` crate which includes tests for storage tag stability across `openmls` versions. This PR was initially opened as a starting point for discussion.

------
As a basis for discussion, this draft pull request proposes a way to address some issues with non-self-describing `serde` serialization that were identified in #2021. Since non-self-describing `serde` formats serialize enum variants using a tag corresponding to the variant's position in the list of variants, when new versions of `openmls` added new variants to some serializable enums (e.g. `CredentialType::Grease(_)`) or removed some from others (e.g. `Proposal::AppAck`), the position of some existing variants was affected, leading to deserialization failures.

To prevent these kinds of deserialization issues in the future, this draft PR proposes pinning a fixed storage tag for each enum variant, so that it is not affected by future removals or additions of variants. The proposed solution is implemented using serialization helper macros that can be used to implement `serde::Serialize` and `serde::Deserialize` traits on these enums. These macros allow specifying a `storage_tag` attribute on each variant, to pin it to a specific index:

```rust
#[derive(
    openmls_serialization_helpers::Serialize,
    openmls_serialization_helpers::Deserialize,
)]
#[allow(missing_docs)]
pub enum Proposal {
    #[storage_tag = 0]
    Add(Box<AddProposal>),
    #[storage_tag = 1]
    Update(Box<UpdateProposal>),
    #[storage_tag = 2]
    Remove(Box<RemoveProposal>),
    #[storage_tag = 3]
    PreSharedKey(Box<PreSharedKeyProposal>),
    #[storage_tag = 4]
    ReInit(Box<ReInitProposal>),
    #[storage_tag = 5]
    ExternalInit(Box<ExternalInitProposal>),
    #[storage_tag = 6]
    GroupContextExtensions(Box<GroupContextExtensionProposal>),
    // AppAck = 7,
    #[storage_tag = 8]
    SelfRemove,
    #[storage_tag = 9]
    Custom(Box<CustomProposal>),
    // # Extensions
    #[cfg(feature = "extensions-draft-08")]
    #[storage_tag = 10]
    AppDataUpdate(Box<AppDataUpdateProposal>),
    #[cfg(feature = "extensions-draft-08")]
    #[storage_tag = 11]
    AppEphemeral(Box<AppEphemeralProposal>),
}
```
Using this approach, new variants can be added to these enums in a way that ensures that their `serde` non-self-describing serialization does not conflict with existing serializations, independent of feature flag annotations on the enum variants. 

In the `openmls` library, in this PR, these `openmls_serialization_helpers` macros are used to implement the `serde::Serialize` and `serde::Deserialize` traits for each of `Proposal`/`ProposalType`, `Extension`/`ExtensionType`, and `CredentialType`. The `serde` trait implementations  using the macros corresponds to the individual implementations of these `serde` traits for these enums proposed in #2021.   

This PR also reorders some enum variants so that enum variants are in ascending order by their `storage_tag` attribute for readability, but this list position no longer affects the index used in non-self-describing `serde` serialization.


wysiwys:
@tylerhawkes

tylerhawkes:
This is more maintainable and easier to reason about than what I had in #2023 and #2026. Thanks!

openmls/src/credentials/mod.rs
```diff
@@ -77,17 +77,31 @@ pub mod errors;
...
     Basic = 1,
+    #[storage_tag = 1]
     /// An X.509 [`Certificate`]
     X509 = 2,
-    /// A GREASE credential type for ensuring extensibility.
-    Grease(u16),
+    #[storage_tag = 2]
     /// Another type of credential that is not in the MLS protocol spec.
     Other(u16),
+    #[storage_tag = 3]
```
franziskuskiefer: We need to make sure to mention these breaking changes and describe why it's needed.

serialization_helpers/src/serialize.rs
```diff
@@ -0,0 +1,82 @@
...
+                }
+                _ => unimplemented!("fields not supported"),
+            }
+        })
+        .collect::<Vec<_>>();
+
+    quote! {
+        impl serde::Serialize for #name {
+            fn serialize<S: serde::Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
+                use serde::ser::SerializeTupleVariant;
```
franziskuskiefer, line 79: This appears to be unused.

openmls/tests/migration_example.rs
```diff
@@ -0,0 +1,65 @@
+#![allow(non_camel_case_types)]
+//! Migration example (0.7.1 -> main)
```
franziskuskiefer: Let's move these tests into openmls to avoid cyclic dependencies and have this closer to home.

serialization_helpers/src/serialize.rs
```diff
@@ -0,0 +1,82 @@
...
+                            #storage_tag,
+                            stringify!(#variant_name),
+                            v
+                        ),
+
+                    }
+                }
+                Fields::Unnamed(FieldsUnnamed { unnamed, .. }) if unnamed.len() == 2 => {
+                    quote! {
+                        #name::#variant_name(v1, v2) => s.serialize_newtype_variant(
```
franziskuskiefer: Is using the newtype for tuples the expected way/what serde is doing? Just wondering. And maybe this should get an explanatory comment.

serialization_helpers/tests/serialization_deserialization.rs
```diff
@@ -0,0 +1,133 @@
...
+    Data([u8; 3]),
+    #[storage_tag = 500]
+    Data2(u64, [u8; 3]),
+    #[storage_tag = 4]
+    Unit3,
+}
+
+struct TestCase {
+    data: TestEnum,
+    expected_tag: u16,
```
franziskuskiefer: I thought this should be a `u32`?

openmls/tests/migration_example.rs
```diff
@@ -0,0 +1,65 @@
+#![allow(non_camel_case_types)]
```
franziskuskiefer: Let's add tests for all changed types.

serialization_helpers/src/deserialize.rs
```diff
@@ -0,0 +1,138 @@
...
+                f.write_str(stringify!(#name))
+            }
+
+            fn visit_enum<A: serde::de::EnumAccess<'de>>(self, data: A) -> Result<#name, A::Error> {
+                use serde::de::VariantAccess;
+                let (storage_tag, access) = data.variant::<u32>()?;
+
+                match storage_tag {
+                    #(#match_arms_non_self_describing)*
+                    _ => Err(serde::de::Error::custom(format!("unexpected storage tag {}", storage_tag))),
```
franziskuskiefer, line 169: Let's test error cases well.

franziskuskiefer (commented):
I think this is the right approach. Let's clean it up and make it mergeable unless someone has a better idea.

openmls/tests/migration_example.rs
wysiwys: I added tests for these types using a custom `serde::Serialize` implementation that retrieves the storage tag for each variant, based on the approach in #2027

serialization_helpers/src/serialize.rs
wysiwys: Indeed, this is not the standard way to serialize and deserialize tuples in `serde`. I updated it to use `Serializer::serialize_tuple_variant()` instead, and to use a dedicated TupleVisitor for deserialization

openmls/tests/migration_example.rs
wysiwys: These storage tag tests are now moved into a new crate in the workspace (`openmls_compat_tests`)

compat_tests/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "openmls_compat_tests"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+serde = "1.0"
+thiserror = "2.0"
```
franziskuskiefer: Should these be workspace dependencies?

compat_tests/tests/test_storage_tag_stability.rs
```diff
@@ -0,0 +1,135 @@
...
+
+const TEST_CASES_EXTENSION_TYPE: &[&str] = &[
+    r#""ApplicationId""#,
+    r#""RatchetTree""#,
+    r#""RequiredCapabilities""#,
+    r#""ExternalPub""#,
+    r#""ExternalSenders""#,
+    r#""LastResort""#,
+    r#"{"Unknown":20}"#,
+];
```
franziskuskiefer: Let's throw these blobs into files to keep it a little nicer here.

serialization_helpers/src/serialize.rs
```diff
@@ -0,0 +1,87 @@
...
+                panic!("storage tag not provided for variant {name}::{variant_name}")
+            });
+            if !storage_tags.insert(storage_tag) {
+                panic!("Duplicate storage tags");
+            }
+
+            // construct a match arm based on the type
+            match &variant.fields {
+                Fields::Unit => quote! {
+                    #name::#variant_name => s.serialize_unit_variant(
```
franziskuskiefer, line 38: You need to reprint the `cfg` attributes everywhere. Both here and in deserialize.

compat_tests/tests/test_storage_tag_stability.rs
```diff
@@ -0,0 +1,135 @@
+//! These tests check storage tag stability for `serde` serializations of `Extension`,
+//!   `ExtensionType`, `Proposal`, `ProposalType`, and `CredentialType`.
+
+use openmls_compat_tests::storage_tag_check::*;
+
+const TEST_CASES_CREDENTIAL_TYPE: &[&str] = &[r#""Basic""#, r#""X509""#, r#"{"Other":20}"#];
+
+const TEST_CASES_EXTENSION_TYPE: &[&str] = &[
```
franziskuskiefer: Shouldn't we test grease values too?

openmls/src/messages/proposals.rs
```diff
@@ -231,25 +255,42 @@ impl From<ProposalType> for u16 {
...
-#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+#[derive(
+    Debug,
+    PartialEq,
+    Clone,
+    openmls_serialization_helpers::Serialize,
+    openmls_serialization_helpers::Deserialize,
+)]
 #[allow(missing_docs)]
-#[repr(u16)]
```
franziskuskiefer, line 236: Just to avoid unintended breakage, maybe we should keep this? It shouldn't make a difference. But it also doesn't hurt.

serialization_helpers/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+documentation = "https://docs.rs/openmls_serialization_helpers"
+repository = "https://github.com/openmls/openmls/tree/main/serialization_helpers"
+readme = "README.md"
+
+
+[lib]
+proc-macro = true
+
+[dependencies]
+serde = { version = "^1.0", features = ["derive"] }
```
franziskuskiefer: I think this is unused?

serialization_helpers/src/attrs.rs
```diff
@@ -0,0 +1,25 @@
...
+            return None;
+        }
+
+        let Expr::Lit(ExprLit {
+            lit: Lit::Int(int), ..
+        }) = &nv.value
+        else {
+            return None;
+        };
+        int.base10_parse().ok()
```
franziskuskiefer: This may return a wrong error when the value is not an integer, i.e. it'll say "storage tag not provided"

serialization_helpers/src/serialize.rs
```diff
@@ -0,0 +1,87 @@
+use proc_macro2::TokenStream;
+use quote::quote;
+use syn::{parse2, Data, DeriveInput, Fields, FieldsUnnamed};
+
+use crate::attrs::extract_storage_tag;
+
+pub(crate) fn serialize(input: TokenStream) -> TokenStream {
+    // parse the input
+    let input: DeriveInput = parse2(input).unwrap();
```
franziskuskiefer: Should this be an expect to give some hint what's going wrong?

serialization_helpers/src/deserialize.rs
```diff
@@ -0,0 +1,166 @@
...
+                }
+            }
+
+        }
+
+        /// A visitor for deserializing the tuple (A,B) contents of an enum variant
+        struct #tuple_visitor<A, B>(std::marker::PhantomData<(A,B)>);
```
franziskuskiefer: We could skip this visitor if there's no such element present. It should be dropped again by the compiler. But it's also unnecessary and may generate warnings.

franziskuskiefer (changes requested):
Thanks, a few nits, then I think this is ready to get merged.

compat_tests/Cargo.toml
wysiwys: I made these dependencies workspace dependencies - there are some other dependencies, like `log`, that are used in different crates in the workspace, and could also be made into workspace dependencies. Is this something that is better to update now (having both types), or possibly as a follow-up to this PR?

compat_tests/tests/test_storage_tag_stability.rs
wysiwys: I moved the test case data into a separate file, as JSON data (for readability)

serialization_helpers/src/serialize.rs
wysiwys: It appears that the `cfg` attributes for enum aren't visible to `derive` macros, meaning variants that aren't enabled are pre-filtered out. I wrote a test that shows that these new proc macros will produce valid input for an enum with a feature-flagged variant, whose feature is not enabled. Do you think this is sufficient to test this case? I think it could be good to find a source that can be referenced in the usage documentation for these macros, too, to indicate how feature flags on enum variants are handled

compat_tests/tests/test_storage_tag_stability.rs
wysiwys: I think one option could be to test that deserializing `Grease` values, and the other enum variants that have been added since `openmls=0.7.1`, results in an error using `openmls=0.7.1`, and that they don't deserialize incorrectly to other variants

serialization_helpers/src/deserialize.rs
wysiwys: The tuple visitor is now skipped if there is no variant that requires it for deserialization

serialization_helpers/src/serialize.rs
franziskuskiefer: Indeed. `cfg` macros are resolved before derive (not before attributes). https://doc.rust-lang.org/reference/attributes.html

compat_tests/Cargo.toml
franziskuskiefer: yeah, no need to push more things in here.

compat_tests/tests/test_storage_tag_stability.rs
franziskuskiefer: Grease values should always be handled fine. I think we just didn't have a way to check for them (which is for testing really). But having a test just to ensure the behaviour is what we expect would be nice.

serialization_helpers/src/deserialize.rs
```diff
@@ -0,0 +1,178 @@
...
+                let (variant_str, access): (String, _) = data.variant()?;
+                match variant_str.as_str() {
+                    #(#match_arms_self_describing)*
+                    _ => Err(serde::de::Error::custom(format!("unexpected variant name \"{}\"", variant_str))),
+                }
+            }
+
+        }
+
+        /// A visitor for non-self-describing enum serialization
```
franziskuskiefer:
```suggestion
        /// A visitor for non-self-describing enum deserialization
```

franziskuskiefer (commented):
Just a few more nits and there are a few open comments.

I think the most important one is on documentation. This is a breaking change for storage when upgrading from 0.8.1. Is there a way to help consumers with that? Is there something to test on that?

serialization_helpers/src/serialize.rs
wysiwys: I added a note about feature-flagged enum variants (without the reference link for now)

openmls/src/credentials/mod.rs
franziskuskiefer: We unbreak it in #2047

compat_tests/tests/test_storage_tag_stability.rs
franziskuskiefer: Addressed in #2047

franziskuskiefer (dismissed):
Let's first get #2047 merged in here. Then this is good to go.

franziskuskiefer (approved):
Let's get this in

PR #2036, opened 2026-05-26
add methods to access nodes in RatchetTree and RatchetTreeIn

keks:
Applications might want that data for doing validation (e.g. check that LeafNodes are valid according to the AS), so we need to provide access.


kkohbrok (dismissed):

keks:
Sorry 'bout that

kkohbrok (approved):

PR #2037, opened 2026-05-27
feat: Implement Safe AAD capabilities from extensions doc

kkohbrok:
This PR implements the Safe AAD functionality as defined in the extensions draft: https://datatracker.ietf.org/doc/draft-ietf-mls-extensions/09/

This PR should be reviewable commit-by-commit.

### DIFF
--- a/openmls/src/framing/validation.rs
+++ b/openmls/src/framing/validation.rs
@@ -36,7 +36,9 @@ use crate::{
 };
 
 #[cfg(feature = "extensions-draft-08")]
-use crate::{component::ComponentId, messages::proposals_in::ProposalOrRefIn};
+use crate::{
+    component::ComponentId, framing::safe_aad::SafeAad, messages::proposals_in::ProposalOrRefIn,
+};
 
 use super::{
     mls_auth_content::AuthenticatedContent,
@@ -282,6 +284,14 @@ pub struct ProcessedMessage {
     authenticated_data: Vec<u8>,
     content: ProcessedMessageContent,
     credential: Credential,
+    /// Parsed Safe AAD prefix, populated only when the message's GroupContext
+    /// required Safe AAD framing. `None` otherwise.
+    #[cfg(feature = "extensions-draft-08")]
+    safe_aad: Option<SafeAad>,
+    /// Length in bytes of the Safe AAD prefix at the start of
+    /// `authenticated_data`. Zero when [`Self::safe_aad`] is `None`.
+    #[cfg(feature = "extensions-draft-08")]
+    safe_aad_prefix_len: usize,
 }
 
 impl ProcessedMessage {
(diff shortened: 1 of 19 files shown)


openmls/src/framing/validation.rs
```diff
@@ -301,9 +311,49 @@ impl ProcessedMessage {
...
+    /// stays `None` and `authenticated_data` is the caller-supplied bytes
+    /// untouched.
+    #[cfg(feature = "extensions-draft-08")]
+    pub(crate) fn try_attach_safe_aad(&mut self) -> Result<(), crate::framing::SafeAadError> {
+        let (safe_aad, prefix_len) =
+            crate::framing::safe_aad::parse_authenticated_data_prefix(&self.authenticated_data)?;
+        self.safe_aad = Some(safe_aad);
+        self.safe_aad_prefix_len = prefix_len;
+        Ok(())
+    }
```
keks, line 333: What is the reason to not having a method that sets the aad in safe aad style?

openmls/src/group/mls_group/mod.rs
```diff
@@ -292,6 +297,28 @@ impl MlsGroup {
...
+    ///
+    /// Ephemeral, like [`Self::set_aad`]: cleared whenever an outgoing message
+    /// is produced.
+    ///
+    /// [`ComponentId`]: crate::component::ComponentId
+    #[cfg(feature = "extensions-draft-08")]
+    pub fn set_safe_aad(&mut self, items: Vec<SafeAadItem>) -> Result<(), SafeAadError> {
+        self.safe_aad = SafeAad::from_items(items)?;
+        Ok(())
+    }
```
keks, line 313: Should this also set the old aad value?

keks (approved):
I think this generally looks good. I have one remark, but I don't think this is blocking anything.

I was a little surprised by the interaction between the old and new aad fields, and that setting the safe aad does not overwrite the plain aad, but that there is a way to "import" the safe aad from the plain aad.

I guess this is related to the general question about whether we should be setting the aad for the next message on the group instead of passing it into the message creation. My impression is that this shouldn't really be state on the MlsGroup, but I don't think this PR is the place to change how that works. Maybe we can discuss it next week.

openmls/src/framing/validation.rs
kkohbrok: I'm not sure what you mean. Something like `try_attach_regular_aad` that appends regular aad behind a potential safe aad?

openmls/src/group/mls_group/mod.rs
kkohbrok: Don't we have the `set_aad` function for that?

kkohbrok (commented):
Yes, appending non-safe aad behind safe aad isn't great, but it's how I read the spec. If you want some proprietary aad, you should probably just define your own component for that. I've filed an issue in the spec repo: https://github.com/mlswg/mls-extensions/issues/95

openmls/src/framing/validation.rs
keks: Ah, okay I think I got this now.

I think an alternative design would be to just set aad and safe aad together, and just encoding and writing both into the existing aad. But I think either are fine.

openmls/src/group/mls_group/mod.rs
keks: As I said in the other comment, I would have expected a method that sets both at once. But I think this is fine as well.

keks (approved):
Thanks for the comments!

kkohbrok:
Thanks for the review again! Let's revisit AADs again when the draft is final and/or when we eventually switch to a builder-provided AAD API for OpenMLS.

PR #2043, opened 2026-06-02
feat: add application PSK type

boxdot:
Add a new application PSK type and the corresponding proposal/staged
commit handling, allowing applications to inject their own pre-shared
keys into the key schedule.

See <https://datatracker.ietf.org/doc/html/draft-ietf-mls-extensions-09#name-pre-shared-keys-psks>.

### DIFF
--- a/openmls/src/group/mls_group/proposal.rs
+++ b/openmls/src/group/mls_group/proposal.rs
@@ -148,6 +148,7 @@ impl MlsGroup {
         ProposalOrRefType::Proposal
     );
 
+    /// Creates proposals to add a non-resumption PSK to the key schedule.
     #[deprecated(
         note = "Renamed to `propose_pre_shared_key`; works for any non-resumption PSK, not just external"
     )]
@@ -160,6 +161,7 @@ impl MlsGroup {
         self.propose_pre_shared_key(provider, signer, value)
     }
 
+    /// Creates proposals to add a non-resumption PSK to the key schedule by value.
     #[deprecated(
         note = "Renamed to `propose_pre_shared_key_by_value`; works for any non-resumption PSK, not just external"
     )]
--- a/openmls/src/group/mls_group/proposal.rs
+++ b/openmls/src/group/mls_group/proposal.rs
@@ -135,18 +135,42 @@ impl MlsGroup {
     );
 
     impl_propose_fun!(
-        propose_external_psk,
+        propose_pre_shared_key,
         PreSharedKeyId,
         create_presharedkey_proposal,
         ProposalOrRefType::Reference
     );
 
     impl_propose_fun!(
-        propose_external_psk_by_value,
+        propose_pre_shared_key_by_value,
         PreSharedKeyId,
         create_presharedkey_proposal,
         ProposalOrRefType::Proposal
     );
+
+    #[deprecated(
+        note = "Renamed to `propose_pre_shared_key`; works for any non-resumption PSK, not just external"
+    )]
+    pub fn propose_external_psk<Provider: OpenMlsProvider>(
+        &mut self,
+        provider: &Provider,
+        signer: &impl Signer,
+        value: PreSharedKeyId,
+    ) -> Result<(MlsMessageOut, ProposalRef), ProposalError<Provider::StorageError>> {
+        self.propose_pre_shared_key(provider, signer, value)
+    }
+
+    #[deprecated(
+        note = "Renamed to `propose_pre_shared_key_by_value`; works for any non-resumption PSK, not just external"
+    )]
+    pub fn propose_external_psk_by_value<Provider: OpenMlsProvider>(
+        &mut self,
+        provider: &Provider,
+        signer: &impl Signer,
+        value: PreSharedKeyId,
+    ) -> Result<(MlsMessageOut, ProposalRef), ProposalError<Provider::StorageError>> {
+        self.propose_pre_shared_key_by_value(provider, signer, value)
+    }
 
     impl_propose_fun!(
         propose_custom_proposal_by_value,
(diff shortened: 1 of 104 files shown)


franziskuskiefer (changes requested):
Can you add some tests for this?

openmls/src/schedule/psk.rs
```diff
@@ -144,6 +146,51 @@ impl ResumptionPsk {
     }
 }
 
+/// Application PSK according to the [draft-ietf-mls-extensions-09]
+///
+/// [draft-ietf-mls-protocol-08]: https://datatracker.ietf.org/doc/html/draft-ietf-mls-extensions-09#name-pre-shared-keys-psks
```
raphaelrobert:
```suggestion
/// [draft-ietf-mls-protocol-09]: https://datatracker.ietf.org/doc/html/draft-ietf-mls-extensions-09#name-pre-shared-keys-psks
```

openmls/src/group/mls_group/staged_commit.rs
```diff
@@ -861,6 +861,7 @@ impl StagedCommit {
         let Some(application_export_tree) = staged_commit.application_export_tree.as_mut() else {
             return Err(StagedSafeExportSecretError::Unsupported);
         };
+        // TODO: Most likely we want to make Secret public
```
raphaelrobert: Is this related?

openmls/src/schedule/psk.rs
```diff
@@ -168,6 +215,10 @@ pub enum Psk {
     /// A resumption PSK derived from the MLS key schedule.
     #[tls_codec(discriminant = 2)]
     Resumption(ResumptionPsk),
+    #[cfg(feature = "extensions-draft-08")]
+    /// An application component PSK.
+    #[tls_codec(discriminant = 3)]
+    Application(ApplicationPsk),
```
raphaelrobert, line 221: You could also add a `::application(component_id, psk_id, nonce)` constructor to match the other variants.

openmls/src/group/mls_group/proposal.rs
```diff
@@ -217,6 +217,16 @@ impl MlsGroup {
...
                 },
+                #[cfg(feature = "extensions-draft-08")]
+                crate::schedule::Psk::Application(_) => match ref_or_value {
+                    ProposalOrRefType::Proposal => {
+                        self.propose_external_psk_by_value(provider, signer, psk_id)
+                    }
+                    ProposalOrRefType::Reference => {
+                        self.propose_external_psk(provider, signer, psk_id)
+                    }
+                },
```
raphaelrobert: The two match arms are identical here and could be collapsed. It's a bit strange to propose_external_psk for an application PSK, the naming is off. Maybe add a comment why this is ok.

raphaelrobert (commented):
This looks good in general, just a few nits

openmls/src/schedule/psk.rs
boxdot: Thanks.

openmls/src/schedule/psk.rs
boxdot: Done.

openmls/src/group/mls_group/staged_commit.rs
boxdot: No. Removed.

openmls/src/group/mls_group/proposal.rs
boxdot: I renamed the `propose_external_psk` function to `propose_pre_shared_key` and deprecated the former (because it is the public API).

raphaelrobert (commented):
Since there are API changes, could you also make an entry in CHANGELOG.md?

boxdot:
> Since there are API changes, could you also make an entry in CHANGELOG.md?

API changes are now in the changelog.

raphaelrobert (dismissed):

openmls/src/schedule/psk.rs
```diff
@@ -144,6 +146,51 @@ impl ResumptionPsk {
...
+    TlsSerialize,
+    TlsSize,
+    Hash,
+)]
+pub struct ApplicationPsk {
+    pub(crate) component_id: ComponentId,
+    pub(crate) psk_id: VLBytes,
+}
+
+#[cfg(feature = "extensions-draft-08")]
```
franziskuskiefer, line 173: Not something we need to change necessarily, but we are on -09 really at this point. We may want to start using a new feature flag.

openmls/src/schedule/psk.rs
```diff
@@ -168,6 +215,10 @@ pub enum Psk {
     /// A resumption PSK derived from the MLS key schedule.
     #[tls_codec(discriminant = 2)]
     Resumption(ResumptionPsk),
+    #[cfg(feature = "extensions-draft-08")]
+    /// An application component PSK.
+    #[tls_codec(discriminant = 3)]
+    Application(ApplicationPsk),
 }
 
 /// ```c
```
franziskuskiefer, line 224: Let's update the comment on top here.

franziskuskiefer (dismissed):
Thanks, lgtm. Just a nit on a comment and thoughts on the feature flag.

openmls/src/schedule/psk.rs
boxdot: Updated to

```
/// // draft-ietf-mls-protocol-19 and draft-ietf-mls-extensions-09
/// enum {
///   reserved(0),
///   external(1),
///   resumption(2),
///   // draft-ietf-mls-extensions-09
///   application(3),
///   (255)
/// } PSKType;
```

I guess this is what you mean.

openmls/src/schedule/psk.rs
raphaelrobert: I noticed that as well. I think it would make sense to drop the version altogether.

boxdot: This is a larger change. I will do it in a separate PR. Most likely we should also just use `extensions-draft` without a suffix to keep it stable between revisions.

raphaelrobert: That's what I meant, drop the draft version suffix. And I agree it should be another PR.

openmls/src/schedule/psk.rs
franziskuskiefer: Ah sorry for not being specific enough here. I meant the `draft-ietf-mls-protocol-19`. This is RFC 9420 now really. We used the draft versions while all this was still in flux but never updated to the final RFC. But mentioning both would make indeed sense.

openmls/src/schedule/psk.rs
franziskuskiefer: yeah that's not for this one. We used draft versions before to allow support for different versions/know where things come from. But that may not be needed here.

openmls/src/schedule/psk.rs
boxdot: Replaced draft-ietf-mls-protocol-19 by RFC 9420

franziskuskiefer (approved):
:shipit:

PR #2045, opened 2026-06-02
feat: allow to bypass KeyPackage/LeafNode validation

boxdot:
Add functions that allow unchecked conversion of `KeyPackageIn`, 
`LeafNodeIn`, `Extensions` into already verified types. Intended 
for callers that have verified the input through another path.

This feature is behind a feature flag `unchecked-conversions`.


keks:
I think this generally makes sense. Any reason not to also have a non-checking function for converting the Extensions without checking?

boxdot:
> I think this generally makes sense. Any reason not to also have a non-checking function for converting the Extensions without checking?

No particular reason. It can be introduced for any In/Out type, however I did not want to do it upfront. Only in case somebody needs it.

boxdot:
Added unchecked conversion for `Extensions` and a feature flag `unchecked-conversions`.

raphaelrobert (approved):
Thanks, let's get this in!

PR #2047, opened 2026-06-03
Add feature flag for `serde` serialization backwards compatibility with `openmls` v0.8.1

wysiwys:
This PR introduces a feature flag to enable storage serialization backwards compatibility with the `openmls` v0.8.1 release. This PR builds on the changes in #2034, which restore storage compatibility with `openmls` v0.7.1, but do not yet address compatibility with `openmls` v0.8.1.

The added `0-8-1-storage-format` feature flag, when enabled, ensures that the `serde` serializations for the affected types (`Extension`, `ExtensionType`, `CredentialType`, `ProposalType`, and `Proposal`) match the formats used in the released `openmls` v0.8.1.

This PR also adds storage compatibility tests for `openmls` v0.8.1, both with and without the `extensions-draft-08` feature,  to the `openmls_compat_tests` crate.

### DIFF
--- a/compat_tests/README.md
+++ b/compat_tests/README.md
@@ -1,0 +1,11 @@
+# Compatibility tests for `openmls`
+
+This crate includes tests for compatibility with previous `openmls` versions.
+- storage format compatibility tests for `openmls=0.7.1`, `openmls=0.8.1`, and `openmls=0.8.1` with the feature `extensions-draft-08` enabled
+
+## Usage
+```
+cargo test -F compat_0_7_1
+cargo test -F compat_0_8_1
+cargo test -F compat_0_8_1_extensions
+```
--- a/compat_tests/tests/test_storage_tag_stability.rs
+++ b/compat_tests/tests/test_storage_tag_stability.rs
@@ -1,5 +1,6 @@
 //! These tests check storage tag stability for `serde` serializations of `Extension`,
 //!   `ExtensionType`, `Proposal`, `ProposalType`, and `CredentialType`.
+#![allow(dead_code)]
 
 use openmls_compat_tests::storage_tag_check::*;
 
@@ -115,4 +116,8 @@ macro_rules! compat_tests {
 }
 
 // check the storage tag stability for openmls=0.7.1 => `main`
+#[cfg(feature = "compat_0_7_1")]
 compat_tests!(test_storage_tags_0_7_1, openmls_0_7_1, openmls);
+// check the storage tag stability for openmls=0.8.1 => `main`
+#[cfg(feature = "compat_0_8_1")]
+compat_tests!(test_storage_tags_0_8_1, openmls_0_8_1, openmls);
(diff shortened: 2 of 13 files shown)


compat_tests/README.md
```diff
@@ -0,0 +1,11 @@
+# Compatibility tests for `openmls`
+
+This crate includes tests for compatibility with previous `openmls` versions.
+- storage format compatibility tests for `openmls=0.7.1`, `openmls=0.8.1`, and `openmls=0.8.1` with the feature `extensions-draft-08` enabled
+
+## Usage
+```
+cargo test -F compat_0_7_1
+cargo test -F compat_0_8_1
+cargo test -F compat_0_8_1_extensions
```
franziskuskiefer, line 10: Let's add this to CI as well. Maybe with a helper script so it's easy to run all of them locally as well.

compat_tests/tests/test_storage_tag_stability.rs
```diff
@@ -1,24 +1,39 @@
 //! These tests check storage tag stability for `serde` serializations of `Extension`,
 //!   `ExtensionType`, `Proposal`, `ProposalType`, and `CredentialType`.
+#![allow(dead_code)]
```
franziskuskiefer, line 3: This seems unnecessary (I don't see warnings if running with any of the features).

openmls/src/credentials/mod.rs
```diff
@@ -77,31 +77,32 @@ pub mod errors;
...
-    PartialEq,
-    Eq,
-    PartialOrd,
-    Ord,
-    openmls_serialization_helpers::Serialize,
-    openmls_serialization_helpers::Deserialize,
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[cfg_attr(
+    feature = "0-8-1-storage-format",
+    derive(serde::Serialize, serde::Deserialize)
```
franziskuskiefer, line 83: Ah, I thought about using the tag in both versions, but keeping the behaviour the same as without for 0.8.1. But this is fine too.

openmls/Cargo.toml
```diff
@@ -46,6 +46,8 @@ zeroize = "1.8.1"
 serde_bytes = "0.11.17"
 
 [features]
+# enable for compatibility with openmls=0.8.1
+0-8-1-storage-format = []
```
franziskuskiefer, line 50: I think this should be enabled by default. Because that's what we have on main and in the latest release. Or maybe we should turn in it around? Either is fine. But let's make sure the default behaviour here doesn't break storage for main -> main.

franziskuskiefer (changes requested):
lgtm with a few nits

compat_tests/tests/test_storage_tag_stability.rs
wysiwys: It's mainly for avoiding build warnings when running with none of the test features enabled (although it doesn't seem to affect the macro definitions)

openmls/Cargo.toml
franziskuskiefer: We decided to keep it this way and updated the changelog to make it clear that this was a bug and the feature may be needed.

franziskuskiefer (approved):

PR #2056, opened 2026-06-15
Virtual clients: More efficient PPRF

kkohbrok:
This PR switches virtual client operations from the 32byte PPRF to a more efficient SecretTree-style PR with one leaf per emulation group member plus on ratchet per operation type.

This PR should be reviewable commit-by-commit.

This PR defers to follow-up PRs:

- KeyPackage creation and Welcome processing
- `GenerationId` usage
- `VirtualClientAction` via SafeAAD
- Backwards-compat pass

Completely out of scope for now:

- Onboarding new emulator clients via state transfer
- Update proposals (in higher-level groups)
- Private handshake messages

### DIFF
--- a/openmls/src/components/vc_operation_tree.rs
+++ b/openmls/src/components/vc_operation_tree.rs
@@ -31,7 +31,7 @@ use std::collections::BTreeMap;
 
 use openmls_traits::{crypto::OpenMlsCrypto, types::Ciphersuite};
 use serde::{Deserialize, Serialize};
-use tls_codec::{Serialize as _, TlsSerialize, TlsSize};
+use tls_codec::{Serialize as _, TlsSerialize, TlsSize, VLBytes};
 
 use crate::{
     binary_tree::{
@@ -178,7 +178,7 @@ impl OperationSecretTree {
             leaf_index,
             generation,
             operation_type,
-            operation_context: operation_context.to_vec(),
+            operation_context: operation_context.to_vec().into(),
         };
         // `operation_generation_secret` is dropped when this function
         // returns, deleting it as required by the spec.
@@ -548,7 +548,7 @@ struct OperationContext {
     leaf_index: LeafNodeIndex,
     generation: u32,
     operation_type: VirtualClientOperationType,
-    operation_context: Vec<u8>,
+    operation_context: VLBytes,
 }
 
 impl OperationContext {
--- a/openmls/src/components/vc_operation_tree.rs
+++ b/openmls/src/components/vc_operation_tree.rs
@@ -160,7 +160,7 @@ impl OperationSecretTree {
     /// consumed generation and reuses key material, the reuse the draft
     /// forbids. See the type-level `# Concurrency` note.
     #[allow(clippy::too_many_arguments)]
-    pub fn derive_operation_secret(
+    pub(crate) fn derive_operation_secret(
         &mut self,
         crypto: &impl OpenMlsCrypto,
         ciphersuite: Ciphersuite,
@@ -199,7 +199,7 @@ impl OperationSecretTree {
     /// observed on the wire. Allocating from a stale copy, or persisting
     /// late, lets two operations share a generation and reuse key material,
     /// the reuse the draft forbids. See the type-level `# Concurrency` note.
-    pub fn next_operation_secret(
+    pub(crate) fn next_operation_secret(
         &mut self,
         crypto: &impl OpenMlsCrypto,
         ciphersuite: Ciphersuite,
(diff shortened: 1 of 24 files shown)


openmls/src/components/vc_operation_tree.rs
```diff
@@ -0,0 +1,1092 @@
...
+    /// # Warning
+    ///
+    /// This mutates the tree: it advances and punctures ratchets. The caller
+    /// MUST persist the mutated tree before any other cycle reads it, as a
+    /// single atomic load-derive-store per emulation epoch. Deriving against
+    /// a stale copy, or not persisting before the next derive, re-serves a
+    /// consumed generation and reuses key material, the reuse the draft
+    /// forbids. See the type-level `# Concurrency` note.
+    #[allow(clippy::too_many_arguments)]
+    pub fn derive_operation_secret(
```
raphaelrobert: pub(crate) should be enough here

openmls/src/components/vc_operation_tree.rs
```diff
@@ -0,0 +1,1092 @@
...
+    ///
+    /// # Warning
+    ///
+    /// This mutates the tree: it advances the own ratchet by one generation.
+    /// The caller MUST persist the mutated tree in the same atomic
+    /// load-derive-store cycle, before the allocated generation can be
+    /// observed on the wire. Allocating from a stale copy, or persisting
+    /// late, lets two operations share a generation and reuse key material,
+    /// the reuse the draft forbids. See the type-level `# Concurrency` note.
+    pub fn next_operation_secret(
```
raphaelrobert: pub(crate) should also be enough here

openmls/src/group/mls_group/mod.rs
```diff
@@ -549,6 +539,19 @@ impl MlsGroup {
     }
 }
 
+/// Error resolving the [`EmulationEpochState`] bound to a group at a given
+/// epoch via [`MlsGroup::vc_emulation_state_at_epoch`]. Callers map it to
+/// their own error type.
+///
+/// [`EmulationEpochState`]: crate::components::vc_derivation_info::EmulationEpochState
+#[cfg(feature = "virtual-clients-draft")]
+pub(crate) enum VcEmulationStateError<StorageError> {
```
raphaelrobert, line 549: This does not derive `thiserror`

traits/src/storage.rs
```diff
@@ -654,37 +667,27 @@ pub trait StorageProvider<const VERSION: u16> {
...
     /// Remove the per-epoch emulation bindings of the given group. Called
     /// when the group is being deleted.
     #[cfg(feature = "virtual-clients-draft")]
     fn delete_vc_emulation_bindings<GroupId: traits::GroupId<VERSION>>(
         &self,
         group_id: &GroupId,
     ) -> Result<(), Self::Error>;
+
+    /// Delete the Virtual Client Operation Secret Tree of the given epoch.
+    /// Called when the emulation epoch's state is being removed.
```
raphaelrobert: This is only called by the application. Either the documentation is not quite accurate, or there's something missing.

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -364,23 +354,26 @@ impl<'a, G: BorrowMut<MlsGroup>> CommitBuilder<'a, Initial, G> {
...
-    /// already-registered emulation epoch (see
-    /// [`MlsGroup::register_vc_emulation_epoch`]). When set, building this
-    /// commit will:
+    /// The application supplies the [`EpochId`] of an already-registered
+    /// emulation epoch (see
+    /// [`MlsGroup::register_vc_emulation_epoch`]). This method loads the
+    /// per-epoch operation secret tree and AEAD key from the storage
+    /// provider, advances the own `LeafNode` operation ratchet by one
+    /// generation, and immediately persists the advanced tree. `build`
+    /// then:
```
raphaelrobert: This ordering can be problematic, because it burns a generation no matter what, while things in the `build()` step can still fail. Shouldn't the persistence happen at a later stage?

raphaelrobert (changes requested):
Thanks, this is pretty comprehensive. Some nits and 2 logic questions.

traits/src/storage.rs
kkohbrok: Good point. It's up to the application to decide when to clean up vc emulation state, but it shouldn't also have to explicitly delete the secret tree.

EDIT: I've unified the endpoints s.t. the application only has to call one.

openmls/src/group/mls_group/commit_builder.rs
kkohbrok: I agree that this isn't ideal and we could write during `build`, but that would complicate the API quite a bit. I'm not too worried for multiple reasons.

- If building fails, we've burned a ratchet generation. That's not great, but also not a big deal.
- Applications will probably make sure that building doesn't fail to begin with. If it does fail, that's probably a bug and the application has other problems.
- Applications will likely run the whole commit building process in a transaction, so in the end it doesn't really matter where exactly we fail.

What I've done to slightly improve the situation is move the leaf validation checks prior to the op secret derivation.

raphaelrobert (approved):
Thanks for making all the changes!

PR #2066, opened 2026-06-17
docs: remove legacy MlsGroup::new doc comment

jazzz:
## Problem 

The doc comment on `MlsGroup::new` incorrectly mentions that private keys are deleted from store when creating a new `MlsGroup`. This appears to be a left over comment from before #1177 when `MlsGroup` required KeyPackages during initialization. 

As this is a high level entry point - this doc comment is used by developers use during debugging. In many cases this is even displayed within IDEs. Having incorrect information adds friction for developers

## Solution

Remove the doc comment as it's no longer correct.


kkohbrok (approved):
Good point, thanks for the PR!

PR #2067, opened 2026-06-18
VC KeyPackage handling

kkohbrok:
Adds KeyPackage building and Welcome processing for virtual clients (behind `virtual-clients-draft`), so a virtual client can publish KeyPackages that a sibling recovers and uses to join higher-level groups.

- **Building:** `KeyPackageBuilder::build_vc_batch` allocates one `key_package` operation generation per batch, derives a per-KeyPackage seed (and from it the init and leaf encryption keys) for each index, and embeds an encrypted `DerivationInfo` in each leaf under `VC_COMPONENT_ID`. Returns a `VcKeyPackageBatch` with the generation and the built bundles.
- **Carrying between siblings:** `assemble_vc_key_package_upload` builds and retains the key material for a received KeyPackage upload message.
- **Welcome:** `find_and_validate_vc_own_leaf` locates the joined leaf by its derived encryption key, decrypts the leaf's `DerivationInfo`, and checks the embedded `(leaf_index, generation, key_package_index)` against the retained material.

### DIFF
--- a/openmls/src/components/vc_derivation_info.rs
+++ b/openmls/src/components/vc_derivation_info.rs
@@ -1,6 +1,7 @@
 use openmls_traits::{
     crypto::OpenMlsCrypto,
     types::{Ciphersuite, CryptoError},
+    OpenMlsProvider,
 };
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
@@ -10,7 +11,7 @@ use tls_codec::{
 
 use crate::{
     binary_tree::{array_representation::TreeSize, LeafNodeIndex},
-    ciphersuite::Secret,
+    ciphersuite::{hash_ref::KeyPackageRef, Secret},
     messages::PathSecret,
     treesync::node::encryption_keys::EncryptionKeyPair,
 };
@@ -23,14 +24,20 @@ use crate::{
 pub const VC_COMPONENT_ID: u16 = 0x0006;
 
 // Operation-secret child labels. Each child is derived from the per-operation
-// secret produced by the per-epoch operation secret tree. Only
-// `Encryption Key` and `Path Generation` are wired up at the moment, which is
-// all the `leaf_node` commit path needs. The spec also defines
-// `Signature Key` and `Init Key` children. Those, together with the
-// `key_package` and `application` operation paths that consume them, are
-// deferred to a follow-up PR.
+// secret produced by the per-epoch operation secret tree. `Encryption Key`
+// and `Path Generation` cover the `leaf_node` commit path, and `Init Key`
+// covers the `key_package` operation path. The spec also defines a
+// `Signature Key` child, which together with the operation paths that consume
+// it is deferred to a follow-up PR.
 const ENCRYPTION_KEY_LABEL: &str = "Encryption Key";
 const PATH_GENERATION_LABEL: &str = "Path Generation";
+const INIT_KEY_LABEL: &str = "Init Key";
+/// `ExpandWithLabel` label for the per-KeyPackage seed secret derived from a
+/// `key_package` operation secret (mls-virtual-clients draft, batch KeyPackage
+/// derivation). One operation secret covers a batch of KeyPackages, and each
+/// KeyPackage's seed is expanded from it using the KeyPackage's index as the
+/// context.
+const KEY_PACKAGE_SEED_LABEL: &str = "key package seed";
 
 /// `ExpandWithLabel` label for the [`DerivationInfoTbe`] AEAD key derived
 /// from the per-epoch [`EpochEncryptionKey`].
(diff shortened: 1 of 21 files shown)


openmls/src/components/vc_derivation_info.rs
```diff
@@ -373,6 +389,264 @@ impl DerivationInfo {
...
+    /// Position of this KeyPackage within the batch.
+    pub key_package_index: u32,
+    /// Per-KeyPackage seed secret from which the init and leaf-encryption keys
+    /// are derived at Welcome time.
+    pub key_package_seed_secret: KeyPackageSeedSecret,
+}
+
+/// Reject a batch whose [`KeyPackageInfo`] entries are not all distinct.
+///
+/// Returns [`VirtualClientsError::DuplicateKeyPackageIndex`] if any
```
boxdot, line 483: This seems to be outdated.

openmls/src/components/vc_derivation_info.rs
```diff
@@ -652,27 +1025,381 @@ pub enum VirtualClientOperationType {
...
+        match self {
+            Self::LeafNode { generation, .. } | Self::KeyPackage { generation, .. } => *generation,
+        }
+    }
+
+    /// Serialize the variant's fields in order, with no variant tag, matching
+    /// the `DerivationInfoTBE` select. The TLS derive macros cannot express a
+    /// tagless select, so this codec is written by hand.
+    fn tls_serialize_detached(&self) -> Result<Vec<u8>, tls_codec::Error> {
+        let mut out = Vec::new();
```
boxdot: Minor: init with `Vec::with_capacity()` inside branches and use the sum of individual field tls sizes.

openmls/src/group/mls_group/creation.rs
```diff
@@ -618,32 +637,216 @@ fn keys_for_welcome<Provider: OpenMlsProvider>(
...
+    };
+    if leaf_index != material.leaf_index
+        || generation != material.generation
+        || key_package_index != material.key_package_index
+    {
+        log::error!("vc: welcome leaf derivation info does not match the retained material");
+        return Err(VirtualClientsError::DerivationInfoMalformed.into());
+    }
+    if own_leaf.encryption_key().as_slice() != derived_encryption_key {
+        log::error!("vc: welcome leaf encryption key does not match the derived key");
```
raphaelrobert: This check shouldn't be needed, because we picked our own leaf by looking at the key in 777.

openmls/src/key_packages/mod.rs
```diff
@@ -573,6 +634,298 @@ impl KeyPackageBuilder {
...
+    generation: u32,
+    operation_secret: crate::components::vc_derivation_info::OperationSecret,
+}
+
+/// Merge a virtual-clients derivation-info blob into `resolved_dictionary`
+/// under [`VC_COMPONENT_ID`] and build the resulting leaf-node extensions.
+/// The dictionary already carries every other component id (notably
+/// `AppComponents`), so injecting the derivation info preserves them.
+#[cfg(feature = "virtual-clients-draft")]
+fn inject_vc_derivation_info(
```
raphaelrobert: We have a very similar function in `commit_builder.rs`, we should merge the two.

openmls/src/key_packages/mod.rs
```diff
@@ -573,6 +634,298 @@ impl KeyPackageBuilder {
...
+        epoch_id: crate::components::vc_derivation_info::EpochId,
+        count: u32,
+    ) -> Result<
+        (
+            u32,
+            Vec<(
+                KeyPackageBundle,
+                crate::components::vc_derivation_info::KeyPackageInfo,
+            )>,
+        ),
```
raphaelrobert: Can we return a struct that gives a name to the `u32` here?

openmls/src/key_packages/mod.rs
```diff
@@ -573,6 +634,298 @@ impl KeyPackageBuilder {
...
+    /// sibling.
+    #[cfg(feature = "virtual-clients-draft")]
+    pub fn build_vc_batch(
+        self,
+        ciphersuite: Ciphersuite,
+        provider: &impl OpenMlsProvider,
+        signer: &impl Signer,
+        credential_with_key: CredentialWithKey,
+        epoch_id: crate::components::vc_derivation_info::EpochId,
+        count: u32,
```
raphaelrobert, line 679: We should return early when the count is 0.

openmls/src/key_packages/mod.rs
```diff
@@ -573,6 +634,298 @@ impl KeyPackageBuilder {
...
+            VirtualClientOperationType::KeyPackage,
+            b"",
+        )?;
+        provider
+            .storage()
+            .write_vc_operation_tree(&epoch_id, &operation_tree)
+            .map_err(|e| {
+                log::error!("vc: persist advanced operation tree in build_vc_batch failed: {e:?}");
+                VirtualClientsError::StorageError
+            })?;
```
raphaelrobert: This is minor, but since we discussed it in the last PR: There's also a risk ok here of burning secrets if things fail later on in `build_vc_key_package_for_index`, like `resolve_vc_leaf_dictionary`.

raphaelrobert (changes requested):
This looks good in general. I found a few nits. It would also be good to have a PR description for posterity.

openmls/src/group/mls_group/creation.rs
kkohbrok: Good catch, thanks!

openmls/src/key_packages/mod.rs
kkohbrok: Good point. It's now returning an error if the count is 0.

openmls/src/key_packages/mod.rs
kkohbrok: Good point. I've done some refactoring to move all storage writes to the end of the function.

openmls/src/components/vc_derivation_info.rs
boxdot: Never mind.

raphaelrobert (approved):

PR #2068, opened 2026-06-18
Add sqlx example with transactions

kkohbrok:


boxdot (approved):

sqlx_storage/examples/transaction.rs
```diff
@@ -0,0 +1,175 @@
+//! Using the SQLx storage provider inside a SQLite transaction.
+//!
+//! The [`SqliteStorageProvider`] borrows a [`SqliteConnection`]. A
+//! [`sqlx::Transaction`] dereferences to a [`SqliteConnection`], so the same
+//! provider can run against an open transaction by passing `&mut *transaction`.
+//! Every write the provider performs then becomes part of that transaction and
+//! either commits or rolls back together with the rest of your application's
+//! writes.
+//!
```
wysiwys, line 9: For the use of SQLite in this example, could it be useful to also include a note about transaction behavior here, e.g. that the transactions in this example use `BEGIN DEFERRED` (and the transaction is not upgraded to a write transaction until the first write)? For some applications, this general topic may be relevant from an error handling perspective - however, it may also be out of scope for this example

wysiwys (approved):
This looks good to me, with one optional nit on the docs for the example!

sqlx_storage/examples/transaction.rs
kkohbrok: Good point, but I would say somewhat orthogonal to the example. I've added a note to the readme, though.

PR #2073, opened 2026-06-24
Expose more From<Out> for In impls without testing feature flag

keks:
These impls mostly existed before, but were behind the test-utils feature flag and had a comment that they are dangerous. However, converting this way is not dangerous, just the other way around: It is safe to convert from a type assumed to be validated to an equivalent type with no such assumption.

We already allow converting this way for several types (e.g. KeyPackage, Commit). This commit adds the same impls for the remaining types, allowing to convert an MlsMessageOut to an MlsMessageIn.

### DIFF
--- a/openmls/src/framing/message_out.rs
+++ b/openmls/src/framing/message_out.rs
@@ -193,6 +193,30 @@ impl MlsMessageOut {
     }
 }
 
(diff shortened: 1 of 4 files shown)


kkohbrok (approved):

franziskuskiefer (approved):
lgtm

PR #2076, opened 2026-06-26
[Draft] Storage migration helpers

wysiwys:
This pull request proposes an `openmls_storage_migration` crate that includes helpers for storage format migrations. 

To enable storage migrations, a `storage_migration` feature flag  is added to `openmls`, that allows compatibility types to be used to convert legacy stored types into the updated types. In particular, these are used for the migration path needed for providers that use non-self-describing `serde` formats, where the addition of `Option<T>` fields annotated with `#[serde(default)]` since `openmls=0.7.0` was not backwards-compatible with legacy serializations. 

`openmls::storage::migration` provides a helper method for this specific migration path, which can be called to migrate all `MlsGroupState` records in the storage provider affected by this issue.

To enable loading of all `GroupId`s used as keys in the storage provider, the `GroupId` trait in `openmls_traits` now also requires the `Entity<_>` trait.

This pull request also updates the tests in `compat_tests` to test the migration of `MlsGroupState` when a non-self-describing serialization is used. In `compat_tests`,  `openmls=0.7.0` is now used instead of `openmls=0.7.1` for the existing storage tag compatibility tests, to avoid dependency resolution issues when using multiple exact-pinned patch versions.

Questions:
- Should the compatibility types in `openmls` be moved to separate modules?
- Should the `StorageMigrationHelper` trait be generic over multiple storage versions at this stage? This could be useful if a new storage version is added later on, to enable a one-time storage migration for the entire provider.
- These changes currently only support some migration paths, in particular `openmls=0.7.0` -> `main` (with different combinations of features enabled). Should different feature-gated paths be provided?
- `openmls::storage::migration` provides a helper method for specific migration paths. Would it be better to move this helper to the `openmls_storage_migration` crate?


storage_migration/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+        StorageMigrationError<
+            <Self as StorageProvider<PREVIOUS_VERSION>>::Error,
+            <Self as StorageProvider<CURRENT_VERSION>>::Error,
+        >,
+    > {
+        for key in keys(self).map_err(StorageMigrationError::Legacy)? {
+            match read_current(self, &key) {
+                Ok(Some(_)) => {}
+                Ok(None) => {}
+                Err(_) => {
```
kkohbrok: Shouldn't we check whether this was actually a deserialization error? Also, are we sure that deserialization will always fail for old formats?

openmls/src/storage/migration.rs
```diff
@@ -0,0 +1,39 @@
+//! Migration helpers for `openmls`
+
+use openmls_storage_migration::*;
+use openmls_traits::storage::{traits, Entity, StorageProvider, CURRENT_VERSION};
+
+/// Migrate all stored `MlsGroupState`s  for each `GroupId` key
+///
+/// NOTE: This migration helper is intended for upgrades from `openmls=0.7.0` -> the current version,
+/// with non-self-describing `serde` storage formats.
```
kkohbrok, line 9: If this is really meant to be specifically for this version transition only, I don't think we need two storage provider versions, right?

kkohbrok (commented):
I like the approach, but it seems very specific to this particular migration. For example, shouldn't the `*Compat` structs be called `*007Compat` or similar? It's totally fine if this PR is just for this particular migration, but we may want to make sure that we can extend it for future migrations.

openmls/src/schedule/message_secrets.rs
```diff
@@ -24,6 +24,31 @@ pub(crate) struct MessageSecrets {
     added_at: Option<SystemTime>,
 }
 
+/// Combined message secrets that need to be stored for later decryption/verification
+#[derive(Serialize, Deserialize)]
```
franziskuskiefer: Let's keep the compat code in separate modules to not pollute the main code.

compat_tests/Cargo.toml
```diff
@@ -5,20 +5,42 @@ edition = "2021"
...
 compat_0_8_1 = ["dep:openmls_0_8_1", "openmls/0-8-1-storage-format"]
 compat_0_8_1_extensions = [
     "compat_0_8_1",
     "openmls_0_8_1/extensions-draft-08",
+    "extensions-draft-08"
+]
+
+# enable features on the underlying `openmls` crate,
+# and on the StorageProvider implementation in `test_storage_provider`
+extensions-draft-08 = [
```
franziskuskiefer: Looks like this needs a rebase. This changed.

book/src/user_manual/storage_migration.md
```diff
@@ -0,0 +1,19 @@
+# Storage migration
+
+[comment]: <> (TODO: add overview)
```
franziskuskiefer, line 3: Yeah some intro would be good. Also talking about the feature.

storage_migration/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+pub enum StorageMigrationError<LegacyError, CurrentError> {
+    #[error(transparent)]
+    Legacy(LegacyError),
+    #[error(transparent)]
+    Current(CurrentError),
+}
+
+/// A trait that enables storage migrations from `PREVIOUS_VERSION` formats to `CURRENT_VERSION`
+/// formats to be performed on a storage provider.
+pub trait StorageMigrationHelper<const PREVIOUS_VERSION: u16, const CURRENT_VERSION: u16>:
```
franziskuskiefer, line 13: We should probably use something other the `const` generic names. Otherwise this may get confused with the storage provider versions? But otherwise I think it's the right shape.

franziskuskiefer (commented):
> Should the compatibility types in openmls be moved to separate modules?

yes, let's keep the migration code separate from the main code.


> Should different feature-gated paths be provided?

I don't think that's needed yet. We can do that later.

> Should the StorageMigrationHelper trait be generic over multiple storage versions at this stage? 

Yes I think that's the right shape.


I think the interaction with the `virtual-clients-draft` feature is broken.

Do you want to turn this into a reviewable PR then?

storage_migration/src/lib.rs
wysiwys: The intent of this trait is actually to refer specifically to these storage provider versions; for this specific migration, these versions are identical, but the idea of this trait in general i that it could be reused later on for migrating to the next storage provider version. One other goal would be to help reduce code duplication in the repository and keep the maintenance footprint of each migration tool as low as possible. However, since there isn't a use case yet for a `VERSION = 1` -> `VERSION = 2` migration, it could also make sense to keep this trait to a single version, and revisit the design later on once it's needed

storage_migration/src/lib.rs
wysiwys: Indeed, in some cases, the deserialization may actually not fail with some self-describing formats, but instead read data into an invalid format. Because of this issue with tolerant deserializations I've replaced the tolerant deserialization logic with a one-time migration.

wysiwys:
I've updated this PR to no longer allow tolerant deserializations of these specific items, and instead to enable a one-time migration of `MlsGroupState` and `MessageSecretsStore`. One further, potential step could also be to make this approach use a `serde_json` bridge (as in #2128) instead of using dedicated compat structs.

PR #2084, opened 2026-06-29
feat: signal own pending commits in processing

### LINKED ISSUE #2077: [Bug] RFC incompliance: Self-generated commits should be possible to decode
# Describe the bug

The library refuses to process commits generated by the same client, in contradiction with [RFC9420 Section 3.2](https://www.rfc-editor.org/rfc/rfc9420.html):

> When client A wants to establish a group with clients B and C, it first initializes a group state containing only itself and downloads KeyPackages for B and C. For each member, A generates an Add proposal and a Commit message to add that member and then broadcasts the two messages to the group. Client A also generates a Welcome message and sends it directly to the new member (there's no need to send it to the group). **Only after A has received its Commit message back from the Delivery Service does it update its state to reflect the new member's addition.**

Right now not only the received back commit is not possible to stage and commit, but it's not possible to even decode the message at all to find out what message it is without getting an error.

# Reproduction

Generate a commit:

```rs
let staged_commit = self.group.commit_builder()
	// ...		
	.build(provider.rand(), provider.crypto(), &signature_key_pair, |_| true)?
	.stage_commit(&provider)?;

let (commit, welcome, groupInfo) = staged_commit.into_contents();
```

(`wire_format_policy` = `PURE_PLAINTEXT_WIRE_FORMAT_POLICY` so `commit` is a `PublicMessage`)

...and then send `commit` over to the DS, and receive it back and try to decode it:

```rs
let message: Vec<u8> = ...;
let protocol_message = MlsMessageIn::tls_deserialize_exact_bytes(message.as_slice())?.try_into_protocol_message()?;
let processed = group.process_message(&provider, protocol_message); // 💥
```

You get [this](https://github.com/openmls/openmls/blob/59e3ac7f34a3fe875d8d436759da0baf44940983/openmls/src/group/errors.rs#L221) error:

```
Error: The Commit was created by this client.
```

# Expected behavior

1. The commit should be possible to decode as long as it is a `PublicMessage`
2. It should be possible to stage and commit the commit

# Use cases

## Use case 1: Preserving group integrity amid network errors

Consider the following scenario:

1. User A clicks "Add User B" button
2. The app generates and stages a commit:
    ```rs
    let staged_commit = self.group.commit_builder()
        .propose_adds(key_package)
        .build(provider.rand(), provider.crypto(), &signature_key_pair, |_| true)?
        .stage_commit(&provider)?;
    let (commit, welcome, groupInfo) = staged_commit.into_contents();
    ```
3. The app sends the `commit` and `welcome` over to the DS to fan out to other members
4. 💥 User A is now in underground tunnel and lost signal, the app never received a confirmation from the DS, but the commit actually went through 💥
5. At this point User A discarded the commit yet all other users have received and accepted it
6. Group state went out of sync

---

It _is_ possible to workaround this with a queue-ish solution as such the commit is retried until it's explicitly approved or rejected by the DS, but:
- In the example above if the commit didn't go through before connection loss, and the user gained back connection after 5 hours, then User B would instantly be added at that time, which is wildly unexpected UX
- Such a workaround is extremely convoluted compared to just relying on the principle that: Committed commits origin explicitly from the DS -- Whether a commit went through or not doesn't matter anymore if it did then it'll be received back and processed at some point, and if it didn't then nothing happens

## Use case 2: Consistent messaging history

If all messages (commits, application messages, etc.) are exclusively sourced from the DS, then knowing that the DS delivers messages in the same order for all clients, all clients get to have all messages received in identical order.

This is not true in an otherwise scenario; If a client is trying to commit something and the network request took 10 seconds, then for the sender the commit happened at N time but for others it happened at N+10, causing messaging history inconsistency

## Use case 3: Eliminating the need for out of order tolerance

As with above, if all messages are sourced from the DS, then you get 100% guarantee that everyone is receiving messages in the same order and out of order tolerance can be set to `0` (= storing previous epoch keys is not needed = much better security for groups that don't frequently commit)

raphaelrobert:
Fixes #2077.

Our own Commit messages can be interpreted as a signal to merge a pending commit when processing messages. We now no longer return an error, but instead return `OwnPendingCommit` if the incoming Commit matches the pending commit.

Caveats:
1. We don't fully verify the content of the incoming Commit, we only verify the signature and the validity of the confirmation tag. The reason is that it is not trivial to verify the rest, because we don't hold the wire format of the Commit as part of the local pending Commit. The risk is minimal though: an attacker would have to have access to the signing key, craft a malformed Commit with a valid confirmation tag and could then create a group split where the own client would merge the pending Commit but other members would reject the malformed Commit.
2. This only works for public messages, a Commit in a private message still cannot be decrypted. This creats a slight asymmetry in the API.


boxdot (approved):

kkohbrok (approved):

openmls/src/group/mls_group/processing.rs
```diff
@@ -765,6 +770,29 @@ impl MlsGroup {
                 #[cfg(not(feature = "virtual-clients-draft"))]
                 let _ = commit;
 
+                let is_own_pending_commit =
+                    matches!(&sender, Sender::Member(member) if member == &self.own_leaf_index());
+                #[cfg(feature = "virtual-clients-draft")]
+                let is_own_pending_commit = is_own_pending_commit && vc_material.is_none();
```
boxdot, line 776: Minor: https://doc.rust-lang.org/std/macro.cfg_select.html

But then you have to bump MSRV to 1.95

raphaelrobert: No objections to bumping again, but that seems out-of-scope for this PR

PR #2088, opened 2026-07-01
fix: align doc generation

raphaelrobert:
I broke the doc generation in GH pages with #2084, and it went unnoticed because the commands were not aligned. This PR fixes that.


keks (approved):
I think that looks good, thanks! It looks like the PR you cite as breaking the docs might be different, though? That one looks unrelated.

raphaelrobert:
> I think that looks good, thanks! It looks like the PR you cite as breaking the docs might be different, though? That one looks unrelated.

Indeed, that was a typo. It's corrected now.

PR #2089, opened 2026-07-01
allow staging of own pathless commits

kkohbrok:
This PR allows clients to process and stage their own commits if they don't have a path _and_ don't match the currently pending commit. This is mostly for use in the virtual clients setting.

### DIFF
--- a/openmls/src/framing/validation.rs
+++ b/openmls/src/framing/validation.rs
@@ -513,13 +513,17 @@ pub enum ProcessedMessageContent {
     ///
     /// This is returned instead of
     /// [`StagedCommitMessage`](Self::StagedCommitMessage) when the processed
-    /// Commit was created by this client. Since this client already holds the
-    /// corresponding pending commit, the incoming Commit is not staged. To
-    /// apply it, merge the pending commit using
+    /// Commit was created by this client and matches the group's pending commit.
+    /// Since this client already holds the corresponding pending commit, the
+    /// incoming Commit is not staged. To apply it, merge the pending commit
+    /// using
     /// [`MlsGroup::merge_pending_commit()`](crate::group::mls_group::MlsGroup::merge_pending_commit()).
+    /// An own Commit that does not match the pending commit is instead returned
+    /// as a [`StagedCommitMessage`](Self::StagedCommitMessage) (if it carries no
+    /// path) or rejected (if it carries a path we cannot decrypt).
     ///
     /// The match against the pending commit is established by comparing the
-    /// confirmation tag of the incoming Commit against one recomputed from the
+    /// confirmation tag of the incoming Commit against the one stored with the
     /// pending commit. The message signature has already been verified, which
     /// authenticates the Commit as ours, and a matching confirmation tag binds
     /// the confirmed transcript hash of the new epoch. We do not otherwise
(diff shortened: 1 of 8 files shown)


openmls/src/framing/validation.rs
```diff
@@ -513,13 +513,17 @@ pub enum ProcessedMessageContent {
...
-    /// corresponding pending commit, the incoming Commit is not staged. To
-    /// apply it, merge the pending commit using
+    /// Commit was created by this client and matches the group's pending commit.
+    /// Since this client already holds the corresponding pending commit, the
+    /// incoming Commit is not staged. To apply it, merge the pending commit
+    /// using
     /// [`MlsGroup::merge_pending_commit()`](crate::group::mls_group::MlsGroup::merge_pending_commit()).
+    /// An own Commit that does not match the pending commit is instead returned
+    /// as a [`StagedCommitMessage`](Self::StagedCommitMessage) (if it carries no
+    /// path) or rejected (if it carries a path we cannot decrypt).
```
raphaelrobert: I don't understand the part about the encrypted path. Do you mean a PrivateMessage? This should also mention UpdatePath again.

openmls/src/group/mls_group/staged_commit.rs
```diff
@@ -139,7 +139,8 @@ impl MlsGroup {
             .map_err(|_| LibraryError::custom("Using the key schedule in the wrong state"))?)
     }
 
-    /// Stages a commit message that was sent by another group member. This
+    /// Stages a commit message. The commit may have been sent by another group
+    /// member or be our own path-less commit. This
```
raphaelrobert: Commit without an UpdatePath

openmls/src/group/public_group/staged_commit.rs
```diff
@@ -252,7 +252,8 @@ impl PublicGroup {
         Ok(())
     }
 
-    /// Stages a commit message that was sent by another group member.
+    /// Stages a commit message. The commit may have been sent by another group
+    /// member or be our own path-less commit.
```
raphaelrobert: Commit without an UpdatePath

openmls/src/group/mls_group/processing.rs
```diff
@@ -770,23 +765,40 @@ impl MlsGroup {
...
+                            emulator_sender_leaf_index,
+                        ));
+                    }
+                    // Not our pending commit. We cannot decrypt a path we
+                    // encrypted to the other members, so a Commit with an
+                    // UpdatePath is unprocessable. A Commit without an
+                    // UpdatePath carries no author-private material and falls
+                    // through to staging (a sibling's Commit without an
+                    // UpdatePath, or our own commit replayed after the pending
+                    // commit was cleared).
```
raphaelrobert, line 799: This comment block sounds a bit confusing. A path cannot be decrypted per-se. The comment should state more clearly what the branching is about.

openmls/src/group/mls_group/processing.rs
```diff
@@ -770,23 +765,40 @@ impl MlsGroup {
...
+                let is_own_commit = is_own_commit && vc_material.is_none();
+
+                if is_own_commit {
+                    let received_tag = content
+                        .confirmation_tag()
+                        .ok_or(StageCommitError::ConfirmationTagMissing)?;
+                    if self.matches_pending_commit(received_tag) {
+                        // The Commit is our pending commit echoed back to us:
+                        // surface `OwnPendingCommit` so the caller merges the
+                        // pending commit instead of staging the echo.
```
raphaelrobert: I see "the echo" in a few places, the term is not very precise. I understand it refers to a Commit that was fanned out by the DS, but we should say just that then.

raphaelrobert (changes requested):
I think this is a nice improvement. The logic looks fine, but the comments need a bit more work.

raphaelrobert (approved):

PR #2090, opened 2026-07-01
virtual clients: allow group creation by siblings

kkohbrok:
This PR allows emulator clients to locally bootstrap groups created by sibling emulator clients. Based on #2087.


raphaelrobert (approved):
LGTM. It might be interesting to scan for code duplication once the VC feature is complete.

PR #2092, opened 2026-07-02
feat: better AppDataUpdate processing API

raphaelrobert:
This PR aligns the API introduced in #1930 with the rest of the processing API.

Before: processing a commit with AppDataUpdate proposals required a special three-step flow, separate from the normal process_message path:

```rust
  // Unprotect only — returns an UnverifiedMessage (signature NOT yet checked)
  let unverified = group.unprotect_message(provider, message)?;

  // Manually inspect raw proposals: validate each one yourself,
  // resolve by-reference proposals from the proposal store yourself
  for p in unverified.committed_proposals().unwrap() { /* validate, resolve, interpret */ }

  // Finish with the special processing function
  let processed = group.process_unverified_message_with_app_data_updates(
      provider, unverified, updater.changes())?;
```

Calling `process_message` on such a commit failed at runtime with `FoundAppDataUpdateProposal` only if the proposals were inline, by-reference ones slipped past and failed later with a confusing staging error.

After: there is only `process_message`. When a commit covers `AppDataUpdate` proposals, it returns a new `ProcessedMessageContent` variant instead of erroring:

```rust
  let processed = group.process_message(provider, message)?;

  if let ProcessedMessageContent::UnresolvedAppDataCommit(unresolved) = processed.into_content() {
      // Signature already verified; references already resolved; sorted by component id
      let mut updater = group.app_data_dictionary_updater();
      for proposal in unresolved.app_data_update_proposals() { /* interpret the diff */ }
      let updates = updater.changes();

      let staged = group.stage_app_data_commit(provider, *unresolved, updates)?;
      group.merge_staged_commit(provider, staged)?;
  }
```

A nice side-effect is that this requires less code.


kkohbrok (approved):

openmls/src/group/mls_group/processing.rs
```diff
@@ -762,7 +777,7 @@ impl MlsGroup {
                     } else {
                         (None, None)
                     };
-                #[cfg(not(feature = "virtual-clients-draft"))]
+                #[cfg(not(feature = "extensions-draft"))]
                 let _ = commit;
```
keks: Why was this changed and what is this doing anyway? Is the plan to drop the commit here?

keks (approved):
I think this looks a lot better than what we had before! I have one question, but it's not super important.

openmls/src/group/mls_group/processing.rs
raphaelrobert: Good catch. This can be cleaned up. We only drop a reference here, the commit is still inside the `content` variable.

PR #2094, opened 2026-07-03
Convert SenderRatchet::DualUse when toggling the virtual-clients-draft feature

gferon:
The `DualUse` variant of a `SenderRatchet` is not compiled when you're not using the `virtual-clients-draft`, meaning that it's not possible to:
- immediately use a client with create_unconfirmed_message / confirm APIs with already persisted data
- stop using the feature flag and use existing groups 

With this change, the ratchet is converted at `serde` time: 
- `EncryptionRatchet` becomes `DualUse` when the feature is on
- or back to `EncryptionRatchet` when it's off

### DIFF
--- a/openmls/src/tree/dual_use_ratchet.rs
+++ b/openmls/src/tree/dual_use_ratchet.rs
@@ -71,6 +71,20 @@ impl DualUsePastSecret {
     }
 }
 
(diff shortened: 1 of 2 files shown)


boxdot (approved):
Thank you!

PR #2095, opened 2026-07-06
feat: allow AppDataUpdate proposals in external commits

boxdot:
Such proposals are allowed in external commits when the `extensions-draft` feature is enabled.

Additionally, add a getter for the group context in the verifiable group info.


kkohbrok (approved):
Looks good! A test would be nice, though.

PR #2099, opened 2026-07-07
feat: process own private messages

### LINKED ISSUE #2078: [Bug] It should be possible to extract non-encrypted data from own `PrivateMessage`s
# Describe the bug

Currently, if you try to process a `PrivateMessage` that was generated by the same client, you get an [CannotDecryptOwnMessage](https://github.com/openmls/openmls/blob/59e3ac7f34a3fe875d8d436759da0baf44940983/openmls/src/group/errors.rs#L472) error.

If I understand correctly, own `PrivateMessage`s cannot be decrypted because the keys that were used to generate it were deleted and the sources of them were ratcheted forwards; However, a `PrivateMessage` is not entirely ciphertext.

Specifically it should be possible to access properties such as:

- `credential` of the sender to determine who sent this `PrivateMessage` at all
- `aad` because AAD is not encrypted, it's just signed data
- Maybe `epoch`, I think this isn't encrypted too?

# To Reproduce

```rs
match MlsMessageIn::tls_deserialize_exact_bytes(message.as_slice())?.extract() {
	MlsMessageBodyIn::PrivateMessage(private_message_in) => {
		let group = ...;

		let processed_message = self.group.process_message(&provider, private_message_in)?;

		// 💥
	},
	_ => ...
}
```

raphaelrobert:
Follow-up to #2092.
Fixes #2078.

This exposes own private messages as yet another variant of `ProcessedMessageContent` and is the logical counterpart of #2084.


openmls/src/framing/validation.rs
```diff
@@ -49,6 +49,35 @@ use super::{
     *,
 };
 
+/// Result of decrypting an inbound PrivateMessage: either content this client
+/// can process further, or a message this client authored itself, which it
+/// cannot decrypt.
+#[derive(Debug)]
+pub(crate) enum InboundDecryptionResult {
+    /// A message from another sender, decrypted and ready for parsing.
```
kkohbrok:
```suggestion
    /// A message from a sibling emulator client sender, decrypted and ready for parsing.
```

openmls/src/framing/validation.rs
```diff
@@ -49,6 +49,35 @@ use super::{
...
+    Decrypted(DecryptedMessage),
+    /// A private message whose sender data claims this client's own leaf.
+    /// Carries the plaintext framing fields needed to build the
+    /// [`ProcessedMessage`], since the content itself cannot be decrypted.
+    #[cfg(not(feature = "virtual-clients-draft"))]
+    OwnPrivateMessage {
+        epoch: GroupEpoch,
+        authenticated_data: Vec<u8>,
```
kkohbrok, line 65: I don't think we authenticated this data. We should be careful how we expose this.

openmls/src/framing/validation.rs
```diff
@@ -49,6 +49,35 @@ use super::{
...
+/// can process further, or a message this client authored itself, which it
+/// cannot decrypt.
+#[derive(Debug)]
+pub(crate) enum InboundDecryptionResult {
+    /// A message from another sender, decrypted and ready for parsing.
+    Decrypted(DecryptedMessage),
+    /// A private message whose sender data claims this client's own leaf.
+    /// Carries the plaintext framing fields needed to build the
+    /// [`ProcessedMessage`], since the content itself cannot be decrypted.
+    #[cfg(not(feature = "virtual-clients-draft"))]
```
kkohbrok: Can't this happen for virtual clients, too? In the emulator group, for example, or in a higher-level group, when the sender gets its own message echoed back.

CHANGELOG.md
```diff
@@ -25,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#1963](https://github.com/openmls/openmls/pull/1963): `MessageEncryptionError` is now part of the public API, but only surfaces if the `virtual-clients-draft` feature flag is enabled.
 - [#2043](https://github.com/openmls/openmls/pull/2043): Renamed and deprecated `MlsGroup::propose_external_psk` to `MlsGroup::propose_pre_shared_key`. Same for the `_by_value` variant.
 - [#2084](https://github.com/openmls/openmls/pull/2084): Renamed `StageCommitError::OwnCommit` to `StageCommitError::OwnCommitMismatch`, now returned only when a Commit authored by this client does not match the pending commit.
+- Removed `ValidationError::CannotDecryptOwnMessage`; processing an own `PrivateMessage` is no longer an error.
```
kkohbrok: Needs a PR #. Same for the previous change.

kkohbrok (changes requested):
I really like this new API shape. A few questions, though.

CHANGELOG.md
franziskuskiefer: Let's also make clear that this is a breaking change.

franziskuskiefer (commented):
Thanks. I think this is the right way to solve this with the nits from Konrad.

CHANGELOG.md
raphaelrobert: I forgot to add the PR number indeed. Almost all of these changes are breaking API changes, that's why we mention them here, so far we have not added any other annotation though.

openmls/src/framing/validation.rs
raphaelrobert: Refactored and discussed offline.

openmls/src/framing/validation.rs
raphaelrobert: This is not publicly exposed. It's just here so we can carry it over to the next step.

kkohbrok (approved):
Thanks!

PR #2105, opened 2026-07-09
virtual clients: surface own commit echo as OwnPendingCommit

kkohbrok:
Processing a commit checked for sibling VC material before checking whether the commit matches the group's pending commit. Since an own commit fanned back by the DS has the same shape as a sibling's commit (Sender::Member(own_leaf) with VC derivation info on the path leaf), load_vc_commit_material re-derived the operation secret the sender had already consumed at send time and processing failed with OperationGenerationConsumed, making OwnPendingCommit unreachable for VC commits.

Check the pending-commit match first, so an own commit echo surfaces as OwnPendingCommit without touching the operation secret tree. Sibling commits never match the pending commit's confirmation tag and still take the VC staging path.

### DIFF
--- a/openmls/src/group/mls_group/processing.rs
+++ b/openmls/src/group/mls_group/processing.rs
@@ -920,15 +920,18 @@ impl MlsGroup {
                 //   auto-Remove targets our previous leaf, so we are the
                 //   sibling being resynced.
                 #[cfg(feature = "virtual-clients-draft")]
-                let vc_commit_material =
-                    if is_sibling_vc_commit(commit, &sender, self.own_leaf_index()) {
-                        self.load_vc_commit_material(provider, commit)?
-                    } else {
-                        None
-                    };
+                let (vc_commit_material, is_own_commit) = {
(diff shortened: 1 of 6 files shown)


openmls/src/group/mls_group/processing.rs
```diff
@@ -924,16 +900,45 @@ impl MlsGroup {
...
+                #[cfg(feature = "virtual-clients-draft")]
+                let vc_commit_material =
+                    if is_sibling_vc_commit(commit, &sender, self.own_leaf_index()) {
+                        self.load_vc_commit_material(provider, commit)?
+                    } else {
+                        None
+                    };
+
+                #[cfg(feature = "virtual-clients-draft")]
+                let is_own_commit = is_own_commit && vc_commit_material.is_none();
```
boxdot: Minor: does it make sense to combine these two lines into a single block guarded by the feature?

boxdot (approved):

raphaelrobert (approved):

PR #2106, opened 2026-07-09
virtual clients: epoch-scoped message confirmation

kkohbrok:
confirm_message deleted the given generation from the current epoch's own application ratchet. If the group advanced an epoch between create_unconfirmed_message and the confirmation, it either silently no-oped, retaining the old secret forever, or deleted a different message's still-unconfirmed secret that shared the generation number in the new epoch, making a lost race against a sibling undecryptable.

confirm_message now takes the creation epoch, carried on UnconfirmedMessage, and deletes exactly the secret its create call retained, or nothing: aged-out epochs and already-consumed generations are no-op successes, future epochs are rejected. The new confirm_handshake_message applies the same semantics to the handshake ratchet, giving retained own proposal and commit secrets a cleanup path. Proposals and commits share the per-epoch handshake ratchet, so one endpoint covers both.

### DIFF
--- a/openmls/src/group/mls_group/application.rs
+++ b/openmls/src/group/mls_group/application.rs
@@ -17,9 +17,13 @@ use super::{errors::CreateMessageError, *};
 pub struct UnconfirmedMessage {
     /// The encrypted application message to fan out.
     pub message: MlsMessageOut,
-    /// The ratchet generation used for encryption. Pass it back to
-    /// [`MlsGroup::confirm_message`] once the DS has accepted the message, to
-    /// delete the retained encryption secret.
+    /// The epoch the message was encrypted in. Pass it together with
+    /// `generation` to [`MlsGroup::confirm_message`] once the DS has accepted
+    /// the message, to delete the retained encryption secret.
+    pub epoch: GroupEpoch,
+    /// The ratchet generation used for encryption. Pass it together with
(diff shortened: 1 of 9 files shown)


openmls/src/group/mls_group/application.rs
```diff
@@ -181,28 +185,99 @@ impl MlsGroup {
...
+    ///
+    /// Returns [`ConfirmMessageError::FutureEpoch`] when `epoch` is newer than
+    /// the group's current epoch.
+    ///
+    /// Only confirm once the DS has accepted exactly this message. After a lost
+    /// race against a sibling (the DS rejected the send because of a generation
+    /// collision), the secret must not be confirmed, since it is what decrypts
+    /// the sibling's winning message at the same generation.
+    #[cfg(feature = "virtual-clients-draft")]
+    pub fn confirm_message<Storage: StorageProvider>(
```
raphaelrobert: This is only application messages, right? Maybe the name of the function should reflect that, given that the other one is called confirm_handshake_message.

raphaelrobert (changes requested):
Looks good, so just a small nit

openmls/src/group/mls_group/application.rs
kkohbrok: The idea was to match the name of the `create_message` endpoint. But `confirm_application_message` works, too.

raphaelrobert (approved):

PR #2109, opened 2026-07-12
feat: early detection of ciphersuite support

raphaelrobert:
We want the crypto provider to be authoritative about which ciphersuites are supported in the future. This PR starts by introducing early detection of ciphersuite support with the existing functions the crypto provider already exposes.
We now fail earlier than before, in case a ciphersuite is not supported, and the error messages are harmonized.
This PR introduces breaking changes as far as error variants are concerned.


kkohbrok (approved):
👍🏻

PR #2113, opened 2026-07-13
feat: add key package batch builder for virtual clients

boxdot:
The newly added `VcKeyPackageBatchBuilder` allows to build heterogeneous batches of key packages. Build the batch on `finalize`. Dropping the builder without calling `finalize` burns no generation.


kkohbrok (approved):
Looks good, thanks! Could you add a small test?

boxdot:
> Looks good, thanks! Could you add a small test?

There is a test already testing `build_vc_batch` which is effectively implemented using the new builder. Should we decide to remove `build_vc_batch` and only provide the builder we will have to adjust the test to use the builder anyways.

PR #2114, opened 2026-07-14
fix(vc): use emulation ciphersuite for init key secret expansion

boxdot:
Use the ciphersuite as specified in the draft for the VC derivation info secret expansions. This bug was not caught by the tests, because the ciphersuite of the emulation group in the tests had the same KDF hash as the higher-level group's ciphersuite, so the expansions were silently compatible.


boxdot:
The spec was amended and introduces a new concept to transfer derived secrets from the domain of the emulation group ciphersuite into the domain of the target ciphersuite, which is much cleaner now: https://github.com/mlswg/mls-virtual-clients/pull/29. The new operation makes this PR obsolete. It will be implemented in a different PR.

PR #2117, opened 2026-07-15
Bump dependencies

franziskuskiefer:
- tls_codec to 0.5
- libcrux and hpke to latest releases
- rand to 0.10 for libcrux
- use hmac-drbg in libcrux provider


franziskuskiefer:
wasm size is too large again 😔

boxdot (approved):

boxdot:
> wasm size is too large again 😔

Only 25 KB

Issue #2104, opened 2026-07-09
PQ ciphersuites draft: Ciphersuites use SHA256 instead of SHAKE256


kkohbrok:
Our current implementations of the ciphersuites specified in the [PQ ciphersuites draft](https://datatracker.ietf.org/doc/draft-ietf-mls-pq-ciphersuites/) are not implemented correctly. The draft says they should use SHAKE256, while we're currently using SHA256. 

Note that the specification is still in flux, so we may want to wait until things have settled: https://github.com/mlswg/mls-pq-ciphersuites/pull/21

franziskuskiefer:
I have a draft for the pq ciphersuites in HPKE that we should be able to merge soon. It'll be behind a feature flag. But that will allow us to start playing with it.

niyamvora:
Looks like this settled: draft-ietf-mls-pq-ciphersuites-06 (July 2026) explicitly chose HKDF — "instead of one-stage KDFs like SHAKE… the cipher suites defined here use the HMAC-based KDF (HKDF) algorithms HKDF-SHA384 and HKDF-SHA256" — and mlswg/mls-pq-ciphersuites#21 was closed without merging. So the current OpenMLS implementation matches the draft, and this can probably be closed.

kkohbrok:
Thanks for the reminder.

PR #2118, opened 2026-07-15
feat: update PQ ciphersuites (draft v5)

boxdot:
The PQ ciphersuites [draft](https://datatracker.ietf.org/doc/draft-ietf-mls-pq-ciphersuites/) added more ciphersuites. It also now supports the previously custom Air ciphersuite.


raphaelrobert (approved):
I think this looks good (modulo the changes we need to make for #2104 later)

PR #2125, opened 2026-07-17
Enforce app_data_dictionary immutability for bare GroupContextExtensions proposals

tylerhawkes:
## What

`validate_app_data_update_proposals_and_group_context` enforces
[draft-ietf-mls-extensions §4.7-6](https://datatracker.ietf.org/doc/html/draft-ietf-mls-extensions#section-4.7-6):
once a group's `required_capabilities` include the `AppDataUpdate` proposal type,
a `GroupContextExtensions` proposal MUST NOT add, remove, or modify the
`app_data_dictionary` extension.

Today that check sits **after** an early return:

```rust
let no_app_data_updates = proposal_queue.app_data_update_proposals().next().is_none();
if no_app_data_updates {
    return Ok(());
}
// ... CannotUpdateDictionaryDirectly check ...
```

So the rule is only enforced when the commit *also* carries at least one
`AppDataUpdate` proposal. A **bare** `GroupContextExtensions` proposal — one with
no accompanying `AppDataUpdate` proposals — skips the check entirely and can
rewrite `app_data_dictionary` directly, bypassing the `AppDataUpdate` machinery
(and any application-level validation layered on it). On a group whose members
all support `AppDataUpdate`, any member can thereby overwrite arbitrary
components of the dictionary.

## Fix

Hoist the dictionary-immutability check **above** the early return, so it binds
every commit that carries a `GroupContextExtensions` proposal regardless of
whether `AppDataUpdate` proposals accompany it.

The rule is treated as active when the group requires `AppDataUpdate` **before or
after** the commit:

- Reading only the *proposed* `required_capabilities` would let a sender evade
  the check by dropping `AppDataUpdate` from `required_capabilities` in the same
  proposal.
- Reading only the *current* `required_capabilities` would incorrectly flag the
  migrating commit that both introduces the dictionary and adds `AppDataUpdate`.

The ordering and duplicate-`AppDataUpdate` checks remain after the early return,
where they belong (they are only meaningful when `AppDataUpdate` proposals are
present).

## Tests

Adds `test_group_context_update_dictionary_without_app_data_update`: a bare
`GroupContextExtensions` proposal that touches the dictionary with no
accompanying `AppDataUpdate` proposal must be rejected with
`CannotUpdateDictionaryDirectly`. The three existing tests in
`app_data_update_proposal_validation.rs` (`update_dictionary`, `wrong_order`,
`after_deactivating`) are unaffected — `validate_group_context_extensions_proposal`
runs first, so the `ExtensionNotInRequiredCapabilities` path is preserved.

Runs under `--features extensions-draft,extensions-draft-test-dependencies`.

### DIFF
--- a/openmls/src/group/public_group/validation.rs
+++ b/openmls/src/group/public_group/validation.rs
@@ -654,11 +654,6 @@ impl PublicGroup {
         &self,
         proposal_queue: &ProposalQueue,
     ) -> Result<(), AppDataUpdateValidationError> {
-        let no_app_data_updates = proposal_queue.app_data_update_proposals().next().is_none();
-        if no_app_data_updates {
-            return Ok(());
-        }
-
         // retrieve the GroupContextExtensions proposal, if available
         let group_context_extension_proposal = proposal_queue
             .filtered_by_type(ProposalType::GroupContextExtensions)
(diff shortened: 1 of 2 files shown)


keks (approved):
Oh, good catch. Thanks!

franziskuskiefer:
@tylerhawkes did you mean to close this? @keks I guess we should still merge this here?

keks:
Yeah, let's merge this

PR #2127, opened 2026-07-20
Interop update and PSK fixes

franziskuskiefer:
With the help from Claude I fixed up the interop runner and a few PSK issues.

This is on top of #1913 for sub-group branching, which is tested here as part of the interop already. ✅ 

After #1913 got merged we can get this one reviewed.
This may already address, or at least interact with #2162.

See commit messages for a full list of changes.

**Commit 1: fix PSK and GREASE handling**
Core changes where PSK handling was incomplete and we didn't correctly check grease support.

**Commit 2: Interop updates**
Everything is documented in the readme and the commit message.
- Focus on docker runs with a new script (`docker/run-interop.py`)
- Fixup MLS++ image and ignore RFC non-conformance of MLS++
- Add mls-rs as interop target. This ignores an RFC non-conformance issue as well.

**Commit 3: CI**
A CI job that can be triggered manually to run interop. We may want to have a lightweight run on PRs again at some point. But we should have at least _a_ way to run this on CI. Note that we can't run this until its merged unfortunately.


kkohbrok (approved):
Looks good, thanks! Could you add a changelog entry?

PR #2128, opened 2026-07-20
`serde_json` migration bridge for the storage provider

wysiwys:
This PR proposes tooling and an approach for migrating the entire storage provider to a self-describing `serde` serialization. The use of non-self-describing formats for `openmls` storage can lead to breakage or invalid deserializations, when upgrading between `openmls` versions; to address this, this PR also proposes that `openmls` explicitly not support non-self-describing serializations after `v0.9.0`. 

In this approach, `serde_json` serialization can be used as a bridge to migrate existing group data (and other data) in storage. The data is first loaded into memory using the original `openmls` version used for storage, then serialized into `serde_json` and deserialized into the new `openmls` version. This `serde_json` serialization is tolerant of layout changes, in particular fields added in newer `openmls` releases and annotated with `#[serde(default)]` that were not present in older ones. 

The migration utilities (in particular the `GroupMigrationBundle`) are available under the `migration-import` feature, and can be used to migrate from older versions of `openmls`, where changes to serialized structs may lead to breakage when using non-self-describing `serde` serializations (see #2021):
- [ ] v0.7.0
- [x] v0.7.4 (using the branch [openmls-v0.7.4-migration-helpers](https://github.com/openmls/openmls/compare/openmls-v0.7.4...openmls-v0.7.4-migration-helpers))
- [x] v0.8.1 (using the branch [openmls-v0.8.1-migration-helpers](https://github.com/openmls/openmls/compare/openmls-v0.8.1...openmls-v0.8.1-migration-helpers))

Each migration helpers branch is based on the release for each version, and includes a new feature, `migration-export`, that allows exporting a `GroupMigrationBundle` for each group.

This PR includes a book page that contains instructions and some initial recommendations for performing the migration.

Also see:
- https://github.com/openmls/openmls/pull/2076

Questions and TODOs:
- [ ] Determine whether to provide an API to wrap the `serde_json` bridge (and maybe also perform some cleanup automatically)
- [ ] Determine if only some `serde` serialization formats should be explicitly supported
- [ ] On the recommendations in the book page for the migration: determine any other requirements or recommendations


wysiwys:
@tylerhawkes I wanted to ask your thoughts if you'd have some time - would this kind of approach (for migrating the whole storage provider to a self-describing format) would be useful for you/would work well for your use case? 

To address the existing breakage that's occurred using non-self-describing serializations, we have already fixed the other issues that are addressable without explicitly migrating stored data. This PR (and #2076) go further and add migration helpers to address the remaining issues (in particular those data items that have added unsupported `#[serde(default)]` fields). In the future, to avoid issues like those that have arisen with non-self-describing serializations, OpenMLS will likely only support self-describing serializations. This PR in particular addresses that case.

#2076, on the other hand, is focused migrating specific items (in particular the `MessageSecretsStore` and the `MlsGroupState`) from a non-self-describing serialization into the same one after a version upgrade, instead of migrating the entire storage provider to a self-describing serialization. I think the approach in #2076 could be more useful as a stopgap, but the approach in this PR (or a similar migration) would ultimately be needed, if only self-describing serializations are supported in the future.

tylerhawkes:
@wysiwys Thanks! Yes this approach works for us. We'd use it. Three clarifications though:

  1. Don't hardcode JSON as the supported end-state. The actual requirement is a tolerance contract: structs as named-field maps, externally-tagged enums, unknown/missing-field tolerance. We'd want a binary self-describing format (e.g. CBOR via ciborium) for the destination store. Openmls values are mostly byte vectors, which serde_json inflates ~3–4× and we deserialize group state on every message processed. Spec the contract and test against at least one binary format alongside JSON.
  2. Make migration-export a portable patch. Our production data was written by a fork (pinned rev), so the prebuilt `*-migration-helpers` branches don't match our persisted layout. We'd need to apply the export side onto our exact rev so ideally it's a small, self-contained commit designed for that.
  3. Support lazy per-group migration. The bundle is per-group and idempotent, so a per-group marker with migrate-on-first-touch (under a group lock) should be valid, not just migrate-all-at-startup. We have clients with hundreds of groups and a startup stall isn't viable. It would be good to acknowledge this in the book page.

openmls/src/group/mls_group/mod.rs
```diff
@@ -1026,6 +1039,53 @@ impl MlsGroup {
         Ok(())
     }
 
+    #[cfg(feature = "migration-import")]
```
franziskuskiefer: Let's move this into a separate `impl` block to better separate these.

openmls/src/group/mls_group/mod.rs
```diff
@@ -1350,3 +1410,50 @@ impl WelcomeKeyMaterial {
         }
     }
 }
+
+/// A group and all of the group-associated data it owns, deserialized from a
+/// previous OpenMLS version for migration. The mirror of the older version's
+/// `MlsGroup::export_for_migration`.
+#[cfg(feature = "migration-import")]
```
franziskuskiefer: Let's put all of these into a separate module for separation.

openmls/src/group/public_group/mod.rs
```diff
@@ -499,6 +501,34 @@ impl PublicGroup {
         Ok(())
     }
 
+    #[cfg(feature = "migration-import")]
```
franziskuskiefer: Same here, let's use a separate `impl`.

compat_tests/tests/test_migration.rs
```diff
@@ -0,0 +1,1439 @@
...
+// ANCHOR_END: serde_json_bridge
+
+// ANCHOR: migration
+/// Migrate a single group from the previous OpenMLS version to the current one.
+///
+/// `old_provider` implements the *previous* version's storage traits and already
+/// holds the group; `new_provider` implements the *current* version's storage
+/// traits and receives the migrated group. Both refer to the same `group_id`.
+///
+/// This requires the `migration-export` feature on the previous version's
```
franziskuskiefer, line 499: 🤔 I don't see a `migration-export` feature anywhere. Is this only on the branches?

franziskuskiefer (commented):
I think this generally looks good. Some early comments.

compat_tests/tests/test_migration.rs
wysiwys: For now, yes, since there are no supported migration paths yet starting from the current version; a note is now included about this in this doc comment

wysiwys:
@tylerhawkes Great! On your points:
1. On `ciborium`/CBOR support: the book page now also includes an example for migration to a CBOR-based target storage provider. The migration tests now also check this approach's support for a `ciborium` target.
2. For the patch for the source version: sure, and could you also let us know the exact rev that you use, that the patch should be against? The changes needed for a patch may have some minor differences between `openmls` versions
3. For per-group migration: the book page and tests also reflect the case now where lazy per-group migration is used, rather than a full migration at startup, and the book pages includes a minimal example (using a separate table to mark group migration state). Could you let us know if this matches up with your use case?

book/src/user_manual/migration.md
```diff
@@ -0,0 +1,294 @@
...
+This flow is performed once per group, observing the invariants in
+[Migration requirements](#migration-requirements); afterwards the group loads normally
+with the current-version `MlsGroup::load`.
+
+## What is not migrated
+
+The migration bundle carries the group and all group-associated data OpenMLS owns
+— group state, queued proposals, a pending commit if stored, and the group's encryption key
+pairs. Application-managed material that OpenMLS does *not* own — signature key
+pairs, PSKs, and key packages — is not group-scoped, and is not part of
```
wysiwys: If the application already holds the [identifier](https://www.rfc-editor.org/rfc/rfc9420.html#section-8.4-6) (in particular the  [`Psk`](https://latest.openmls.tech/doc/openmls/schedule/psk/enum.Psk.html) struct) for each PSK, each PSK could be migrated one-by-one by iterating over this list. Otherwise, the list of identifiers could be retrieved directly from the storage provider, instead. This may require the `Psk` identifier struct to implement the [`Entity`](https://docs.rs/openmls_traits/latest/openmls_traits/storage/trait.Entity.html) trait, to allow loading it from storage

tylerhawkes:
@wysiwys That all sounds great. We just updated to the latest main a couple of days ago, so we can handle any patches to the patch. It should be relatively minor, so no need to do anything special for us.

book/src/user_manual/migration.md
```diff
@@ -0,0 +1,294 @@
...
+
+The storage migration is a re-encoding of a member's own stored, local state:
+it doesn't change anything on the wire, and doesn't need coordination with other group members.
+
+This migration approach requires the `migration-export` feature on the previous version's `openmls`
+crate and the `migration-import` feature on the current one.
+
+## When to use this
+
+The main migration use case is switching to a self-describing `serde` storage format,
```
franziskuskiefer, line 16: Add a half sentence that openmls didn't define so far what type of storage has been supported and which haven't.

book/src/user_manual/migration.md
```diff
@@ -0,0 +1,294 @@
...
+between `openmls` versions that introduce breaking changes when using non-self-describing formats.
+
+The approach described here can be used to migrate data serialized using `0.7.0`, `0.7.4` or `0.8.1`
+into the format used by `0.9.0`, which requires a self-describing format. 
+
+**NOTE**: As of this version, non-self-describing formats are no longer supported.
```
franziskuskiefer: mdbook supports `> [!NOTE]`. Lets use that.

book/src/user_manual/migration.md
```diff
@@ -0,0 +1,294 @@
...
+- **The two stores coexist for the whole support window.** You cannot discard the
+  old store promptly or swap it out, because un-accessed groups still live there.
+  Clean up each group's old data after it has been migrated (via the previous
+  version's `MlsGroup::delete`, see [Cleaning up the old data](#cleaning-up-the-old-data))
+  . Track how many groups remain
+  un-migrated, so you know when it is safe to retire the old-version code.
+- **Rollback is per-group.** The “rollback only before first use” rule applies to
+  each group independently: once a lazily-migrated group has been used, its old copy
+  is stale and must not be reverted to.
+<!-- TODO: determine guidelines on binary size -->
```
franziskuskiefer: Binary size is an issue for many mobile apps. So if you want to make it complete, you can mention that there is an impact (there's more openmls, so it's larger).

openmls/src/group/public_group/mod.rs
```diff
@@ -67,6 +67,8 @@ mod validation;
 
 /// This struct holds all public values of an MLS group.
 #[derive(Debug)]
+#[cfg_attr(feature = "migration-import", derive(serde::Deserialize))]
+#[cfg_attr(feature = "migration-test-utils", derive(serde::Serialize))]
```
franziskuskiefer: Do we really need a new feature here? Couldn't we use `test-utils`?

openmls/src/group/public_group/mod.rs
```diff
@@ -585,3 +587,38 @@ impl PublicGroup {
         )
     }
 }
+
+mod migration_import {
+
+    use super::*;
+
+    impl PublicGroup {
+        #[cfg(feature = "migration-import")]
```
franziskuskiefer: Shouldn't this feature go onto the module?

openmls/src/group/public_group/mod.rs
```diff
@@ -585,3 +587,38 @@ impl PublicGroup {
         )
     }
 }
+
+mod migration_import {
```
franziskuskiefer: Let's put the module into it's own file.

openmls/src/group/mls_group/mod.rs
```diff
@@ -1350,3 +1363,103 @@ impl WelcomeKeyMaterial {
         }
     }
 }
+
+#[cfg(feature = "migration-import")]
+pub(crate) mod migration_import {
```
franziskuskiefer: Let's put this in its own file.

compat_tests/tests/test_migration.rs
```diff
@@ -0,0 +1,2033 @@
...
+    let provider = T::openmls_provider(&target);
+
+    // The external PSK is application-managed: it is not part of the group migration
+    // bundle, so committing the migrated PreSharedKey proposal would fail until the
+    // application re-stores the secret it tracks into the new store.
+    let psk_id = openmls_current::schedule::PreSharedKeyId::external(
+        b"migration-test-psk".to_vec(),
+        vec![0u8; 32],
+    );
+    migrate_psk(&psk_id, &[7u8; 32], &provider);
```
wysiwys, line 1819: Would it be possible for this example to also cover the case where a previous, application-managed `PreSharedKeyId` is converted, instead of creating a new one? Some applications may also only store the PSK itself in the storage provider. To cover this case, is there a way to illustrate loading it and then re-storing it in the new provider?

franziskuskiefer (approved):
Let's get this in.

book/src/user_manual/migration.md
```diff
@@ -0,0 +1,310 @@
...
+
+This migration approach requires the `migration-export` feature on the previous version's `openmls`
+crate and the `migration-import` feature on the current one.
+
+## When to use this
+
+The main migration use case is switching to a self-describing `serde` storage format,
+between `openmls` versions that introduce breaking changes when using non-self-describing formats.
+
+The approach described here can be used to migrate data serialized using `0.7.0`, `0.7.4` or `0.8.1`
```
kkohbrok: Does this support migration from `0.7.0`? I only see helpers for `0.7.4`. Are those identical?

openmls/Cargo.toml
```diff
@@ -47,8 +47,12 @@ zeroize = "1.8.1"
 serde_bytes = "0.11.17"
 
 [features]
+default = ["migration-import"]
```
kkohbrok: I don't think this should be a default feature.

kkohbrok (commented):
Looks good generally, except that I'd prefer for this not to be a default feature.

openmls/Cargo.toml
franziskuskiefer: 🙈

kkohbrok (approved):
Thanks!

keks (approved):
Also looks good to me :+1:

PR #2129, opened 2026-07-21
Implement group reinitialization (ReInit)

franziskuskiefer:
Add full support for the ReInit proposal and reinitialization flow per RFC 9420 §11.2 / §12.1.5.
This makes the interop matrix complete. It's a draft for now until we have the other PRs merged.

This is on top of
* #2127 and
* #1913.

Proposals & senders:
- Make `Proposal::re_init` public and add `ReInitProposal::new` plus accessors (group_id/version/ciphersuite/extensions).
- `MlsGroup::propose_reinit` for member-initiated ReInit proposals and `ExternalProposal::new_reinit` for external senders (§12.1.8.2); process incoming external ReInit proposals in `process_message`.

Commit & suspension:
- Validate ReInit commits: a ReInit must be the only proposal in the commit (valn0309) and must not downgrade the protocol version (valn0901).
- Suspend the group on merging a ReInit commit (becomes Inactive), so it can only seed the successor group.
- `ProposalQueue::reinit_proposal` / `StagedCommit::reinit_proposal` and keep only the first ReInit when filtering.

Successor group:
- `CommitBuilder::reinit` adds the resumption PSK (usage Reinit) sampled at the successor ciphersuite's KDF.Nh and injects the old group's resumption secret at the old (reinit) epoch.
- `StagedWelcome::new_from_reinit` validates the successor's parameters against the ReInit proposal, enforces epoch 1, and optionally checks that every member matches a member of the old group.
- Generalize resumption-PSK injection to (epoch, secret) and allow Reinit alongside Branch in the epoch-0 PSK check.

Also export PSK types from the prelude, add the reinit RPC to the interop client, and add tests, KATs, and a book chapter.


niyamvora:
We run a production MLS deployment (our app, on OpenMLS, moving to 0.9.0 and the PQ ciphersuites) and will need ReInit to move live groups from the classical suite to `MLKEM768X25519`. If it helps land this, I can test the branch against real groups on real devices or add tests for the ciphersuite-change case — just say what would be most useful.

book/src/user_manual/reinit.md
```diff
@@ -0,0 +1,72 @@
...
+
+A member proposes the reinitialization with
+[`MlsGroup::propose_reinit`](https://docs.rs/openmls/latest/openmls/group/struct.MlsGroup.html),
+describing the successor group's parameters:
+
+```rust,no_run,noplayground
+{{#include ../../../openmls/tests/book_code.rs:reinit_propose}}
+```
+
+## Committing the ReInit (suspending the old group)
```
nhanff, line 31: "suspended" sounds like its own state on the group, but it is actually the same as `Inactive`. I feel like the book could be more clear about that, although it might be an implementation detail.

book/src/user_manual/reinit.md
```diff
@@ -0,0 +1,72 @@
...
+
+Every other member processes and merges the commit, which suspends their view of
+the group as well:
+
+```rust,no_run,noplayground
+{{#include ../../../openmls/tests/book_code.rs:reinit_process}}
+```
+
+The ReInit may also be committed *by value* by adding the proposal directly to
+the commit with `CommitBuilder::add_proposal(Proposal::re_init(..))` instead of
```
nhanff, line 49: Are there tests for this path?

openmls/src/group/mls_group/commit_builder.rs
nhanff, line 812:
```suggestion
        // Also validates branch PSK proposals: a resumption PSK of usage `Branch` or `Reinit`
        // is only accepted at epoch 0 (i.e. in the initial commit of a subgroup),
        // see `validate_pre_shared_key_proposals`.
```

interop_client/docker/test-runner/Dockerfile
nhanff:
```suggestion
# The entrypoint fails when starting with `docker compose` directly.
# This is intended, the container should be used with `docker compose run <args>`, such that `test-runner` is called with `<args>`.
```

interop_client/docker/test-runner/main.go.patch
nhanff, line 27: The runner is strange. It should only involve a client in a test if it supports the cipher_suite, but when running it without this patch I see openmls be used with other suites as well, failing instantly.

interop_client/docker/debug-scenario.py
nhanff, line 38: This might mislead users to use regexes. Could be renamed to `--filter` or similar.

interop_client/docker/run-interop.py
nhanff:
```suggestion
                       OpenMLS does not implement.
```

interop_client/docker/run-interop.py
```diff
@@ -25,7 +25,7 @@
 Usage:
   ./run-interop.py [--peers mls-rs,mlspp] [--configs a.json,b.json]
                    [--timeout SECONDS] [--suite N] [--mode both|public|private]
-                   [--no-build] [--no-readme] [--no-clean] [--exit-zero] [--verbose]
+                   [--no-build] [--no-readme] [--no-clean] [--exit-zero]
 
 Defaults: both peers, all config files in the test-runner image, all
 ciphersuites, both handshake modes, build images, 900s per-run timeout, update
```
nhanff, line 31:
```suggestion
ciphersuites OpenMLS supports, both handshake modes, build images, 900s per-run timeout, update
```

interop_client/docker/run-interop.py
nhanff, line 52:
```suggestion
Requires: docker, docker compose plugin, and Python 3.8+.
```

nhanff, line 19:
```suggestion
  * unsupported (🚫) — the error carries `code = Unimplemented`; a feature
                       OpenMLS does not implement.
```

nhanff, line 284:
```suggestion
#   also list in that leaf's `capabilities`. RFC 9420 §10 requires every
#   non-default extension type in a  `keypackage.extensions` to appear in the
#   leaf's `capabilities`, with no GREASE exemption; mls-rs and OpenMLS both
```

nhanff, line 281:
```suggestion
# https://github.com/cisco/mlspp/issues/470
# - MLS++ sprinkles GREASE values into `key_package.extensions` that it does not
```

nhanff, line 308:
```suggestion
    - (https://github.com/awslabs/mls-rs/issues/386) mls-rs reuses its ReInit join path for branch and asserts the subgroup's
```

nhanff, line 314:
```suggestion
      mismatch is never masked.
```

interop_client/docker/run-interop.py
```diff
@@ -473,6 +386,15 @@ def acc(config, col):
...
                     fe["genuine_sample"] = sample_of(r)
+        # second pass: per (col, script) totals for "n/N failed" context
+        for script, results in data.get("scripts", {}).items():
+            for r in results:
+                col = bucket(r.get("actors", {}), peer)
+                if col is None:
+                    continue
+                fe = fails[col].get((name, script))
+                if fe is not None:
+                    fe["total"] += 1
```
nhanff, line 397: This whole function is simultaneously very important to not miss failures, and very messy. I can't positively assert to have verified it, but I'm OK with moving simplification to a follow-up issue.

interop_client/README.md
nhanff, line 33:
```suggestion
Simply navigate to the `./docker` folder and run `docker compose up --build`. This will build the images, start the services, and run the test.
```

`up` alone only builds the first time, it does not rebuild on change.

nhanff, line 110:
```suggestion
Requires `docker`, `docker compose`, and Python 3.8+. It exits non-zero only when there are *genuine* interop failures (🚫 / 🔶 / not-completed
```

interop_client/docker/run-interop.py
nhanff, line 302: Follow-up issue / clean up interop runner:

🔶 indicates both non-conformances and over-strictness of peers. This semantic differentiation is both added and removed in this file, I think it makes sense to unify them, also simplifying some logic.

Also, the matchers for peer problems should receive the peer name (possibly all actor assignments) and ciphersuite so they can be most specific.
For example, we only want to mark a peer limitation if mls-rs plays a role that receives an openmls-created `branch` Welcome:

+ script matches `branch/with_extensions`
+ actors match `[openmls, *others]`; `mlsrs in others`
+ `step == 1`
+  `err` matches the specific string containing `ReInitExtensionsMismatch`.

interop_client/README.md
```diff
@@ -157,68 +157,39 @@ The per-script failures behind the table above, grouped by pairing, each with a
...
+`external_add` / `joiner_signed_add` sub-scripts fail 🔶 **only** against MLS++, from its GREASE
+non-conformance (see below); they pass against mls-rs and OpenMLS-self.
 
 ### Unimplemented (🚫, reported as gRPC `Unimplemented`, never counted as a failure)
 
-- **ReInit** — `re_init_proposal`, `re_init_commit`, `handle_*_re_init_*`, `re_init_welcome`, and the
-  `external_signer_proposal` `reinit` sub-type. ReInit is not implemented in OpenMLS.
+_None across the configs exercised above._ Both member-initiated and external-sender **ReInit** are now
+implemented (`re_init_proposal`, `re_init_commit`, `handle_*_re_init_*`, `re_init_welcome`, and the
+`external_signer_proposal` `reinit` sub-type).
```
nhanff, line 192:
```suggestion
```

LLM describing what it did; not useful for documentation.

interop_client/README.md
nhanff, line 213:
```suggestion
**MLS++ GREASE non-conformance**: [MLS++ issue](https://github.com/cisco/mlspp/issues/470)
MLS++ stamps random GREASE values into `key_package.extensions` without also
listing them in that leaf's `capabilities`. RFC 9420 §10 requires every non-default extension in a KeyPakage's
`extensions` to appear in `leaf_node.capabilities` (no GREASE exemption).
The interop client cannot fix a peer-minted key package, so this error
(`A key package extension is not supported in the leaf's capabilities`) is excluded.

**mls-rs supports no extension change on branch**: [mls-rs issue](https://github.com/awslabs/mls-rs/issues/386)
On `branch / with_extensions`, mls-rs rejects with `ReInitExtensionsMismatch`
when OpenMLS creates the subgroup and mls-rs joins it.
In the same issue, mls-rs also ignores the `extensions` field of `branch` actions.
```

Shorter, more specific and correct, references issues for tracking.

nhanff, line 217: (Review note to self: Are there tests showcasing that? Does openMLS expose an ergonomic interface for the application to check?)

interop_client/src/main.rs
```diff
@@ -120,51 +141,114 @@ impl MlsClientImpl {
...
+        // Advertise support for the successor ciphersuite (plus the common
+        // interop suites) so the successor group accepts this leaf.
+        let mut ciphersuites = vec![
+            Ciphersuite::MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519,
+            Ciphersuite::MLS_128_DHKEMP256_AES128GCM_SHA256_P256,
+            Ciphersuite::MLS_128_DHKEMX25519_CHACHA20POLY1305_SHA256_Ed25519,
+        ];
+        if !ciphersuites.contains(&successor_ciphersuite) {
+            ciphersuites.push(successor_ciphersuite);
+        }
```
nhanff, line 209: Is this really required? The test runner should only switch to a cipher suite that is supported by all clients.

https://github.com/mlswg/mls-implementations/blob/cfd450286d1bfd9cd2519b95c80f9771f94a5b1a/interop/test-runner/main.go#L509-L544

interop_client/src/main.rs
```diff
@@ -120,51 +141,114 @@ impl MlsClientImpl {
...
+        };
 
-        Ok((key_package, credential, signature_keys, crypto_provider))
+        // Advertise support for the successor ciphersuite (plus the common
+        // interop suites) so the successor group accepts this leaf.
+        let mut ciphersuites = vec![
+            Ciphersuite::MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519,
+            Ciphersuite::MLS_128_DHKEMP256_AES128GCM_SHA256_P256,
+            Ciphersuite::MLS_128_DHKEMX25519_CHACHA20POLY1305_SHA256_Ed25519,
+        ];
```
nhanff, line 206: TODO: Get suites from provider

interop_client/src/main.rs
nhanff, line 1242: Aren't external commits with proposal references illegal? (https://www.rfc-editor.org/rfc/rfc9420.html#section-12.4.3.2-11.2)

This should be rejected by the group instead of the interop client, but this comment confuses me.

nhanff, line 1331: I guess this is supposed to be about what the field should be named on the proposal? If so, the interop runner protobuf calls it `epoch_id`.

nhanff, line 1760: Remove outdated comment

nhanff, line 2160: We don't comment the exact feature on other methods, probably an LLM artifact here.

nhanff, line 2455: I'm confused: Is this method for creating an Add proposal or a join via external commit?

openmls/src/extensions/mod.rs
nhanff, line 5: Follow-up: 
```suggestion
//! - In [`KeyPackages`](`crate::key_packages`), and [`LeafNode`](`crate::treesync::node::leaf_node`) to describe client capabilities
```

openmls/src/extensions/mod.rs
```diff
@@ -22,7 +22,6 @@
 //! - [`ExternalPubExtension`] (GroupInfo extension)
```
nhanff, line 22: I think ExternalSenders is also supported by now, at least it is defined in `ExtensionType`.

openmls/src/extensions/mod.rs
nhanff, line 163: This method returns `bool`, no `Option`. Stale comment?

openmls/src/extensions/mod.rs
```diff
@@ -173,8 +172,8 @@ impl ExtensionType {
             | ExtensionType::ExternalSenders => false,
             // GREASE may appear as an extension type in `leaf_node.extensions`
             // and must be tolerated there. It is still subject to the normal rule
-            // that it must be declared in `capabilities` (checked separately),
-            // so this only permits the type, it does not exempt it from that check.
+            // that it be declared in `capabilities` (checked separately), so this
+            // only permits the type, it does not exempt it from that check.
             ExtensionType::Grease(_) | ExtensionType::Unknown(_) | ExtensionType::ApplicationId => {
```
nhanff, line 177: The point of GREASE, as I understand it, is injecting unknown values to ensure clients ignore them. The reason a dedicated GREASE section exists in the extension space is to prevent  allocating them for real extensions. In that sense, treating GREASE differently from UNKNOWN is against the point.

[RFC 9420, Section 13.5](https://www.rfc-editor.org/rfc/rfc9420.html#section-13.5-8):

> Clients MUST NOT implement any special processing rules for how to handle these values when receiving them, since this negates their utility for detecting extensibility failures.

> GREASE values MUST be handled using normal logic for processing unsupported values.

[RFC 8701 Section 3.2](https://www.rfc-editor.org/info/rfc8701/#section-3.2-1):

> When processing a ClientHello, servers MUST NOT treat GREASE values differently from any unknown value.

4.2:

> When processing a CertificateRequest or NewSessionTicket, clients MUST NOT treat GREASE values differently from any unknown value.

Same for all methods in this `impl`.

The failure case is that we accidentally allow GREASE but not UNKNOWN in some context, defeating the purpose.

I believe the correct way to implement this is to *not* declare `ExtensionType::Grease`. No code apart from tests and comments should even mention GREASE unless it injects it, in particular no validation code. The GREASE values is only needed for constructing new greased values.

openmls/src/extensions/mod.rs
nhanff, line 840: Is there GREASE-aware validation (of incoming values) in the RFC? That would contradict 13.5 as noted above.

nhanff, line 919: Is there GREASE-aware validation? If so, it is an anti-feature or conflicting spec with 13.5

openmls/src/schedule/psk.rs
```diff
@@ -376,16 +376,12 @@ impl PreSharedKeyId {
...
                         // We can't check anything in here since we need more
-                        // information about the commit. We do this check
-                        // on the outside.
+                        // information about the commit. Reinit and Branch PSKs
+                        // are only valid in the initial commit of the successor
+                        // resp. sub-group (epoch 0); this is checked on the
+                        // outside in `validate_pre_shared_key_proposals`.
                     }
                 }
             }
```
nhanff, line 387: Now this whole match statement doesn't do anything anymore.
Which is not surprising as valn0801/0802 need at least the commit context. I'd suggest removing it to avoid creating the impressions that it validates anything other than valn0803.

openmls/src/schedule/psk.rs
nhanff, line 617: Is storing the PSK in epoch 0 a hacky workaround? I see potential for confusion as the PSK in the Welcome has group ID and epoch ID pointing to the old group.

openmls/src/group/mls_group/proposal_store.rs
nhanff, line 534: Stale comment, this is now 12.3 and the operations changed.

nhanff, line 550: This is redundant with the quote from the RFC (if it were up-to-date) and the implementation itself.

openmls/src/group/mls_group/proposal_store.rs
```diff
@@ -606,7 +620,19 @@ impl ProposalQueue {
...
+                    // sufficient here.
+                    // TODO: like the other arms here, this silently drops
+                    // additional (here: duplicate ReInit) proposals rather than
+                    // surfacing an error. Silently dropping proposals hides
+                    // malformed commits from the caller; this filtering should
+                    // eventually return an error instead of quietly discarding.
+                    if !contains_reinit {
+                        valid_proposals.add(queued_proposal.proposal_reference());
+                        contains_reinit = true;
+                    }
```
nhanff, line 635: valn0309

> It contains a ReInit proposal together with any other proposal. If the committer has received other proposals during the epoch, they SHOULD prefer them over the ReInit proposal, allowing the ReInit to be resent and applied in a subsequent epoch.

Currently, the reinit would be added regardless of others. That should fail validation.
Only adding Reinit if no other proposals exist might on the other hand block Reinit indefinitely if the  application is high-latency and proposal-heavy.

openmls/src/group/mls_group/commit_builder.rs
nhanff, line 786: We should add `validate_reinit_proposals` here to catch valn0309 (ReInit is not alone) before sending the commit.

In general, there is lots of duplication between `CommitBuilder::build_internal` and `PublicGroup::validate_commit`.

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -359,37 +335,31 @@ impl<'a> CommitBuilder<'a, Initial, &mut MlsGroup> {
         self
     }
 
-    /// Branches from a parent group into this (freshly created) group to form a
+    /// Branches from `parent` into this (freshly created) group to form a
```
nhanff, line 338: We should check that the group is freshly created, or only allow this step in a group builder.

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -407,6 +377,57 @@ impl<'a> CommitBuilder<'a, Initial, &mut MlsGroup> {
         Ok(self)
     }
 
+    /// Reinitializes `old_group` into this (freshly created) successor group, as
```
nhanff, line 380: Again assert in code that the group is freshly created, to produce early error instead of waiting for commit validation.

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -359,37 +335,31 @@ impl<'a> CommitBuilder<'a, Initial, &mut MlsGroup> {
...
-    /// The parent group's parameters are provided via `branch_info`, which the
-    /// parent exports with
-    /// [`MlsGroup::branch_info`](crate::group::MlsGroup::branch_info).
-    ///
     /// This adds a resumption [`PreSharedKeyId`] of usage `Branch` to the initial
     /// commit, with a freshly sampled `psk_nonce` of length KDF.Nh, and injects
     /// the parent group's resumption PSK secret so it is mixed into this
     /// subgroup's key schedule.
     ///
+    /// This must be called on a new group created with the same parameters (i.e.
```
nhanff, line 346: This should also be checked early to prevent creating invalid proposals.

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -348,8 +325,7 @@ impl<'a> CommitBuilder<'a, Initial, &mut MlsGroup> {
     /// list of proposals to be committed.
     ///
     /// Note that this should not be used for sub-group branching, as those PSKs
-    /// are not allowed in regular proposals. Please use
-    /// [`MlsGroupBuilder::branch`](crate::group::MlsGroupBuilder::branch) instead.
+    /// are not allowed in regular proposals. Please use [`Self::branch()`] instead.
```
nhanff, line 328: What should it be used for positively? Presumably external PskIds? If so, state explicitly.

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -398,7 +368,7 @@ impl<'a> CommitBuilder<'a, Initial, &mut MlsGroup> {
         // The branch PSK secret comes from a different group, so we clear this
         // group's resumption PSK store and inject it at the sentinel epoch 0,
         // where `load_psks` looks it up for branch usage.
-        let secret = branch_info.resumption_psk_secret().clone();
+        let secret = parent.resumption_psk_secret().clone();
         self.group.borrow_mut().resumption_psk_store.clear();
```
nhanff, line 372: For my understanding: The resumption psk store should be empty anyway in the fresh group, right?

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -407,6 +377,57 @@ impl<'a> CommitBuilder<'a, Initial, &mut MlsGroup> {
...
+                ResumptionPskUsage::Reinit,
+                old_group.group_id().clone(),
+                old_group.epoch(),
+            )),
+        )
+        .map_err(LibraryError::unexpected_crypto_error)?;
+        self = self.propose_psks([psk_id]);
+
+        // The reinit PSK secret comes from the old group. Reinit PSKs are looked
+        // up by their own epoch (the old group's final, reinit epoch), so we
```
nhanff, line 418: Now this is definitely hacky. The PskId includes the old group ID, so this expects the lookup to ignore it.

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -407,6 +377,57 @@ impl<'a> CommitBuilder<'a, Initial, &mut MlsGroup> {
...
 
+    /// Reinitializes `old_group` into this (freshly created) successor group, as
+    /// described in [RFC 9420 §11.2].
+    ///
+    /// `old_group` must have merged a commit containing a ReInit proposal, so it
+    /// is suspended (inactive); see [`MlsGroup::propose_reinit`]. The caller is
+    /// responsible for creating this successor group with the parameters from
+    /// that ReInit proposal (group id, protocol version, ciphersuite and group
+    /// context extensions) and for not reusing a suspended group to seed more
+    /// than one successor.
```
nhanff, line 388: Same as above; check or enforce that the group is fresh.

openmls/src/group/mls_group/creation.rs
```diff
@@ -483,88 +587,145 @@ impl StagedWelcome {
...
-    ///
-    /// If the receiver does not yet know which parent epoch the branch was taken
-    /// from (its own view of the parent group may have advanced), use
-    /// [`StagedWelcome::process_branch_welcome`] to read the parent reference
-    /// from the `Welcome` first and then pick the matching `branch_info`. That
-    /// path decrypts the `Welcome` only once. This one-shot method is a
-    /// convenience for callers that already know the parent epoch; it also
-    /// decrypts only once, via the same carrier.
+    /// credential equality for equivalent identifiers when `check_members` is
+    /// `true`.
```
nhanff, line 604: Add a doc snippet on how to check that in the application.

Probably comparing the returned `staged_welcome.members().map(Member::credential)` with the parent group's `parent.members()`

openmls/src/group/mls_group/creation.rs
```diff
@@ -483,88 +587,145 @@ impl StagedWelcome {
...
+        provider: &Provider,
         mls_group_config: &MlsGroupJoinConfig,
         welcome: Welcome,
-        branch_info: BranchInfo,
-    ) -> Result<JoinBuilder<'a, Provider>, WelcomeError<Provider::StorageError>> {
-        Self::process_branch_welcome(provider, mls_group_config, welcome)?
-            .build_from_branch(provider, branch_info)
+        ratchet_tree: Option<RatchetTreeIn>,
+        parent: &MlsGroup,
+        check_members: bool,
```
nhanff, line 613: Maybe rename to `check_members_by_credential` for better clarity.

openmls/src/group/mls_group/creation.rs
```diff
@@ -483,88 +587,145 @@ impl StagedWelcome {
...
     }
 
-    /// Decrypt a subgroup-branch `Welcome` once so its parent reference can be
-    /// inspected before selecting the matching [`BranchInfo`].
+    /// Creates a [`StagedWelcome`] for a group reinitialized from `old_group`, as
+    /// described in [RFC 9420 §11.2].
+    ///
+    /// `old_group` must have merged the commit containing `reinit_proposal` (so
+    /// it is suspended). The caller supplies that same `reinit_proposal` (it was
+    /// covered by the ReInit commit both members merged) so the successor's
```
nhanff, line 657: Who are `both` members?

openmls/src/group/mls_group/creation.rs
nhanff, line 604: Follow-up issue: Introduce a consistent doc comment section about "Application Responsibilities". It should come before any implementation details.

openmls/src/group/mls_group/creation.rs
```diff
@@ -483,88 +587,145 @@ impl StagedWelcome {
...
     ///
-    /// The branch PSK's parent group and epoch live inside the `Welcome`'s
-    /// encrypted `GroupSecrets`, so reading them requires decryption. This
-    /// returns a [`PendingBranchWelcome`] holding the decrypted state: call
-    /// [`PendingBranchWelcome::parent`] to read the parent `(group_id, epoch)`
-    /// the branch derives from (RFC 9420 §8.4), select the `BranchInfo` for that
-    /// parent epoch (e.g. from a sliding window of recent epochs), then finish
-    /// with [`PendingBranchWelcome::build_from_branch`] — without decrypting the
-    /// `Welcome` a second time.
+    /// * the successor's group id, protocol version, ciphersuite and extensions
```
nhanff, line 666: Reference valn1413, this implements it in part.

openmls/src/group/mls_group/creation.rs
```diff
@@ -483,88 +587,145 @@ impl StagedWelcome {
...
         provider: &Provider,
         mls_group_config: &MlsGroupJoinConfig,
         welcome: Welcome,
-    ) -> Result<PendingBranchWelcome, WelcomeError<Provider::StorageError>> {
-        let (resumption_psk_store, key_material, group_secrets) =
-            decrypt_group_secrets(provider, mls_group_config, &welcome)?;
+        ratchet_tree: Option<RatchetTreeIn>,
+        old_group: &MlsGroup,
+        reinit_proposal: &ReInitProposal,
+        check_members: bool,
```
nhanff, line 683: Same comments on `check_members` as for `new_from_branch`

openmls/src/group/mls_group/creation.rs
```diff
@@ -483,88 +587,145 @@ impl StagedWelcome {
...
+    ) -> Result<Self, WelcomeError<Provider::StorageError>> {
+        // The reinit PSK is looked up by the old group's (reinit) epoch.
+        let reinit_epoch = old_group.epoch();
```
nhanff, line 687: Check that `old_group` matches the `group_id` from the `reinit_proposal` to catch invalid calls early.

openmls/src/group/public_group/staged_commit.rs
```diff
@@ -260,6 +257,38 @@ impl PublicGroup {
...
+        if proposal_queue.queued_proposals().count() != 1 {
+            return Err(ProposalValidationError::ReInitProposalNotAlone);
+        }
+
+        // https://validation.openmls.tech/#valn0901
+        let version_number = |v: ProtocolVersion| match v {
+            ProtocolVersion::Mls10 => 1u16,
+            ProtocolVersion::Other(v) => v,
+        };
+        if version_number(reinit.version()) < version_number(self.group_context.protocol_version())
```
nhanff, line 284: We could simplify this using `PartialOrd` on `ProtocolVersion`.

nhanff:
Follow-up issue: Add viable validations to `ProposalIn::validate` for ReInit, PreSharedKey?
These proposals are also verified in other places, so not a direct issue, just unexpected.

https://github.com/openmls/openmls/blob/ce54fc100ab29180ebdac808183f117f92e0f6d4/openmls/src/messages/proposals_in.rs#L124-L125

openmls/src/group/mls_group/creation.rs
```diff
@@ -483,88 +587,145 @@ impl StagedWelcome {
...
+        ratchet_tree: Option<RatchetTreeIn>,
+        parent: &MlsGroup,
+        check_members: bool,
+    ) -> Result<Self, WelcomeError<Provider::StorageError>> {
+        let processed_welcome = ProcessedWelcome::new_from_welcome_inner(
+            provider,
+            mls_group_config,
+            welcome,
+            // Branch PSKs are injected at the sentinel epoch 0.
+            Some((0.into(), parent.resumption_psk_secret().clone())),
```
nhanff, line 620: Does this work when the parent advanced in the mean time?
The parent's epoch should be looked up from the PskId of the parent; The group_id should also be checked.

openmls/src/group/errors.rs
```diff
@@ -133,10 +128,18 @@ pub enum WelcomeError<StorageError> {
...
+    #[error("The successor group's parameters do not match the ReInit proposal.")]
+    ReInitParameterMismatch,
+    /// The successor group is not at epoch 1, as required for reinitialization
+    /// (RFC 9420 §11.2).
+    #[error("The successor group is not at epoch 1.")]
+    ReInitEpochInvalid,
+    /// A member of the successor group does not match any member of the old
+    /// group (RFC 9420 §11.2).
+    #[error("A member of the successor group does not match any member of the old group.")]
+    ReInitLeafMismatch,
```
nhanff, line 142: The RFC is ambiguous here: Is it required that the sets of members is equal, that is, every old member must also be a new member? We currently don't check this, but e.g. [mls-rs does](https://github.com/awslabs/mls-rs/blob/d4c5b19f4bbf8a931face27fc9e30f2db33ce83e/mls-rs/src/group/resumption.rs#L347-L351).

[11.2](https://www.rfc-editor.org/rfc/rfc9420.html#section-11.2-1):
> A group may be reinitialized by creating a new group with the same membership and different parameters, and linking it to the old group via a resumption PSK.

openmls/tests/book_code.rs
```diff
@@ -2034,3 +2034,202 @@ fn external_commit_builder() {
...
+    .unwrap()
+    .into_group(bob_provider)
+    .unwrap();
+    // ANCHOR_END: reinit_join
+
+    assert_eq!(
+        alice_successor.confirmation_tag(),
+        bob_successor.confirmation_tag()
+    );
+}
```
nhanff, line 2235: TODO: Add a similar test for subgroup branching.

openmls/src/messages/external_proposals.rs
```diff
@@ -2,7 +2,10 @@
 //!
 //! Contains the types and methods to build external proposal to add/remove a client from a MLS group
 //!
-//! `ReInit` is not yet implemented
+//! External `ReInit` proposals are supported via
+//! [`ExternalProposal::new_reinit`]. See
+//! [`MlsGroup::propose_reinit`](crate::group::MlsGroup::propose_reinit) for
+//! member-initiated ReInit proposals.
```
nhanff, line 8:
```suggestion
```

Update comment to reflect that context extensions, psk, and reinit are supported, but keep verbosity at the level of add/remove.

openmls/tests/reinit.rs
```diff
@@ -0,0 +1,316 @@
...
+fn reinit_by_value_same_ciphersuite() {
+    run_reinit_flow::<Provider>(ciphersuite, ciphersuite, true);
+}
+
+/// Happy path: reinit into a different ciphersuite (by reference), when an
+/// alternative ciphersuite with the same signature scheme is available.
+#[openmls_test]
+fn reinit_cross_ciphersuite() {
+    let provider = Provider::default();
+    let new_ciphersuite = alternate_ciphersuite(ciphersuite, &provider);
```
nhanff, line 314:
```suggestion
    let new_ciphersuite = alternate_ciphersuite(ciphersuite, &provider);
    assert_ne!(ciphersuite, new_ciphersuite);
```
I don't like to implicitly silence the case where no alternative was found. I'd rather the test returns early than re-test for the same ciphersuite.

openmls/tests/reinit.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    (
+        (alice_credential, alice_signer),
+        (bob_credential, bob_signer),
+        alice_group,
+        bob_group,
+    )
+}
+
+/// Pick a ciphersuite different from `ciphersuite` but with the same signature
```
nhanff, line 89: We should also test switching the signature scheme and replacing credentials.

openmls/tests/reinit.rs
```diff
@@ -0,0 +1,316 @@
...
+/// scheme (so the same credentials can be reused) that the provider supports.
+/// Falls back to `ciphersuite` if no such alternative is available.
+fn alternate_ciphersuite(ciphersuite: Ciphersuite, provider: &impl OpenMlsProvider) -> Ciphersuite {
+    const CANDIDATES: &[Ciphersuite] = &[
```
nhanff, line 93: Why this specific list, as opposed to `provider.crypto().supported_ciphersuites()`?

openmls/tests/reinit.rs
```diff
@@ -0,0 +1,316 @@
...
+
+use openmls::{prelude::*, test_utils::single_group_test_framework::generate_credential};
+use openmls_test::openmls_test;
+use openmls_traits::signatures::Signer;
+
+/// Set up a group with Alice and Bob, where Bob joins via the regular welcome
+/// path. Returns Alice's and Bob's credentials/signers and their two group
+/// views.
+#[allow(clippy::type_complexity)]
+fn setup_group(
```
nhanff, line 17: This is almost identical to `test_utils::setup_alice_bob_group`, which is not visible in this crate. Should we make it?

openmls/tests/sub-groups.rs
```diff
@@ -603,91 +326,19 @@ fn subgroup_branch_carrier_rejects_wrong_epoch() {
     let welcome: MlsMessageIn = welcome.into();
     let welcome = welcome.into_welcome().unwrap();
 
-    let pending = StagedWelcome::process_branch_welcome(
+    // Bob checks the subgroup against his view of the parent group and must
+    // reject it because Dave has no matching leaf there.
+    let result = StagedWelcome::new_from_branch(
```
nhanff, line 331: Not required by the RFC, I would have expected `CommitMessageBuilder::build` to also reject. This would however make testing more complicated.

openmls/tests/sub-groups.rs
```diff
@@ -603,91 +326,19 @@ fn subgroup_branch_carrier_rejects_wrong_epoch() {
...
-
-    assert_eq!(
-        pending.parent(),
-        None,
-        "a plain welcome has no branch parent reference"
+    assert!(
+        matches!(result, Err(WelcomeError::SubgroupLeafMismatch)),
+        "expected a subgroup with a non-parent member to be rejected, got {result:?}"
     );
 }
```
nhanff, line 344: Add a test where some of the clients process another commit in the parent group before processing the branched Welcome. This would likely fail currently because `StagedWelcome::new_from_branch` uses the current `resumption_psk_secret` of the parent instead of properly looking up the psk.

nhanff:
General remark: Applications that receive a Welcome for branch/reinit but have not yet applied the commit that introduces the parent/old epoch that is being referenced by the PSK must store the Welcome, wait for the commit, and then apply it.
We may want to document this and expose some ergonomic way to find out whether a Welcome is resuming, for which group and which epoch.

nhanff:
Closed in favor of #2231, where changes have been cherry-picked to.

PR #2130, opened 2026-07-22
Fix `openmls_serialization_helpers` enum deserialization for self-describing binary codecs

wysiwys:
This PR fixes the `Deserialize` impl generated for enums by `openmls_serialization_helpers`. Previously the generated code branched on `Deserializer::is_human_readable()` and, for non-human-readable codecs, expected the enum variant to be encoded as a numeric storage tag. That assumption holds for index-based binary codecs like `postcard`, but not for self-describing binary codecs such as `ciborium` (CBOR), where `is_human_readable() == false` yet enum variants are encoded by name, so deserialization previously failed.

The fix removes the `is_human_readable()` branch and instead deserializes the variant identifier on its own first, via `deserialize_identifier`, into a dedicated helper that accepts either an integer tag or a string name. A single visitor (replacing the `non_human_readable_visitor` and `human_readable_visitor`) then dispatches on the result: an integer is matched against the variants' storage tags, a name against the variants' names. This mirrors how serde's own derived `Deserialize` resolves enum variant identifiers, so the generated impl now round-trips through human-readable, index-based, and self-describing-binary codecs alike.

This PR also extends the `openmls_serialization_helpers` crate tests to include a test for `ciborium` support.

### DIFF
--- a/serialization_helpers/src/deserialize.rs
+++ b/serialization_helpers/src/deserialize.rs
@@ -179,7 +179,7 @@ pub(crate) fn deserialize(input: TokenStream) -> TokenStream {
         }
 
 
-        /// A visitor for enum deserialization, codec-agnostic (see #variant_id).
+        /// A visitor for enum deserialization, codec-agnostic.
         struct #enum_visitor;
 
         impl <'de> serde::de::Visitor<'de> for #enum_visitor {
--- a/serialization_helpers/src/deserialize.rs
+++ b/serialization_helpers/src/deserialize.rs
@@ -126,16 +126,28 @@ pub(crate) fn deserialize(input: TokenStream) -> TokenStream {
                 f.write_str("variant storage tag or name")
             }
 
+            // NOTE: only tags that can be converted to a valid `u32` are accepted
             fn visit_u64<E: serde::de::Error>(self, v: u64) -> Result<Self::Value, E> {
-                Ok(#variant_id::Tag(v as u32))
+                u32::try_from(v).map(#variant_id::Tag).map_err(|_| {
+                    E::invalid_value(
+                        serde::de::Unexpected::Unsigned(v),
+                        &"a variant storage tag that fits in a u32",
+                    )
+                })
             }
 
             fn visit_u32<E: serde::de::Error>(self, v: u32) -> Result<Self::Value, E> {
                 Ok(#variant_id::Tag(v))
             }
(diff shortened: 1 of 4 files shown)


serialization_helpers/src/deserialize.rs
```diff
@@ -106,54 +107,70 @@ pub(crate) fn deserialize(input: TokenStream) -> TokenStream {
...
 
+            fn expecting(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+                f.write_str("variant storage tag or name")
             }
-        }
 
-        /// A visitor for self-describing enum deserialization
-        struct #human_readable_visitor;
+            fn visit_u64<E: serde::de::Error>(self, v: u64) -> Result<Self::Value, E> {
+                Ok(#variant_id::Tag(v as u32))
```
franziskuskiefer: This looks like we truncate larger values. This doesn't look right. Same fir `i64`.

serialization_helpers/src/deserialize.rs
```diff
@@ -106,54 +107,70 @@ pub(crate) fn deserialize(input: TokenStream) -> TokenStream {
...
+
+        impl <'de> serde::Deserialize<'de>  for #name {
+            fn deserialize<D: serde::Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
+                const VARIANT_NAMES: &[&'static str] = &[#(#variant_names), *];
+                d.deserialize_enum(stringify!(#name), VARIANT_NAMES, #enum_visitor)
+            }
+        }
+
+
+        /// A visitor for enum deserialization, codec-agnostic (see #variant_id).
```
franziskuskiefer: We don't want `#variant_id` in the doc comment I guess.

franziskuskiefer (changes requested):
Generally lgtm with one issue.

serialization_helpers/src/deserialize.rs
wysiwys: This method now includes error handling, using `u32::try_from` to convert the tag to a `u32` first, and returning an error that is constructed using [`serde::de::Error::invalid_value`](https://docs.rs/serde/latest/serde/de/trait.Error.html#method.invalid_value) when the conversion fails

franziskuskiefer (approved):
Thanks

PR #2134, opened 2026-07-24
Reject trailing bytes in known structured extension payloads

erskingardner:
## Summary

Require exact TLS decoding of the bounded `extension_data` payload for every known structured MLS extension.

OpenMLS deserializes each known extension's opaque byte vector into a typed Rust value. A typed decoder can successfully consume a valid prefix while leaving trailing bytes in that bounded payload. This change uses `tls_codec::Deserialize::tls_deserialize_exact` at the existing extension dispatch point so all known structured extension types consistently reject trailing data.

The protected extension types are:

- `ApplicationId`
- `RatchetTree`
- `RequiredCapabilities`
- `ExternalPub`
- `ExternalSenders`
- `LastResort`
- `AppDataDictionary` with the `extensions-draft` feature

Unknown and GREASE extensions remain opaque and preserve their payload bytes unchanged.

## Tests

- Round-trip every known structured extension.
- Reject trailing payload bytes for every known structured extension.
- Verify `AppDataDictionary` follows the same exact-decoding path.
- Verify unknown and GREASE extensions round-trip arbitrary payload bytes unchanged.

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p openmls --no-default-features`
- `cargo test -p openmls --no-default-features --features extensions-draft,extensions-draft-test-dependencies`
- `cargo clippy -p openmls --tests --benches --examples --no-default-features -- -D warnings`
- `cargo clippy -p openmls --tests --benches --examples --no-default-features --features extensions-draft,extensions-draft-test-dependencies -- -D warnings`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- [#2134](https://github.com/openmls/openmls/pull/2134): Known structured extension payloads now reject trailing bytes during decoding instead of silently ignoring them.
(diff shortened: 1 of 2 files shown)


kkohbrok (approved):
Good catch, thanks!

PR #2137, opened 2026-07-28
refactor(treemath): simplify [level] into [trailing_ones]

clementblaudeau:
This PR offers to use the intrinsic instead of a somewhat subtle loop. Found during the verification campaign (at Cryspen).

### DIFF
--- a/openmls/src/binary_tree/array_representation/treemath.rs
+++ b/openmls/src/binary_tree/array_representation/treemath.rs
@@ -280,15 +280,8 @@ fn log2(x: u32) -> usize {
 }
 
 pub fn level(index: u32) -> usize {
-    let x = index;
-    if (x & 0x01) == 0 {
-        return 0;
-    }
-    let mut k = 0;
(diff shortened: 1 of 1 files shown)


keks:
Thanks!

Before Clement opened the PR, I validated that this really does the same thing using this code:

```rust
/// copied from openmls
pub fn level(index: u32) -> usize {
    let x = index;
    if (x & 0x01) == 0 {
        return 0;
    }
    let mut k = 0;
    while ((x >> k) & 0x01) == 1 {
        k += 1;
    }
    k
}

fn main() {
    for i in 0..u32::MAX {
        assert_eq!(level(i) as u64, i.trailing_ones() as u64);
    }
}
```

This runs for about 40s on my machine and does not panic.

Later I realized that this does not include `u32::MAX` itself. When I tested that, I realized that our current function panics for that value :grimacing: 

So yeah, this is actually a fix!

keks (approved):
Thank you!

PR #2138, opened 2026-07-29
perf: don't create errors in happy path

boxdot:
In particular, this improves the performance in debug builds because LibraryError::custom captures a backtrace in such builds.


kkohbrok (approved):
Thanks!

PR #2139, opened 2026-07-30
fix: libcrux crypto provider rejected some supported ciphersuites

boxdot:
Crypto provider `supports` methods rejected ciphersuites that where advertised as supported by `supported_ciphersuites` method.


franziskuskiefer (approved):
ugh, nice catch

PR #2140, opened 2026-07-30
fix(bench): restore benchmarks and compare crypto providers

boxdot:
The bench target declared `libcrux-provider` as a required feature, and every benchmark only ran against the libcrux provider. Since CI did not enable the feature, `cargo bench` measured nothing at all.

Benchmarks now run against both the RustCrypto and libcrux providers, grouped per operation with provider and ciphersuite as benchmark id dimension, so criterion group summary compares them directly. CI enables `libcrux-provider`, and criterion html reports are turned on.


franziskuskiefer (approved):
lgtm

PR #2143, opened 2026-08-01
fix: sign the GroupInfo with the new signer when a commit rotates the key

### LINKED ISSUE #2009: [Bug] Old signer used for group info when producing welcome message or exported group info
**Describe the bug**

`build_internal` consumes both a new_signer and an old_signer and sets up a commit that transitions a member to use the new_signer. If a welcome message is produced by this commit, it will use a old_signer (which doesn't match the updated group state post-commit) to sign the group info in the welcome message. The same occurs when building exported_group_info. 

When consuming a welcome message, a joiner uses the post-commit group state to verify the signature on the group info (which now contains the new signing key), and this will fail to verify because the group info is not signed by the key in the current group state (it is signed by the old, pre-commit key).

The fix seems to be to use the new signer for the group info signatures if it is provided. `build_internal` should also reject and fail to commit if there is no new signer available and the current member consumes a proposal to update their own signer in the commit (since it won't be possible to sign the group info with the correct key). 

**To Reproduce**
```rust
#[openmls_test::openmls_test]
fn welcome_group_info_signature_broken_under_committer_rotation() {
    let alice_party = CorePartyState::<Provider>::new("alice");
    let bob_party = CorePartyState::<Provider>::new("bob");
    let charlie_party = CorePartyState::<Provider>::new("charlie");

    let alice_pre_group = alice_party.generate_pre_group(ciphersuite);
    let bob_pre_group = bob_party.generate_pre_group(ciphersuite);
    let charlie_pre_group = charlie_party.generate_pre_group(ciphersuite);

    let mls_group_create_config = MlsGroupCreateConfig::builder()
        .ciphersuite(ciphersuite)
        .use_ratchet_tree_extension(true)
        .build();
    let mls_group_join_config = mls_group_create_config.join_config().clone();

    // Alice + Bob in a group.
    let group_id = GroupId::from_slice(b"welcome-rotation-bug");
    let mut group_state =
        GroupState::new_from_party(group_id, alice_pre_group, mls_group_create_config).unwrap();

    group_state
        .add_member(AddMemberConfig {
            adder: "alice",
            addees: vec![bob_pre_group],
            join_config: mls_group_join_config.clone(),
            tree: None,
        })
        .expect("Could not add Bob");

    // Alice mints a fresh signer + credential to rotate to.
    let new_pre_group_state = alice_party.generate_pre_group(ciphersuite);
    let new_signature_key = new_pre_group_state
        .credential_with_key
        .signature_key
        .clone();

    let charlie_key_package = charlie_pre_group.key_package_bundle.key_package().clone();

    // Alice drives the CommitBuilder directly: one Add(Charlie) proposal PLUS
    // a signer rotation via `build_with_new_signer`. The returned Welcome
    // should be verifiable by Charlie, but today it isn't.
    let (commit_msg, welcome) = {
        let [alice] = group_state.members_mut(&["alice"]);
        let provider = &alice.party.core_state.provider;

        let new_signer = NewSignerBundle {
            signer: &new_pre_group_state.signer,
            credential_with_key: new_pre_group_state.credential_with_key.clone(),
        };

        let bundle = alice
            .group
            .commit_builder()
            .propose_adds(Some(charlie_key_package))
            // Add-only commits normally omit the update path, which would
            // silently skip the rotation even when a new signer is supplied.
            // Force the path so the new leaf actually carries the new key.
            .force_self_update(true)
            .load_psks(provider.storage())
            .expect("load psks")
            .build_with_new_signer(
                provider.rand(),
                provider.crypto(),
                &alice.party.signer, // old signer (envelope + existing leaf)
                new_signer,          // new signer (new leaf self-sig)
                |_| true,
            )
            .expect("build commit with new signer")
            .stage_commit(provider)
            .expect("stage commit");

        let (commit, welcome, _group_info) = bundle.into_contents();
        alice
            .group
            .merge_pending_commit(provider)
            .expect("merge own commit");

        (commit, welcome.expect("Add must produce a Welcome"))
    };

    // Bob applies the commit. This path works fine: he verifies the envelope
    // against Alice's OLD key (still in his pre-commit tree) and then
    // processes the UpdatePath + Add normally.
    group_state
        .deliver_and_apply_if(commit_msg.into(), |s| s.party.core_state.name == "bob")
        .expect("bob apply commit");

    // Sanity: Alice and Bob both see Alice's NEW signature key on her leaf.
    for name in ["alice", "bob"] {
        let [m] = group_state.members_mut(&[name]);
        let sk = m
            .group
            .members()
            .find(|x| x.index == LeafNodeIndex::new(0))
            .expect("alice leaf present")
            .signature_key;
        assert_eq!(
            sk.as_slice(),
            new_signature_key.as_slice(),
            "{name} sees stale key after rotation commit",
        );
    }

    // Charlie builds his state from the Welcome + the post-commit ratchet
    // tree. The joiner-side code in `PublicGroup::from_ratchet_tree` pulls
    // the verification key from `treesync.leaf(signer).signature_key()`,
    // which is Alice's NEW key (the tree is already post-commit). The
    // GroupInfo signature was produced with `old_signer`, so verification
    // fails.
    let ratchet_tree = {
        let [alice] = group_state.members_mut(&["alice"]);
        alice.group.export_ratchet_tree()
    };

    let join_result = StagedWelcome::new_from_welcome(
        &charlie_party.provider,
        &mls_group_join_config,
        welcome,
        Some(ratchet_tree.into()),
    );

    assert!(
        join_result.is_ok(),
        "Wrong key: Charlie rejected the Welcome with: {:?}",
        join_result.err(),
    );
}
```

munzzyy:
Closes #2009.

`CommitBuilder::build_internal` takes an `old_signer` and an optional `NewSignerBundle`. When a new signer is present and a path is computed, `compute_path` puts the new signature key in the committer's leaf. But both GroupInfo signatures further down were still produced with `old_signer`:

- the GroupInfo inside the Welcome (commit_builder.rs:1012)
- the exported GroupInfo (commit_builder.rs:1071)

A receiver looks up the verification key in the *post-commit* ratchet tree. `PublicGroup::from_ratchet_tree` does `treesync.leaf(verifiable_group_info.signer()).signature_key()`, which is already the new key. So the signature never verifies and the joiner gets `InvalidGroupInfoSignature`.

Repro on unpatched main (209a824), test included in this PR:

```
$ cargo test -p openmls --lib commit_with_new_signer_signs_group_info_with_new_signer

running 1 test
test group::mls_group::tests_and_kats::tests::mls_group::commit_with_new_signer_signs_group_info_with_new_signer_rustcrypto_MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519 ... FAILED

---- ...commit_with_new_signer_signs_group_info_with_new_signer... stdout ----
panicked at openmls/src/group/mls_group/tests_and_kats/tests/mls_group.rs:4289:6:
Charlie failed to process the Welcome: PublicGroupError(InvalidGroupInfoSignature)

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 270 filtered out
```

With the Welcome assertion removed, the second half fails the same way, so the two sites are independently broken:

```
panicked at openmls/src/group/mls_group/tests_and_kats/tests/mls_group.rs:4303:6:
Charlie failed to track the group from the exported GroupInfo: InvalidGroupInfoSignature
```

## The change

`build_internal` now returns `(path_computation_result, new_signer_used)` from the path branch. `new_signer_used` is `Some(signer)` only on the branch that actually handed the new signer to `compute_path`. Both GroupInfo sites match on it and fall back to `old_signer` otherwise.

The Commit's own signature (commit_builder.rs:918, `AuthenticatedContent::commit`) still uses `old_signer`, and that is deliberate. Existing members verify the commit against the sender's leaf in their *pre-commit* tree, which holds the old credential. Signing the commit with the new key would break verification for every member. The two signatures have different verification contexts, so they need different keys.

## Why this is keyed on the path, not on the argument

The obvious version is `if new_signer.is_some() { new } else { old }`. That is wrong. `build_internal` only applies the new signer when a path is computed:

```rust
if apply_proposals_values.path_required
    || contains_own_updates
    || cur_stage.force_self_update
    || !cur_stage.leaf_node_parameters.is_empty()
```

An Add-only commit doesn't require a path, so `build_with_new_signer` on such a commit leaves the old key in the leaf. Signing the GroupInfo with the new key there would create the mirror image of this bug. Tracking whether the path actually consumed the new signer keeps the GroupInfo signature tied to what is really in the tree, whatever that pre-existing behaviour ends up being. My test passes `force_self_update(true)` for that reason.

The other option I looked at was hoisting the choice into a single `&dyn Signer`, which reads a bit nicer at the two call sites. It would have meant either an object-safety commitment on `Signer` or an extra allocation, so I stuck with the two-arm match. `Option<&S>` is `Copy`, so both closures capture it without restructuring.

## Side effects

- No public API change. `build`, `build_with_new_signer`, and `self_update_with_new_signer` keep their signatures.
- Only commits built through `build_with_new_signer` (and `self_update_with_new_signer`, which routes through it) behave differently. Every other commit passes `None` and hits the `old_signer` arm, byte-identical to before.
- The reporter also asked for `build_internal` to fail when the committer commits their own Update proposal without supplying a new signer. Not in this PR: @kkohbrok pointed out RFC 9420 §12.2 already forbids a commit containing an Update proposal by the committer, and that discussion moved to Zulip.

## Verification

```
$ cargo test -p openmls --lib commit_with_new_signer_signs_group_info_with_new_signer
test ...commit_with_new_signer_signs_group_info_with_new_signer... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 270 filtered out; finished in 0.18s

$ cargo test -p openmls --lib
test result: ok. 268 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 59.33s

$ cargo test -p openmls --lib -F extensions-draft,extensions-draft-test-dependencies
test result: ok. 313 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 63.12s

$ cargo test -p openmls --lib -F virtual-clients-draft,virtual-clients-draft-test-dependencies
test result: ok. 349 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 62.21s

$ cargo test -p openmls
# every target green, including the RFC 9420 KAT vectors and the passive-client vectors

$ cargo fmt --check     # clean
$ cargo clippy -p openmls --all-targets     # no warnings
```

For the before/after I checked out `commit_builder.rs` at 209a824 with the new test in place and ran it (fails, output above), then restored the fix and reran it (passes). The three feature-set runs are there because the path-computation block I touched has `virtual-clients-draft` and `extensions-draft` cfgs around it.

The test sits next to the four `propose_self_update_with_new_signer_*` tests in `mls_group.rs` and uses the same single-group framework. Alice adds Charlie and rotates her signature key in one commit. Bob applies the commit, which confirms the old signer is still right for the framing. Then Charlie joins from the Welcome, and separately tracks the group through `PublicGroup::from_external` with the exported GroupInfo, so both signing sites are covered.

The CHANGELOG entry links to #2143; happy to renumber if that's off.


kkohbrok (approved):
Nice catch! Thanks for the PR.

PR #2144, opened 2026-08-03
fix: set_configuration apply past epoch policy

franziskuskiefer:
`set_configuration` replaced the whole `MlsGroupJoinConfig` but never resized or persisted the message secrets store, so changing the retention policy through this API had no effect on which past epochs could actually be decrypted. `set_past_epoch_deletion_policy` already did this correctly; `set_configuration` now mirrors it.


kkohbrok (approved):
Thanks!

PR #2152, opened 2026-08-04
fix: cleanup after ciphersuite related changes

kkohbrok:
This PR adds a KEM specific check for the `supports` function of the libcrux provider.

It also adds two missing entries to the storage stability test vector.


openmls/test_vectors/storage-stability.json
franziskuskiefer, line 1: Is this change intentional?

franziskuskiefer (approved):

openmls/test_vectors/storage-stability.json
franziskuskiefer: nevermind

PR #2158, opened 2026-08-07
feat: support AppEphemeral proposals in external commits

kkohbrok:


openmls/src/messages/mod.rs
```diff
@@ -252,6 +252,13 @@ pub(crate) struct CommitIn {
 }
 
 impl CommitIn {
+    /// Returns the proposals covered by this commit. The commit has not been
+    /// validated at this point.
+    #[cfg(feature = "extensions-draft")]
+    pub(crate) fn proposals(&self) -> &[ProposalOrRefIn] {
```
boxdot: Should this method also be called `unverified_proposals` symmetrically to `unverified_credential`?

boxdot (approved):

openmls/src/messages/mod.rs
kkohbrok: Yes, it probably should. Thanks!

PR #2159, opened 2026-08-07
virtual clients: two-step builder for VC sibling external-commit joins

kkohbrok:
Replace MlsGroup::vc_join_via_sibling_external_commit with VcExternalCommitJoinBuilder. process_commit rebuilds the prior-epoch group and verifies the commit, returning a PendingVcExternalCommitJoin that exposes the verified AppDataUpdate and AppEphemeral proposals plus a dictionary updater seeded with the prior-epoch group context. join stages and merges the commit with the application-resolved AppDataUpdates, mirroring the resolve_app_data_commit contract of regular processing.

Previously a commit carrying an AppDataUpdate proposal failed at staging with MissingAppDataUpdates, making the entry point unusable for applications that bump a group-context component on every commit.

The consume-once operation secret derivation now happens in join, and missing updates are rejected before it, so an incomplete join can be retried by processing the same commit again.

### DIFF
--- a/openmls/src/group/mls_group/creation.rs
+++ b/openmls/src/group/mls_group/creation.rs
@@ -27,6 +27,14 @@ use crate::{
 
 use crate::key_packages::KeyPackage;
 
+#[cfg(feature = "virtual-clients-draft")]
+use crate::{
+    component::ComponentId,
+    framing::mls_auth_content::AuthenticatedContent,
+    group::mls_group::processing::{AppDataDictionaryUpdater, AppDataUpdates},
+    messages::proposals::{AppDataUpdateProposal, AppEphemeralProposal},
+};
+
 impl MlsGroup {
     // === Group creation ===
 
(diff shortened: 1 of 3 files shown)


openmls/src/group/mls_group/creation.rs
```diff
@@ -1269,6 +1144,352 @@ impl MlsGroup {
...
+    }
+
+    /// Completes the join. `app_data_updates` must be `Some` exactly when the
+    /// commit covers AppDataUpdate proposals. Absent updates are rejected
+    /// before the consume-once operation secret generation is spent, so the
+    /// join can be started over. Updates that do not reproduce the committing
+    /// sibling's dictionary fail with a confirmation tag mismatch after the
+    /// generation is consumed, so they should be computed deterministically
+    /// from the proposals rather than guessed.
+    pub fn join<Provider: OpenMlsProvider>(
```
raphaelrobert: This could be `into_group()` to align it with the rest of the API.

openmls/src/group/mls_group/creation.rs
```diff
@@ -1269,6 +1144,352 @@ impl MlsGroup {
...
+        self,
+        provider: &Provider,
+        app_data_updates: Option<AppDataUpdates>,
```
raphaelrobert: We should have a `with_app_data_dictionary_updates()` for that I think

openmls/src/group/mls_group/creation.rs
```diff
@@ -1269,6 +1144,352 @@ impl MlsGroup {
...
+/// help of [`Self::app_data_dictionary_updater`]) and pass the resulting
+/// [`AppDataUpdates`] to [`Self::join`], exactly as it would for
+/// [`MlsGroup::resolve_app_data_commit`] when processing the same commit as
+/// a member.
+///
+/// Dropping this value discards the join. No operation secret generation is
+/// consumed before [`Self::join`], so a discarded join can be started over
+/// by processing the same commit again.
+#[cfg(feature = "virtual-clients-draft")]
+pub struct PendingVcExternalCommitJoin {
```
raphaelrobert: `Pending...` might be the wrong prefix here. Isn't this more like `StagedVcExternalCommitJoin` or `UnresolvedVcExternalCommitJoin`?

openmls/src/group/mls_group/creation.rs
```diff
@@ -1269,6 +1144,352 @@ impl MlsGroup {
...
+    group: MlsGroup,
+    /// The verified commit content.
+    content: AuthenticatedContent,
+    /// The by-value AppDataUpdate proposals covered by the commit, sorted by
+    /// component id.
+    app_data_update_proposals: Vec<AppDataUpdateProposal>,
+}
+
+#[cfg(feature = "virtual-clients-draft")]
+impl PendingVcExternalCommitJoin {
```
raphaelrobert: This object only exposes very few functions for inspection, given that it contains a whole group. We should expose the group context and group members at a minimum.

openmls/src/group/mls_group/creation.rs
```diff
@@ -1269,6 +1144,352 @@ impl MlsGroup {
...
+
+    /// Returns the AppEphemeral proposals for `component_id` that the commit
+    /// carries by value, in the order they appear in the commit, for example
+    /// to decide how to follow the join. Unlike
+    /// [`PublicMessageIn::unverified_app_ephemeral_proposals`], the returned
+    /// data is authenticated: the commit's signature has been verified.
+    ///
+    /// [`PublicMessageIn::unverified_app_ephemeral_proposals`]:
+    ///     crate::framing::PublicMessageIn::unverified_app_ephemeral_proposals
+    pub fn app_ephemeral_proposals(
```
raphaelrobert: Shouldn't this be called `app_ephemeral_proposals_for_component_id`? That's what we call in in the `ProposalQueue`.

raphaelrobert (changes requested):
Thanks, looks good in general, some comments on the API.

raphaelrobert (approved):

PR #2161, opened 2026-08-10
virtual clients: secret management

kkohbrok:
This PR is the first step towards implementing the secret management strategy introduced to the virtual clients spec in this PR: https://github.com/mlswg/mls-virtual-clients/pull/30

Concretely, this PR renames emulation epoch -> derivation epoch and introduces implicit derivation of emulation secrets. Groups derive and store emulation secrets automatically for emulation groups when one of the following hold

- It's the group's first epoch
- Group membership changed
- The commit that begins the epoch marks the epoch as a derivation epoch


openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -400,30 +418,59 @@ impl<'a, G: BorrowMut<MlsGroup>> CommitBuilder<'a, Initial, G> {
...
     /// Implies that a self-update takes place: the commit will always have
     /// a path even if no other proposals are queued.
-    ///
-    /// [`MlsGroup::register_vc_emulation_epoch`]: crate::group::MlsGroup::register_vc_emulation_epoch
     #[cfg(feature = "virtual-clients-draft")]
     pub fn vc_emulation<Crypto: OpenMlsCrypto, Storage: StorageProvider>(
+        self,
+        crypto: &Crypto,
+        storage: &Storage,
+        emulation_group: &MlsGroup,
```
raphaelrobert: If we pass the whole group here, we should call `is_emulation_group()`, otherwise we could just pass the group id.

openmls/src/group/mls_group/builder.rs
```diff
@@ -60,11 +67,8 @@ impl MlsGroupBuilder {
     ///
     /// [`MlsGroup::vc_join_at_creation`]: crate::group::MlsGroup::vc_join_at_creation
     #[cfg(feature = "virtual-clients-draft")]
-    pub fn vc_emulation(
-        mut self,
-        epoch_id: crate::components::vc_derivation_info::EpochId,
-    ) -> Self {
-        self.vc_epoch_id = Some(epoch_id);
+    pub fn vc_emulation(mut self, emulation_group: &MlsGroup) -> Self {
```
raphaelrobert: If we pass the whole group here, we should call `is_emulation_group()`, otherwise we could just pass the group id.

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -482,6 +529,35 @@ impl<'a, G: BorrowMut<MlsGroup>> CommitBuilder<'a, Initial, G> {
...
+    /// state. Operations that reference a derivation epoch keep using the
+    /// newest derivation epoch of that input state, including operations
+    /// carried by this very commit.
+    ///
+    /// The group has to be configured as an emulation group and its
+    /// GroupContext has to require Safe AAD framing. Otherwise `build` fails
+    /// with [`CreateCommitError::NewDerivationEpochOutsideEmulationGroup`] or
+    /// [`CreateCommitError::NewDerivationEpochWithoutSafeAad`].
+    #[cfg(feature = "virtual-clients-draft")]
+    pub fn new_derivation_epoch(mut self) -> Self {
```
raphaelrobert: I think this should take a `bool` as an argument for consistency, and should lose the `new_` prefix, which sounds like a constructor.

openmls/src/group/mls_group/config.rs
```diff
@@ -214,6 +214,10 @@ pub struct MlsGroupJoinConfig {
     pub(crate) use_ratchet_tree_extension: bool,
     /// Sender ratchet configuration
     pub(crate) sender_ratchet_configuration: SenderRatchetConfiguration,
+    /// Flag marking this group as an emulation group of a virtual client.
+    #[cfg(feature = "virtual-clients-draft")]
+    #[serde(default)]
```
raphaelrobert: I'm not sure whether feature gating the flag is a good idea. I'm also not sure we should put more things in the persisted config, since we have vague agreement that we want to get rid of that someday. Can you find a way to avoid that?

openmls/src/group/mls_group/config.rs
```diff
@@ -521,6 +563,14 @@ impl MlsGroupCreateConfigBuilder {
         self
     }
 
+    /// Marks the group as an emulation group of a virtual client. See
+    /// [`MlsGroupJoinConfigBuilder::emulation_group`] for what that entails.
+    #[cfg(feature = "virtual-clients-draft")]
+    pub fn emulation_group(mut self, emulation_group: bool) -> Self {
```
raphaelrobert, line 568: Unless I missed something, there is no getter for this setter.

memory_storage/src/lib.rs
```diff
@@ -260,11 +260,11 @@ const GROUP_CONTEXT_LABEL: &[u8] = b"GroupContext";
 #[cfg(feature = "extensions-draft")]
 const APPLICATION_EXPORT_TREE_LABEL: &[u8] = b"ApplicationExportTree";
 #[cfg(feature = "virtual-clients-draft")]
-const VC_EMULATION_EPOCH_STATE_LABEL: &[u8] = b"VcEmulationEpochState";
+const VC_DERIVATION_EPOCH_STATE_LABEL: &[u8] = b"VcEmulationEpochState";
```
raphaelrobert: The name and the content are no longer the same here.

openmls/src/group/mls_group/config.rs
```diff
@@ -363,6 +374,37 @@ impl MlsGroupJoinConfigBuilder {
...
+    ///
+    /// An emulation group derives the virtual client's secrets from its
+    /// derivation epochs. The initial epoch is a derivation epoch, and so is
+    /// the output epoch of every commit that changes membership or that carries
+    /// a `new_derivation_epoch` action in its virtual-clients Safe AAD item
+    /// (see [`CommitBuilder::new_derivation_epoch`]). OpenMLS derives and
+    /// persists the derivation-epoch state itself at group creation, at a
+    /// Welcome join, and when such a commit is merged. Applications should wrap
+    /// merge calls in a storage transaction, since those writes happen
+    /// alongside the merge's own writes.
```
raphaelrobert: We don't really know which merges need to be wrapped, so should we wrap all of them? Just asking because it's not entirely clear from the comment

raphaelrobert (changes requested):
This looks good, I made a few comments about the API.

memory_storage/src/lib.rs
kkohbrok: This is for backwards compatibility, but for the memory storage provider that probably doesn't matter.

openmls/src/group/mls_group/config.rs
kkohbrok: I think it's generally okay to add flags to the config until we have something better, but in this case, we can infer whether a group is an emulation group from storage, so I'll do that instead.

openmls/src/group/mls_group/commit_builder.rs
kkohbrok: Good point, I'll adjust it for consistency. But we should move away from that eventually. With `false` as the default, it's only ever going to be called with true. The same goes for `force_self_update`.

openmls/src/group/mls_group/config.rs
kkohbrok: Yes, all merges should be wrapped. There's now a comment in the book chapter on virtual clients.

openmls/src/group/mls_group/config.rs
kkohbrok: There is `emulation_group` a few lines up.

raphaelrobert (approved):
Thanks for making the changes

PR #2162, opened 2026-08-11
fix: classify GREASE extensions as ExtensionType::Grease

### LINKED ISSUE #2156: [Bug] GREASE extensions are rejected even when listed in the leaf node capabilities
**Describe the bug**

We ran into this with a KeyPackage produced by a client built on Cisco [mlspp](https://github.com/cisco/mlspp): OpenMLS rejects it, although it GREASEs exactly as RFC 9420 Section 13.5 prescribes.

`Extension::extension_type()` returns `ExtensionType::Unknown(x)` for a GREASE value, while `impl From<u16> for ExtensionType`, used when decoding `Capabilities`, returns `ExtensionType::Grease(x)`. Validation compares the two with `Vec::contains`, so a GREASE extension never matches its own capabilities entry and is treated as unsupported.

Section 13.5 lists `KeyPackage.extensions` and `LeafNode.extensions` as fields a client SHOULD populate with GREASE values, and requires the leaf node ones to be reflected in `LeafNode.capabilities.extensions`.

**To Reproduce**

The two sides disagree on the same wire value:

```rust
let from_capabilities: ExtensionType = 0x8A8Au16.into();
let from_extension = Extension::Unknown(0x8A8A, UnknownExtension(vec![])).extension_type();
assert_eq!(from_capabilities, from_extension); // Grease(35466) != Unknown(35466)
```

An unmodified mlspp key package carries `0x8A8A` in `KeyPackage.extensions`, `0x4A4A` in `LeafNode.extensions`, and both in `LeafNode.capabilities.extensions`:

```rust
let bytes = general_purpose::STANDARD.decode(MLSPP_KEY_PACKAGE).unwrap();
let key_package = MlsMessageIn::tls_deserialize(&mut bytes.as_slice())
    .unwrap()
    .into_keypackage()
    .unwrap();

let key_package_in: KeyPackageIn = key_package.into();
key_package_in.validate(provider.crypto(), ProtocolVersion::Mls10)
// Err(UnsupportedExtension)
```

<details>
<summary>base64 of that key package</summary>

```
AAEABQABAAEgYy4COfvC/iouTT8qF4VgcVfgyKb3hColTcqem8FVyW0gFguUpDuSLZVKAR6u/sZ0Mw70+/CLlWpEga/Ey+lu8GEgqMxyAzg/BldGAdYr/BqpcdyWuaJd/mcy02KTBtmrS6wAAQtjbGllbnRfMl9pZAIAAQ4AATo6AAIAAwAEAAUABgSKikpKAtraCjo6AAEAAv4A/wABAAAAAAAAAAD//////////wZKSgMQK+BAQF7MDrBfiiW90n87xjd1NuTI3RQeOvGvb3/cl+hNb+aZ98nB4tE2Es/J1Eja4MfREPahuniYDkgeJdGcNilTywIDiooAQEDmxlDXAclxqFv+qBIUoRE7fWEP1jVzuoGFV/U2l6WHSmzJb0paNLsM2xHT7FBO6E4QKaBIHWffNT3JQzE6mucH
```

</details>

A GREASE value in `LeafNode.extensions` takes a second path. The key package builds, but adding that member fails:

```rust
let bob_key_package = KeyPackage::builder()
    .leaf_node_capabilities(
        Capabilities::builder()
            .extensions(vec![ExtensionType::Grease(0x8A8A)])
            .build(),
    )
    .leaf_node_extensions(
        Extensions::single(Extension::Unknown(0x8A8A, UnknownExtension(vec![]))).unwrap(),
    )
    .build(ciphersuite, &bob_provider, &bob_signer, bob_credential)
    .unwrap();

alice_group.add_members(&provider, &alice_signer, &[bob_key_package.key_package().clone()])
// Err(CreateCommitError(ProposalValidationError(LeafNodeValidation(UnsupportedExtensions))))
```

The same builder with `.key_package_extensions(...)` instead reproduces the first error without the mlspp key package.

**Expected behavior**

Both validate. The GREASE value is listed in `LeafNode.capabilities.extensions`, so it is supported.

**Actual behavior**

`KeyPackageVerifyError::UnsupportedExtension` and `LeafNodeValidationError::UnsupportedExtensions`.

**Platform**

- macOS 15 (Darwin 24.3.0), aarch64
- `main` @ fbbf0362f and 0.8.1, rustc 1.89.0

**Additional context**

#1900 added `ExtensionType::Grease` and updated `impl From<u16>`, but not `Extension::extension_type()`. That looks like where the receiving side (#1337) regressed.

The key package path runs through `KeyPackageIn::validate`, the leaf node path through `LeafNode::validate_locally`, reached from `validate_leaf_node_capabilities` when adding a member and joining from a Welcome.

Possibly relevant nearby: `is_valid_in_leaf_node`, `is_valid_in_key_package` and `is_valid_in_group_info` all reject `Grease(_)`. Those arms seem unreachable today, since no extension ever reports `Grease`.

`Extensions::unknown()` looks up an `ExtensionType::Unknown`, so it returns `None` for a GREASE extension that is present in the set.

`GroupContext.extensions` currently accepts GREASE values, which the same section forbids sending. Out of scope here, but fixing the classification changes it.

Affects 0.8.0 through `main`. 0.7.x is fine, because without the `Grease` variant both sides produce `Unknown(x)`, so we stayed on 0.7.4.

Happy to open a PR: we have a branch with a fix and regression tests for both paths, including the mlspp key package.

fatihergin:
Closes #2156.

A note up front: I am not an MLS expert. I ran into this while processing a key package from an mlspp-based client and worked out most of the protocol details by following the failure. I'm confident about the classification fix, less so about the `GroupContext` decision below. I leaned towards more test coverage rather than less, so say the word if you want any of it dropped.

`Extension::extension_type()` returns `ExtensionType::Unknown(x)` for a GREASE value, while `ExtensionType::from(u16)`, which is used when decoding capabilities, returns `ExtensionType::Grease(x)`. Validation compares the two with `Vec::contains`, so a GREASE extension never matches its own capabilities entry. A key package that GREASEs the way [RFC 9420 Section 13.5](https://www.rfc-editor.org/rfc/rfc9420.html#section-13.5) describes is rejected with `KeyPackageVerifyError::UnsupportedExtension`, and adding such a member fails with `LeafNodeValidationError::UnsupportedExtensions`.

`Extension::extension_type()` now classifies GREASE code points as `Grease`. I remapped only GREASE rather than calling `ExtensionType::from(*kind)`, which would turn a hand-built `Unknown(0x0001)` into `ApplicationId`.

That alone is not enough: three of the four `is_valid_in_*` predicates had an explicit `Grease(_)` arm that rejected it, unreachable while no extension ever reported `Grease`. Leaving them would only have swapped one rejection for another. GREASE now takes the same path as any other unknown extension type in all four; the fourth, `is_valid_in_group_context`, is below. `Extensions::unknown()` also finds GREASE extensions now.

The GREASE chapter in the book only covered capabilities, so it gained a short section on sending GREASE as an actual extension.

Nothing changes on the wire or in storage: `From<ExtensionType> for u16` maps both `Grease(v)` and `Unknown(v)` to `v`, and `extension_type()` is computed rather than persisted.

## GroupContext

`is_valid_in_group_context` accepts `Unknown(_)`, so a GREASE extension is accepted there today. Once GREASE classifies as `Grease(_)` it would fall through to the catch-all and start being rejected, so I added `Grease(_)` to the accepting arm. That keeps the current behaviour rather than changing it.

Section 13.5 does forbid GREASE in `GroupContext.extensions`, so rejecting it is tempting. I did not, because the rule binds senders while this predicate also runs when a received GroupContext is deserialized. Rejecting there would turn a sender-side MUST NOT into a parse failure on the receiving side, and would treat the GREASE code points more strictly than any other unknown value. Extensions the group does not support are still turned away by the usual capability checks, since GREASE is not a default extension type.

Enforcing the sender-side rule properly would mean separating send validation from receive validation, which felt out of scope here.

This is also why `grease_extension_validators_match_unknown_handling` asserts that `Extensions::<GroupContext>::single(grease)` succeeds, which would otherwise look wrong.

## Verification

The existing GREASE tests cover GREASE values in capabilities; these cover GREASE sent as an actual extension.

- classification over every reserved GREASE value, plus `0xFAFA` as a negative
- the four per-context validators, which also run on deserialization
- the `Extensions::unknown()` getter
- the capability comparison that actually broke
- end to end through the `openmls_test` harness (`KeyPackageIn::validate`, `add_members`, joining from the Welcome), so it runs for every ciphersuite the test build enables
- an unmodified key package generated with Cisco's mlspp, taken from the issue report, so there is a real interoperability case.


openmls/tests/grease.rs
```diff
@@ -475,3 +475,34 @@ fn test_grease_injection_in_groups_via_with_grease() {
...
+/// A KeyPackage generated with Cisco's mlspp (<https://github.com/cisco/mlspp>),
+/// taken from <https://github.com/openmls/openmls/issues/2156>. It carries
+/// GREASE extension 0x8A8A in `KeyPackage.extensions` and 0x4A4A in
+/// `LeafNode.extensions`, both listed in the leaf node capabilities.
+#[test]
+fn test_mlspp_key_package_with_grease_extensions_validates() {
+    use base64::{engine::general_purpose, Engine};
+    use openmls::prelude::tls_codec::Deserialize as _;
+
+    const MLSPP_KEY_PACKAGE: &str = "AAEABQABAAEgYy4COfvC/iouTT8qF4VgcVfgyKb3hColTcqem8FVyW0gFguUpDuSLZVKAR6u/sZ0Mw70+/CLlWpEga/Ey+lu8GEgqMxyAzg/BldGAdYr/BqpcdyWuaJd/mcy02KTBtmrS6wAAQtjbGllbnRfMl9pZAIAAQ4AATo6AAIAAwAEAAUABgSKikpKAtraCjo6AAEAAv4A/wABAAAAAAAAAAD//////////wZKSgMQK+BAQF7MDrBfiiW90n87xjd1NuTI3RQeOvGvb3/cl+hNb+aZ98nB4tE2Es/J1Eja4MfREPahuniYDkgeJdGcNilTywIDiooAQEDmxlDXAclxqFv+qBIUoRE7fWEP1jVzuoGFV/U2l6WHSmzJb0paNLsM2xHT7FBO6E4QKaBIHWffNT3JQzE6mucH";
```
fatihergin, line 488: is a test vector like this welcome in the source file, or would you rather keep test data elsewhere?

keks:
Thanks for the fix! I had a look at the issue and your PR. I think I like the general approach here, and I think your concern that parsing and validation are coupled too closely is something we should consider.

However, currently all `Extensions` validation is done during parsing, so I'd rather do it during parsing than not at all. Does that make sense, or am I misunderstanding/overlooking something?

fatihergin:
> Thanks for the fix! I had a look at the issue and your PR. I think I like the general approach here, and I think your concern that parsing and validation are coupled too closely is something we should consider.
> 
> However, currently all `Extensions` validation is done during parsing, so I'd rather do it during parsing than not at all. Does that make sense, or am I misunderstanding/overlooking something?

That's fair @keks. The piece I think I weighed wrong is that there's no separate send validation, so keeping the check out of the parser leaves it largely unenforced.

Two things. The "largely" is deliberate: as far as I can tell `Grease` is not a default extension type, so the capability checks already turn it away in `validate_add_proposals` (valn0502), `validate_update_proposals` (valn0602), `validate_commit` (valn1210) and `into_staged_welcome_inner` (valn1415). Does that change your read?

The other is what dropping the arm does in practice. GREASE classifies as `Unknown(x)` today and `is_valid_in_group_context` accepts `Unknown(_)`, so a received GroupContext carrying GREASE currently parses. Afterwards it won't parse at all, and since `Extensions<T>`'s deserialize impl maps every validator error to `DecodingError("Found duplicate extensions")`, that is the error a caller would see. A group whose members do advertise that value also becomes unjoinable. `Unknown(0xF042)` stays accepted in the same field. Either way I'll note it in the CHANGELOG.

The tightening itself seems reasonable to me. I'd just rather it landed as its own change than as a side effect here, but it's your call. If you'd like it in the parser now, I'll drop the arm, fix that error mapping so a rejected extension type doesn't surface as a duplicate-extension one (`InvalidExtensionError` already distinguishes them), and rework the tests that pin the current behaviour. One is named after GREASE matching unknown handling, so it needs renaming rather than flipping.

keks:
@fatihergin I didn't notice it earlier, but we already PR #2127 open, which seems to (beside other things) also make very similar changes as you made here. Sorry about that conflict. I think there is one bug that #2127 didn't fix, which is here:

https://github.com/openmls/openmls/blob/1b02263b19b94952e3d717416f9fc4531ba7f800/openmls/src/extensions/mod.rs#L727-L739

We still don't match on Grease(_) here, so we won't find them when looking them up. Realistically that won't happen, but we might use it for testing.

I also appreciate your book section. I think now that we merged the other PR, the APIs might be a little different.

If you are up for it, please rebase, introduce the fix and update the book code (if needed). I'll review and make sure this gets merged.

Thanks for the effort you already put in!

fatihergin:
Thanks @keks, rebased. I also did a few more minor updates to the book.

keks (approved):
Thank you for the fix and the patience!

PR #2163, opened 2026-08-11
Fix `MemoryStorage::clear_proposal_queue` leaving orphaned proposal data

freedom-winds:
## Summary

Fix `MemoryStorage::clear_proposal_queue` so that queued proposal values are actually removed from the backing store.

## Problem

The queue stores proposal values under a key containing the storage label and version suffix. `clear_proposal_queue` attempted to remove only the serialized proposal identifier, so the stored values remained as unreachable entries after the reference list was deleted. Repeated queue cleanup could therefore cause unbounded memory growth and persisted storage bloat.

## Changes

- Build the complete `QUEUED_PROPOSAL_LABEL` storage key before removing each proposal.
- Add a regression assertion that the backing `MemoryStorage` is empty after queue cleanup.

## Testing

`cargo test -p openmls_memory_storage --test proposals read_write_delete`

### DIFF
--- a/memory_storage/src/lib.rs
+++ b/memory_storage/src/lib.rs
@@ -874,7 +874,8 @@ impl StorageProvider<CURRENT_VERSION> for MemoryStorage {
         for proposal_ref in proposal_refs {
             // Delete all proposals.
             let key = serde_json::to_vec(&(group_id, proposal_ref))?;
-            values.remove(&key);
+            let storage_key = build_key_from_vec::<CURRENT_VERSION>(QUEUED_PROPOSAL_LABEL, key);
+            values.remove(&storage_key);
         }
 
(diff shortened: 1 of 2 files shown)


niyamvora:
Verified locally on top of current `main`: the added assertion fails against the existing `clear_proposal_queue` (`assertion failed: storage.values.read().unwrap().is_empty()`) and passes with this change. The key mismatch is only in `clear_proposal_queue` — `remove_proposal` already goes through `delete(label, key)` — so this covers the one broken site. Rebases cleanly. We hit this in production: we persist `MemoryStorage` on every save, so every orphaned proposal body ends up in every backup. Would be great to see this in a 0.9.x point release.

kkohbrok (approved):
Thanks!

PR #2164, opened 2026-08-11
Use the leftmost removed leaf for external Commit sender index calculation

freedom-winds:
## Summary

Use the leftmost removed leaf when determining the sender position of an external Commit.

## Problem

The external Commit builder considers all `Remove` proposals and selects the smallest available leaf index. Commit validation on the receiving side previously inspected only the first `Remove` proposal. When multiple Remove proposals were present in an order that did not match their leaf indices, sender and receiver selected different UpdatePath positions, causing valid external Commits to be rejected or applied against the wrong tree position.

## Changes

- Inspect every `Remove` proposal during external Commit validation.
- Select the minimum removed leaf index, matching external Commit construction.

## Testing

The relevant OpenMLS library tests should be run with:

`cargo test -p openmls --lib`

### DIFF
--- a/openmls/src/messages/mod.rs
+++ b/openmls/src/messages/mod.rs
@@ -295,12 +295,12 @@ impl CommitIn {
                     self_removes_in_store,
                 } => {
                     // We need to determine if it is a a resync or a join.
-                    // Find the first remove proposal and extract the leaf index.
-                    let former_sender_index = proposals.iter().find_map(|p| {
+                    // Use the leftmost removed leaf, matching external-commit construction.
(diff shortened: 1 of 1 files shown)


kkohbrok:
According to RFC9420 external commits can only contain one remove proposal (see [here](https://www.rfc-editor.org/rfc/rfc9420.html#name-joining-via-external-commit)), which I believe we check in our validation functions. If you still think this is a bug, it would be helpful if you could provide an integration test that triggers it. Otherwise I'm inclined to close this issue.

kkohbrok:
Closing for now. We'll reopen if it turns our we've overlooked something.

PR #2166, opened 2026-08-11
Correct TLS serialized length calculation for `HashSet<Vec<u8>>`

freedom-winds:
## Summary

Correct the TLS serialized length calculation for `HashSet<Vec<u8>>` in `ds-lib`.

## Problem

The custom `tls_serialized_len` implementation used the number of set elements as the vector payload length. TLS vector lengths represent serialized bytes, including each element's own length prefix, so the reported size was incorrect for non-empty or variable-length values.

## Changes

- Calculate the payload length from each element's serialized length.
- Select the outer TLS length-prefix width from the payload byte length.
- Add a regression test comparing the reported size with the actual encoded size.

## Testing

`cargo test -p ds-lib`

### DIFF
--- a/delivery-service/ds-lib/src/lib.rs
+++ b/delivery-service/ds-lib/src/lib.rs
@@ -34,7 +34,7 @@ mod hashset_codec {
         io::{Read, Write},
     };
 
-    use crate::tls_codec::{self, Deserialize, Serialize};
+    use crate::tls_codec::{self, Deserialize, Serialize, Size};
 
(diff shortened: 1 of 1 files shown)


delivery-service/ds-lib/src/lib.rs
```diff
@@ -37,14 +37,17 @@ mod hashset_codec {
     use crate::tls_codec::{self, Deserialize, Serialize};
```
fatihergin:
```suggestion
    use crate::tls_codec::{self, Deserialize, Serialize, Size};
```

PR #2046, opened 2026-06-02
feat: P-384, ML-DSA and general PQ ciphersuite support for RustCrypto provider

boxdot:
Add post-quantum and P-384 ciphersuite support to the RustCrypto
provider, hidden behind the `draft-ietf-mls-pq-ciphersuites` feature
flag:

- `MLS_192_MLKEM1024_AES256GCM_SHA384_P384` (ML-KEM1024 + ECDSA P-384)
- `MLS_256_MLKEM1024_AES256GCM_SHA512_MLDSA87` (ML-KEM1024 + ML-DSA87)
- the remaining PQ ciphersuites supported by RustCrypto, plus the custom
  `AIR_128_MLKEM768_AES256GCM_SHA384_Ed25519` ciphersuite

Wire ML-DSA 0.1 signing into basic_credential, propagate the feature
flag across crates, and adjust KAT/storage tests for the new
ciphersuites and ciphersuite-dependent hash lengths.

Co-authored-by: Konrad Kohbrok <konrad.kohbrok@datashrine.de>

PR #2170, opened 2026-08-16
feat: ML-DSA signatures in the libcrux provider

### LINKED ISSUE #2169: libcrux provider: wire ML-DSA signatures (draft-ietf-mls-pq-ciphersuites)
#2046 added `SignatureScheme::MLDSA44/65/87` and the ML-DSA ciphersuites behind `draft-ietf-mls-pq-ciphersuites`, but only the RustCrypto provider (`ml-dsa` crate) implements the signing side. `openmls_libcrux_crypto` still returns `UnsupportedSignatureScheme` for anything but Ed25519, so none of

- `MLS_128_MLKEM768X25519_CHACHA20POLY1305_SHA384_MLDSA44`
- `MLS_192_MLKEM768_AES256GCM_SHA384_MLDSA65`
- `MLS_256_MLKEM1024_AES256GCM_SHA384_MLDSA87`

are usable with the (verified) libcrux backend.

`libcrux-ml-dsa` 0.0.10 is on crates.io with matching `libcrux-sha3 =0.0.10` / `libcrux-intrinsics =0.0.8` / `libcrux-secrets =0.0.6` pins (already in `Cargo.lock`), and `hpke-rs-libcrux` 0.7 supports pure ML-KEM behind `draft-connolly-cfrg-hpke-mlkem`.

Proposal (I'll send a PR):
- add `libcrux-ml-dsa` (optional, enabled by `draft-ietf-mls-pq-ciphersuites`) and turn on `hpke-rs-libcrux/draft-connolly-cfrg-hpke-mlkem` under the same feature
- implement MLDSA44/65/87 in `signature_key_gen` / `sign` / `verify_signature`, using the same wire format as the RustCrypto provider (private key = 32-byte seed, public key = encoded verification key, empty context) so keys/signatures interoperate across providers
- accept `MlKem768`/`MlKem1024` KEMs and the three suites in `supports()` / `supported_ciphersuites()` so `openmls_test` picks them up in the nightly PQ matrix
- CHANGELOG entry

Happy to adjust scope (e.g. leave pure-ML-KEM suites out and only enable the X-Wing+ML-DSA-44 suite) if you'd rather keep the hpke-rs feature off.

niyamvora:
Closes #2169.

#2046 added `SignatureScheme::MLDSA44/65/87` and the ML-DSA ciphersuites, but only the RustCrypto provider could sign with them. This wires `libcrux-ml-dsa` into `openmls_libcrux_crypto` (behind `draft-ietf-mls-pq-ciphersuites`) so the libcrux backend supports:

- `MLS_128_MLKEM768X25519_CHACHA20POLY1305_SHA384_MLDSA44`
- `MLS_192_MLKEM768_AES256GCM_SHA384_MLDSA65`
- `MLS_256_MLKEM1024_AES256GCM_SHA384_MLDSA87`

### Changes
- `libcrux_crypto/Cargo.toml`: `libcrux-ml-dsa = 0.0.10` (optional, pulled in by the PQ feature). Its pins (`libcrux-sha3 =0.0.10`, `libcrux-intrinsics =0.0.8`, `libcrux-secrets =0.0.6`) already match `Cargo.lock`, so the only lockfile delta is the new crate. Also enables `hpke-rs-libcrux/draft-connolly-cfrg-hpke-mlkem` under the same feature so the pure ML-KEM KEMs used by the MLDSA65/87 suites are available.
- `crypto.rs`: ML-DSA arms in `signature_key_gen` / `sign` / `verify_signature`, in a small `ml_dsa` module (one macro instantiation per parameter set, since libcrux exposes each as its own module with fixed-size types). `supports()` accepts the ML-DSA schemes and `MlKem768/1024`; the three suites are added to `supported_ciphersuites()` so `openmls_test` picks them up in the nightly PQ matrix.
- **Wire format matches the RustCrypto provider** — private key = 32-byte seed, public key = encoded verification key, empty context, hedged signing with fresh DRBG randomness — so keys and signatures interoperate across providers.
- `tests/ml_dsa.rs`: round-trip + cross-provider (libcrux ↔ RustCrypto, both key origins) for all three parameter sets. Adds `openmls_rust_crypto` as a dev-dependency of the libcrux provider for that test — happy to move it into `openmls_test` if you'd rather not have that edge.
- CHANGELOG entry.

### Testing
- `cargo test -p openmls_libcrux_crypto -F draft-ietf-mls-pq-ciphersuites` — 2/2 green.
- `cargo test --release -p openmls --no-default-features -F all-ciphersuites,libcrux-provider,draft-ietf-mls-pq-ciphersuites -- libcrux_MLS_ MLDSA` — 2100 passed / 0 failed locally (238 tests per new libcrux suite × 3, plus the RustCrypto ML-DSA and other libcrux suites matched by the filter).
- clippy `-D warnings` clean with and without the feature.

### Notes / questions for maintainers
- Signing regenerates the key pair from the seed on every call (same approach as `ml_dsa::SigningKey::from_seed` in the RustCrypto provider). If you'd prefer to store the expanded signing key instead, that changes the wire format for both providers — flagging rather than deciding.
- If enabling `draft-connolly-cfrg-hpke-mlkem` on `hpke-rs-libcrux` is unwanted, I can drop the pure-ML-KEM suites and keep only the X-Wing + MLDSA44 one.


libcrux_crypto/src/crypto.rs
```diff
@@ -251,6 +261,15 @@ impl OpenMlsCrypto for CryptoProvider {
     }
 
     fn signature_key_gen(&self, alg: SignatureScheme) -> Result<(Vec<u8>, Vec<u8>), CryptoError> {
+        #[cfg(feature = "draft-ietf-mls-pq-ciphersuites")]
+        if let Some(mldsa) = ml_dsa::Variant::from_scheme(alg) {
```
franziskuskiefer: Having a bunch of `if let`s here won't scale nicely. How about unifying the variants into a match?

libcrux_crypto/src/crypto.rs
```diff
@@ -651,3 +683,117 @@ mod tests {
...
+    pub(super) enum Variant {
+        MlDsa44,
+        MlDsa65,
+        MlDsa87,
+    }
+
+    /// Generates `key_gen`, `sign` and `verify` for one parameter set, since
+    /// libcrux exposes each variant as its own module with distinct fixed-size
+    /// key/signature types.
+    macro_rules! variant_fns {
```
franziskuskiefer: Is there a reason to have `impl Variant` as well as free functions?

franziskuskiefer (changes requested):
Thanks. This generally looks ok, but there's some cleanup that would be nice.

Also, the basic credential as defined right now doesn't actually use the provider, hence doesn't use the signature schemes added here. If the goal is to make use of this in the credential, we also need to use the provider in the basic credential. Or do you have a credential where you use this?

libcrux_crypto/src/crypto.rs
niyamvora: Agreed — folded into the existing matches (Ed25519 arm, feature-gated ML-DSA arm, _ => Unsupported) in all three functions. Pushed in d86a965.

libcrux_crypto/src/crypto.rs
niyamvora: No good reason — it was two layers doing one job. Collapsed into a single with_parameter_set! dispatch macro; the ml_dsa module is now just verification_key / sign / verify matching on the scheme. Same commit.

niyamvora:
We use openmls_basic_credential (A messenger on OpenMLS 0.8.1). We're moving to the PQ ciphersuites in production — X-Wing now, ML-DSA when the draft gets code points — and want the verified libcrux path end-to-end, so you're right that this PR only covers the verify side plus any provider-based credential. I've kept this PR provider-only and opened #2172 for the credential: SignatureKeyPair::generate(crypto, scheme) generating via provider.crypto().signature_key_gen(), plus SignatureKeyPair::signer(crypto) returning a small provider-backed Signer that signs through provider.crypto().sign(); the existing constructor and Signer impl are unchanged, and the test checks that keys from either path are interchangeable on both providers. If you'd rather have it in this PR, or a different shape, let me know and I'll adjust.

Cleanups: done and pushed — the three signature functions are single matches now, and the Variant enum + free functions are collapsed into one dispatch macro. While at it I also fixed a feature-unification problem where the dev-dependency enabled the traits' PQ variants unconditionally, which would have broken a plain cargo test -p openmls_libcrux_crypto; the dev-dep's PQ feature is now enabled from this crate's PQ feature instead.

Thanks for the quick review.

niyamvora:
@franziskuskiefer — both cleanups are in (single matches, one dispatch macro), and I've rebased this and #2172 onto main after the 0.9.0 release so they're mergeable again. Happy to adjust anything.

libcrux_crypto/src/crypto.rs
```diff
@@ -68,14 +68,20 @@ impl OpenMlsCrypto for CryptoProvider {
 
         match ciphersuite.signature_algorithm() {
             SignatureScheme::ED25519 => Ok(()),
+            #[cfg(feature = "draft-ietf-mls-pq-ciphersuites")]
+            SignatureScheme::MLDSA44 | SignatureScheme::MLDSA65 | SignatureScheme::MLDSA87 => {
```
franziskuskiefer, line 72: This is still using the old codepoints. Let's update them to the current draft. They are also implementing that now.

libcrux_crypto/Cargo.toml
```diff
@@ -43,6 +44,13 @@ extensions-draft = [
 draft-ietf-mls-pq-ciphersuites = [
     "openmls_traits/draft-ietf-mls-pq-ciphersuites",
     "openmls_memory_storage/draft-ietf-mls-pq-ciphersuites",
+    "dep:libcrux-ml-dsa",
+    # NOTE: on hpke-rs 0.7.0 this only makes ML-KEM available in the libcrux
```
franziskuskiefer: Let's switch to the git version of hpke-rs for now until the next release is out (should happen soon).

libcrux_crypto/src/crypto.rs
```diff
@@ -651,3 +689,89 @@ mod tests {
...
+        };
+    }
+
+    /// The encoded verification key derived from `seed`.
+    pub(super) fn verification_key(
+        alg: SignatureScheme,
+        seed: &[u8; KEY_GENERATION_RANDOMNESS_SIZE],
+    ) -> Result<Vec<u8>, CryptoError> {
+        with_parameter_set!(alg, m => Ok(
+            m::generate_key_pair(*seed).verification_key.as_slice().to_vec()
```
franziskuskiefer: If we take ownerhip of the `seed` here, we shouldn't take a reference. Same for `sign`.

franziskuskiefer (changes requested):
Thanks, I think this mostly looks good.
I have a branch with a bunch more changes here, but we can get this one in first with those few things addressed.

niyamvora:
All three addressed and pushed:

- `HpkeKemType::XWingKemDraft6` now uses `0x647a` (the code point `draft-ietf-hpke-pq` registers for `MLKEM768-X25519`) instead of the obsolete `0x004D`. The ML-DSA (`0x0904–0x0906`) and ML-KEM (`0x0041/0x0042`) values already matched the current drafts, so that was the only stale one I found — happy to adjust if you meant more than that.
- The workspace now patches the four hpke crates to the celabshq/libcrux git repo and the libcrux provider uses `draft-ietf-hpke-pq` instead of the deprecated flag, so ML-KEM and X-Wing get the labeled `DeriveKeyPair`. I put the pin in `[patch.crates-io]` so it's one thing to delete when the release is out.
- Seed and signing randomness are taken by value now.

Full PQ suite through the libcrux provider is green on the new derivation (3375/0 in release, X-Wing suites included).

libcrux_crypto/src/crypto.rs
niyamvora: Done — `0x647a` via `draft-ietf-hpke-pq`, see the new commit.

libcrux_crypto/Cargo.toml
niyamvora: Done — switched to the git repo via `[patch.crates-io]`.

libcrux_crypto/src/crypto.rs
niyamvora: Done, by value now.

libcrux_crypto/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- [#2170](https://github.com/openmls/openmls/pull/2170): ML-DSA-44/65/87 signatures via `libcrux-ml-dsa`, and pure ML-KEM KEMs via `hpke-rs-libcrux`'s `draft-connolly-cfrg-hpke-mlkem`, so the ML-DSA ciphersuites from `draft-ietf-mls-pq-ciphersuites` are now supported by the libcrux provider (behind the `draft-ietf-mls-pq-ciphersuites` feature flag). Key and signature encodings match the RustCrypto provider.
```
franziskuskiefer: This needs an update to reference `draft-ietf-hpke-pq` instead.

franziskuskiefer: I guess we should mention the hpke-pq draft here now.

```suggestion
- [#2170](https://github.com/openmls/openmls/pull/2170): ML-DSA-44/65/87 signatures via `libcrux-ml-dsa`, and pure ML-KEM KEMs via `hpke-rs-libcrux`'s `draft-ietf-hpke-pq`, so the ML-DSA ciphersuites from `draft-ietf-mls-pq-ciphersuites` are now supported by the libcrux provider (behind the `draft-ietf-mls-pq-ciphersuites` feature flag). Key and signature encodings match the RustCrypto provider.
```

franziskuskiefer (approved):
Thanks! lgtm now with that one nit.

PR #2172, opened 2026-08-18
feat(basic_credential): generate and sign through the crypto provider

niyamvora:
Addresses #1755. Follow-up to #2170, addressing the review comment there: `openmls_basic_credential` generates keys and signs with RustCrypto directly, so a provider's `signature_key_gen`/`sign` are never used by the credential.

This adds an opt-in provider path without changing existing behaviour:

- `SignatureKeyPair::generate(crypto: &impl OpenMlsCrypto, scheme)` — key generation via `OpenMlsCrypto::signature_key_gen`.
- `SignatureKeyPair::signer(&self, crypto) -> ProviderSigner<'_, C>` — a `Signer` that signs via `OpenMlsCrypto::sign`.
- `SignatureKeyPair::new` and the existing `Signer` impl on `SignatureKeyPair` are untouched.

Key encodings are the same on both paths (Ed25519 32/32, P-256/P-384 scalar + uncompressed point, ML-DSA seed + encoded verification key), so provider-generated and RustCrypto-generated key pairs are interchangeable — that's asserted in `tests/provider_signer.rs`, which runs against both `openmls_rust_crypto` and `openmls_libcrux_crypto` (with and without `draft-ietf-mls-pq-ciphersuites`). Once #2170 lands, the libcrux half of that test automatically covers ML-DSA too.

Not in this PR (happy to do either if wanted): switching `openmls_test`/the test suite over to `generate` + `signer` so every provider's signature path is exercised, and any deprecation of the RustCrypto-direct constructor.

Testing: `cargo test -p openmls_basic_credential` and `-F draft-ietf-mls-pq-ciphersuites` green; clippy `-D warnings` clean in both configs.


franziskuskiefer (approved):
Thanks, lgtm.

PR #2174, opened 2026-08-19
perf: various reductions of computational complexity

kkohbrok:
This PR reduces computational complexity involved in lookups in various places by using HashSets. Also uses binary search for unmerged leaves.

### DIFF
--- a/openmls/src/group/public_group/mod.rs
+++ b/openmls/src/group/public_group/mod.rs
@@ -239,7 +239,14 @@ impl PublicGroup {
                                 // None would be blank, and we don't care about those
                                 if let Some(intermediate_node) = treesync
                                     .parent(*intermediate_index) {
-                                    if !intermediate_node.unmerged_leaves().contains(leaf_index) {
+                                    // The list of unmerged leaves is sorted,
+                                    // which UnmergedLeaves enforces when it is
+                                    // deserialized.
+                                    if intermediate_node
+                                        .unmerged_leaves()
+                                        .binary_search(leaf_index)
+                                        .is_err()
+                                    {
                                         return Err(CreationFromExternalError::<StorageError>::IntermediateNodeMissingUnmergedLeaf);
                                     }
                                 }
--- a/openmls/src/group/public_group/mod.rs
+++ b/openmls/src/group/public_group/mod.rs
@@ -241,7 +241,7 @@ impl PublicGroup {
                                     .parent(*intermediate_index) {
                                     // The list of unmerged leaves is sorted,
                                     // which UnmergedLeaves enforces when it is
-                                    // deserialized.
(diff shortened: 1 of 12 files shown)


openmls/src/treesync/node/leaf_node.rs
```diff
@@ -512,17 +514,32 @@ impl LeafNode {
         &self,
         extensions: &[ExtensionType],
     ) -> Result<(), LeafNodeValidationError> {
-        for required in extensions.iter() {
-            if !self.supports_extension(required) {
-                log::error!(
-                    "Leaf node does not support required extension {:?}\n
+        let mut required = extensions.iter().filter(|e| !e.is_default()).peekable();
+
+        // Most groups require no non-default extensions. Skip building the
```
franziskuskiefer: I'm not sure if these comments about "most groups" hold up with everything that is being built on top of the main RFC. The code is fine. But maybe update the comments?

openmls/src/group/public_group/mod.rs
```diff
@@ -239,7 +239,14 @@ impl PublicGroup {
                                 // None would be blank, and we don't care about those
                                 if let Some(intermediate_node) = treesync
                                     .parent(*intermediate_index) {
-                                    if !intermediate_node.unmerged_leaves().contains(leaf_index) {
+                                    // The list of unmerged leaves is sorted,
+                                    // which UnmergedLeaves enforces when it is
+                                    // deserialized.
```
franziskuskiefer: Is this always the case? Even for trees previously stored?

franziskuskiefer (changes requested):
Thanks. lgtm with two nits/questions.

openmls/src/group/public_group/mod.rs
kkohbrok: We do check that on TLS deserialization, but not in serde deserialization. So if someone messes with a RatchetTree while on disk, this doesn't necessarily hold anymore, but I think that's not something we need to anticipate.

openmls/src/treesync/node/leaf_node.rs
kkohbrok: Good point. Fixed.

openmls/src/group/public_group/mod.rs
franziskuskiefer: That depends on our model of local attacks or corruptions. I think this is fine, but I also think we should document it somehow so that it is not a surprise. Maybe just add the caveat here to the comment, that storage deserialized trees are expected to have the same properties, but we don't check that.

franziskuskiefer (approved):
Let's get it in, maybe update that comment before.

I'd say let's merge the tests for now and look at them when they make trouble.

PR #2175, opened 2026-08-20
Add `with_aad` method to CommitBuilder

keks:
This PR add a method `with_aad` to the commit builder. This way, the caller can set the AAD for the commit, which completely bypasses the value controlled by `MlsGroup::set_aad`.

Even before this change, external commits had a way to set the AAD this way. The external commit builder code was extended to also use the new member `bare_aad` of the CommitBuilder phase structs.

I don't think we need to rush this in before the release.

### DIFF
--- a/openmls/src/group/mls_group/commit_builder.rs
+++ b/openmls/src/group/mls_group/commit_builder.rs
@@ -97,7 +97,6 @@ use super::HandshakeConfirmationData;
 
 #[derive(Debug)]
 struct ExternalCommitInfo {
-    aad: Vec<u8>,
     /// The authoritative credential and signature key for the external
     /// committer's leaf. `build_internal` folds it into the leaf node
     /// parameters and rejects parameters that pin a different credential.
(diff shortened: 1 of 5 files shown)


franziskuskiefer (changes requested):
CI is not happy

franziskuskiefer (approved):
Thanks. lgtm. 

It needs rebasing and a rustfmt.

Can you add a line to the changelog to tell people about the new API?

openmls/src/group/mls_group/commit_builder.rs
```diff
@@ -120,6 +119,9 @@ pub struct Initial {
     leaf_node_parameters: LeafNodeParameters,
     external_commit_info: Option<ExternalCommitInfo>,
 
+    /// The bare AAD to use for this commit. If it is set, the one from the group is ignored and left untouched.
```
nhanff, line 122: As I see it, "the one from the group is [...] left untouched" is not implemented, at least not tested. `stage_commit` resets the group's AAD no matter what.

https://github.com/openmls/openmls/blob/89b27c5e0030630d775ad25fef24e6795b4e4733/openmls/src/group/mls_group/commit_builder.rs#L1356-L1399

PR #2179, opened 2026-08-21
virtual clients: derivation epoch liveness tracking for groups

kkohbrok:
This PR adds some bookkeeping to allows OpenMLS to track which epoch ID is bound to which group. This additional tracking allows OpenMLS to clean up key material from unused, old derivation epochs. That clean-up (bound to an application-defined policy) comes in a follow up.

### DIFF
--- a/memory_storage/src/lib.rs
+++ b/memory_storage/src/lib.rs
@@ -263,8 +263,16 @@ const APPLICATION_EXPORT_TREE_LABEL: &[u8] = b"ApplicationExportTree";
 const VC_DERIVATION_EPOCH_STATE_LABEL: &[u8] = b"VcDerivationEpochState";
 #[cfg(feature = "virtual-clients-draft")]
 const VC_EMULATION_BINDING_LABEL: &[u8] = b"VcEmulationBinding";
+// The binding record is opaque, so the epochs it binds are kept next to it as
+// a separate entry that the epoch-liveness scan can read.
+#[cfg(feature = "virtual-clients-draft")]
+const VC_BOUND_EPOCHS_LABEL: &[u8] = b"VcBoundDerivationEpochs";
 #[cfg(feature = "virtual-clients-draft")]
 const REGISTERED_VC_DERIVATION_EPOCH_LABEL: &[u8] = b"RegisteredVcDerivationEpoch";
+// The registration record is opaque, so the epoch it names is kept next to it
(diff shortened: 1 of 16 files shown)


openmls/src/components/vc_derivation_info.rs
```diff
@@ -1005,6 +1005,31 @@ impl VcEmulationBindings {
...
+    /// order of first appearance. A derivation epoch stays bound while the
+    /// virtual-client LeafNode it produced is active, so the same epoch id
+    /// usually appears under several higher-level epochs.
+    pub fn bound_epoch_ids(&self) -> Vec<EpochId> {
+        let mut epoch_ids = Vec::with_capacity(self.bindings.len());
+        for (_, epoch_id) in &self.bindings {
+            if !epoch_ids.contains(epoch_id) {
+                epoch_ids.push(epoch_id.clone());
+            }
+        }
```
boxdot: I think it makes sense to avoid n**2 dedup here. I would propose to add Ord to EpochId and use sort_unstable + dedup. This has constant memory and good enough n log(n) runtime.

memory_storage/src/lib.rs
```diff
@@ -1142,26 +1172,41 @@ impl StorageProvider<CURRENT_VERSION> for MemoryStorage {
         &self,
         group_id: &GroupId,
     ) -> Result<(), Self::Error> {
-        self.delete::<CURRENT_VERSION>(
-            VC_EMULATION_BINDING_LABEL,
-            &serde_json::to_vec(group_id).unwrap(),
-        )
+        let serialized_group_id = serde_json::to_vec(group_id)?;
+        self.delete::<CURRENT_VERSION>(VC_EMULATION_BINDING_LABEL, &serialized_group_id)?;
+        self.delete::<CURRENT_VERSION>(VC_BOUND_EPOCHS_LABEL, &serialized_group_id)
```
boxdot: The write lock is acquired 2x here. Should we do the same as in `write_vc_emulation_bindings` and acquire it only once? Otherwise, the second entry is visible after the first one is gone.

openmls/tests/virtual_clients.rs
boxdot, line 1: No test writes a group's bindings twice with a shrunk bound_epochs and checks the dropped epoch becomes deletable.

boxdot (approved):

PR #2135, opened 2026-07-27
specs: specify treemath functions

clementblaudeau:
This PR adds hax_lib specs to the treemath functions. The proofs are PR-ed separately, in the following PR of the stack.

## Overview

Functions are annotated with `requires` and `ensures`. For instance : 
```rust
#[requires(x.valid() && x.u32() != MAX_ROOT_INDEX)]
#[ensures(|r| r.valid() && level(r.to_tree_index()) == level(x.u32()) + 1)]
fn parent(x: TreeNodeIndex) -> ParentNodeIndex {
    let x = x.u32();
    let k = level(x);
    let b = (x >> (k + 1)) & 0x01;
    let index = (x | (1 << k)) ^ (b << (k + 1));
    ParentNodeIndex::from_tree_index(index)
}
```

Those specs help make the requirements explicit.

## Details

This PR introduces specs some more top-level explicit bounds : 

```rust
const MAX_TREE_INDEX: u32 = MAX_TREE_SIZE - 1;
const MAX_LEAF: u32 = MAX_TREE_INDEX / 2;
const MAX_PARENT: u32 = MAX_LEAF - 1;
const MAX_LEAF_COUNT: u32 = MAX_LEAF + 1;
const MAX_ROOT_INDEX: u32 = MAX_TREE_SIZE / 2;
```

Besides, we define `valid` methods for `LeafNode`, `ParentNode`, `TreeNode`, `TreeSize`. For instance:

```rust
#[attributes]
impl TreeSize {
    /// Used only for verification.
    fn valid(&self) -> bool {
        MIN_TREE_SIZE <= self.0
            && self.0 <= MAX_TREE_SIZE
            // the size is of the form 2^k - 1, where k = log2(self.0) + 1
            && self.0 == 2u32.pow(log2(self.0) as u32 + 1) - 1
    }
}
```

They are used in pre/post conditions.

PR #2182, opened 2026-08-22
fix+test: enforce treemath invariants

clementblaudeau:
This PR sits on top of https://github.com/openmls/openmls/pull/2181. Part of the verification effort consisted in adding explicit specs for treemath functions. An AI-audit of the call sites found two cases were the stated invariants could be violated. 

- `is_node_in_tree` was not checking that the index was valid, which could led to accepting impossible nodes and panic.
- unmerged leaves indexes were not checked. This could lead to an infinite loop.

This PR adds two tests and fixes.

 A wider verification effort could allow us to show that the treemath library is correctly used throughout the whole project.


clementblaudeau:
Should be reopened after merging of #2135

PR #2186, opened 2026-08-24
Fix minor treemath issues

keks:
We had two cases where treemath-related functions performed operations without failing. This led to OpenMLS sometimes accepting commits from invalid senders.

This PR is based on #2182 by @clementblaudeau from @cryspen.


kkohbrok (approved):
Thanks!

PR #2189, opened 2026-08-26
virtual clients: derivation epoch retention policy

kkohbrok:
This PR introduces a retention policy to derivation epochs similar to the existing policies for message secrets of past epochs. This PR also introduces some facilities for applications that already use the virtual clients feature to migrate existing state.


openmls/src/components/vc_derivation_info.rs
```diff
@@ -772,38 +780,170 @@ pub(crate) struct VcWelcomeMaterial {
...
+}
+
+/// Per-emulation-group log of the derivation epochs the group registered, in
+/// registration order with the newest at the back. The newest entry is the
+/// derivation epoch all new virtual-client operations of the group resolve to,
+/// which may be older than the group's current epoch.
+#[derive(Debug, Default)]
+pub(crate) struct VcDerivationEpochLog {
+    // In registration order, oldest at the front.
+    entries: std::collections::VecDeque<VcDerivationEpochLogEntry>,
```
boxdot: Nit: for readability I think it makes sense to import such types.

openmls/src/components/vc_derivation_info.rs
```diff
@@ -772,38 +780,170 @@ pub(crate) struct VcWelcomeMaterial {
...
+        debug_assert!(
+            entries
+                .iter()
+                .map(|entry| &entry.epoch_id)
+                .collect::<std::collections::BTreeSet<_>>()
+                .len()
+                == entries.len(),
+            "duplicate derivation epoch id in log"
+        );
+        Ok(Self {
```
boxdot, line 852: Technically, since you sorted, you can also dedup after to be more forgiving to faulty implementations of the storage provider. Feel free to ignore.

openmls/src/components/vc_derivation_info.rs
```diff
@@ -772,38 +780,170 @@ pub(crate) struct VcWelcomeMaterial {
...
-/// registration and cannot be re-evaluated).
-///
-/// Not folded into [`VcEmulationBindings`]: bindings are per higher-level group
-/// and are carried forward to the new epoch when a merged commit installs no
-/// virtual-client leaf, so they cannot say which derivation epoch is newest.
+/// One registration in an emulation group's log of derivation epochs, stored
+/// as its own row keyed by `(group_id, epoch_id)`.
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-pub(crate) struct RegisteredVcDerivationEpoch {
+pub struct VcDerivationEpochLogEntry {
```
boxdot, line 787: I think it makes sense to add a test pinning the serde layout of this type.

openmls/src/components/vc_derivation_info.rs
```diff
@@ -907,29 +1053,28 @@ pub(crate) fn register_vc_derivation_epoch<
...
     let export_tree =
         export_tree.ok_or(RegisterVcDerivationEpochError::MissingApplicationExportTree)?;
 
-    let registered: Option<RegisteredVcDerivationEpoch> = storage
-        .registered_vc_derivation_epoch(group_id)
-        .map_err(|e| {
-            log::error!("vc: load newest derivation epoch before registration failed: {e:?}");
-            RegisterVcDerivationEpochError::Storage(e)
-        })?;
+    let mut log = VcDerivationEpochLog::load(storage, group_id).map_err(|e| {
```
boxdot, line 1062: With the policy `KeepAll` this will load epochs into memory. Is this acceptable? 

It seems that we actually need the newest epoch. The history is only there to be pruned. For sql based providers it seems that loading all log entries for pruning only is more expensive, than pruning directly in the db. 

I investigated a little, the API might look like this:

- `write_vc_derivation_epoch_log_entry(group_id, epoch_id, sequence: u64, registered_at: Duration, entry)`
  `sequence` and `registered_at` become columns so the provider can order without decoding the blob.
- `newest_vc_derivation_epoch_log_entry(group_id) -> Option<Entry>`
  Point read by highest sequence; replaces the whole-log read on the hot path.
- `prune_vc_derivation_epoch_log_to(group_id, keep: usize) -> Vec<EpochId>`
  Delete all but the `keep` newest, return what was dropped. `keep >= 1`.
- `prune_vc_derivation_epoch_log_superseded_before(group_id, cutoff: Duration) -> Vec<EpochId>`
  Delete entries whose successor (smallest greater sequence in the group) registered before `cutoff`, return what was dropped. Newest never drops.
- `delete_vc_derivation_epoch_log(group_id)` unchanged.

Independently, whether we change the provider API or keep it as is, it seems that loading of logs can be avoid most of the time.

boxdot: Discussed privately: we can do this optimization on top in a follow-up. Since VC feature is not released yet, we can break the provider API.

openmls/src/group/mls_group/creation.rs
```diff
@@ -686,6 +689,25 @@ impl StagedWelcome {
...
+        // A join through a virtual client's KeyPackage binds the joined epoch
+        // to the KeyPackage's derivation epoch, which also keeps the epoch
+        // referenced after `keys_for_welcome` consumed the retained material.
+        // Written before the group itself, so an error between the writes
+        // cannot leave a loadable group without a binding (a bound group is
+        // required for the reuse-guard MUST).
+        #[cfg(feature = "virtual-clients-draft")]
+        if let Some(epoch_id) = self.key_material.vc_epoch_id() {
+            let max_entries = mls_group.message_secrets_store.max_epochs.saturating_add(1);
+            crate::components::vc_derivation_info::write_vc_emulation_binding_with_pruning(
```
boxdot, line 700: If am not mistaken the key package is deleted before, which means the epoch is not kept alive anymore, and so it can be deleted before this call happens. I think the test welcome_join_takes_over_epoch_reference_from_retained_material actually covers it but the gap is so small, that it does not trigger.

boxdot (approved):
I think there is a race in create group otherwise everything looks fine.

openmls/src/components/vc_derivation_info.rs
kkohbrok: Wouldn't we want that bug to surface rather than hide it? Feel free to change it in a follow-up, but I'm leaving it as is for now.

openmls/src/group/mls_group/creation.rs
kkohbrok: Good catch! Should be fixed now.

Issue #2180, opened 2026-08-22
[Feature request] A supported way to carry a rejoining member's own past-epoch secrets across an external commit


z0hrab:
# [Feature request] A supported way to carry a rejoining member's own past-epoch secrets across an external commit

## Summary

When a member rejoins its own group via an external commit (RFC 9420 §12.4.3.2), there is no supported way to carry over the `max_past_epochs` of message secrets that same member was already holding. The rejoining device loses the ability to decrypt application messages it could have decrypted a moment earlier — messages from epochs it legitimately participated in.

I don't think this is a forward-secrecy question, and I've tried to argue that carefully below. I'd rather be told I'm wrong about that than have the request taken at face value.

## Why the secrets are lost

The two group states collide in the storage provider. `ExternalCommitBuilder::finalize` produces an `MlsGroup` for the *same* `GroupId` the consumer already holds, and the storage keys are derived from that id — so a consumer cannot keep both. One has to be discarded, and it's the old one, which is the only one holding the retained `MessageSecretsStore`.

In our code (openmls 0.8.1, `max_past_epochs(16)`) that discard is explicit:

```rust
pub fn adopt(self, real: &Provider) -> Result<Group, Error> {
    if let Some(stale) = Group::load(real, &self.group_id)? {
        stale.delete(real)?;          // ← every retained past-epoch secret
    }
    // graft the scratch provider's entries into the live keystore
}
```

We build the external commit in a throwaway provider first, because `finalize` merges immediately and admission is our server's call — so we can't let it touch live state before we know the commit won. That isolation is what makes the discard unavoidable rather than incidental.

`MessageSecretsStore` and `MlsGroup::set_max_past_epochs` are `pub(crate)`, so there's no way to move the retained epochs from the outgoing state into the new one.

### "Couldn't you just keep the old state in a second provider?"

Technically yes — that is how we build the external commit in the first place, and it is the obvious workaround, so it is worth saying why it isn't a solution.

[...]

PR #2194, opened 2026-08-26
fix: decide own-sender by the secret tree's leaf, not the current one

### LINKED ISSUE #2192: [Bug] virtual clients: after a sibling-resync external commit, pre-resync messages are misclassified
**Describe the bug**

`DecryptedMessage::from_inbound_ciphertext` decides whether an inbound `PrivateMessage` is this client's own by comparing the sender's leaf index against `group.own_leaf_index()`:

```rust
// openmls/src/framing/validation.rs
let (message_secrets, _old_leaves) = group.message_secrets_and_leaves(ciphertext.epoch())?;
let sender_data = ciphertext.sender_data(message_secrets, crypto, ciphersuite)?;
let own_sender = sender_data.leaf_index == group.own_leaf_index();
```

`own_leaf_index()` is the index in the *current* tree, but the message being decrypted belongs to whatever epoch it was sent in. The two stop agreeing as soon as the client's own leaf moves.

With `virtual-clients-draft` it moves. A sibling-resync external commit installs the joiner's leaf as our own (`staged_commit.rs`, `if let Some(new_idx) = state.new_own_leaf_index`, added in #2031). Every retained epoch from before the resync keeps a secret tree built for the *old* leaf, so from that point on the own-sender question is answered against the wrong leaf for all of them — in both directions.

This only affects builds with `virtual-clients-draft`. The assignment above is the only one to `MlsGroup::own_leaf_index` after group creation, and it is cfg-gated, so elsewhere the two indices always agree.

**To Reproduce**

The resync has to actually move the leaf, which needs a blank to the left of the virtual client — so the group is created by a member who is removed before the resync:

```text
  leaf 0      leaf 1      leaf 2
  Dave        Alice (VC)  Bob      epoch 1: Dave sends, message left unread
  -           Alice (VC)  Bob      epoch 2: Dave removed; Alice and Bob send, unread
  Alice (VC)  -           Bob      epoch 3: resync, Alice moves leaf 1 -> leaf 0
```

1. Dave creates a group with `PastEpochDeletionPolicy::MaxEpochs(n)`, `n > 0`, and adds the virtual client (leaf 1) and Bob (leaf 2).
2. Dave sends an application message from leaf 0. Alice's client does not read it yet.
3. Bob removes Dave, blanking leaf 0.
4. Bob and the virtual client each send an application message. Alice's client does not read them yet.
5. Alice's second client joins by sibling-resync external commit. Its auto-`Remove` takes the virtual client's leaf 1, and the joiner lands on the blank leaf 0. Alice's first client processes and merges that commit, so `own_leaf_index` moves from 1 to 0.
6. Alice's first client now reads the three messages from steps 2 and 4.

**Expected behavior**

| message | expected |
| --- | --- |
| Dave's, from leaf 0 in epoch 1 | `ProcessedMessageContent::ApplicationMessage` |
| Bob's, from leaf 2 in epoch 2 | `ProcessedMessageContent::ApplicationMessage` |
| the virtual client's own echo, epoch 2 | `ProcessedMessageContent::OwnPrivateMessage` |

**Actual behavior**

Bob's message, the control, is fine — his leaf never moved. The other two are both wrong, and in opposite directions:

- **Dave's message is silently swallowed.** He sent from the leaf the client has since moved onto, so `own_sender` is wrongly `true` and the message short-circuits into `ProcessedMessageContent::OwnPrivateMessage`. No decryption is attempted, no error is raised, and the application never learns the message existed. This is the worse of the two, because it is quiet.

- **The client's own echo becomes a hard error.** The delivery service echoes own sends back, and an own send has already consumed its generation on the sending ratchet, so the echo cannot be decrypted — which is why the own-sender check exists at all. After the resync `own_sender` is wrongly `false` for the client's own pre-resync messages, the consumed generation is reached, and `SecretTreeError::SecretReuseError` surfaces to the caller instead of `OwnPrivateMessage`.

**Screenshots or debug log**

Reading Dave's message:

```
expected Dave's application message, got OwnPrivateMessage
```

Reading the own echo:

```
ERROR openmls::framing::private_message_in >   Ciphertext generation out of bounds 0
        SecretReuseError

ValidationError(UnableToDecrypt(SecretTreeError(SecretReuseError)))
```

- OS & OS version: macOS 15 (Darwin 25.5.0); not platform-specific
- Architecture: aarch64
- Version: `main` @ e725f587, built with `--no-default-features -F virtual-clients-draft,virtual-clients-draft-test-dependencies`
- Ciphersuite: `MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519`, `openmls_rust_crypto`

**Additional context**

`MessageSecrets` already holds the answer the check needs: the epoch's secret tree records the leaf it was built for, which is the group's own leaf index at the time the epoch was entered. Comparing against that instead of `group.own_leaf_index()` makes the question epoch-local, and is a no-op wherever the leaf never moved.

I have a patch and a regression test for the scenario above, and will open a PR referencing this issue.

mtrossbach:
Fixes #2192.

## The change

`DecryptedMessage::from_inbound_ciphertext` asks whether an inbound `PrivateMessage` came from this client by comparing the sender's leaf index against `group.own_leaf_index()`. That is the index in the *current* tree, while the message being decrypted belongs to whatever epoch it was sent in — so the comparison is only correct while the client's own leaf has never moved.

Under `virtual-clients-draft` it moves: a sibling-resync external commit installs the joiner's leaf as our own (`staged_commit.rs`, `if let Some(new_idx) = state.new_own_leaf_index`). Every retained epoch from before the resync keeps a secret tree built for the old leaf, so from that point on the own-sender question is answered against the wrong leaf for all of them, in both directions:

- **wrongly "mine"** — a message another member sent from the leaf the client has since moved onto short-circuits into `ProcessedMessageContent::OwnPrivateMessage`: no decryption attempted, no error, the caller never learns it existed;
- **wrongly "not mine"** — the client's own pre-resync echo no longer short-circuits, reaches the already-consumed generation on its own sending ratchet, and fails with `SecretTreeError::SecretReuseError`.

`MessageSecrets` already knows the right answer: the epoch's secret tree records the leaf it was built for, which is the group's own leaf index at the time the epoch was entered. The comparison now uses that, through a new `SecretTree::own_index` accessor surfaced as `MessageSecrets::own_index`.

The production change, without the comment that explains it:

```rust
+        let own_index = message_secrets.own_index();
         let sender_data = ciphertext.sender_data(message_secrets, crypto, ciphersuite)?;
-        let own_sender = sender_data.leaf_index == group.own_leaf_index();
+        let own_sender = sender_data.leaf_index == own_index;
```

## Side effects and compatibility

No public API change; both accessors are `pub(crate)`.

Builds **without** `virtual-clients-draft` are unaffected. The only assignment to `MlsGroup::own_leaf_index` after group creation is the cfg-gated sibling-resync one, so outside that feature the current own index and the epoch's secret-tree index are always the same value and the substitution is behaviour-preserving.

Two neighbouring cases I checked explicitly:

- `kat_message_protection` deliberately forces `set_own_leaf_index(sender_index)` while installing a secret tree built for that same index, so the two stay consistent there and the KAT is unaffected.
- The `ExternalCommitBuilder` seeds the pre-join epoch's `MessageSecrets` with a "fake own index of 0" (`external_commits.rs`, tracked in #767), which is the one place outside `virtual-clients-draft` where the two indices can differ. That epoch's `sender_data_secret` is derived from an all-zero epoch secret rather than the group's real one, so `ciphertext.sender_data(..)` fails its AEAD open before the own-sender line is reached — with or without this change.

## Alternatives considered

- **Compare against `own_leaf_index()` only for the current epoch, and skip the check for past epochs.** That reintroduces the `SecretReuseError` for own echoes of past-epoch sends, which is the symptom `OwnPrivateMessage` exists to avoid.
- **Record the own leaf index alongside each retained epoch in `MessageSecretsStore`.** This is the same information the epoch's secret tree already carries, so it would be a second source of truth to keep in sync.
- **Fix it in `MlsGroup::merge_staged_commit` by rewriting the retained epochs' trees on a resync.** The retained trees are correct as they stand — they belong to the leaf they were built for. Only the question asked about them was wrong.

## Verification

`openmls/tests/virtual_clients.rs::vc_past_epoch_read_survives_sibling_resync` builds the tree in which the resync actually moves the leaf (a blank to the left of the virtual client, so the group is created by a member who is removed before the resync) and then reads three pre-resync messages: the removed member's, an unaffected member's as a control, and the client's own echo. `test_boundaries` in `test_secret_tree.rs` gains one line asserting the new accessor reports the index the tree was built for.

What I ran, on `main` @ e725f587 + this commit:

```
$ cargo test -p openmls --no-default-features \
    -F virtual-clients-draft,virtual-clients-draft-test-dependencies \
    --test virtual_clients vc_past_epoch_read_survives_sibling_resync
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 67 filtered out
```

The same test with only `openmls/src/framing/validation.rs` reverted to `main`, everything
else unchanged, fails on the first of the two symptoms:

```
panicked at openmls/tests/virtual_clients.rs:6292:18:
expected Dave's application message, got OwnPrivateMessage
```

and, with just that first assertion relaxed so the run reaches the third message, on the
second:

```
ERROR openmls::framing::private_message_in >   Ciphertext generation out of bounds 0
        SecretReuseError

panicked at openmls/tests/virtual_clients.rs:6319:10:
the echo of a message the virtual client sent from its old leaf, in an epoch whose secret
tree belongs to that same old leaf, must still be recognised as its own rather than failing
to decrypt: ValidationError(UnableToDecrypt(SecretTreeError(SecretReuseError)))
```

Bob's message, the control, reads correctly in every run that gets as far as it.

No regressions, full suite both ways, plus the checks CI runs:

```
$ cargo test -p openmls --no-default-features \
    -F virtual-clients-draft,virtual-clients-draft-test-dependencies
ok, 0 failed  (373 lib + 24 integration binaries + doctests)

$ cargo test -p openmls --no-default-features
ok, 0 failed  (285 lib + integration binaries + doctests)

$ cargo clippy -p openmls --tests -- -D warnings
clean

$ cargo fmt -- --check
clean
```

Running the whole `virtual_clients` binary with `validation.rs` reverted gives `67 passed; 1 failed` — the new test is the only one in the suite that depends on the change.

## User-facing description

Fixed: with the `virtual-clients-draft` feature, after a sibling-resync external commit moved the client to a different leaf, messages from epochs before the resync were classified against the wrong leaf — another member's message could be silently reported as this client's own, and the client's own echo could fail to decrypt.

### DIFF
--- a/openmls/src/framing/validation.rs
+++ b/openmls/src/framing/validation.rs
@@ -157,8 +157,9 @@ impl DecryptedMessage {
         let (message_secrets, _old_leaves) = group
             .message_secrets_and_leaves(ciphertext.epoch())
             .map_err(MessageDecryptionError::SecretTreeError)?;
+        let own_index = message_secrets.own_index();
         let sender_data = ciphertext.sender_data(message_secrets, crypto, ciphersuite)?;
-        let own_sender = sender_data.leaf_index == group.own_leaf_index();
+        let own_sender = sender_data.leaf_index == own_index;
         // If we are the sender, the content cannot be decrypted and the
         // signature cannot be verified: the own sender ratchet only produces
         // encryption keys. Return early before touching any ratchet state so
--- a/openmls/src/schedule/message_secrets.rs
+++ b/openmls/src/schedule/message_secrets.rs
@@ -101,6 +101,12 @@ impl MessageSecrets {
             ..self
         }
     }
+
+    /// The leaf index this epoch's secret tree was built for, i.e. the group's
+    /// own leaf index at the time the epoch was entered.
+    pub(crate) fn own_index(&self) -> LeafNodeIndex {
+        self.secret_tree.own_index()
+    }
 }
 
 // Test functions
--- a/openmls/src/tree/secret_tree.rs
+++ b/openmls/src/tree/secret_tree.rs
@@ -184,6 +184,11 @@ impl SecretTree {
         secret_tree
     }
 
+    /// The leaf index this tree was built for.
+    pub(crate) fn own_index(&self) -> LeafNodeIndex {
+        self.own_index
+    }
+
     /// Get current generation for a specific SenderRatchet
     #[cfg(test)]
     pub(crate) fn generation(&self, index: LeafNodeIndex, secret_type: SecretType) -> u32 {
--- a/openmls/tests/virtual_clients.rs
+++ b/openmls/tests/virtual_clients.rs
@@ -5985,3 +5985,227 @@ fn new_derivation_epoch_requires_emulation_group() {
         "unexpected error: {err:?}"
     );
 }
+
+/// Reading a *past* epoch after a sibling-resync external commit.
+///
+/// The resync moves `own_leaf_index` to the joiner's new leaf
+/// (`staged_commit.rs`, `if let Some(new_idx) = state.new_own_leaf_index`).
+/// The secret trees of epochs before the resync were built for the *old*
+/// index, so from that point on "the group's current own leaf" and "the leaf a
+/// past epoch's secret tree belongs to" are two different things.
+///
+/// `DecryptedMessage::from_inbound_ciphertext` decides whether an inbound
+/// message is the caller's own by comparing the sender's leaf index against
+/// one of those two. Comparing against the current one misclassifies every
+/// message in every pre-resync epoch that came from the leaf the client has
+/// since moved onto -- here Dave's, who was removed before the resync and
+/// whose leaf the joiner then reuses.
+///
+/// The tree is arranged so that the resync actually moves the leaf, which
+/// needs a blank to the left of the virtual client:
+///
+/// ```text
+///   leaf 0      leaf 1      leaf 2
+///   Dave        Alice (VC)  Bob      epoch 1: Dave sends
+///   -           Alice (VC)  Bob      epoch 2: Dave removed; Alice and Bob send
+///   Alice (VC)  -           Bob      epoch 3: resync, Alice moves 1 -> 0
+/// ```
+#[openmls_test]
+fn vc_past_epoch_read_survives_sibling_resync() {
+    use openmls::group::PastEpochDeletionPolicy;
+    use openmls::prelude::LeafNodeIndex;
+
+    let dave_provider = Provider::default();
+    let alice_a_provider = Provider::default();
+    let alice_b_provider = Provider::default();
+    let bob_provider = Provider::default();
+
+    let (vc_signer, vc_credential) =
+        shared_vc_identity(ciphersuite, &alice_a_provider, &alice_b_provider);
+
+    let join_config = MlsGroupJoinConfig::builder()
+        .wire_format_policy(PURE_PLAINTEXT_WIRE_FORMAT_POLICY)
+        .use_ratchet_tree_extension(true)
+        .set_past_epoch_deletion_policy(PastEpochDeletionPolicy::MaxEpochs(10))
+        .build();
+
+    // Higher-level group. Dave creates it and takes leaf 0, so that removing
+    // him later leaves a blank to the left of the virtual client -- without
+    // that blank the joiner reuses the virtual client's own leaf and
+    // `own_leaf_index` never moves.
+    //
+    // The past-epoch buffer has to be large enough to still hold the epochs
+    // the messages below are sent in; the default `MaxEpochs(0)` would drop
+    // them at the next commit and the reads would fail for an unrelated
+    // reason.
+    let group_config = MlsGroupCreateConfig::builder()
+        .wire_format_policy(PURE_PLAINTEXT_WIRE_FORMAT_POLICY)
+        .ciphersuite(ciphersuite)
+        .use_ratchet_tree_extension(true)
+        .set_past_epoch_deletion_policy(PastEpochDeletionPolicy::MaxEpochs(10))
+        .capabilities(vc_capabilities())
+        .with_leaf_node_extensions(vc_leaf_extensions())
+        .expect("attach leaf-node extensions on higher-level config")
+        .build();
+    let (dave_credential, dave_signer) =
+        new_credential(&dave_provider, b"Dave", ciphersuite.signature_algorithm());
+    let mut dave_main = MlsGroup::new(&dave_provider, &dave_signer, &group_config, dave_credential)
+        .expect("dave create higher-level group");
+    assert_eq!(dave_main.own_leaf_index(), LeafNodeIndex::new(0));
+
+    // Dave adds the virtual client (leaf 1) and Bob (leaf 2).
+    let alice_vc_kp = KeyPackage::builder()
+        .key_package_extensions(Extensions::empty())
+        .leaf_node_capabilities(vc_capabilities())
+        .leaf_node_extensions(vc_leaf_extensions())
+        .build(
+            ciphersuite,
+            &alice_a_provider,
+            &vc_signer,
+            vc_credential.clone(),
+        )
+        .expect("alice VC KP build")
+        .key_package()
+        .to_owned();
+    let (bob_kp, bob_signer) = vc_key_package(ciphersuite, &bob_provider, b"Bob");
+    let (_, welcome, _) = dave_main
+        .add_members(&dave_provider, &dave_signer, &[alice_vc_kp, bob_kp])
+        .expect("dave adds alice and bob");
+    dave_main
+        .merge_pending_commit(&dave_provider)
+        .expect("dave merge add");
+    let ratchet_tree = dave_main.export_ratchet_tree();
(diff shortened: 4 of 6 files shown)


openmls/src/tree/secret_tree.rs
```diff
@@ -184,6 +184,12 @@ impl SecretTree {
         secret_tree
     }
 
+    /// The leaf index this tree was built for. Its ratchets are encryption
+    /// ratchets; every other leaf gets a decryption ratchet.
```
kkohbrok:
```suggestion
    /// The leaf index this tree was built for.
```

openmls/src/schedule/message_secrets.rs
```diff
@@ -101,6 +101,13 @@ impl MessageSecrets {
             ..self
         }
     }
+
+    /// The leaf index this epoch's secret tree was built for, i.e. the group's
+    /// own leaf index at the time the epoch was entered. See the note in
+    /// `DecryptedMessage::from_inbound_ciphertext`.
```
kkohbrok:
```suggestion
    /// The leaf index this epoch's secret tree was built for, i.e. the group's
    /// own leaf index at the time the epoch was entered.
```
It's a bit odd for function-level doc comments to reference inline-code comments.

openmls/src/framing/validation.rs
```diff
@@ -157,8 +157,16 @@ impl DecryptedMessage {
         let (message_secrets, _old_leaves) = group
             .message_secrets_and_leaves(ciphertext.epoch())
             .map_err(MessageDecryptionError::SecretTreeError)?;
+        // Which leaf is "ours" is a question about the epoch the message was
+        // sent in, not about the current one. The secret tree of that epoch
+        // records the leaf it was built for, which is the group's own leaf
+        // index at the time; `group.own_leaf_index()` is the current one, and
+        // the two differ for every epoch before the client's leaf last moved.
+        // Under `virtual-clients-draft` a sibling-resync external commit moves
+        // it (see `MlsGroup::merge_staged_commit`).
```
kkohbrok: Please either boil this down to 1-2 lines or just remove it completely.

openmls/tests/virtual_clients.rs
```diff
@@ -5985,3 +5985,347 @@ fn new_derivation_epoch_requires_emulation_group() {
...
+/// needs a blank to the left of the virtual client:
+///
+/// ```text
+///   leaf 0      leaf 1      leaf 2
+///   Dave        Alice (VC)  Bob      epoch 1: Dave sends
+///   -           Alice (VC)  Bob      epoch 2: Dave removed; Alice and Bob send
+///   Alice (VC)  -           Bob      epoch 3: resync, Alice moves 1 -> 0
+/// ```
+#[openmls_test]
+fn vc_past_epoch_read_survives_sibling_resync() {
```
kkohbrok, line 6014: I think quite a bit of the setup in this test function is covered by helpers in this file. Could you use those? That would make the test a bit leaner.

kkohbrok (changes requested):
Thanks for catching the bug and filing the PR! The fix is also relevant should we implement a solution for #2180.

I left a few comments, mostly to clean up overly verbose LLM output. I'm guilty of just merging those myself in the past, but I think it's worth not letting them get out of hand.

Virtual clients is still quite experimental both on the spec and on the implementation level. If you're playing around with it I (and the other draft authors) would appreciate feedback if you have any :-)

openmls/src/tree/secret_tree.rs
mtrossbach: Done.

openmls/src/schedule/message_secrets.rs
mtrossbach: Done, reference dropped with it.

openmls/src/framing/validation.rs
mtrossbach: Removed it. `own_index()` already documents what the index is, and if someone changes the comparison back to `group.own_leaf_index()` the regression test fails.

openmls/tests/virtual_clients.rs
mtrossbach: That trimmed it from 318 to 198 lines: `shared_vc_identity`, `join_sibling_emulator`, `vc_key_package` and `process_and_merge_commit`.

Two I left alone:

* `new_vc_main_group_with_policy` founds the group on the VC's leaf. This setup needs a third party on leaf 0 who gets removed later, otherwise the joiner reuses the VC's leaf and `own_leaf_index` never moves.
* `send_and_process_app_message` processes immediately. These messages have to stay unread across an epoch change.

mtrossbach:
Force-pushed. Both suggestions applied as-is. The comment in `validation.rs` is gone. The test is down from 318 to 198 lines using the helpers you pointed at.

kkohbrok (approved):
Thanks!

PR #2195, opened 2026-08-27
feat: carry a rejoining member's past epoch secrets across an external commit

### LINKED ISSUE #2180: [Feature request] A supported way to carry a rejoining member's own past-epoch secrets across an external commit

z0hrab:
Fixes #2180.

## The change

`ExternalCommitBuilder` gains `retain_past_epochs_from(group: MlsGroup)`, the builder option discussed in #2180. When a member rejoins a group it is already a member of via an external commit, the group produced by the builder replaces the state the member already holds — and with it the past epoch message secrets that state retained, so application messages the member could still have decrypted a moment before the rejoin become undecryptable. Passing the previous group state to the builder carries those secrets over into the new group.

The semantics, and the decisions behind them:

- **Bounded against the rejoin epoch.** Only epochs within the join config's `PastEpochDeletionPolicy` window, measured against the epoch the external commit creates, are carried over — exactly the window a member that had stayed caught up would have. A member that missed more epochs than the policy retains has no valid past epochs left, and nothing is carried over. The count-based eviction in `add_past_epoch_tree` relies on epochs advancing one at a time and would let epochs outlive the window after a rejoin gap, so the transfer prunes by epoch distance explicitly. (The default policy is `MaxEpochs(0)`, which retains nothing; the docs say so explicitly so nobody wires this up and wonders why it is a no-op.)
- **The previous state's current secrets are folded in.** They are its most recent (and most valuable) retained secrets. A state that had fallen behind the group contributes them as one more past epoch, together with the member list of that epoch, mirroring what `merge_commit` does when an epoch ends normally.
- **The pre-rejoin epoch's placeholder secrets are replaced when the member has the real ones.** `build_group` seeds the pre-join epoch's `MessageSecrets` from an all-zero epoch secret with a fake own index (#767), because an external joiner cannot know the real ones. A rejoining member often does: a state exactly at the `GroupInfo`'s epoch (with a matching group context — see below) contributes its current secrets, and a state that had moved *past* a stale `GroupInfo` contributes its retained tree for that epoch. Either way the real secrets take the placeholder's slot rather than sitting next to the entry the merge pushes for the same epoch, so no duplicate epoch entries ever exist in the store. Nothing reads the current message secrets between `build_group` and the merge (`content_to_mls_message` computes the membership tag only for `Member` senders, and an external commit is a `NewMemberCommit`; the commit's own key schedule derives from the init secret), so the substitution is inert until the merge folds the secrets in through the normal path. Messages from epochs the member never reached still fail to decrypt — with an AEAD error rather than `TooDistantInThePast`, which the docs and a test now pin.
- **Epochs beyond the `GroupInfo`'s epoch are dropped.** A state ahead of the group state the commit is built against has moved past a stale `GroupInfo` or sits on a fork; its secrets for newer epochs cannot decrypt anything on the branch being joined and would collide with the epoch numbers the rejoined group moves through.
- **The state must belong to the member doing the rejoin.** Three checks, three error variants: group id, ciphersuite, and the signature key of the state's own leaf against the credential the commit is built with. The last one matters because the retained secrets answer the own-message question with their epoch's own leaf (#2194) — another member's state would silently misclassify that member's messages as the client's own. A same-epoch state whose group context does not match the `GroupInfo` (a fork at the same epoch number) is not installed over the placeholder; decryption of that epoch then fails closed exactly as it would without the transfer.
- **No panics.** If a (malformed) input store would exceed the new store's capacity after pruning, the oldest entries are dropped, the same way `add_past_epoch_tree` trims at capacity — there is no assert to trip.

The transfer itself is one `pub(crate)` method, `MessageSecretsStore::inherit_past_epochs` in `past_secrets.rs`, where the store invariants live and the `EpochTree` internals stay private; `build_group` validates and calls it after constructing the new store. `finalize` already persists the store before the merge and `merge_commit` writes it again after, so no storage changes are needed. Inherited secrets keep their original timestamps (no re-stamping), so time-based deletion sees their true age.

## Why this does not weaken forward secrecy

The rejoining member already holds these secrets — retained deliberately via the past epoch deletion policy, for epochs in which it was a member. Carrying them across the rejoin grants the device no access it did not have one instruction earlier. A client without a previous group state has nothing to carry over, so nothing changes for a genuinely new joiner, and the retention bound stays the consumer's existing policy, measured against the current epoch as usual. (Argued in more detail in #2180.)

## Side effects and compatibility

- Not using the new option changes nothing: the builder field defaults to `None` and `build_group` behaves exactly as before.
- `MessageSecretsStore`'s fields are unchanged, so its serialized form and the `0-8-1-storage-format` compatibility path are untouched.
- The three new `ExternalCommitBuilderError` variants are a minor breaking change for exhaustive matches, like the variant additions in 0.9.0.
- The deprecated `join_by_external_commit` cannot set the option; its `From<ExternalCommitBuilderError>` conversion maps the new variants like the other variants that path cannot produce.
- The new builder field is feature-independent; the cfg-gated `virtual-clients-draft` fields and flows in the builder are untouched.

## Alternatives considered

- **An explicit transfer method on `MlsGroup`** (option 2 in #2180): sequencing it around the consumer's own storage swap invites misuse — the transfer has a correct moment relative to storing the new state and deleting the old one, and a free-standing method leaves finding that moment to every consumer. The builder option keeps the transfer inside the one operation that already rebuilds the state.
- **Exposing `set_max_past_epochs` and the retained trees** (option 3 in #2180): hands secret material to consumer code that never touches it today.
- **Taking the previous group by reference.** `MessageSecretsStore` deliberately derives `Clone` only under test cfgs, so `&MlsGroup` would mean deriving `Clone` on secret material in production. By value it is: the caller is discarding that state anyway, storage stays untouched until `finalize`, and on a `build_group` error the state can be reloaded via `MlsGroup::load` (documented).
- **Rejecting a state ahead of the `GroupInfo` with an error** instead of using its retained pre-rejoin tree: a stale `GroupInfo` is delivery-service-served input and a realistic condition on exactly this recovery path; using the retained real secrets for the epochs at or before the `GroupInfo` is strictly better than erroring, and wrong-branch secrets can only fail AEAD-open — they cannot produce wrong plaintext.

## Verification

Nine `#[openmls_test]` tests in `external_commit_builder.rs` (with shared fixtures), plus a compile-tested book example (`book_code.rs`, anchored into the external-commit chapter):

- `external_commit_retains_past_epochs` — a member with four epochs of history falls one epoch behind and rejoins with a retention window of 3: epochs 3 and 4 decrypt (epoch 4 from the folded current secrets), epochs 1 and 2 fail `TooDistantInThePast` exactly like epochs that aged out normally (a two-sided boundary: epoch 3 sits exactly on the keep edge, epoch 2 exactly one outside), a message from the never-reached epoch 5 fails with `AeadError` (the placeholder), the member's own pre-rejoin echo surfaces as `OwnPrivateMessage` (exercising #2194's per-epoch own index across a rejoin), and the group is reloaded from storage before the assertions to cover persistence.
- `external_commit_rejoin_beyond_retention_window` — the empty-window case from the issue discussion: nothing is carried over, old messages fail as they do today.
- `external_commit_rejoin_keep_all_past_epochs` — `KeepAll` carries everything across an arbitrary gap.
- `external_commit_retains_current_epoch_secrets` — the fully-caught-up rejoin: a message sealed in the pre-join epoch decrypts after the rejoin (the #767 placeholder case).
- `external_commit_retain_past_epochs_default_policy` — the default `MaxEpochs(0)` policy carries nothing; documented and pinned.
- `external_commit_rejoin_from_state_ahead_of_group_info` — a rejoin against a stale `GroupInfo`: the retained tree for the pre-rejoin epoch replaces the placeholder (exactly one entry for that epoch — asserted), a pre-`GroupInfo` message decrypts, an epoch beyond the `GroupInfo` fails `AeadError`.
- `external_commit_chained_rejoins_from_stale_group_info` — two rejoins in a row against stale `GroupInfo`s with a tightening retention window: no duplicate entries accumulate, and the old message still decrypts after both.
- `external_commit_retain_past_epochs_from_forked_state` — a same-epoch fork passes the id/ciphersuite/signature-key checks but not the group-context comparison; its secrets are not installed and decryption fails closed.
- `external_commit_retain_past_epochs_rejects_wrong_group` — all three guard variants, including a state belonging to a different member.

As a negative control, with `inherit_past_epochs` stubbed to a no-op the four decrypt-success tests fail (`retains_past_epochs`, `retains_current_epoch_secrets`, `rejoin_keep_all_past_epochs`, `rejoin_from_state_ahead_of_group_info`) and the guard/same-as-today tests still pass — no test passes for the wrong reason.

What I ran, on `main` @ 00ef45baa + this commit — every test binary reports `0 failed`:

```
$ cargo test -p openmls --no-default-features
ok

$ cargo test -p openmls
ok

$ cargo test -p openmls --no-default-features \
    -F virtual-clients-draft,virtual-clients-draft-test-dependencies
ok

$ cargo check -p openmls --all-features --all-targets
clean

$ cargo check -p openmls --target wasm32-unknown-unknown -F js
clean

$ cargo clippy -p openmls --tests --benches --examples -- -D warnings
clean

$ cargo doc -p openmls --all-features --no-deps
no warnings

$ cargo fmt -- --check
clean
```

## User-facing description

Added: `ExternalCommitBuilder::retain_past_epochs_from` carries a rejoining member's own past epoch message secrets across an external commit, bounded by the join config's past epoch deletion policy measured against the epoch the rejoin creates — so rejoining a group no longer makes application messages from epochs the member participated in undecryptable.


franziskuskiefer:
@z0hrab was there a reason to close this? I meant to review this and it seems like a useful addition.

PR #2197, opened 2026-08-27
fix: keep the newest past epochs when the retention shrinks

### LINKED ISSUE #2196: [Bug] Shrinking the past-epoch retention keeps the wrong epochs
**Describe the bug**

`MessageSecretsStore::resize` rotates the store before truncating it, so that `truncate` drops the oldest entries. It rotates by the difference between the old and the new limit, which equals the number of entries to drop only while the store is full (`openmls/src/group/mls_group/past_secrets.rs:85`):

```rust
let old_size = self.max_epochs;
self.max_epochs = max_past_epochs;
if old_size > max_past_epochs {
    let num_epochs_out = old_size - max_past_epochs;
    self.past_epoch_trees
        .rotate_left(num_epochs_out.min(self.past_epoch_trees.len()));
    self.past_epoch_trees.truncate(max_past_epochs);
}
```

Below that the rotation overshoots. Working through the algorithm gives three failure modes, all of which leave the store out of order: a window that is not contiguous, the oldest `n` where `min(len)` clamps the overshoot to a full turn, and, where nothing needs dropping, the right epochs reversed.

**To Reproduce**

Through `MlsGroup::set_past_epoch_deletion_policy`, or `set_configuration` with a changed policy. The store starts empty on creation and on every join, so any shrink before it has filled is affected; under `KeepAll` its limit is `isize::MAX`, so it never fills.

1. `MaxEpochs(6)`, advance 5 epochs, shrink to 3.
2. `KeepAll`, advance 30 epochs, shrink to 3.
3. `MaxEpochs(3)`, advance 2 epochs, shrink to 2, then commit once.

**Expected behavior**

The `n` most recent past epochs, in order.

**Actual behavior**

| case | retained before | retained after | expected |
| --- | --- | --- | --- |
| `MaxEpochs(6)`, 5 epochs, shrink to 3 | `0 … 4` | `0, 3, 4` | `2, 3, 4` |
| `KeepAll`, 30 epochs, shrink to 3 | `0 … 29` | `0, 1, 2` | `27, 28, 29` |
| `MaxEpochs(10)`, store full, shrink to 2 | `2 … 11` | `10, 11` | `10, 11` ✓ |

A shrink that drops nothing is not harmless either. With `MaxEpochs(3)` and 2 epochs, shrinking to 2 removes neither, but the store is left reversed, so the next commit evicts epoch 0 instead of epoch 1 and leaves `0, 2` where `1, 2` is correct.

The store is written through to the storage provider, so the loss is permanent. Messages from the missing epochs fail with `SecretTreeError::TooDistantInThePast`, the same error #2180 reports for a different cause.

**Screenshots or debug log**

Nothing is logged; the truncation is silent. Measured by listing the epochs for which `message_secrets_and_leaves` returns `Ok`.

- OS & OS version: macOS 15 (Darwin 25.5.0); not platform-specific
- Architecture: aarch64
- Version: `main` @ 00ef45ba, default features
- Ciphersuite: `MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519`, `openmls_rust_crypto`

**Additional context**

This is a regression rather than a decision. `resize` began as `while len > max { pop_front() }`. 28a0a45c replaced that with the rotation and substituted `old_size - max_past_epochs` for the excess, and #1972 added the `.min(len)` clamp when it made `resize` reachable at runtime, where the overshoot would otherwise panic.

`add_past_epoch_tree` is unaffected, its excess is always exactly one, and `delete_past_epoch_secrets` takes its from `len()`. Doing the same in `resize` fixes every case:

```rust
let excess = self.past_epoch_trees.len().saturating_sub(max_past_epochs);
self.past_epoch_trees.drain(0..excess);
```

`test_update_policy` shrinks only from a full store, which is the case the current code gets right, and asserts on the count. A PR with this fix and tests that assert on the surviving epochs follows.

mtrossbach:
Fixes #2196.

## The change

`resize` rotated the store by the difference between the old and the new limit before truncating. That difference is the number of entries to drop only while the store is full; below that the rotation overshoots, and the store ends up in the wrong order as well as missing the wrong epochs. A shrink that needs to drop nothing still reverses it, so the following commit evicts the newer of the two epochs it holds.

```rust
let excess = self.past_epoch_trees.len().saturating_sub(max_past_epochs);
self.past_epoch_trees.drain(0..excess);
```

This is what the `max_past_epochs` bound in `delete_past_epoch_secrets` already does, and what `resize` did before 28a0a45c replaced its `while len > max { pop_front() }` loop with the rotation. `old_size` and the surrounding condition go with it: growing yields an excess of zero, so `drain` is a no-op.

Stores written by an earlier version keep the rotated order, so the change above doesn't reach them. `past_epoch_trees` is sorted on deserialization instead, which covers `MlsGroup::load` and the `migration-import` path.

```rust
#[serde(deserialize_with = "epoch_trees_oldest_first")]
past_epoch_trees: VecDeque<EpochTree>,
```

There is no `serialize_with`, so what a current version writes is unchanged.

## Side effects

No API change. `add_past_epoch_tree` computes an excess of exactly one, and `delete_past_epoch_secrets` takes its from `len()`, so `resize` was the only way to produce the rotation.

The rotation affects more than the eviction order. `delete_past_epoch_secrets` with a duration or a timestamp locates the last entry past the cutoff and drains everything up to it, which assumes the same ordering. With a cutoff that should remove epoch 0 alone:

```
ordered [0, 1]  ->  [1]
rotated [1, 0]  ->  []
```

Both variants are public, and the normalization covers them.

## Alternatives considered

Keeping `rotate_left` and `truncate` and only correcting the rotation distance. Equivalent, but longer, and it leaves a condition that no longer decides anything.

Normalizing inside `MlsGroup::load` rather than on deserialization. That would miss the `migration-import` path, which deserializes an `MlsGroup` directly.

## Verification

`shrink_policy_keeps_newest_epochs` asserts on the epochs that survive rather than on their number, across four shrinks. `shrink_policy_keeps_the_store_ordered` covers a shrink that drops nothing and then commits, where the damage only shows in the eviction that follows. `rotated_store_is_normalized_on_load` applies the previous `resize` verbatim, writes the store out, loads it back and checks the order and the next eviction.

```
$ cargo test -p openmls --no-default-features --lib
288 passed; 0 failed

# with both fixes in past_secrets.rs reverted
285 passed; 3 failed

$ cargo clippy -p openmls --tests -- -D warnings
clean

$ cargo fmt -- --check
clean
```

## User-facing description

Fixed: reducing a group's `PastEpochDeletionPolicy` at runtime discarded the wrong past epochs and left the store out of order, so recently delayed messages stopped decrypting. Groups that hit this before upgrading are repaired when their state is next loaded.


kkohbrok (changes requested):
Thanks for the PR! 

The new resize logic assumes past_epoch_trees is already oldest-first. However, the previous implementation persisted out-of-order queues. For example, shrinking [0, 1] from limit 3 to 2 stored [1, 0]. After upgrading, the next commit’s existing eviction logic transforms that into [0, 2], still discarding newer epoch 1 instead of epoch 0. Thus affected groups remain unable to decrypt some delayed messages and retain older secrets longer than intended. Missing epochs cannot be recovered, but a migration/load-time normalization should restore ordering and prevent further incorrect evictions. Add a regression test that loads a legacy-corrupted store and verifies that it's correct.

mtrossbach:
Sorry for the slow reply, I didn't get to this on Friday. Thanks for working through the upgrade path, that's a real gap. You're right, and I reproduced it on `main`, printing the queue in physical order:

```
after 2 commits, MaxEpochs(3):   [0, 1]
after shrinking to MaxEpochs(2): [1, 0]
reloaded from storage:           [1, 0]
after the next commit:           [0, 2]
```

The same state breaks the two time-based deletion paths too, which I hadn't looked at. Both `delete_past_epoch_secrets_older_than_duration` and `delete_past_epoch_secrets_before_timestamp` find the last entry past the cutoff and drain everything up to it, so they also assume an ordered queue. Same setup, cutoff picked so only epoch 0 should go:

```
ordered [0, 1]  ->  [1]
rotated [1, 0]  ->  []
```

The fresh epoch goes with it. Both are reachable through the public `MlsGroup::delete_past_epoch_secrets`, so this isn't only about eviction order.

I'd rather sort at the deserialization boundary than inside `load`. That way the `migration-import` path is covered too, and there's no call site to forget. `EpochTree` already stores the epoch, so the order is unambiguous.

Should the repair be written back, or is fixing it on load enough? `load` only takes `&Storage` today, so persisting would change that. If it stays in memory, the queue on disk keeps its rotation until the group is next written, which works but isn't really a migration.

I'd put the test in `past_secrets.rs`, not in `past_secrets_storage_compatibility.rs`. That file is only compiled with `sqlite-provider` and `libcrux-provider` enabled, and `tests.yml` doesn't enable them, so a test there would only run nightly. It would rotate a populated queue, write it out and read it back, then check that the order is restored and that the next eviction takes the oldest epoch.

kkohbrok:
Sounds good! Fixing it on load should be enough.

mtrossbach:
Pushed. `past_epoch_trees` is now sorted on deserialization, so there's no write-back and the `migration-import` path is covered by the same attribute. The regression test applies the previous `resize` verbatim, so the state it builds is the one that was actually persisted.

The PR description now also covers the two time-based deletion paths, which the same rotation breaks.

kkohbrok (approved):
Thanks for making the changes! Looks good to me now.

PR #2202, opened 2026-08-28
feat: zeroize exported secrets on drop

kkohbrok:


openmls/src/group/mls_group/exporting.rs
```diff
@@ -14,6 +20,90 @@ use crate::{
...
+        self.secret.as_slice()
+    }
+}
+
+// The inner [`Secret`] is zeroized when dropped.
+impl<T> ZeroizeOnDrop for ExportedSecret<T> {}
+
+impl<T> Debug for ExportedSecret<T> {
+    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
+        f.debug_struct("ExportedSecret")
```
franziskuskiefer, line 94: I don't think we should print secrets unconditionally. We can do that with the crypto tracing feature on. But otherwise this should be an `***`.

openmls/src/group/mls_group/exporting.rs
```diff
@@ -14,6 +20,90 @@ use crate::{
...
+            .field("secret", &self.secret)
+            .finish()
+    }
+}
+
+impl<T, U> PartialEq<ExportedSecret<U>> for ExportedSecret<T> {
+    // Constant time comparison.
```
franziskuskiefer, line 101: We should look at that comparison at some point. I'm not sure it's always constant time. Not for this PR though.

franziskuskiefer (commented):
lgtm with one comment

openmls/src/group/mls_group/exporting.rs
kkohbrok: We're not printing any secrets here. The Debug impl is only there to avoid the Debug bound on the paramter and to omit the PhantomData field. It otherwise delegates to the Debug impl of `Secret`, which reducts its content as you suggest.

openmls/src/group/mls_group/exporting.rs
kkohbrok: Good point. Let's handle that in a follow-up.

openmls/src/group/mls_group/exporting.rs
franziskuskiefer: 🙈 ah right, it's a secret in there. Never mind

franziskuskiefer (approved):
:shipit:

PR #2203, opened 2026-08-30
fix: make clear_proposal_queue actually delete the queued proposal bodies

### LINKED ISSUE #2188: openmls_memory_storage: clear_proposal_queue removes a key nothing was stored under, leaking every proposal body
## Summary

`MemoryStorage::clear_proposal_queue` builds the per-proposal key differently
from the way `queue_proposal` wrote it, so it removes a key nothing was ever
stored under. The `ProposalQueueRefs` list is deleted — making the queue *look*
empty — while every serialised `QueuedProposal` body stays in the map, and
nothing ever removes it.

Present in `openmls_memory_storage` **0.5.0** and **0.6.0**, and on `main` at the
time of writing.

## The mismatch

`queue_proposal` writes the body through the `write` helper, which prefixes the
label and suffixes the storage version via `build_key_from_vec`:

```rust
let key = serde_json::to_vec(&(group_id, proposal_ref))?;
let value = serde_json::to_vec(proposal)?;
self.write::<CURRENT_VERSION>(QUEUED_PROPOSAL_LABEL, &key, value)?;
```

`clear_proposal_queue` removes the raw serialisation instead, with no label and
no version (`src/lib.rs`, 0.6.0 line 876):

```rust
for proposal_ref in proposal_refs {
    // Delete all proposals.
    let key = serde_json::to_vec(&(group_id, proposal_ref))?;
    values.remove(&key);           // <-- no label prefix, no version suffix
}
```

So the key passed to `remove` can never match the key `write` inserted.

## Reproduction

```toml
[dependencies]
openmls_memory_storage = "0.6.0"
openmls_traits = "0.6.0"
serde = { version = "1", features = ["derive"] }
serde_json = "1"
```

```rust
use openmls_memory_storage::MemoryStorage;
use openmls_traits::storage::{traits, Entity, Key, StorageProvider, CURRENT_VERSION};

#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
struct Gid(Vec<u8>);
impl Key<CURRENT_VERSION> for Gid {}
impl traits::GroupId<CURRENT_VERSION> for Gid {}

#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
struct Body(Vec<u8>);
impl Entity<CURRENT_VERSION> for Body {}
impl traits::QueuedProposal<CURRENT_VERSION> for Body {}

#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
struct Ref(u32);
impl Key<CURRENT_VERSION> for Ref {}
impl Entity<CURRENT_VERSION> for Ref {}
impl traits::ProposalRef<CURRENT_VERSION> for Ref {}

#[test]
fn clear_proposal_queue_leaves_the_proposal_bodies_behind() {
    let storage = MemoryStorage::default();
    let group = Gid(b"group".to_vec());

    storage.queue_proposal(&group, &Ref(1), &Body(vec![0xA1])).unwrap();
    storage.queue_proposal(&group, &Ref(2), &Body(vec![0xA2])).unwrap();

    // 2 proposal bodies + 1 refs list
    assert_eq!(storage.values.read().unwrap().len(), 3);

    storage.clear_proposal_queue::<Gid, Ref>(&group).unwrap();

    // The refs list is gone, so the queue looks empty...
    assert!(storage.queued_proposal_refs::<Gid, Ref>(&group).unwrap().is_empty());

    // ...but both bodies survive.
    assert_eq!(storage.values.read().unwrap().len(), 0);
}
```

Result:

```
assertion `left == right` failed: expected an empty store after clearing the
queue; 2 entries survived
  left: 2
 right: 0
```

## Impact

Every commit that consumes proposals clears the queue, so a long-lived client
accumulates orphaned serialised proposals for the lifetime of the store. In the
in-memory provider that is unbounded growth; for anyone using this provider as
the reference for a persistent `StorageProvider` — which the crate is commonly
used as — the same key-construction bug is easy to carry across into a store
that survives restarts, where the orphans are permanent.

There is also a correctness edge: because the bodies are still present under
their real keys, re-queuing the same `(group_id, proposal_ref)` after a clear
overwrites rather than inserts, so the stale body is only masked, never removed.

## Suggested fix

Route the deletion through the same helper the other deleters use, so the key is
built the way it was written:

```rust
for proposal_ref in proposal_refs {
    let key = serde_json::to_vec(&(group_id, &proposal_ref))?;
    self.delete::<CURRENT_VERSION>(QUEUED_PROPOSAL_LABEL, &key)?;
}
```

(That takes the lock inside `delete`, so the surrounding `values.write()` guard
needs to move or be dropped first.)

## A related nit in the same file

`queued_proposals` unwraps the per-reference read:

```rust
let proposal = self.read(QUEUED_PROPOSAL_LABEL, &key)?.unwrap();
```

A reference present in the queue whose body is missing is a corrupt store — and
it is reachable, because `queue_proposal` performs **two** writes (the body, then
the reference) and the `StorageProvider` trait has no transaction, so a crash
between them leaves exactly that state. Today it becomes a panic in the middle of
message processing. Returning an error would let the caller report the corruption
instead of aborting.

munzzyy:
Closes #2188.

`MemoryStorage::queue_proposal` writes each proposal body through the `write` helper, which builds the storage key as `label + key + version` via `build_key_from_vec`:

```rust
let key = serde_json::to_vec(&(group_id, proposal_ref))?;
let value = serde_json::to_vec(proposal)?;
self.write::<CURRENT_VERSION>(QUEUED_PROPOSAL_LABEL, &key, value)?;
```

`clear_proposal_queue` removed the same bytes without the label or version:

```rust
let key = serde_json::to_vec(&(group_id, proposal_ref))?;
values.remove(&key);
```

`values.remove` was never handed the key `write` actually inserted, so the removal was a no-op. `queued_proposal_refs` still comes back empty afterward, because the refs list itself is deleted correctly, further down, through `build_key`. `queued_proposals` iterates that now-empty refs list, so it reports an empty queue too, but every serialized proposal body is still sitting in the map under its real key. There's no code path that ever removes it: a long-lived provider accumulates one orphaned entry per proposal, for as long as the process runs, and the existing test in `memory_storage/tests/proposals.rs` didn't catch it because it only checks what `queued_proposals` reports, not what's actually left in the store.

Repro on unpatched main, using the test added here (before the fix, run with the old `values.remove` body):

```
$ cargo test -p openmls_memory_storage --test proposals clear_proposal_queue_removes_the_proposal_bodies

running 1 test
test clear_proposal_queue_removes_the_proposal_bodies ... FAILED

---- clear_proposal_queue_removes_the_proposal_bodies stdout ----

thread 'clear_proposal_queue_removes_the_proposal_bodies' panicked at memory_storage/tests/proposals.rs:102:5:
assertion `left == right` failed: clear_proposal_queue left orphaned proposal bodies in the store
  left: 2
 right: 0
```

## The change

`clear_proposal_queue` now routes both deletions through `delete`, the same helper every other single-key deleter in this file uses, instead of reaching into `self.values` directly. `delete` builds the key the same way `write` did, so it matches what's actually stored. The manual `values.write()` lock that spanned the loop is gone with it; `delete` takes and releases its own lock per key, same as elsewhere in the file.

## Side effects

No public API change. `MemoryStorage` is a reference/testing storage provider; this only affects what it does internally with the `values` map. Behavior for every other method is unchanged.

## Tests

Added `clear_proposal_queue_removes_the_proposal_bodies` next to the existing `read_write_delete` test in `memory_storage/tests/proposals.rs`. It queues two proposals, checks the raw `storage.values` map has three entries (two bodies plus the refs list), clears the queue, and asserts the map is empty, rather than going through `queued_proposals`, which is exactly the check that let this ship.

## Verification

```
$ cargo test -p openmls_memory_storage
running 2 tests
test clear_proposal_queue_removes_the_proposal_bodies ... ok
test read_write_delete ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo test -p openmls --lib
test result: ok. 286 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 65.33s

$ cargo clippy -p openmls_memory_storage --all-targets
Finished, no warnings

$ cargo fmt --check -p openmls_memory_storage
clean
```

I didn't touch the `.unwrap()` in `queued_proposals` mentioned in the issue as a related nit. That's a real gap (a body missing for a stored ref becomes a panic instead of a reportable error), but it's a different failure mode with its own design question about the error type, so I left it out of this fix to keep the change to the one bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- [#2203](https://github.com/openmls/openmls/pull/2203): `MemoryStorage::clear_proposal_queue` now actually deletes the queued proposal bodies. It built the per-proposal key by hand instead of going through the `label + key + version` scheme `queue_proposal` used to store them, so the delete never matched and every proposal body outlived the group's proposal-ref list that was supposed to track it. `queued_proposals` looked empty afterward because it only reads what the ref list points to, but the bodies stayed in the store, growing without bound for the lifetime of the provider.
 - [#2194](https://github.com/openmls/openmls/pull/2194): With the `virtual-clients-draft` feature, whether an inbound `PrivateMessage` is this client's own is now decided against the leaf index the epoch's secret tree was built for, rather than the group's current own leaf index. The two differ for every epoch before a sibling-resync external commit moved the client's leaf. Previously a message another member sent from the leaf the client had since moved onto was silently returned as `ProcessedMessageContent::OwnPrivateMessage`, and the echo of the client's own pre-resync message failed with `SecretTreeError::SecretReuseError`.
 
 ## 0.9.0 (2026-08-25)
--- a/memory_storage/src/lib.rs
+++ b/memory_storage/src/lib.rs
@@ -878,16 +878,19 @@ impl StorageProvider<CURRENT_VERSION> for MemoryStorage {
         // Get all proposal refs for this group.
         let proposal_refs: Vec<ProposalRef> =
             self.read_list(PROPOSAL_QUEUE_REFS_LABEL, &serde_json::to_vec(group_id)?)?;
-        let mut values = self.values.write().unwrap();
         for proposal_ref in proposal_refs {
-            // Delete all proposals.
+            // Delete all proposals. `queue_proposal` wrote each body through
+            // `write`, which prefixes the label and suffixes the version via
+            // `build_key_from_vec`, so the removal has to go through the same
+            // helper (`delete`) rather than a bare `values.remove` on the raw
+            // `serde_json` key, or nothing gets removed.
             let key = serde_json::to_vec(&(group_id, proposal_ref))?;
-            values.remove(&key);
+            self.delete::<CURRENT_VERSION>(QUEUED_PROPOSAL_LABEL, &key)?;
         }
 
         // Delete the proposal refs from the store.
-        let key = build_key::<CURRENT_VERSION, &GroupId>(PROPOSAL_QUEUE_REFS_LABEL, group_id);
-        values.remove(&key);
+        let key = serde_json::to_vec(group_id)?;
+        self.delete::<CURRENT_VERSION>(PROPOSAL_QUEUE_REFS_LABEL, &key)?;
 
         Ok(())
     }
--- a/memory_storage/tests/proposals.rs
+++ b/memory_storage/tests/proposals.rs
@@ -75,3 +75,33 @@ fn read_write_delete() {
     let proposals_read: Vec<(ProposalRef, Proposal)> = storage.queued_proposals(&group_id).unwrap();
     assert!(proposals_read.is_empty());
 }
+
+/// `queued_proposals` only reports what's reachable through the refs list, so
+/// an emptied refs list makes the queue look clear even if the proposal
+/// bodies are still sitting in the store under their own keys. Check the raw
+/// store directly.
+#[test]
+fn clear_proposal_queue_removes_the_proposal_bodies() {
+    let group_id = TestGroupId(b"TestGroupId".to_vec());
+    let storage = MemoryStorage::default();
+
+    storage
+        .queue_proposal(&group_id, &ProposalRef(0), &Proposal(vec![0xA1]))
+        .unwrap();
+    storage
+        .queue_proposal(&group_id, &ProposalRef(1), &Proposal(vec![0xA2]))
+        .unwrap();
+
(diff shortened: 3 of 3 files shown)


kkohbrok:
Dupe of #2163. I'm planning on merging that one first. If you think that leaves a gap, please let me know.

munzzyy:
Checked the merged diff of #2163. It builds the labeled key the same way write does, and the assertion it added to read_write_delete checks the map is actually empty afterwards, which is the same thing my dedicated test verified. No gap. Closing this one.

Issue #2079, opened 2026-06-29
Move In types (and Franken* in testing) into an all-pub crate for wire types


keks:
This reduces code duplication between our In an Frankenstein types and lets users access all the fields of the messages.

kkohbrok:
Part of the idea with `In` variants was that applications can only access message content after verifying the message. Between this and allowing things like `_into_unchecked`, maybe we should revisit our strict regime on splitting between `In` and `Out` types.

Issue #1112, opened 2022-11-21
[Bug] AAD should not be exposed before signature verification


raphaelrobert:
- [x] Remove `aad()` from `UnverifiedMessage` (should be covered by #1088)
 - [x] Remove the AAD from `UnverifiedMlsPlaintext`

PR #2204, opened 2026-09-02
Tests: add more aad tests

### LINKED ISSUE #1622: More tests for resetting the AAD
The AAD in `MlsGroup` is reset in quite a few places and not all of them are covered by tests.

nhanff:


nhanff:
I'm keeping the PR at this size for now.
For full coverage, at least these variants are missing:

* Test both `PublicMessage` and `PrivateMessage` wire types.
* Validate AAD from same and different member.
* Test the remaining functions generated from the `impl_propose_fun!` macro.
* Test code paths passing through `propose_dispatch`.
* Test functions that are only available with features, such as `create_targeted_message`. 
* Test and improve documentation that AAD is only reset on success.
     * I believe this is not strictly true in the current code base, there is occasional error handling after calling `reset_add`.

openmls/src/group/tests_and_kats/tests/aad.rs
```diff
@@ -194,7 +208,603 @@ fn test_add_member_with_aad() {
...
+        let [alice] = group_state.members_mut(&["alice"]);
+
+        let aad: Vec<u8> = b"Test AAD".to_vec();
+        alice.group.set_aad(aad);
+
+        assert_eq!(alice.group.aad(), b"Test AAD");
+    }
+}
+
+// Tests for indivudual functions that call reset_add()
```
keks:
```suggestion
// Tests for individual functions that call reset_add()
```

openmls/src/group/tests_and_kats/tests/aad.rs
```diff
@@ -194,7 +208,603 @@ fn test_add_member_with_aad() {
...
+        let mut group_state = GroupState::new_from_party(
+            group_id,
+            alice_party.generate_pre_group(config.ciphersuite),
+            config.clone(),
+        )
+        .unwrap();
+
+        let [alice] = group_state.members_mut(&["alice"]);
+
+        let aad: Vec<u8> = b"Test AAD".to_vec();
```
keks: Let's make `b"Test AAD"` a constant that is shared by all the tests.

keks (changes requested):
Looks pretty good, just a few nits. Let's also see what to do about the missing bits, i.e. what should be addressed here and what in a followup.

keks:
Thanks!

> * Test both PublicMessage and PrivateMessage wire types.

Is this because ~PrivateMessageIn~ PublicMessageIn doesn't provide access to the AAD? I think the answer here would be to also process the message and then read it from the processed message.

> * Test and improve documentation that AAD is only reset on success.
> 
>     * I believe this is not strictly true in the current code base, there is occasional error handling after calling `reset_add`.

Oh, interesting, but it looks like you are right. I found the following occurrences of this:

https://github.com/openmls/openmls/blob/84a3e80c1922dbea765c893cbad8dd99e96690b2/openmls/src/group/mls_group/commit_builder.rs#L1441-L1448

https://github.com/openmls/openmls/blob/84a3e80c1922dbea765c893cbad8dd99e96690b2/openmls/src/group/mls_group/commit_builder/external_commits.rs#L444-L462

https://github.com/openmls/openmls/blob/84a3e80c1922dbea765c893cbad8dd99e96690b2/openmls/src/group/mls_group/membership.rs#L346-L349

https://github.com/openmls/openmls/blob/84a3e80c1922dbea765c893cbad8dd99e96690b2/openmls/src/group/mls_group/membership.rs#L106-L108

Do you think these are reasonably easy to test? Do you want to do that in here or open a followup?

From the others, I think the dispatch one is the most interesting, but no strong opinions.

openmls/src/group/tests_and_kats/tests/aad.rs
nhanff: :ballot_box_with_check:

nhanff:
The AAD is directly and publicly available on `PrivateMessage` (`PrivateMessageIn::aad`) where it is reframed, but not on `PublicMessage` where it is contained in the inner `FramedContent`.

According to #1112, it should be exposed on neither, right?

nhanff:
> Do you think these are reasonably easy to test? Do you want to do that in here or open a followup?

Is resetting the AAD before final error handling even intended rather than accidental? If so, that would require detailed documentation and I believe could be unexpected.
I personally prefer the API of #2175 as opposed to the group state that must be set before committing, as it avoids edge cases: Code that does not intend to set the AAD might still do accidentally because some earlier code called `set_aad`  and either failed or did not create a commit. `reset_aad()` is not public, so code calling openMLS must remember to `set_aad(b"")` before reusing a group object.

keks:
> The AAD is directly and publicly available on PrivateMessage (PrivateMessageIn::aad) where it is reframed, but not on PublicMessage where it is contained in the inner FramedContent.

Ah yes, PublicMessage :facepalm:  That's what I meant.

> According to https://github.com/openmls/openmls/issues/1112, it should be exposed on neither, right?

Yeah, that is a topic that we'll have to discuss again. I think there are good reasons to e.g. let the DS access the plaintext fields of the messages, but they will not be able to do any validation on them, because they don't have the keys. So while that rule was well-intended to provide strong validation guarantees, it also might not be entirely practical. That also came up in https://github.com/openmls/openmls/issues/2079.

> Is resetting the AAD before final error handling even intended rather than accidental? If so, that would require detailed documentation and I believe could be unexpected.

I think these are probably accidents that should be fixed. Can you confirm @franziskuskiefer?

> I personally prefer the API of https://github.com/openmls/openmls/pull/2175 as opposed to the group state that must be set before committing, as it avoids edge cases: Code that does not intend to set the AAD might still do accidentally because some earlier code called set_aad and either failed or did not create a commit. reset_aad() is not public, so code calling openMLS must remember to set_aad(b"") before reusing a group object.

Yes, me too. But we don't have builder-style message creation for application messages and proposals yet, so we wanted to test what is there, before starting to refactor things.

openmls/src/group/tests_and_kats/tests/aad.rs
```diff
@@ -194,7 +209,725 @@ fn test_add_member_with_aad() {
...
+
+    let extensions =
+        Extensions::from_vec(vec![Extension::Unknown(1, UnknownExtension(Vec::new()))]).unwrap();
+
+    let proposals = [
+        Propose::Add(charlie_pkb.key_package.clone()),
+        Propose::Update(LeafNodeParameters::default()),
+        Propose::Remove(1), // Remove Bob
+        // Propose::RemoveCredential(),
+        // Propose::PreSharedKey(),
```
keks: I think not testing these is probably fair. But we should remove these comments.

openmls/src/group/tests_and_kats/tests/aad.rs
```diff
@@ -194,7 +213,771 @@ fn test_add_member_with_aad() {
             )
             .expect("Error handling message");
 
-        // Test the AAD was set correctly
         assert_eq!(bob_processed_message.aad(), &aad);
     }
 }
+
+const AAD: &[u8] = b"Test AAD";
```
keks: Tiny nitpick, but I have a small preference for calling this TEST_AAD.

openmls/src/group/tests_and_kats/tests/aad.rs
```diff
@@ -194,7 +213,771 @@ fn test_add_member_with_aad() {
...
+        .unwrap();
+
+    assert_eq!(alice_processed_message.aad(), AAD);
+}
+
+#[openmls_test::openmls_test]
+fn test_aad_error_commit() {
+    use crate::group::tests_and_kats::utils::storage_error::{
+        TestProvider, TestStorageError, TestStorageProvider,
+    };
```
keks: Is there a reason against moving these to the top of the file?

openmls/src/group/tests_and_kats/utils.rs
keks, line 1: I like this :) But can you add some module-level doc-comments that explain what this is doing, and why? Maybe also move the submodule to it's own file, it's pretty big.

keks (commented):
I think this generally looks really good! I like the helper for injecting errors into the storage provider responses.

openmls/src/group/tests_and_kats/tests/aad.rs
nhanff: :ballot_box_with_check:

openmls/src/group/tests_and_kats/tests/aad.rs
nhanff: :ballot_box_with_check:

openmls/src/group/tests_and_kats/tests/aad.rs
nhanff: Moved them up :ballot_box_with_check:

openmls/src/group/tests_and_kats/utils.rs
nhanff: :ballot_box_with_check:

keks (approved):
Thanks, I think this looks good now!

Issue #1915, opened 2025-12-05
Add support for custom (user-provided) ciphersuites


notABot101010:
Hello and thank your for all your work on OpenMLS!

I would like too experiment with custom ciphersuites, which are highly unlikely to be accepted in this repository (e.g. using BLAKE3).

Therefore, this openmls would need to be able to be used with user-provided ciphersuites.

One possible implementation would be to change the [Ciphersuite](https://docs.rs/openmls/latest/openmls/prelude/enum.Ciphersuite.html) enum to a trait (but I'm not sure exactly what would be the methods), and require to update the  `CryptoProvider` trait and associated types (e.g. [HashType](https://docs.rs/openmls/latest/openmls/prelude/enum.HashType.html)).

Another implementation, potentially less invasive, would be to extend the enums (CipherSuite, HashType...) with a `Custom` variant that would enable user-provided algorithms. This is pretty similar to how the `http` crate deals with `HttpMethod` with base variants for common HTTP verbs, and variants for custom HTTP verbs (see https://docs.rs/http/latest/src/http/method.rs.html#53).

From my limited understanding of the project, I think that the second implementation is preferable.

Thank you!

notABot101010:
@kkohbrok Any update? I would love to help make this move forward.

Even if the solution above are not perfect, we can ship, get feedback, and potentially make breaking changes later if problems arise.

PR #2205, opened 2026-09-02
fix: check provider support in KeyPackageIn::validate

mtrossbach:
`KeyPackageIn::validate` doesn't check whether the crypto provider supports the key package's ciphersuite before it verifies the signatures. If the provider doesn't support it, signature verification fails and the caller gets `InvalidLeafNodeSignature`, even though the signature is valid. I ran into this with the libcrux provider, which only does Ed25519: validating a P-256 key package returns `InvalidLeafNodeSignature`. For something like a delivery service that validates key packages on its own this is misleading, because it looks like a bad signature and not like an unsupported ciphersuite.

#2109 added a `supports()` check and `UnsupportedCiphersuite` errors to group creation, `StagedWelcome` and the external commit builder. `KeyPackageIn::validate` is another public entry point that takes a ciphersuite from the wire, and it didn't get the check. This PR adds it as the first step of `validate`, plus a `KeyPackageVerifyError::UnsupportedCiphersuite(Ciphersuite)` variant. The test uses the `RestrictedProvider` from `test-utils` and fails without the change.

Two smaller things in the same area:

- `AddProposalIn::validate` did the full key package validation (both signature checks) before comparing the key package's ciphersuite with the group's. Now the comparison comes first. The test corrupts the key package signature and checks that the ciphersuite mismatch is reported, not the signature.
- A test that a GREASE value in the ciphersuite field of a KeyPackage is rejected when deserializing. This already works, GREASE values aren't in the `Ciphersuite` table, but there was no test for it.

Context: I've been looking into how crypto providers could bring their own ciphersuites (#1915) and what that would touch in the code base, and I'll probably have something to propose on that at some point. This came up while going through the ciphersuite checks. It's independent of that though.


mtrossbach:
Resubmitted as #2206, sorry for the noise.

PR #2208, opened 2026-09-03
feat: the hybrid ML-KEM/P-curve ciphersuites of draft-ietf-mls-pq-ciphersuites

mtrossbach:
The ciphersuite table of draft-ietf-mls-pq-ciphersuites-06 has eleven entries. openmls has eight of them behind `draft-ietf-mls-pq-ciphersuites`; the three missing ones are the suites on the hybrid ML-KEM/P-curve KEMs of draft-ietf-hpke-pq:

| draft | suite | HPKE KEM |
| --- | --- | --- |
| TBD3 | MLS_128_MLKEM768P256_AES128GCM_SHA256_P256 | MLKEM768-P256, 0x0050 |
| TBD4 | MLS_128_MLKEM768P256_AES256GCM_SHA384_P256 | MLKEM768-P256, 0x0050 |
| TBD5 | MLS_192_MLKEM1024P384_AES256GCM_SHA384_P384 | MLKEM1024-P384, 0x0051 |

We need TBD4 and TBD5, in a setting where the P-curve half of the KEM is a requirement, and run them from a crypto provider of our own. For that openmls has to know the suites and the two KEMs, which is what this PR adds: the three `Ciphersuite` variants, the two `HpkeKemType` variants (`0x0050`, `0x0051`), and the KEM mappings in the two bundled providers made fallible, since hpke-rs 0.7 has no counterpart for these KEMs and the mappings were exhaustive matches. It follows how #2046 and #2118 added the other draft suites.

Parameters follow draft -06, which moved the KDF of every suite from SHAKE256 (-05) to HKDF; that is what the existing suites here use as well. The draft has no code points yet. I took `0x0053` to `0x0055` as provisional ones in draft order, following on from the `0x004D..0x0052` block the existing draft suites occupy; `0x0050` and `0x0051` as suite code points are already taken by TBD7 and TBD10 here. If you prefer other values, say so and I change them. The new variants are appended to both enums, so the serde indices of the existing ones do not move.

One thing is different from the earlier suite PRs, and I want to name it rather than have you find it: neither bundled provider can run these suites yet. hpke-rs 0.7 has no hybrid KEMs; they are on hpke-rs `main` since celabshq/libcrux#1539 but not in a release. So this adds the definitions before the providers here can use them. With a KEM that has no hpke-rs counterpart, an HPKE call now returns `UnsupportedCiphersuite`, `supports()` reports the suites as unsupported, and a provider that implements the KEMs advertises them through `supports()` and `supported_ciphersuites()`, from which `Capabilities::for_provider` builds the leaf node capabilities.

Why I think adding them now is still the better order:

- The definitions are the part that needs a release of `openmls_traits`, because they are new enum variants. Provider support is a change inside a provider crate and can follow in any release, without touching the traits again. With the variants in the next `openmls_traits` release, the provider side after the hpke-rs release is a small patch, the way #2118 and #2152 did it, and nothing waits for a second traits release.
- Nothing changes for anyone who does not use the suites. They are behind the feature, not in the default capabilities, and a provider has to claim them in `supports()` before openmls will touch them. Defining suites the bundled providers cannot run has precedent here: `0x0004` to `0x0007` from RFC 9420 are in the table and neither provider supports them either.
- Every row of the draft's table is then in openmls, which makes it usable for anyone who runs these suites from a provider outside this repository, not only us.

If you would rather wait and take suites and provider support in one PR after the hpke-rs release, that works for me too; I convert this to a draft until then.

No test vectors, since no bundled provider can run the suites. A unit test pins the parameters and the code points.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- [#2208](https://github.com/openmls/openmls/pull/2208): Added the three ciphersuites of draft-ietf-mls-pq-ciphersuites on the hybrid ML-KEM/P-curve KEMs, `MLS_128_MLKEM768P256_AES128GCM_SHA256_P256` (TBD3), `MLS_128_MLKEM768P256_AES256GCM_SHA384_P256` (TBD4) and `MLS_192_MLKEM1024P384_AES256GCM_SHA384_P384` (TBD5), with provisional code points `0x0053` to `0x0055`, and the KEMs `HpkeKemType::MlKem768P256` (`0x0050`) and `HpkeKemType::MlKem1024P384` (`0x0051`) from draft-ietf-hpke-pq, behind `draft-ietf-mls-pq-ciphersuites`. Neither bundled crypto provider implements these KEMs; they report the ciphersuites as unsupported and their HPKE operations return `UnsupportedCiphersuite` for them.
 - [#2184](https://github.com/openmls/openmls/pull/2184): Added `PreSharedKeyProposal::psk`, a getter for the `PreSharedKeyId` of a `PreSharedKey` proposal.
 - [#2167](https://github.com/openmls/openmls/pull/2167): Added `PublicGroup::validate_key_package_for_add`, which checks whether a single `KeyPackage` is eligible to be added to the group without building a commit. Applications adding several members at once can use it to filter out candidates that a commit would reject, and to report which candidate was rejected and why. Uniqueness of the signature, init and encryption keys is not covered, since it can only be decided for a full set of proposals.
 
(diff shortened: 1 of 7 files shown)


franziskuskiefer:
Thanks! There are a couple parallel versions for PQ ciphersuites in flight right now that need to get aligned. I hope we get to that next week.

- #2170 
- some branches and I think another PR I can't find right now

franziskuskiefer:
Thanks for the contribution again and apologies for the slow round trip.
We merged #2170 and thus can support some of the pq ciphersuites. P256 support for libcrux is in #2215 and should land soon as well.
I'd really like to merge something here that actually implements the ciphersuites as well. With those PRs we should be able to do that. Would you mind rebasing onto main and adding actual implementations?

Note that this was primarily blocked on the capabilities changes that are really needed to avoid using leaf nodes with capabilities the caller doesn't actually want. We should get that merged soon in #2230.

PR #2211, opened 2026-09-04
fix: joining a group with virtual-clients enabled

gferon:
A device that joins a group via a VC key package `Welcome` never got a `VcEmulationBindings` entry for the joined epoch.

On my buggy (current) setup, `decrypt_message` sees the own leaf index, finds no emulator context and returns `OwnPrivateMessage`, so a sibling's messages look like our own echo and are discarded by sibling clients.

Commits are still flowing, which is why the group otherwise looks fine and only own messages go missing.

Also: add a helper method to return the derivation info in four places.


openmls/src/group/mls_group/creation.rs
```diff
@@ -656,13 +656,53 @@ impl StagedWelcome {
...
+                    epoch_id,
+                    max_entries,
+                );
+                bindings
+                    .store(provider.storage(), group_id)
+                    .map_err(WelcomeError::StorageError)?;
+            } else {
+                log::warn!(
+                    "vc: own welcome leaf carries derivation info for an unknown derivation epoch; not binding"
+                );
```
kkohbrok, line 702: Shouldn't we error out here?

openmls/src/treesync/node/leaf_node.rs
```diff
@@ -498,6 +498,35 @@ impl LeafNode {
         &self.payload.extensions
     }
 
+    /// The virtual-client derivation this leaf might contain.
```
kkohbrok, line 501:
```suggestion
    /// The virtual-client derivation info this leaf might contain.
```

kkohbrok (approved):
Looks good! Just two comments, but generally good to merge.

openmls/src/group/mls_group/creation.rs
```diff
@@ -656,13 +656,53 @@ impl StagedWelcome {
...
-                        .clone(),
                 ),
             )?;
         }
 
+        // A join via a virtual-client KeyPackage lands on a leaf shared with
+        // sibling emulator clients. Bind the joined epoch to the leaf's
+        // derivation epoch, as an external-commit join does, so that messages
+        // siblings send from the shared leaf can be deprotected instead of
+        // being taken for our own echo.
```
raphaelrobert:
```suggestion
        // being mistaken for our own echo.
```

openmls/tests/virtual_clients.rs
```diff
@@ -2635,195 +2628,104 @@ fn retained_material_welcome<P: OpenMlsProvider>(
...
-        .storage()
-        .vc_derivation_epoch_state(&epoch_id)
-        .expect("read epoch state after group delete");
-    assert!(
-        state.is_some(),
-        "the retained material must keep the epoch state alive"
-    );
+    process_vc_key_package_upload(&alice_b_provider, &upload).expect("alice_b process upload");
 
+    // Bob founds a higher-level group and adds the virtual client.
```
raphaelrobert:
```suggestion
    // Bob creates a higher-level group and adds the virtual client.
```

openmls/src/group/mls_group/creation.rs
```diff
@@ -663,6 +663,44 @@ impl StagedWelcome {
...
+        #[cfg(feature = "virtual-clients-draft")]
+        if let Some(derivation_info) = self
+            .public_group
+            .leaf(self.own_leaf_index)
+            .map(LeafNode::vc_derivation_info)
+            .transpose()?
+            .flatten()
+        {
+            use crate::components::vc_derivation_info::VcDerivationEpochState;
+            use crate::components::vc_derivation_info::VirtualClientsError;
```
raphaelrobert: We already have feature gated imports at the beginning of the file.

raphaelrobert: We already have feature gated imports at the beginning of the file. Any reason why we have it here as well?

raphaelrobert (approved):
Just one more nit

openmls/src/group/mls_group/creation.rs
gferon: Both styles were used in the file, I ended up normalizing all imports from the existing code.

openmls/tests/virtual_clients.rs
gferon: This is already the term used in `vc_sibling_joins_higher_level_group_via_key_package_welcome`, I'll fix both.

raphaelrobert (approved):

PR #2215, opened 2026-09-11
feat: ECDSA P-256 signatures in the libcrux provider

### LINKED ISSUE #1895: Add EcDSA P256 ciphersuite to libcrux provider

niyamvora:
Closes #1895.

The KEM side of P-256 was already supported via hpke-rs-libcrux; this adds the signature side through `libcrux-ecdsa` (with `libcrux-ecdh` for public-key derivation) and enables `MLS_128_DHKEMP256_AES128GCM_SHA256_P256` in `supported_ciphersuites()`, removing the TODO.

Wire format matches the RustCrypto provider — raw 32-byte scalar private key, uncompressed SEC1 public key, DER-encoded signatures — so keys and signatures interoperate across providers. `libcrux-ecdsa` produces raw `(r, s)`, so the provider carries a minimal strict DER encoder/decoder for `ECDSA-Sig-Value`; it's unit-tested on the integer edge cases and cross-checked against the RustCrypto provider's DER handling in `tests/p256.rs` (both signing directions, both key origins, repeated to exercise varying integer lengths). Happy to swap the hand-rolled DER for a crate if you'd rather not carry it.

Key generation samples the scalar through the provider DRBG (matching the Ed25519 arm, so reseed failures propagate); the ECDSA nonce is rejection-sampled inside libcrux from the OS rng.

Testing: full `openmls` suite over the new libcrux P-256 ciphersuite in release (~275 tests green), cross-provider signature tests, clippy clean with and without `draft-ietf-mls-pq-ciphersuites`.

### DIFF
--- a/libcrux_crypto/src/crypto.rs
+++ b/libcrux_crypto/src/crypto.rs
@@ -67,7 +67,7 @@ impl OpenMlsCrypto for CryptoProvider {
         }?;
 
         match ciphersuite.signature_algorithm() {
-            SignatureScheme::ED25519 => Ok(()),
+            SignatureScheme::ED25519 | SignatureScheme::ECDSA_SECP256R1_SHA256 => Ok(()),
             #[cfg(feature = "draft-ietf-mls-pq-ciphersuites")]
             SignatureScheme::MLDSA44 | SignatureScheme::MLDSA65 | SignatureScheme::MLDSA87 => {
                 Ok(())
@@ -115,8 +115,7 @@ impl OpenMlsCrypto for CryptoProvider {
             Ciphersuite::MLS_192_MLKEM768_AES256GCM_SHA384_MLDSA65,
             #[cfg(feature = "draft-ietf-mls-pq-ciphersuites")]
             Ciphersuite::MLS_256_MLKEM1024_AES256GCM_SHA384_MLDSA87,
-            // TODO: enable
-            //Ciphersuite::MLS_128_DHKEMP256_AES128GCM_SHA256_P256,
+            Ciphersuite::MLS_128_DHKEMP256_AES128GCM_SHA256_P256,
         ]
     }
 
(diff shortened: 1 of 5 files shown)


franziskuskiefer (changes requested):
Thanks! This was on my todo list as well 😄 We just merged https://github.com/celabshq/libcrux/pull/1585 in libcrux, which allows you to drop the der bits here. This will need to depend on git for now. But we'll do a libcrux release soon.

I'll have to look at the rest, but at first glance that looks good.

niyamvora:
Done — dropped the hand-rolled DER and switched to `Signature::to_der`/`from_der` from celabshq/libcrux#1585, with `libcrux-ecdsa` joining the existing `[patch.crates-io]` git pin until the release. Two nice side effects of the new rev: `PrivateKey::public_key()` let me drop the `libcrux-ecdh` dependency, and the fallible-rng API from celabshq/libcrux#1584 means the ECDSA nonce is now drawn through the provider DRBG instead of the OS rng, so reseed failures propagate like the Ed25519 arm. Cross-provider DER tests still green against RustCrypto; full P-256 + PQ suites re-run on the advanced pin.

libcrux_crypto/src/crypto.rs
```diff
@@ -60,14 +60,41 @@ impl CryptoProvider {
     }
 }
 
+/// A fallible rng over the provider's HMAC-DRBG, for libcrux APIs that take a
+/// [`rand::TryCryptoRng`]. Randomness failures surface as errors instead of
+/// panics.
+struct DrbgRng<'a>(&'a CryptoProvider);
```
franziskuskiefer, line 66: This shouldn't be necessary. The Rng we have here can just be used. Just get a lock on `self.rng` and pass it in.

libcrux_crypto/src/crypto.rs
```diff
@@ -288,6 +314,27 @@ impl OpenMlsCrypto for CryptoProvider {
...
+                        // private key is the raw 32-byte scalar, the public key
+                        // the uncompressed SEC1 point.
+                        let pk = sk.public_key().map_err(|_| CryptoError::SigningError)?;
+                        let mut pk_sec1 = Vec::with_capacity(65);
+                        pk_sec1.push(0x04);
+                        pk_sec1.extend_from_slice(pk.as_ref());
+                        return Ok((candidate.to_vec(), pk_sec1));
+                    }
+                }
+                Err(CryptoError::SigningError)
```
franziskuskiefer, line 336: This shouldn't be necessary. We can just map the error.
```
let sk = p256::PrivateKey::random(&mut drbg).map_err(|_| CryptoError::SigningError)?;
```

franziskuskiefer (changes requested):
Thanks I think this mostly looks good. But we can simplify it a bit.

PR #2219, opened 2026-09-12
fix: derive implicit capabilities from crypto provider

### LINKED ISSUE #2218: Derive implicit capabilities from the configured crypto provider

When callers omit explicit capabilities, several OpenMLS builder paths use
`Capabilities::default()`. That default contains a global ciphersuite list and
does not reflect the ciphersuites actually supported by the configured crypto
provider.

This can cause a client using a custom or deliberately restricted provider to
advertise ciphersuites it cannot use, or omit additional ciphersuites it does
support.

[Section 7.2 of RFC 9420](https://datatracker.ietf.org/doc/html/rfc9420#section-7.2)
states that there are no default values for the versions, ciphersuites, and
credential types in a capabilities object, and that clients must list the
values they support. The capabilities carried by a
[KeyPackage](https://datatracker.ietf.org/doc/html/rfc9420#section-10)
also participate in parameter negotiation and downgrade protection.

Provide a provider-aware capabilities constructor that derives its ciphersuite
list from the configured `OpenMlsCrypto` implementation.

Use those provider-derived capabilities when a caller omits capabilities while
creating a group, creating a KeyPackage, or joining through an external commit.
Explicit capabilities supplied by the caller must remain unchanged.

Update the generic provider test matrix so each test uses the ciphersuite
selected for that provider instead of silently reverting to a global default.

Requiring every caller to provide explicit capabilities would avoid the
mismatch, but it would make the existing builder defaults substantially less
useful and would be a compatibility burden.

Expanding the global default would not solve the underlying problem because no
single global list can accurately describe every custom provider.

**Additional context**

The proposed behavior does not select a ciphersuite on behalf of the caller or
broaden a provider's supported surface. It only makes implicit capability
advertisement reflect the provider that will perform the cryptographic
operations.

vanderbr:
Closes #2218.

## Summary

- Add `Capabilities::for_provider` to derive advertised ciphersuites from the
  configured crypto provider.
- Use provider-derived capabilities only when group, KeyPackage, or external
  join callers have not supplied explicit capabilities.
- Preserve all explicit caller capability settings.
- Correct the generic provider test matrix so tests exercise the ciphersuite
  selected for each provider.

## Specification context

[Section 7.2 of RFC 9420](https://datatracker.ietf.org/doc/html/rfc9420#section-7.2)
requires a client to list the protocol versions, ciphersuites, and credential
types it supports. In addition,
[Section 10](https://datatracker.ietf.org/doc/html/rfc9420#section-10)
explains that KeyPackage capabilities are used for parameter negotiation and
downgrade protection.

A global default cannot accurately represent an arbitrary crypto provider.
Deriving omitted capabilities from the configured provider keeps the advertised
ciphersuite list aligned with the implementation that will be asked to use it.

## Design and alternatives

The provider-aware behavior is applied at the builder boundary, where the
provider is available. `Capabilities::default()` remains available, and
explicit capabilities continue to take precedence.

Requiring explicit capabilities for every builder invocation was considered,
but would impose a compatibility and usability cost on existing callers.
Expanding the global default was rejected because it would remain inaccurate
for restricted and custom providers.

## Effects

Applications that rely on implicit capabilities may advertise a different,
narrower ciphersuite list when using a restricted provider. This is intentional:
the resulting list reflects what that provider reports it can support.

There is no change for callers that set capabilities explicitly.

## Verification

- `cargo fmt --all -- --check`
- Focused tests for implicit and explicit group capabilities.
- Focused tests for implicit and explicit KeyPackage capabilities.
- Provider-matrix coverage for group operations, external joins, and virtual
  clients.
- Clippy with warnings denied on the affected test configurations.

### DIFF
--- a/openmls/src/group/mls_group/builder.rs
+++ b/openmls/src/group/mls_group/builder.rs
@@ -1,6 +1,8 @@
 use openmls_traits::{crypto::OpenMlsCrypto, signatures::Signer, types::Ciphersuite};
 use tls_codec::Serialize;
 
+use super::config::CapabilitiesSource;
+
 #[cfg(feature = "extensions-draft")]
 use crate::schedule::application_export_tree::ApplicationExportTree;
 use crate::{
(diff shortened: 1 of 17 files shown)


franziskuskiefer (commented):
The default values in capabilities are a general issue right now that need to get fixed more comprehensively. Do you want to take a stab at that? I have some options for that as well on a branch that you can take as a starting point.

